PR #367, opened 2018-09-02
Add a list of all users with ability to unset expiration for admins.

oponder:
A convenience page for admins. This lists all users, and shows a little `x` button next to the expiration for removing the expiration date from that user.

Could be worked out further for adding and removing users, as well as a place for an eventual  'invite user' button.

But for now, this is a convenience since some accounts are expiring now and I want to be able to quickly unexpire them.


oponder:
Just a little saturday/sunday coding. 

Admin only, and possibly a jumping off point for a user facing feature that can be worked out better.

Not meaning to cut corners in any processes.

marians:
My two cents: This is a lot of UI for something we could do via a script.

PR #368, opened 2018-09-09
Add a delete user button to the users screen.

oponder:
This will allow admins to delete users easily.


oponder:
delete

src/actions/userActions.js
```diff
@@ -286,7 +286,7 @@ export function usersLoad() {
 }
 
 
-// removeExpiration
+// userRemoveExpiration
```
oponder: unrelated comment fix

PR #369, opened 2018-09-09
Request installation specific audience on SSO login.

oponder:
This adds the installation name to the configuration and uses it to request a installation specific audience claim during SSO.

Towards: https://github.com/giantswarm/giantswarm/issues/4156

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -21,7 +21,8 @@
     passageEndpoint: 'http://localhost:5001',
     environment: 'development',
     version: 'development',
-    ingressBaseDomain: 'k8s.sample.io'
+    ingressBaseDomain: 'k8s.sample.io',
+    installationName: 'development',
   };
   window.config = config;
   </script>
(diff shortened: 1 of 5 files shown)


src/index.html
```diff
@@ -21,7 +21,8 @@
     passageEndpoint: 'http://localhost:5001',
     environment: 'development',
     version: 'development',
-    ingressBaseDomain: 'k8s.sample.io'
+    ingressBaseDomain: 'k8s.sample.io',
+    installationName: 'development',
```
oponder: This is the 'default' installation name. It will get overwritten on deployment in the cluster because of the start.sh script which boots up Happa and replaces this value with something present in the ENV variable INSTALLATION_NAME

oponder:
Adjusted to request the api endpoint url instead of `giantswarm-api-INSTALLATIONCODENAME`

PR #370, opened 2018-09-12
Add mail domain column to user lists

marians:
The rationale here is to improve the "scannability" of the user tables and make it visible which domains (in many cases equalling companies) a user comes from. Check for example https://happa.g8s.gollum.westeurope.azure.gigantic.io/users/ for an installation with many users.

In a next step, these tables could be made sortable. Then having this column would allow to sort by email domain.

One thing I am unsure of: Would it be better to truncate the content of the `Email` column to the first part only? Like this:

| Email | Email Domain
|------|------|
| foo@ | example.com |
| bar@ | example.com |

This kind of defaces the email address as such.

### Preview

image

image

### DIFF
--- a/src/components/organizations/detail.js
+++ b/src/components/organizations/detail.js
@@ -93,6 +93,7 @@ class OrganizationDetail extends React.Component {
                         <thead>
                           <tr>
                             <th>Email</th>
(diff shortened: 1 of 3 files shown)


marians:
All right, thanks

PR #372, opened 2018-09-13
Testing sortable tables

marians:
This is the attempt to make tables sortable by using [react-bootstrap-table2](https://github.com/react-bootstrap-table/react-bootstrap-table2).

Preview for the clusters table in org details:

gxvvwxvpcs

-----

Current state:

image

image

image

So this also changes the action buttons and aligns them more.

Note that I also added the creation date to the users table.


oponder:
Love it!

> My understanding of the bind keyword is too limited here

Wanna pair about it?

marians:
See updated main comment

oponder:
We'll have to update this for the changes in the user overview screen after the invitation feature has been added. But I'd like to get this in Happa sooner rather than later. Still feel like this is a good UX improvement @marians ?

marians:
Totally forgot about this

marians:
This is currently broken. The file `src/components/users/index.js` makes no sense, as I just accepted both the branches changes and those from master. I did that to simply have my changes committable.

To make the branch useable again, the users table will have to be built based on a data structure combined from invites and users.

oponder:
Needs a bit of CSS polishing to get the table widths right, or removing the `created` column, which was new. 

But I got this branch up to date again and working with the invitations in the list.

screen shot 2018-10-31 at 16 57 51

marians:
Thanks @oponder for fixing this! What do you think, should we merge?

oponder:
@marians the table is a bit cramped now, and getting it to resize nicely is giving me some trouble..

How do you feel about removing the `Created` column? that would give some space.

Also this PR changes the icons in the table to a `delete` button, is that intentional? I feel like it gets a bit busy.

marians:
Yes, the delete (user) and also remove (org mamber) button change was intentional. I am not happy with the icons, as I think they are not explicit enough.

Maybe we need something in-between. We could improve the fading on mouse-over, so that the column of buttons doesn't look as heavy as it does now. Should we iterate in another PR?

oponder:
Ok, I did a bit of CSS finagling, I'm ok with this now.

marians:
Superb! Will take a look now

Issue #134, opened 2017-04-12
The loading spinner is not vertically centered on the account creation form


oponder:

oponder:
Fixed

PR #375, opened 2018-09-24
Fix valid signup form despite TOS being unchecked.

oponder:
There appears to have been a regression in the sign up form. When reaching the final step, where the TOS has to be checked,
the button would appear green instead of disabled.

Clicking on the green button without checking the TOS would put the user in a invalid state.

This ensures the form does not go 'green' when reaching the final step.


oponder:
Also fixes: https://github.com/giantswarm/happa/issues/134 (The loading spinner is not vertically centered on the account creation form)

PR #376, opened 2018-09-25
Add the list of pending invitations to the user overview screen.

oponder:
Towards: https://github.com/giantswarm/giantswarm/issues/4259

Now that passage can list pending invites, this PR surfaces that information so that it is visible in Happa.

Related: https://github.com/giantswarm/passage/pull/75


oponder:
What it looks like:

screen shot 2018-09-25 at 17 17 29

src/components/users/user_row.js
```diff
@@ -33,6 +33,14 @@ class UserRow extends React.Component {
             </div>
           }
         </td>
+        <td>
```
oponder: Since we added the `STATUS` column, this also adds some info for normal users there.

PR #377, opened 2018-09-25
Add a way for admins to invite users.

oponder:
Towards: https://github.com/giantswarm/giantswarm/issues/4259

The 'invite user' button is added to the top right of the User overview screen, and displays this modal:

screen shot 2018-09-25 at 22 23 44

WIP because I want to add the 'success' state for the modal where I could show the invitation accept link, in case the user did not check the 'send email' checkbox.


src/actions/invitationActions.js
```diff
@@ -51,3 +51,46 @@ export function invitationsLoad() {
...
+        class: 'success',
+        ttl: 3000
+      }));
+
+      dispatch(invitationsLoad());
+
+      return result;
+    })
+    .catch(error => {
+      console.error(error);
```
pipo02mix: is for debugging or you want to leave a message in the console?

src/components/users/index.js
```diff
@@ -235,6 +316,76 @@ class Users extends React.Component {
...
+                      <BootstrapModal.Header closeButton>
+                        <BootstrapModal.Title>Invite a New User</BootstrapModal.Title>
+                      </BootstrapModal.Header>
+
+                      <BootstrapModal.Body>
+                        <form>
+                        <p>Creating an invitation is the way to get new people onto this installation.</p>
+                        <p>You can chose whether or not to send the invitee an email, and which organization the invitee should become a member of once they accept their invitation.</p>
+                        <p>Invitations are valid for 48 hours, after which you&apos;ll have to re-invite the person.</p>
+
```
pipo02mix: `invitee` -> `invite`

src/components/users/user_row.js
```diff
@@ -36,8 +36,8 @@ class UserRow extends React.Component {
         <td>
           <small>
           {
-            expirySeconds < 0 ?
-            "EXPIRED" : "ACTIVE"
+            expirySeconds < 0 && this.props.user.expiry != NEVER_EXPIRES?
+            'EXPIRED' : 'ACTIVE'
           }
```
pipo02mix: nitpick but space before `?`

src/lib/helpers.js
```diff
@@ -103,12 +103,12 @@ export function validateOrRaise(validatable, constraints) {
...
 export function formatDate(ISO8601DateString) {
   // http://momentjs.com/docs/#/displaying/
-  return moment(ISO8601DateString).utc().format('D MMM YYYY, HH:mm z');
+  return moment.utc(ISO8601DateString).format('D MMM YYYY, HH:mm z');
 }
 
 export function relativeDate(ISO8601DateString) {
   var formatedDate = formatDate(ISO8601DateString);
-  var relativeDate = moment(ISO8601DateString).utc().fromNow();
+  var relativeDate = moment(ISO8601DateString).fromNow();
```
pipo02mix: does this change affect other callers that expect `formatDate` return value to be in UTC?

src/components/users/index.js
```diff
@@ -235,6 +316,76 @@ class Users extends React.Component {
...
+                          </DropdownButton>
+                        </div>
+
+                        <div className='textfield'>
+                          <label>Send Email:</label>
+                          <div className='checkbox'>
+                            <label htmlFor='sendEmail'>
+                              <input type='checkbox' onChange={this.handleSendEmailChange.bind(this)} id='sendEmail' checked={this.state.invitationForm.sendEmail} />
+                              Send the invitee an e-mail with the accept invitation link.
+                            </label>
```
pipo02mix: `invitee` -> `invite`

src/lib/helpers.js
oponder: No it was wrong everywhere I think, this should fix it correctly. The thing is that moment was parsing the timestamp as if it were in the local timezone.

The change I did ensures `moment` thinks it is in UTC timezone, which all our timestamps are in.

src/components/users/index.js
oponder: invitee in this case is right I think, it means an invited person.

src/components/users/index.js
oponder: same as before: invitee in this case is right I think, it means an invited person.

src/components/users/index.js
```diff
@@ -235,6 +316,76 @@ class Users extends React.Component {
...
+                  case 'inviteUser':
+                    return <BootstrapModal className='create-key-pair-modal' show={this.state.modal.visible} onHide={this.closeModal.bind(this)}>
+                      <BootstrapModal.Header closeButton>
+                        <BootstrapModal.Title>Invite a New User</BootstrapModal.Title>
+                      </BootstrapModal.Header>
+
+                      <BootstrapModal.Body>
+                        <form>
+                        <p>Creating an invitation is the way to get new people onto this installation.</p>
+                        <p>You can chose whether or not to send the invitee an email, and which organization the invitee should become a member of once they accept their invitation.</p>
```
marians: > You can chose whether or not to send the invitee an email

That part is really obvious from the checkbox down there. It should also be the case for 99% of invites. So I'd remove this.

> and which organization the invitee should become a member of

This is also really obvious. I'd remove the paragraph as a whole.

src/components/users/index.js
```diff
@@ -235,6 +316,76 @@ class Users extends React.Component {
...
+                        <p>Invitations are valid for 48 hours, after which you&apos;ll have to re-invite the person.</p>
```
marians: > after which you&apos;ll have to re-invite the person

I think this can be guessed. Also not relevant in the situation the user is right now.

src/components/users/index.js
pipo02mix: sorry I was reading in diagonal

src/components/users/index.js
```diff
@@ -235,6 +316,76 @@ class Users extends React.Component {
...
+                          <label>Organization:</label>
+                          <DropdownButton id="organizationDropdown" title={this.state.invitationForm.organization}>
+                            {
+                              _.map(_.sortBy(this.props.organizations.items, 'id'), (organization) =>
+                                <MenuItem key={organization.id} onClick={this.handleOrganizationChange.bind(this, organization.id)}>
+                                  {organization.id}
+                                </MenuItem>
+                              )
+                            }
+                          </DropdownButton>
```
marians: Is the currently selected organization pre-selected in this dropdown?

marians:
A little bit too much text for my taste, otherwise good!

src/components/users/index.js
oponder: Currently it's just the first organization in the list. Makes sense for it to be the currently selected org. I'll change that.

src/actions/invitationActions.js
oponder: Yeah in this case I want to log the error, otherwise it gets swallowed completely. It's kind of the pattern i've been following so far.

oponder:
RFR, code is running on `ginger`

PR #380, opened 2018-09-27
Update dependencies

marians:
This updates a few dependencies with patch or minor upgrades.


marians:
tested only locally in dev mode so far, will deploy next

PR #381, opened 2018-09-30
Fix relative time text

oponder:
It was off for me since it was parsing the timestamp in the local timezone as well.

### DIFF
--- a/src/lib/helpers.js
+++ b/src/lib/helpers.js
@@ -108,7 +108,7 @@ export function formatDate(ISO8601DateString) {
 
 export function relativeDate(ISO8601DateString) {
   var formatedDate = formatDate(ISO8601DateString);
-  var relativeDate = moment(ISO8601DateString).fromNow();
+  var relativeDate = moment.utc(ISO8601DateString).fromNow();
 
   return <OverlayTrigger placement="top" overlay={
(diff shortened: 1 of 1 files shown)


src/lib/helpers.js
```diff
@@ -108,7 +108,7 @@ export function formatDate(ISO8601DateString) {
 
 export function relativeDate(ISO8601DateString) {
   var formatedDate = formatDate(ISO8601DateString);
-  var relativeDate = moment(ISO8601DateString).fromNow();
+  var relativeDate = moment.utc(ISO8601DateString).fromNow();
```
pipo02mix: So you pass the date in local time, then move it to UTC and get the time passed. But I was thinking `fromNow` use local time to calculate, isnt it?

oponder: Hmm... my thinking is that once we add the .utc() it'll be in 'utc' timezone and everything after it should be UTC as well?

The issue is that the timestamp we get in has no timezone information in it. So moment.utc should force to parse it in the UTC timezone.

marians:
I don't understand the problem, so can't review.

src/lib/helpers.js
pipo02mix: ok I see

PR #384, opened 2018-10-08
Show BYOC credentials in the organization details context

marians:
For https://github.com/giantswarm/giantswarm/issues/3987

The title says it.


marians:
Previews

image

image

(Account ID redacted)

marians:
@oponder One question. Given this SSO scenario:

1. Some page in happa opened
2. Not doing anything for > 5 minutes to let the token expire
3. Go to `Details for <orgname>` in menu to open organization details

This will now create an authentication error with the according flash message that credential details could not be loaded. What do I have to change to prevent this?

src/components/organizations/credentials.js
```diff
@@ -0,0 +1,125 @@
...
+          var parts = this.props.credentials.items[0].aws.roles.awsoperator.split(':');
+          var accountID = parts[4];
+          return (
+            <div>
+              {providerWarning}
+              <table className='table resource-details' id={'credential-' + this.props.credentials.items[0].id}>
+                <tbody>
+                  <tr key='account_id'>
+                    <td>AWS account ID</td>
+                    <td className='value code'>{ accountID } <a href={'https://' + accountID + '.signin.aws.amazon.com/console'} target='_blank' rel='noopener noreferrer'><i className='fa fa-external-link'></i></a></td>
```
pipo02mix: nice

src/components/organizations/credentials.js
```diff
@@ -0,0 +1,125 @@
...
+                    <td className='value code'>{ this.props.credentials.items[0].aws.roles.admin }</td>
+                  </tr>
+                </tbody>
+              </table>
+            </div>
+          );
+        
+        // Azure credential
+        } else if (typeof this.props.credentials.items[0].azure !== 'undefined') {
+          console.log(this.props.credentials.items[0].azure);
```
pipo02mix:
```suggestion
```
Removing?

src/components/organizations/credentials.js
```diff
@@ -0,0 +1,125 @@
...
+          return (
+            <div>
+              {providerWarning}
+              <table className='table resource-details'>
+                <tbody>
+                  <tr key='account_id'>
+                    <td>Azure subscription ID</td>
+                    <td className='value'>{ this.props.credentials.items[0].azure.credential.subscription_id }</td>
+                  </tr>
```
pipo02mix: can we link to azure portal too?

src/components/organizations/detail.js
```diff
@@ -31,7 +32,30 @@ class OrganizationDetail extends React.Component {
     this.props.dispatch(push('/organizations/' + this.props.organization.id + '/clusters/' + cluster));
   }
 
+  // determine whether the component should deal with BYOC credentials
+  // (not relevant on KVM)
+  canCredentials = (provider) => {
+    if (provider === 'aws' || provider === 'azure') {
```
pipo02mix: nit
`canCredentials ` -> `hasCredentials ` ?

oponder:
> This will now create an authentication error with the according flash message that credential details could not be loaded. What do I have to change to prevent this?

Hmm, this shouldn't happen. It should refresh the token before using it. Maybe there was a different underlying error? Will test myself as well.

src/components/organizations/credentials.js
marians: No, haven't found a way

oponder:
> Hmm, this shouldn't happen. It should refresh the token before using it. Maybe there was a different underlying error? Will test myself as well.

Ok I managed to recreate it.. that's not the intention. Would like to treat it as a bug, not in scope of this PR. So as far as I can see you've set up things correctly, and something needs to be changed in the way happa manages the refresh process and state.

src/components/organizations/credentials.js
```diff
@@ -0,0 +1,124 @@
...
+      );
+    } else {
+      if (this.props.credentials.items.length === 0) {
+
+        var gsctlInfo = <p>You can set credentials to use your own subscription using <a href='https://docs.giantswarm.io/reference/gsctl/update-org-set-credentials/' target='_blank' rel='noopener noreferrer'>our CLI</a></p>;
+
+        if (provider === 'azure') {
+          return (
+            <div>
+              <p>No credentials set. Clusters of this organization will be created in the default tenant cluster subscription configured for this installation.</p>
```
JosephSalisbury: > No credentials set.
This reads a bit odd/worrying to me. There are _some_ credentials set (the default ones).

Maybe something more like 'No additional credentials available'?

marians: Interesting. The last thing I want is to worry non-BYOC users that something is wrong. How about "No specific provider credentials set. (...)"?

JosephSalisbury: I think that's an improvement. We can iterate with user feedback if needed.

src/components/organizations/detail.js
marians: This function should return whether the current context (installation) supports credentials.

PR #385, opened 2018-10-10
Add validation for the CN prefix in key pair creation

marians:
For https://github.com/giantswarm/giantswarm/issues/4275

### Preview

cn-prefix-validation

### DIFF
--- a/src/components/organizations/cluster_key_pairs.js
+++ b/src/components/organizations/cluster_key_pairs.js
@@ -57,7 +57,7 @@ class ClusterKeyPairs extends React.Component {
       var regex = /^[a-z0-9][a-z0-9-]*[a-z0-9]$/g;
       var found = e.target.value.match(regex);
       if (found === null) {
-        error = 'Please use characters a-z 0-9 and -';
+        error = 'Please use characters a-z, 0-9 or -';
       }
     }
 
--- a/src/components/organizations/cluster_key_pairs.js
+++ b/src/components/organizations/cluster_key_pairs.js
@@ -25,6 +25,7 @@ class ClusterKeyPairs extends React.Component {
       expireTTL: 720,
       description: this.defaultDescription(props.user.email),
       cn_prefix: '',
+      cn_prefix_error: null,
       certificate_organizations: '',
       modal: {
         visible: false,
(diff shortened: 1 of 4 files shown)


src/components/organizations/cluster_key_pairs.js
```diff
@@ -51,8 +52,18 @@ class ClusterKeyPairs extends React.Component {
   }
 
   handleCNPrefixChange(e) {
+    var error = null;
+    if (e.target.value !== '') {
+      var regex = /^[a-z0-9][a-z0-9-]*[a-z0-9]$/g;
+      var found = e.target.value.match(regex);
+      if (found === null) {
+        error = 'Please use characters a-z 0-9 and -';
+      }
```
pipo02mix: Nitpick 
`Please use characters a-z, 0-9 or -`

src/styles/components/_form.sass
```diff
@@ -26,6 +26,8 @@ form
   .text-field-hint
     font-size: 12px
     color: #ccc
+    .error
+      color: #771d1d
```
pipo02mix: To me the contrast is not fine. I would select another color tone which highlights from the background

marians:
Totally agreeing. What was I thinking?

Unfortunately, red on blue doesn't work very well in any case. And I don't want to make it pink. So I am using the "gold" color we use for the "upgrade available" notification. WDYT?

image

marians:
Contrast is still not very good. This is also because we use a lighter background in modals.

Do we need that lighter background, especially as we place it above a dark glass pane?

Here is how modals would look with our standard background color:

image

This is how they look currently:

image

marians:
Opinions? Should we make the modal darker?

oponder:
I think the gold on the lighter modal looks ok!

Darker modal is also ok, but need to do a bit of exploratory testing and make sure it looks ok for other modals too.

pipo02mix:
Agree with Oli

src/components/organizations/cluster_key_pairs.js
marians: done

src/styles/components/_form.sass
marians: I settled for the darker blue in the modal

marians:
I changed the modal background to the darker blue and clicked through the other dialogs. Looks good to me. Actually this bring more contrast to the modals over all, which is a good thing I think.

PR #387, opened 2018-10-16
Relax CN prefix validation to allow for email addresses

marians:
This should fix https://github.com/giantswarm/giantswarm/issues/4445


marians:
Can be tested on gollum

marians:
No, certain characters are not allowed in the end or beginning.

PR #389, opened 2018-10-17
Throw clusterLoadDetails error so that the frontend can actually handle it.

oponder:


marians:
Is there an example where this is handled? Looks like there was a flash message before, but now there isn't?

oponder:
I moved the flash message so that other calls to load cluster detail will also throw the flash message.

Since the homepage also loads cluster details.

marians:
Ah, yes, My bad eyes.

PR #393, opened 2018-10-31
Remove componentWillMount

marians:
I came accros some uses of `componentWillMount`, which is to be deprecated in React 16.3. See https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#initializing-state

This is so far only tested locally. Needs testing in SSO/OAuth context.

This also uses the most simple way to initialize the component state.


marians:
Checked on godsmack and gorgoth using SSO

PR #397, opened 2018-11-02
SASS and dependency boyscouting

marians:
- Update node in dev to 10
- Update node-sass
- Replace SCSS with SASS for consistency and simplicity
- Adds a missing dependency js-base64
- Change syntax to import CSS in JS to use import over require
- Update eslint
- Update webpack
- Update webpack-dev-server


marians:
Had to fix something I broke through merging in master. Sorry for the noise.

PR #398, opened 2018-11-05
add m5.12xlarge

jgsqware:
Signed-off-by: Julien Garcia Gonzalez <julien@giantswarm.io>


marians:
What I see in the diff doesn't do what the title indicates.

I guess the goal has changed in-between. I recommend updating the title.

oponder:
@jgsqware is this still required by the customer? They want `r5.4xlarge` now right? Best to change the titles.

Related: https://github.com/giantswarm/cluster-service/pull/344
Related: https://github.com/giantswarm/installations/pull/550

PR #401, opened 2018-11-06
Set credentials on org details page

marians:
For https://github.com/giantswarm/giantswarm/issues/3987

This PR adds the capability to set provider credentials (BYOC) in the organization details page. Before this, we had a link to the gsctl reference.

Can be tested currently on

- gorgoth (Azure)
- ginger (AWS)
- godsmack (KVM, so form should not appear)


marians:
Did I ask for a review yet? If yes, then that was by accident. I am still working on this.

marians:
Status update: On azure, setting credentials works, but after that the reading of credentials fails. Error:

`TypeError: Cannot read property 'subscription_id' of undefined`

Possibly that error has already existed in master. Anyway, I want to fix it before merging.

marians:
Just removed `babel-preset-stage-2` as it is in fact no longer needed.

marians:
In case you want to test:

- Open the developer console of your browser
- Open happa in one of the installations named above
- Create a new organization
- Go to `Details of <new-organization>`
- There should be no provider credentials.
- it the `Set credentials` button.
- Add some credentials into the form. See the [API docs](https://docs.giantswarm.io/api/#operation/addCredentials) for example data.
- Submit the form.
- You should now see credentials on the organization details page.

marians:
Seems like no-one can review this properly ow. Merging without approval. 🙏

PR #402, opened 2018-11-06
Dynamic truncation for the key pairs table

marians:
The key pairs table with the static truncation didn't really use the available space for the best usability.

This PR makes the truncation dynamic, to let the table use the entire column width. In addition, some columns are hidden entirely in smaller viewports, to be able to render a recognizable table as such.

## Preview

screen-recording-2018-11-06-at-14 10

### DIFF
--- a/src/components/organizations/cluster_key_pairs.js
+++ b/src/components/organizations/cluster_key_pairs.js
@@ -218,7 +218,7 @@ class ClusterKeyPairs extends React.Component {
         </div>
 
         <div className='row'>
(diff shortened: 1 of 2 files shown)


marians:
Sorry @xh3b4sd, my workflow for recording a screen section to an animated GIF is still not very straightforward. I saved the work. But as it would be nice to have it for the archive, here is the current way:

kt88wtuk4g

The major difference of the current way is the line breaks in the description field, which make the table look weird, plus truncated columns not making good use of the available space.

PR #405, opened 2018-11-07
Remove use of deprecated componentWillReceiveProps

marians:
Setting eslint to the right React version yields complaints about us using `componentWillReceiveProps`, which is deprecated since 16.3.0.

Details: https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops

This replaces the former usage of `componentWillReceiveProps` in the following components:

- `InputField`, used in cluster creation to set worker machine/instance type, for both Azure and AWS.
- `SignUp`, which is the signup form.
- `BarChart` which is not in use.

### DIFF
--- a/src/components/home/bar_chart.js
+++ b/src/components/home/bar_chart.js
@@ -39,7 +39,7 @@ class BarChart extends React.Component {
   componentDidUpdate(prevProps) {
     if (this.props.percentage !== prevProps.percentage) {
       this.setState({
-        percentage: this.props.percentage
+        percentage: this.props.percentage,
       });
     }
   }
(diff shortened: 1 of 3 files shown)


marians:
Still have to test the signup. And I have merged conflicting changes in another branch where I simply renamed `componentWillReceiveProps` to `UNSAFE_componentWillReceiveProps` to silence webpack/eslint.

oponder:
Yeah I saw that, but didn't feel too great about it though. I'd rather see the deprecation errors so we don't forget. But its ok. We'll figure this branch out and then it won't matter anymore.

marians:
Signup is not working. Will fix later.

oponder:
Rebased, and fixed the broken signup screen.

This should be good to go now. Deployed to ginger for test in case you want to look around.

marians:
Just tested the signup again. Thanks for making this work! ❤️

PR #406, opened 2018-11-12
Add dependency cache

marians:
For https://github.com/giantswarm/giantswarm/issues/4672

This PR attempts to speed up builds in CI. It does so by

- activating a cache on CircleCI for the `node_modules` folder
- using `yarn` to install dependencies
- maintaining a `yarn.lock` file in the repo which represents the exact dependency versions we are using
- Removing executing of `ncu` (dependency update check)
- Using our own [base image](https://github.com/giantswarm/happa-build) for the container running `grunt`, instead of using node:slim and installing dependencies every time.

A CI run now takes ~1:15 when dependencies have not changed. Most of this time goes into `grunt build` (0:40) and building the final docker image (0:20).

The `make develop` for local development is also quite fast. To test it out, please remove existing `node_modules*` directories and then run it twice. First run should be ~1 minute, depending on network, second run should be ~10 seconds (both only counting yarn install, not accounting for webpack).


marians:
Oh, idea: I need to actually execute something in the container, otherwise it will close immediately. :o

.circleci/config.yml
```diff
@@ -5,6 +5,23 @@ jobs:
     steps:
     - checkout
 
+    - run:
+        name: Save the current month to a temp file (for restore_cache)
```
JosephSalisbury: Month and year, right?

.circleci/config.yml
```diff
@@ -5,6 +5,23 @@ jobs:
...
+        command: date +"%Y-%m" > /tmp/month
+  
+    - restore_cache:
+          keys:
+            # Find a cache corresponding to this specific package.json checksum
+            # created in this month.
+            - v3-npm-deps-{{ checksum "package.json" }}-{{ checksum "/tmp/month" }}
+    
+    - run:
+        name: Inspect cache
```
JosephSalisbury: This is just to check the modules?

.circleci/config.yml
```diff
@@ -4,7 +4,21 @@ jobs:
     machine: true
     steps:
     - checkout
-
+  
+    - restore_cache:
+          keys:
+            # Find a cache corresponding to this specific package.json checksum
+            # created in this month.
+            - v4-npm-deps-{{ checksum "package.json" }}-{{ checksum "yarn.lock" }}
```
oponder: I think the checksum of just yarn.lock should be sufficient no?

PR #407, opened 2018-11-20
Add and run prettier

oponder:
So that we have a standardised code style.

We could do a git commit hook that runs prettier automatically, but I'm not a big fan of that. 

I'd rather the developers set up 'prettier' to run on save in their editors, or remember to manually run prettier.

Regardless, I do want to add some CI step which checks if the code is 'prettified', kind of like we do for our golang projects.


oponder:
@marians WDYT? This PR adds prettier and a validation step to the CI.

It is up to us to configure our dev environment so that prettier runs before pushing to the repo.

I set it up to run on save of an individual file, so code should always be correctly styled.

src/actions/invitationActions.js
```diff
@@ -15,43 +15,58 @@ export function invitationsLoad() {
...
+            class: 'danger',
+          })
+        );
+
+        dispatch({
+          type: types.INVITATIONS_LOAD_ERROR,
+        });
+
+        console.error(error);
+        throw error;
```
marians: What happened here?

marians: Oh, that's by adding indentation.

marians:
I like it! Most changes are obvious improvements and the rest might take getting used to.

Should we add some explanation to the README?

For VS code users: https://github.com/prettier/prettier-vscode

marians:
Are you still working on this, or do you want formal approval?

oponder:
I'm still a Sublime Text user 😅 

I'll add something to the readme, since prettier does take a few config params, we do have to know what params we are all using.

README contains that, as well as the scripts section of `package.json` which the true source of truth, since that's also what CI ends up running.

oponder:
Formal approval would be nice!

PR #411, opened 2018-11-23
Take capability info from installations repo.

oponder:
Towards: https://github.com/giantswarm/giantswarm/issues/4508

The remaining capability info is now just to support the dev environment, so that we can see some instance types / vm sizes.

The real capability info gets added at run time to the `window.config` object when happa is deployed to an installation. This should achieve a single source of truth now for capability info.


src/index.html
```diff
@@ -21,7 +21,9 @@
     passageEndpoint: 'http://localhost:5001',
     environment: 'development',
     version: 'development',
-    ingressBaseDomain: 'k8s.sample.io'
+    ingressBaseDomain: 'k8s.sample.io',
+    awsCapabilitiesJSON: '',
+    azureCapabilitiesJSON: '',
   };
   window.config = config;
   </script>
```
pipo02mix: Out of curiosity, why do you define the config in the html?

oponder: I don't have a better way to get config into happa.

Happa is just static html files behind nginx, nothing dynamic going on.

We have a `start.sh` which then replaces these values with things it finds in the ENV.

If happa had some serverside component, then I could pass it in as parameters to that binary, but that is not the case.

marians:
cool!

Issue #125, opened 2017-04-03
Show a "something might be wrong" message when things take longer to load than normal on initial app load


oponder:
When the app is loaded for the first time, and organizations can't be loaded, we just show a infinite spinner instead of some nice message to the user.

oponder:
done

PR #413, opened 2018-11-27
Give user some feedback if startup requests fail, and why.

oponder:
This will give people some feedback if happa encounters an error while starting up.

If the API returns a 500 error, or is simply not reachable, a flash message will appear above the spinner.

If the API returns a 401 error, Happa will redirect to the login page and show an error message asking the user to log in again.

screenshot 2018-11-27 at 15 07 34

screenshot 2018-11-27 at 15 08 30

### DIFF
--- a/src/components/layout.js
+++ b/src/components/layout.js
@@ -53,22 +53,26 @@ class Layout extends React.Component {
         .catch(error => {
           if (error.status === 401) {
             this.props.flashActions.flashAdd({
-              message: 'Please log in again, your previously saved credentials appear to be invalid.',
+              message:
+                'Please log in again, your previously saved credentials appear to be invalid.',
               class: 'warning',
             });
 
             this.props.dispatch(push('/login'));
           } else {
             this.props.flashActions.flashAdd({
-              message: (<div>
-                <strong>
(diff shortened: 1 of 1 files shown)


oponder:
Related to: https://github.com/giantswarm/happa/issues/125

src/components/layout.js
```diff
@@ -40,13 +40,42 @@ class Layout extends React.Component {
...
+              message: (
+                <div>
+                  <strong>
+                    Something went wrong while trying to load user and
+                    organization information.
+                  </strong>
+                  <br />
+                  {error.body
+                    ? error.body.status_text
+                    : 'Perhaps our servers are down, please try again later or contact support: support@giantswarm.io'}
```
xh3b4sd: I would like to say something else. That message here implies the worst unproven assumption possible. Some more positive wording would be good. 
```suggestion
                    : 'In case the connection issue persists, please try again later or contact support: support@giantswarm.io'}
```

oponder: ah indeed... 

How about we go for just "Please try again later or contact support: support@giantswarm.io"

PR #415, opened 2018-11-27
shows AZ in cluster details page

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/pull/2202.

### DIFF
--- a/src/components/organizations/cluster_detail.js
+++ b/src/components/organizations/cluster_detail.js
@@ -419,6 +419,14 @@ class ClusterDetail extends React.Component {
                                 </td>
                               </tr>
                               <tr>
+                                <td>Number of availablility zones</td>
+                                <td className='value'>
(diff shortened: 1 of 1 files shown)


src/components/organizations/cluster_detail.js
```diff
@@ -212,6 +212,18 @@ class ClusterDetail extends React.Component {
       }
     }
 
+    var availabilityZonesOrNothing = null;
+    if (this.props.cluster.availability_zones) {
```
oponder: Q: does the API omit availability zones if there are none? 
A: yes it does (https://github.com/giantswarm/api/blob/master/server/endpoint/cluster/searcher/response.go#L20)

So, thats ok.

xh3b4sd:
screen shot 2018-11-27 at 15 57 40

screen shot 2018-11-27 at 15 57 31

screen shot 2018-11-27 at 15 57 12

PR #416, opened 2018-11-27
aligns div widths in cluster details view

xh3b4sd:
Without the change the last two boxes appear to be wider then the others. 

screen shot 2018-11-27 at 15 23 18

With the change the alignment is fixed. Note that the padding on the right side is off regardless. I do not intend to fix this here. 

screen shot 2018-11-27 at 15 23 03


marians:
Nice-y!

PR #417, opened 2018-11-28
Optimize bundle size

oponder:
Reduces happa's gzipped bundle from 580KB to ~383KB~ 471KB by only including parts of react-bootstrap and moment.js that we actually use.


oponder:
Current: 
screenshot 2018-11-28 at 13 09 51

After:
screenshot 2018-11-28 at 13 15 45

PR #421, opened 2018-11-30
Faster admin login

oponder:
This makes the `/admin-login` page a bit smarter. If credentials already exist, it attempts to use them. And otherwise it will redirect to auth0 as before.


oponder:
Go ahead and give `opsctl open -a happa -c ginger` a try!

related to: https://github.com/giantswarm/opsctl/pull/302

marians:
Great, works for me!

PR #423, opened 2018-11-30
Number picker: set fixed width

marians:
Towards https://github.com/giantswarm/giantswarm/issues/4789

This extracts the CSS for NumberPicker into a separate module. It also sets a fixed width.

image

image


marians:
Little reminder where the NumberPicker is in use:

- Cluster creation: availability zones (AWS)
- Cluster creation: KVM node options
- Scaling (all providers)

PR #424, opened 2018-12-04
feature auto scaling

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/pull/2206. This PR is the final feature branch in which we merge multiple separate changes. At the end of the feature implementation we merge this PR in order to be done.

PR #425, opened 2018-12-04
adds first UI changes

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/pull/2206. Note that this PR is based on [`feature-auto-scaling`](https://github.com/giantswarm/happa/pull/424). This is pretty much WIP as of now. I just played around to get the UI kind of aligned with the specced mockups. I need to pair a bit to sort out some quirks in the actual functionality.

screenshot 2018-12-04 at 17 48 15

### DIFF
--- a/src/components/create_cluster/index.js
+++ b/src/components/create_cluster/index.js
@@ -441,7 +441,7 @@ class CreateCluster extends React.Component {
                       numbers to the same value.
                     </p>
                     <div className='col-3'>
-                      <p>Minimum number of nodes.</p>
+                      <p>Minimum</p>
                       <NumberPicker
                         label=''
                         stepSize={1}
@@ -453,7 +453,7 @@ class CreateCluster extends React.Component {
                       />
                     </div>
                     <div className='col-3'>
-                      <p>Maximum number of nodes.</p>
+                      <p>Maximum</p>
                       <NumberPicker
                         label=''
                         stepSize={1}
--- a/src/components/create_cluster/index.js
+++ b/src/components/create_cluster/index.js
@@ -291,7 +291,8 @@ class CreateCluster extends React.Component {
                   <p>
                     Give your cluster a name so you can recognize it in a crowd.
                   </p>
-                  <input className='col-4'
+                  <input
+                    className='col-4'
                     ref={i => {
                       this.cluster_name = i;
                     }}
@@ -440,9 +441,7 @@ class CreateCluster extends React.Component {
                       numbers to the same value.
                     </p>
                     <div className='col-3'>
-                      <p>
-                        Minimum number of nodes.
-                      </p>
+                      <p>Minimum number of nodes.</p>
                       <NumberPicker
                         label=''
                         stepSize={1}
@@ -454,9 +453,7 @@ class CreateCluster extends React.Component {
                       />
                     </div>
                     <div className='col-3'>
-                      <p>
-                        Maximum number of nodes.
-                      </p>
+                      <p>Maximum number of nodes.</p>
                       <NumberPicker
                         label=''
                         stepSize={1}
--- a/src/components/create_cluster/index.js
+++ b/src/components/create_cluster/index.js
@@ -273,8 +273,7 @@ class CreateCluster extends React.Component {
                 >
                   <div>
                     <p>
-                      To have full control over cluster scaling, set both
-                      numbers to the same value.
+                      To disable autoscaling, set both numbers to the same value
                     </p>
                     <div className='col-3'>
                       <p>Minimum</p>
--- a/src/components/create_cluster/index.js
+++ b/src/components/create_cluster/index.js
@@ -4,12 +4,7 @@ import React from 'react';
 import { connect } from 'react-redux';
 import DocumentTitle from 'react-document-title';
 import Button from '../button';
-import _ from 'underscore';
 import { clusterCreate } from '../../actions/clusterActions';
-import update from 'react-addons-update';
-import NewKVMWorker from './new_kvm_worker.js';
-import NewAWSWorker from './new_aws_worker.js';
-import NewAzureWorker from './new_azure_worker.js';
 import NumberPicker from './number_picker.js';
 import ReleaseSelector from './release_selector.js';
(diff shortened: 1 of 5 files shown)


oponder:
Happy to pair!

I'm not a fan of the merge multiple PR's into a another PR thing though. Let's talk about that too.

src/components/create_cluster/index.js
```diff
@@ -441,25 +422,50 @@ class CreateCluster extends React.Component {
...
-                    <div className='new-cluster--low-worker-warning'>
-                      We recommend that you have at least three worker nodes in
-                      a cluster
+                  <div>
+                    <p>
+                      To have full control over cluster scaling, set both
+                      numbers to the same value.
+                    </p>
+                    <div className='col-3'>
+                      <p>Minimum number of nodes.</p>
```
marians:
```suggestion
                      <p>Minimum</p>
```

src/components/create_cluster/index.js
```diff
@@ -441,25 +422,50 @@ class CreateCluster extends React.Component {
...
+                        onChange={this.updateAvailabilityZones}
+                        readOnly={false}
+                      />
                     </div>
-                  ) : (
-                    ''
-                  )}
-                </div>
+                    <div className='col-3'>
+                      <p>Maximum number of nodes.</p>
```
marians:
```suggestion
                      <p>Maximum</p>
```

src/components/create_cluster/index.js
marians: A label doesn't need punctuation. Also truncated it to remove duplication with the section label.

src/components/create_cluster/index.js
```diff
@@ -380,8 +363,8 @@ class CreateCluster extends React.Component {
 
             <div>
               <div className='row section'>
-                <div className='col-12'>
-                  <h3 className='table-label'>Worker Node Configuration</h3>
+                <div className='col-3'>
+                  <h3 className='table-label'>Instance Type</h3>
```
marians: "Instance Type" is AWS-specific terminology. Fo Azure it would be "VM Size". For KVM we could leave it at "Worker Node Configuration".

src/components/create_cluster/index.js
xh3b4sd: You may notice that the implementation is based on the mockups we agreed on.

src/styles/components/_create_cluster.sass
```diff
@@ -130,10 +130,10 @@
 .new-cluster--instance-type-selector-button
   cursor: pointer
   position: absolute
-  top: 0px
+  top: 1px
```
xh3b4sd:
```suggestion
  top: 0px
```

src/styles/components/_create_cluster.sass
```diff
@@ -130,10 +130,10 @@
...
   right: 0px
   background-color: #3a5f7b
-  height: 39px
+  height: 41px
```
xh3b4sd:
```suggestion
  height: 39px
```

src/components/shared/input_field.js
```diff
@@ -124,7 +124,7 @@ class InputField extends React.Component {
         {// If it is readOnly, don't show validation errors
 
         this.props.readOnly ? null : (
-          <span className='message'>{this.state.validationError}&nbsp;</span>
+          <span className='message'>{this.state.validationError}</span>
```
xh3b4sd:
```suggestion
          <span className='message'>{this.state.validationError}&nbsp;</span>
```

src/components/create_cluster/index.js
```diff
@@ -378,88 +261,47 @@ class CreateCluster extends React.Component {
...
-                        onWorkerUpdated={this.updateWorker.bind(this, index)}
+              <div className='col-9'>
+                <form
+                  onSubmit={e => {
+                    e.preventDefault();
+                  }}
+                >
+                  <div>
+                    <p>
+                      To have full control over cluster scaling, set both
```
oponder: This makes me think.

I don't understand really what will happen here.

In a previous mockup, i saw some radio buttons, that was pretty clear to me.

The actual behaviour now is not so explicit to me. Perhaps the label needs to be improved.

ping @marians, since I hear from @xh3b4sd that this was your labelling suggestion. 

How do these numbers interact with the 'scale cluster' button? Should that be disabled when these numbers are not equal to eachother?

oponder:
For the reviewers, this is a small part of a larger UI change.

Together with @xh3b4sd I limited the scope of this PR to just getting the scaling values through to the API correctly, and having the UI behaviour work correctly there.

Next step will be to add worker configuration (so instance type selection for aws, vmsize for azure, and cpu/ram/storage for kvm). 

Since that was more complicated than it needed to be, we decided to remove that, and add it back in a new PR. Instead of trying to refactor the current functionality and bring in the scaling buttons at the same time.

src/components/create_cluster/index.js
marians: The radio button design (modal switch between current manual scaling and new autoscaling) has been discarded, as it's actually simpler overall to have the autoscaler in place everywhere. Setting min equal to max can be seen as a trick to make the set a static cluster size.

My current status is that the scaling function as it is (modifying the nodes array length in the cluster details via PATCH) will no longer work from the release that introduces autoscaling. But team spirit has the latest news on this.

I am not aware that a final decision has been made regarding what this means for the scaling function in happa (and gsctl, either). Where and how should users access and manipulate the min/max limits for a cluster?

xh3b4sd: I think we could just extend the scaling modal with min and max and go from there. 

screenshot 2018-12-05 at 15 04 34

xh3b4sd: I like having a separate endpoint for scaling, just like status. The scaling subresource in k8s is also a separate one.

oponder: How about 'To disable autoscaling, set both numbers to the same value' as the description?

marians: Works for me

xh3b4sd: Added the suggestion.

PR #428, opened 2018-12-10
introduces aws instance type selection

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/pull/2206. Note that this PR is based on [`feature-auto-scaling`](https://github.com/giantswarm/happa/pull/424). This is pretty much WIP as of now. Oli wanted to take over to fix some React issue we couldn't figure out during pairing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,12 +47,21 @@ jobs:
         command: ./architect build
 
     - run:
-        name: Test the image we just built
+        name: Launch container for tests
         command: |
           docker run --name testcontainer -p 8000:8000 -d quay.io/$CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME:$CIRCLE_SHA1
           CURL_OUTPUT=$(curl -s -k https://localhost:8000)
           echo "${CURL_OUTPUT}" | grep Happa
-          docker kill testcontainer
+
+    - run:
+        name: Lighthouse audits
+        command: |
+          docker run --rm \
+            -v $PWD:/out \
+            -v $PWD/dev-shm:/dev/shm \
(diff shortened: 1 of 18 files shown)


marians:
Not that the base branch set for this PR (master) does not match what you wrote up there (feature-auto-scaling)

xh3b4sd:
When selecting the instance type there is an error I am not sure how to fix it. 

screenshot 2018-12-13 at 14 43 28

xh3b4sd:
There is also the problem of the "Create Cluster" button not being disabled in case the selection is not valid. @oponder had an idea there but I am not into that and am not sure how to solve this myself now.

oponder:
Should be fixed now if you want to take a look at the diff, we we're missing a correct assignment of the value / valid attributes on the 'selectInstanceType' method of the instance type picker. (that's the one that fires when someone selects an instance type from the modal).

```
   selectInstanceType = () => {
-    this.props.onChange(this.state.preSelectedInstanceTypeName);
+    this.props.onChange({
+      value: this.state.preSelectedInstanceTypeName,
+      valid: true
+    });
     this.closeModal();
```

So what was happening was the input got assigned to undefined (`this.aws.instanceType.value` was undefined after this), and that converts it from a 'controlled' to a 'uncontrolled' input. Which React does not like.

Controlled vs Uncontrolled refers to who is handling the state of that input field. When its controlled the `value` attribute of the input field is being set by a variable that we control in our state.

Uncontrolled is what a input normally behaves like, i.e. the browser takes care of its state. And we have to use a ref to get the current value of the input field.

See: https://reactjs.org/docs/uncontrolled-components.html
And: https://reactjs.org/docs/forms.html

PR #431, opened 2018-12-13
show current number of nodes

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/pull/2206. Note that this PR is based on [`feature-auto-scaling`](https://github.com/giantswarm/happa/pull/424). This is pretty much WIP as of now since I am rather lost in the react magic. Marian wanted to check it out by chance. Let's see.


marians:
@tuommaki Feel free to delete the taylorbot comments if you find them disturbing. These have been introduced by a PR I merged today.

tuommaki:
CI Green. Now this should show the current desired & present number of nodes correctly.

@giantswarm/sig-ux PTAL :)

src/reducers/appReducer.js
```diff
@@ -49,6 +49,9 @@ export default function appReducer(
         provider: '',
       },
     },
+    status: {
```
tuommaki: Hmm... This is probably not needed? Or should I do something properly here and change this?

src/reducers/clusterReducer.js
```diff
@@ -179,6 +179,24 @@ export default function clusterReducer(
...
+
+      return {
+        lastUpdated: state.lastUpdated,
+        isFetching: false,
+        items: items,
+      };
+
+    case types.CLUSTER_LOAD_STATUS_ERROR:
+      return {
+        numberOfNodes: 0,
```
tuommaki: Should this be something different? I don't think my code per se uses this, so what would be correct here?

pipo02mix:
```suggestion

        lastUpdated: state.lastUpdated,
        isFetching: false,
        items: items,
```

Indeed I would return the state without changes. Same as we do with other errors

src/reducers/appReducer.js
pipo02mix: I dont think it is needed. AFAIK `status` belongs to cluster state

xh3b4sd:
Do you want me to test that locally? CI is red.

src/reducers/clusterReducer.js
tuommaki: Roger. I checked example from load cluster details error and it's same here now.

src/reducers/appReducer.js
tuommaki: ACK. Removed.

oponder:
BTW I made a PR to add the DCO and License here: https://github.com/giantswarm/happa/pull/445

PR #432, opened 2018-12-14
Update min max scaler for new number picker

oponder:
I made some changes to the number picker, this updates the feature-auto-scaling branch so it works as expected.

### DIFF
--- a/src/components/create_cluster/index.js
+++ b/src/components/create_cluster/index.js
@@ -23,18 +23,20 @@ class CreateCluster extends React.Component {
     clusterName: 'My cluster',
     scaling: {
       min: 3,
+      minValid: true,
       max: 3,
+      maxValid: true,
     },
     submitting: false,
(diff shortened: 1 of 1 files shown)


src/components/create_cluster/index.js
```diff
@@ -23,27 +23,31 @@ class CreateCluster extends React.Component {
...
       },
     });
   };
 
   updateScalingMin = n => {
     this.setState({
       scaling: {
-        min: n,
+        min: n.value,
+        minValid: n.valid,
```
pipo02mix: Out of curiosity, why we use here `n.valid` instead of `this.state.scaling.minValid` ?

oponder: This is the callback that gets called from the minimum scaling number picker when a value changes.

it returns the value, and whether or not that value is valid. So we want to update the state, instead of re-use what was already in the state.

pipo02mix: ah ok, so actually the number picker is made of two inputs.

PR #433, opened 2018-12-17
Add azure vm size selector

oponder:
Brings back the azure vm size selector, now that the create cluster form has changed somewhat.


oponder:
FYI, this is still part of the larger `feature-auto-scaling` branch and will be merged into that branch.

PR #434, opened 2018-12-17
Add kvm worker config

oponder:
Last step in the story, brings back a way for the worker to be configured on kvm installations.


src/components/create_cluster/index.js
```diff
@@ -428,14 +575,22 @@ CreateCluster.propTypes = {
...
   var minAvailabilityZones = state.app.info.general.availability_zones.default;
   var maxAvailabilityZones = state.app.info.general.availability_zones.max;
   var selectedOrganization = state.app.selectedOrganization;
   var provider = state.app.info.general.provider;
-  var defaultInstanceType = 'm3.large';
+  var defaultInstanceType = 'm3.large'; // TODO
+  var defaultVMSize = 'Standard_A2_v2'; // TODO
+  var defaultCPUCores = 1; // TODO
+  var defaultMemorySize = 1; // TODO
+  var defaultDiskSize = 1; // TODO
```
marians: So what about these TODOs?

oponder: I'll wire them up in another PR as far as is possible.

PR #435, opened 2018-12-17
Set default values

oponder:
Set default values from the info endpoint.

For KVM,  we don't have default values for cpu, memory and storage from the api info endpoint, so setting them to what they always were in Happa.

### DIFF
--- a/src/components/create_cluster/index.js
+++ b/src/components/create_cluster/index.js
@@ -441,7 +441,7 @@ class CreateCluster extends React.Component {
                         <br />
 
                         <NumberPicker
-                          label='Memory'
+                          label='Memory (GB)'
                           unit='GB'
                           stepSize={1}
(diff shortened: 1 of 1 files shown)


src/components/create_cluster/index.js
```diff
@@ -586,11 +586,30 @@ function mapStateToProps(state) {
...
+  ) {
+    defaultVMSize = state.app.info.workers.vm_size.default;
+  } else {
+    defaultVMSize = 'Standard_D2s_v3';
+  }
+
   var defaultCPUCores = 1; // TODO
   var defaultMemorySize = 1; // TODO
-  var defaultDiskSize = 1; // TODO
+  var defaultDiskSize = 20; // TODO
```
pipo02mix: why `20`?

oponder: I think that was the default it was originally.

PR #436, opened 2018-12-18
Add NodeCountSelector component for cluster creation

tuommaki:
NodeCountSelector provides number picker[s] and related labels for node
count selection and appropriate logic to provide either single node
count or min/max values for scaling limits.

This is also now wired for cluster creation view and there is a function
that decides based on provider and release version whether cluster
supports automatic scaling limits or just a static node count.

**NOTE:** This is just a change for cluster creation. I still need to do changes for existing cluster details page & scaling view.

**NOTE2:** Please, have a critical eagle eye on reasonability / code style of this change. :smile: 

Towards https://github.com/giantswarm/giantswarm/pull/2206

### DIFF
--- a/src/components/create_cluster/index.js
+++ b/src/components/create_cluster/index.js
@@ -176,7 +176,7 @@ class CreateCluster extends React.Component {
 
     // In order to have support for automatic scaling and therefore for scaling
     // limits, provider must be AWS and cluster release >= 6.1.0.
-    return (cmp(releaseVer, '6.1.0') === 1);
+    return cmp(releaseVer, '6.1.0') === 1;
   }
 
   selectRelease = releaseVersion => {
@@ -486,7 +486,10 @@ class CreateCluster extends React.Component {
 
             <div className='row section'>
               <NodeCountSelector
-                autoscalingEnabled={this.isScalingAutomatic(this.props.provider, this.state.releaseVersion)}
+                autoscalingEnabled={this.isScalingAutomatic(
+                  this.props.provider,
+                  this.state.releaseVersion
+                )}
                 scaling={this.state.scaling}
                 readOnly={false}
                 onChange={this.UpdateScaling}
--- a/src/components/create_cluster/index.js
+++ b/src/components/create_cluster/index.js
@@ -174,6 +174,8 @@ class CreateCluster extends React.Component {
       return false;
     }
 
+    // In order to have support for automatic scaling and therefore for scaling
+    // limits, provider must be AWS and cluster release >= 6.1.0.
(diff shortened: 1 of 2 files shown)


src/components/create_cluster/index.js
```diff
@@ -178,6 +169,14 @@ class CreateCluster extends React.Component {
     this.cluster_name.select();
   }
 
+  isScalingAutomatic(provider, releaseVer) {
+    if (provider != 'aws') {
+      return false;
+    }
+
+    return (cmp(releaseVer, '6.1.0') === 1);
```
marians: Since the `cmp` function is so un-intuitive, I'd like a comment line above to explain what this means (releaseVer >= 6.1.0 I assume)

tuommaki: Added.

src/components/create_cluster/node_count_selector.js
```diff
@@ -0,0 +1,165 @@
...
+        max: numberPicker.value,
+        maxValid: numberPicker.valid,
+      },
+    });
+  };
+
+  updateNodeCount = numberPicker => {
+    this.setState({
+      scaling: {
+        automatic: false,
```
pipo02mix: What about aligning this? I know `updateNodeCount` it is only used  when it is false but it could lead to mistakes in the future
```suggestion
        automatic: this.state.scaling.automatic,
```

tuommaki: Oh, that's leftover. It should be actually removed. Thanks for pointing it out!

src/components/create_cluster/node_count_selector.js
```diff
@@ -0,0 +1,162 @@
...
+              onSubmit={e => {
+                e.preventDefault();
+              }}
+            >
+              <div>
+                <p>
+                  To disable autoscaling, set both numbers to the same value
+                </p>
+                <div className='col-3'>
+                  <p>Minimum</p>
```
teemow: Why isn't this a <label>? Same for `Maximum` below.

tuommaki:
Regarding that usability improvement:

I do agree that this would be nice to have, but due to the way how min/max limit validation is currently implemented and how `NumberPicker` is embedded in `NodeCountSelector` - I'd either need help to design a correct structure so that this gets properly done or alternatively suggest to create a TODO-issue for this and add this later. I tried to implement this but I just broke the validation and can't figure a good way to do it.

@marians want to pair on this or should we postpone this improvement to be added/implemented later?

marians:
I don't want to block the PR. Created https://github.com/giantswarm/giantswarm/issues/4944 to keep this on the radar.

PR #442, opened 2019-01-09
Remove VERSION file

marians:
Fixes https://github.com/giantswarm/giantswarm/issues/4897

The VERSION file was once maintained by `builder`. It is no longer updated, and it is also not exposed anywhere really. Even the nginx headers didn't come up in my tests. (We would probably have to enable them via ingress configuration)

This PR should remove it all.


oponder:
screenshot 2019-01-09 at 15 32 49

Not sure about these delta icons btw. Shouldn't we prefer some smaller numbers here? or is it all on some index of 1 to 100.

Also looks like its not fully deterministic, because I don't see how removing a version file would impact these metrics. Still worth having it though.

marians:
@oponder All scores fall between 0 and 100. The results are indeed questionable when this change (which shouldn't affect performance) creates such big differences. Especially since both tests are executed on the same (virtual) machine.

PR #446, opened 2019-01-15
Explain a helpful pre commit hook to avoid prettier noise in CI.

oponder:
Since we have a CI step that checks prettier, I have often pushed code that is bound to fail it.

Instead of going through that dance, I've added a pre-commit hook that will stop me from committing code that is bound to fail this step in the CI.

It might be helpful for others, so I am documenting it.


oponder:
Ideally btw you set up your editor to run prettier on the file you just saved.

Once I figure that out for sublime, I'll document that too. But not everyone uses sublime. So I thought I'd start with this.

PR #452, opened 2019-01-18
First step towards managed services in Happa.

oponder:
This gives us some markup and a screen to actually talk about.

It's based on Marian's work previously.

Questions that come to mind are:

Do we have these apps already represented as ~CRD's~ CR's so that I can try and wire this up instead of having it as static info?

If not, is this list complete and correct? Does it apply to all installations?


oponder:
screenshot

puja108:
Currently we do not have anything for them in the CP and for some of them we might never have, or at least need to think of a trick there at least for those that are not managed by chart-operator.

As for the list (AWS 6.2.0 release):

Essentials:
- Kubernetes Components
- RBAC and PSP defaults
- Calico
- CoreDNS
- metrics-server
- cluster-autoscaler

Management:
- chart-operator
- Tiller
- kube-state-metrics
- node-exporter
- net-exporter
- cert-exporter

Ingress:
- NGINX IC

See

```
giantswarm    chart-operator-74b4dd4786-r6rjz                                     1/1     Running   0          24h
giantswarm    tiller-deploy-7c65f8d958-rm759                                      1/1     Running   0          24h
kube-system   calico-kube-controllers-59cb45589f-p2v4c                            1/1     Running   0          24h
kube-system   calico-node-7cvbr                                                   2/2     Running   0          24h
kube-system   calico-node-nhzl5                                                   2/2     Running   0          24h
kube-system   calico-node-wt4d6                                                   2/2     Running   0          24h
kube-system   calico-node-znxxc                                                   2/2     Running   0          24h
kube-system   cert-exporter-5pvg2                                                 1/1     Running   0          24h
kube-system   cert-exporter-77sx6                                                 1/1     Running   0          24h
kube-system   cert-exporter-jhh8t                                                 1/1     Running   0          24h
kube-system   cert-exporter-z2gh6                                                 1/1     Running   0          24h
kube-system   cluster-autoscaler-59ccb597c6-ffbnc                                 1/1     Running   0          24h
kube-system   coredns-7996df49d7-kg25g                                            1/1     Running   0          24h
kube-system   coredns-7996df49d7-msr6n                                            1/1     Running   0          24h
kube-system   default-http-backend-86bc996747-qzd9s                               1/1     Running   0          24h
kube-system   default-http-backend-86bc996747-shvqb                               1/1     Running   0          24h
kube-system   k8s-api-server-ip-10-1-1-47.eu-central-1.compute.internal           1/1     Running   0          24h
kube-system   k8s-controller-manager-ip-10-1-1-47.eu-central-1.compute.internal   1/1     Running   0          24h
kube-system   k8s-scheduler-ip-10-1-1-47.eu-central-1.compute.internal            1/1     Running   0          24h
kube-system   kube-proxy-2gqsz                                                    1/1     Running   0          24h
kube-system   kube-proxy-2rtvs                                                    1/1     Running   0          24h
kube-system   kube-proxy-4f4kv                                                    1/1     Running   0          24h
kube-system   kube-proxy-mxhrg                                                    1/1     Running   0          24h
kube-system   kube-state-metrics-798b58fcb9-2cwrf                                 1/1     Running   0          24h
kube-system   metrics-server-5c9b5587f5-r9kwk                                     1/1     Running   0          24h
kube-system   net-exporter-4b5mz                                                  1/1     Running   0          24h
kube-system   net-exporter-dcq97                                                  1/1     Running   0          24h
kube-system   net-exporter-f7srq                                                  1/1     Running   0          24h
kube-system   net-exporter-tffdj                                                  1/1     Running   0          24h
kube-system   nginx-ingress-controller-5cdc998974-jt8nx                           1/1     Running   0          24h
kube-system   node-exporter-5hj59                                                 1/1     Running   0          24h
kube-system   node-exporter-gmfd2                                                 1/1     Running   0          24h
kube-system   node-exporter-gnpmk                                                 1/1     Running   0          24h
kube-system   node-exporter-tkskb
```

Additionally, there's a version for each of those, and maybe sub-components once you click into them? but maybe we start with the list and then go deeper

teemow:
The release CRs contain the list of operators. The operators expose their components through the version bundle endpoint. I guess there is no way around of having two distinct sources for the components.

Maybe it even makes sense to distinguish between components in the control plane, core components managed in a tenant cluster and additional components.

Control plane operator versions (Release and App CR):
 * aws-operator 1.2.3
 * cluster-operator 2.3.4
 * cert-operator 0.5.0

Core components (no CR yet):
 * Containerlinux 1234.4
 * Docker 19.01.1
 * Etcd 3.4.0

Early morning braindump: The information about the core components is also known at build time. If we want to keep the source of truth in the operators we could move the core components configuration into some yaml. Preferably something that gets automatically installed with the operator version. Maybe a CR in the operator helm chart, that contains something like: (Core) Component Name, Version, Managed By.

pipo02mix:
> Maybe a CR in the operator helm chart, that contains something like: (Core) Component Name, Version, Managed By.

why not just a simple config map with a yaml inside?

rossf7:
Moving core components to YAML makes sense to me. Currently the release CRD has the authorities which map to operators.

I think we will want to add apps to the release CRD. e.g. kube-state-metrics. As for me a release also has mandatory apps we need to install. For the future we will also have optional apps e.g. Nginx Ingress Controller.

Could these core components and versions also be in the release CRD? So it becomes the single source of truth.

oponder:
Im a bit confused to be honest.. this gets complicated for my brain.

Timo said:
> Core components (no CR yet):
> Containerlinux 1234.4
> Docker 19.01.1
> Etcd 3.4.0

We do have information about these core components in the Release though right? That's where I'll be getting version numbers for now in Happa.

See `gsctl list releases` containts etcd for example:

```
Version: 5.2.2
Created: 2018 Dec 03, 22:56 UTC
Active: true
Components:
  aws-operator: 4.2.1
  calico: 3.2.3
  cert-operator: 0.1.0
  chart-operator: 0.4.0
  cluster-operator: 0.7.1
  containerlinux: 1745.4.0
  coredns: 1.1.1
  docker: 18.03.1
  kubernetes: 1.11.5
  etcd: 3.3.8
  kube-state-metrics: 1.3.1
  nginx-ingress-controller: 0.15.0
  node-exporter: 0.15.1
  vault: 0.10.3
Changelog:
  chart-operator: Add your changes here.
  cluster-operator: Fixed attempt to create already existing chartconfig and configmap.
  kubernetes: Enabled admission plugins: DefaultTolerationSeconds, MutatingAdmissionWebhook, ValidatingAdmissionWebhook.
  kubernetes: Update to 1.11.5 (CVE-2018-1002105).
  vault: Vault version updated.
```

teemow:
Since the configmap vs CR discussion keeps coming up. 

https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#should-i-use-a-configmap-or-a-custom-resource

oponder:
Alright, code has been updated, these are the apps we're showing now:

```
Essentials:
  rbac and psp defaults
  calico
  containerlinux
  coredns
  docker
  kubernetes
  etcd
  
Management:
  kube-state-metrics
  node-exporter
  chart-operator
  cert-exporter
  net-exporter

Ingress:
  nginx-ingress-controller
```

The version data comes from the release endpoint, but we have control over what components we show and where. So source of truth isn't happa, but the extra metadata is still in Happa.

oponder:
screenshot 2019-01-28 at 15 46 39

pipo02mix:
Nitpick: IMO `RBAC and PSP` would be in mgmt column as it is not an essential part and we/customers use it to manage the authZ.

I see exporters run in the latest version, any ideas how to get the version on the screen?

oponder:
@pipo02mix 
> I see exporters run in the latest version, any ideas how to get the version on the screen?

They have to somehow show up in the releases endpoint. I'm not sure what microservice would best take those on though. Right now they are added in after the fact by Happa and hardcoded. If we could make a release that includes those components then the version can show up as well.

Some goes for RBAC & PSP defaults. If we can add that as a 'component' so that it shows up in the releases endpoint, then I can make it show up in Happa.

I'd actually also like to make a proposal to add the metadata I am currently hardcoding the the releasebundle data structure somehow. So components should gain a `category` and a `logoUrl`. Then I can remove some of this weird hardcoded metadata in Happa.

Issue #316, opened 2018-03-27
cn_prefix validation when creating a key pair in happa/gsctl


oponder:
The cn prefix can't contain special characters. The UI should warn about attempt to do so, and the error message should be clearer.

## TODO

- [x] Improve error message in happa that shows during entry to not fire too early during entry

webwurst:
After changing the [allowed characters](https://github.com/giantswarm/happa/pull/387/files#diff-4e02126a1acfe6068199a589817e540dR57) the error message in the UI is derailed: When entering an unallowed char, it still shows `Please use characters a-z, 0-9 or -`.

This also showes up when you enter `.`, `-`, `@` at the end, which just happens while you type in a name that contains these in the middle.

This does not influence the function of the input fields, just might be irritating for the user.

marians:
Yeah, that happened when I tried to quickly fix something. :/

webwurst:
Yeah, and I just wanted to point this out since I noticed it. Not super critical to me in contrast to the quick fix. That was needed :)

marians:
No problem. Please keep pointing out annoying things! We have to fix this. My current idea is using a delay, so that the warning only appears when the user stops typing for like a second. The next best thing would be a specific notice like "The CN prefix must and with a-z, A-Z, 0-9".

pipo02mix:
done via https://github.com/giantswarm/happa/pull/453

PR #453, opened 2019-01-22
Add debounce function for CN prefix validation

pipo02mix:
Towards https://github.com/giantswarm/happa/issues/316

ezgif com-crop 1


marians:
Deploying to godsmack for testing

PR #456, opened 2019-01-24
Gracefully handle missing cluster status

tuommaki:
When creating a new cluster, there's cluster status missing for few minutes
before the operators get to the point that there's something to report. When
API returns 404 for cluster status, handle it gracefully.


MarcelMue:
Just an idea: could you check for the cluster version instead and not even make the request if the version is below the one which has `status`? Also which version did we introduce `status`?

tuommaki:
Cluster status exists for all of our clusters. It's provider CRs status field, but now that you mentioned, I just realized that some fields in it are not present in all providers and need some adjustments. Working on this next...

MarcelMue:
I decided to check for `6.3.0` in my code as I am only interested in the desired number of nodes from the status while scaling. Not sure what else could be needed / makes sense for the details view and so on.

tuommaki:
Excluded desired number of workers on non-AWS providers.

PR #457, opened 2019-01-24
Fix number of nodes in home view

tuommaki:
Worker array doesn't contain entry for each node any more. Instead one should
use node entries found from cluster status.

### DIFF
--- a/src/components/home/cluster_dashboard_item.js
+++ b/src/components/home/cluster_dashboard_item.js
@@ -48,6 +48,36 @@ class ClusterDashboardItem extends React.Component {
     return c;
   }
 
+  getNumberOfNodes() {
+    if (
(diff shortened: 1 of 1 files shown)


src/components/home/cluster_dashboard_item.js
```diff
@@ -48,6 +48,36 @@ class ClusterDashboardItem extends React.Component {
...
+    ) {
+      var nodes = this.props.cluster.status.cluster.nodes;
+      if (nodes.length == 0) {
+        return 0;
+      }
+
+      var workers = 0;
+      nodes.forEach(node => {
+        if (Object.keys(node).includes('labels')) {
+          if (node.labels['role'] === 'worker') {
```
xh3b4sd: We do not have the `role=worker` labels AFAIK. How does that work?

PR #459, opened 2019-01-25
Refresh cluster details after successful scaling

tuommaki:
Once cluster scaling succeeds, cluster details should be refreshed to
correspond done changes. Load cluster state fom backend to get possible other
changes to cluster state (e.g. changes in desired number of workers).

### DIFF
--- a/src/components/cluster_detail/scale_cluster_modal.js
+++ b/src/components/cluster_detail/scale_cluster_modal.js
@@ -112,7 +112,7 @@ class ScaleClusterModal extends React.Component {
           .then(patchedCluster => {
             this.close();
 
-            this.props.clusterActions.clusterLoadDetailsSuccess(patchedCluster);
+            this.props.clusterActions.clusterLoadDetails(patchedCluster.id);
             this.props.flashActions.flashAdd({
               message: 'Successfully scaled cluster',
               class: 'success',


src/components/cluster_detail/scale_cluster_modal.js
```diff
@@ -112,7 +112,7 @@ class ScaleClusterModal extends React.Component {
           .then(patchedCluster => {
             this.close();
 
-            this.props.clusterActions.clusterLoadDetailsSuccess(patchedCluster);
+            this.props.clusterActions.clusterLoadDetails(patchedCluster.id);
```
xh3b4sd: Why was the success thingy done before? Why doesn't it matter now? It makes sense to me to load the details again though in order to refresh the state.

oponder: The way this worked was that the call to patch a cluster actually returns the full representation of the updated cluster. So you can call `clusterLoadDetailsSuccess` and pass in the patchedCluster, which will update happa's state with that cluster.

However now, I assume @tuommaki wants to trigger a full refresh also of the cluster status, which comes from a different endpoint.

xh3b4sd: I think the call before would be more performant. Loading the status directly after initiating the scaling does not do anything.

tuommaki: Ah, right. Now that you explained this it makes sense. Yeah, I changed this because the previous implementation didn't refresh the scaling values and my change fixed that problem, but when it was because of `scaling` missing from response then it explains this. Therefore once the API response is changed, this can be discarded.

oponder: btw if this scaling info isn't coming back in Happa, but it's been added already in the API, then it its probably because we have to regenerate the js client, but based on the not yet merged branch: https://github.com/giantswarm/api-spec/pull/121

tuommaki:
This should have been fixed with added scaling values in api response. Dropping this.

PR #461, opened 2019-01-29
Refactor: Extract Navigation into its own component.

oponder:
Doing some refactorings in Happa.


src/components/layout.js
```diff
@@ -102,164 +99,16 @@ class Layout extends React.Component {
     } else {
       return (
         <DocumentTitle title='Giant Swarm'>
-          <div>
+          <React.Fragment>
```
oponder: This is new in React 16 and we can start applying it where possible.

`React.Fragment` lets us group components without adding extra unnecessary `<div>`s to the DOM.

oponder:
> Is there actually a way of writing useful unit tests for the components?

Yes! I think for some components that more complex it can be useful. Like the number pickers and such. Mostly the point is to ensure that a given state results in the component looking as expected.

xh3b4sd:
Then let's start doing that where it makes sense so we get an idea of how to do that and what is actually useful. I have no idea how to do that. So this would be cool.

PR #464, opened 2019-01-30
Don't show cluster apps when we don't have any release information.

oponder:
ClusterApps fails if for some reason we do not have release information for the cluster.

This is a rare edge case, but worth to guard against.

### DIFF
--- a/src/components/cluster_detail/index.js
+++ b/src/components/cluster_detail/index.js
@@ -230,7 +230,11 @@ class ClusterDetail extends React.Component {
                         credentials={this.props.credentials}
                         release={this.props.release}
                       />
-                      <ClusterApps release={this.props.release} />
+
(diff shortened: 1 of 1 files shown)


src/components/cluster_detail/index.js
```diff
@@ -230,7 +230,13 @@ class ClusterDetail extends React.Component {
                         credentials={this.props.credentials}
                         release={this.props.release}
                       />
-                      <ClusterApps release={this.props.release} />
+
+                      {this.props.release ? (
+                        <ClusterApps release={this.props.release} />
+                      ) : (
+                        undefined
+                      )}
```
pipo02mix: hmm what about
```
{this.props.release &&
    <ClusterApps release={this.props.release} />
}
```
to avoid the `undefined`?

PR #465, opened 2019-01-30
Replace underscore's map with native javascript map.

oponder:
In Happa we occasionally used underscore to map over an array and other times just normal native javascript.

This is an effort to standardize and reduce dependencies at the same time.


marians:
+1 - the PR description doesn't seem to match what's happening here

oponder:
Sorry, I branched wrong. But fixed it.

The diff should now relate to the title.

PR #470, opened 2019-02-01
Calculate total CPUs/RAM/Storage by running nodes

tuommaki:
When number of nodes changes, calculate total amount of CPU cores, RAM and
storage space by number of running nodes, not by scaling limits.


src/components/cluster_detail/cluster_detail_table.js
```diff
@@ -23,17 +23,6 @@ class ClusterDetailTable extends React.Component {
     return null;
   }
 
-  getMemoryTotal() {
```
tuommaki: Moved this below to be closer with `getCpusTotal()` and `getStorageTotal()`.

marians:
Late :)

PR #471, opened 2019-02-01
Add metrics server to managed apps

pipo02mix:
Towards https://github.com/giantswarm/giantswarm/issues/5250


src/components/cluster_detail/cluster_apps.js
```diff
@@ -14,46 +14,41 @@ import PropTypes from 'prop-types';
...
       name: 'etcd',
       logoUrl: '/images/app_icons/etcd@2x.png',
       category: 'essentials',
     },
+    'kubernetes': {
+      name: 'kubernetes',
+      logoUrl: '/images/app_icons/kubernetes@2x.png',
+      category: 'essentials',
+    },
     'kube-state-metrics': {
```
pipo02mix: Sorting by category and then alphabetically

oponder:
Hey fer, you gotta run prettier to make the CI pass. `make run-prettier` should do the changes you need.

pipo02mix:
Tested on ginger 

screenshot 2019-02-01 at 13 17 35

src/components/cluster_detail/cluster_apps.js
```diff
@@ -64,11 +59,21 @@ class ClusterApps extends React.Component {
       logoUrl: '/images/app_icons/chart_operator@2x.png',
       category: 'management',
     },
+    'metrics-server': {
+      name: 'metrics-server',
+      logoUrl: '/images/app_icons/metrics_server@2x.png',
+      category: 'management',
```
puja108: IMO metrics-server is essential (e.g. for HPA) and we do not use it for our mgmt

sidenote, it seems we are not clear abotu essential vs mgmt, for me latter should be only about the things that are for our mgmt and not for the user, for me RBAC is essential, without it there's no security, we cannot remove that without influencing the customer, if we add or remove anything from mgmt that would not influence the customer directly.

pipo02mix: agreed and committed

PR #472, opened 2019-02-02
Bump kubectl download links to 1.11.7 (from 1.7.5)

foben:
The kubectl version referenced  in current Happa download links is outdated.
This PR updates links to 1.11. version, which matches the latest Kubernetes version available through happa.


oponder:
@foben thanks so much for the contribution indeed! 

FYI to the giantswarm team ping (@giantswarm/sig-ux), our CI won't run this branch because we've disabled that for forks, but the contribution checks out to me. 

I'd like to go ahead and merge it unless there's any veto.

PR #477, opened 2019-02-07
Add cluster autoscaler and icon.

oponder:
fyi @mazzy89 @teemow @puja108


teemow:
Do we have any other idea for an icon? I am ok with the up/down arrows but don't find this explaining auto scaling very well. 

This might be a better icon: https://www.fortinet.com/content/dam/fortinet/images/icons/benefits-extra/icon-benefits-scalable.svg

marians:
Screenshot would be appreciated :)

oponder:
Old icon:
screenshot 2019-02-07 at 23 34 25

New icon after feedback from timo:

screenshot 2019-02-07 at 23 34 55

mazzy89:
<3 I like the second icon. The first one if I don't make a mistake is the one used for inbound/outbound network connections.

PR #481, opened 2019-02-12
Update dependencies

oponder:
Updates various dependencies, including react and redux, to their latest versions.

Here is what's left to upgrade:

```
Package          Current Wanted Latest       Package Type URL
bootstrap        3.4.0   3.4.0  4.3.0        dependencies https://getbootstrap.com/
giantswarm-v4    4.0.0   exotic exotic       dependencies git+https://github.com/giantswarm/giantswarm-js-client.git#master
immutable        3.8.2   3.8.2  4.0.0-rc.12  dependencies https://facebook.github.com/immutable-js
react-bootstrap  0.32.4  0.32.4 1.0.0-beta.5 dependencies https://react-bootstrap.github.io/
react-datepicker 1.6.0   1.6.0  2.1.0        dependencies https://github.com/Hacker0x01/react-datepicker
```

I think I can get rid of the immutable.js dependency. The datepicker requires a bit of work to change the type of date it uses from moment.js dates to native dates.


src/reducers/clusterReducer.js
```diff
@@ -179,78 +134,6 @@ export default function clusterReducer(
         items: items,
       };
 
-    case types.CLUSTER_LOAD_METRICS:
```
oponder: This and the other removals here was dead code, the newer version of babel detected this and raised errors about it.

oponder:
I'm trying to understand why the lighthouse audit is mostly zero's suddenly. Ping @marians any idea?

oponder:
Looking into `branch.json` I'm seeing now: `Something went wrong with recording the trace over your page load. Please run Lighthouse again. (NO_FCP)`

oponder:
Ok it was a problem with the redux devtools conditional setup. Everything worked fine for me since I have redux devtools installed, but for people that don't (including the audit), the site simply wouldn't load.

PR #488, opened 2019-02-20
Use sorting of allowed instances array

marians:
For https://github.com/giantswarm/giantswarm/issues/5305

This makes use of the sorting of the allowed instance types/VM sizes array to sort the selection table.


marians:
This can currently be tested on gauss and godsmack

oponder:
Nice work!

PR #493, opened 2019-02-21
Add icons to buttons on cluster details page

marians:
### Preview

image

and as it's not visible in the screenshot:

image


marians:
Agree. To make those buttons work with icons, we'd have to make them larger in general. Also the Scaling icon could be tweaked a bit. I'll look at it again tomorrow.

marians:
Updated the icons

External monitor

image

Retina display

image

PR #494, opened 2019-02-21
Add a footer with version number. Adjust release process.

oponder:
This adds a little footer to Happa with links to the github releases page.

screenshot 2019-02-22 at 01 36 36

As for the actual release process, I'm thinking something like:

1. Bump the version in package.json
1. git commit -m "Release ${VERSION}"
1. git tag -a ${VERSION}
1. git push
1. CI takes over and makes the release draft

And we change our CI to only continuously deploy tags. Master can still be continuously deployed to our own Giant Swarm installations though.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,7 +19,7 @@ workflows:
           filters:
             branches:
               only:
-                - versioning
+                - master
 
 jobs:
   build:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,26 @@
 version: 2
+
+workflows:
+  version: 2
+  build-and-deploy:
+    jobs:
+      - build
+      - deploy-to-all:
+          requires:
+            - build
+          filters:
+            tags:
+              only: /^[0-9]+\.[0-9]+\.[0-9]+$/
+            branches:
+              ignore: /.*/
+      - deploy-to-testing:
+          requires:
+            - build
+          filters:
+            branches:
+              only:
+                - versioning
+
 jobs:
   build:
     machine: true
@@ -34,7 +56,7 @@ jobs:
           fi
 
     - run:
-        name: Valdiate code style using prettier
+        name: Validate code style using prettier
         command: |
           make validate-prettier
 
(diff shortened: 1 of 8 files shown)


oponder:
If we're ok with the release process, I'll try to make the necessary changes to circle config.yml, and document it in a RELEASE.md


And then I'll try to use the release process to draft a `0.1.31` release, and add some text like:

-----------

## Happa v0.1.31 - Start versioning properly

Exciting times! New features are coming to Happa and we want a way to keep you informed. Up until now we haven't really officially versioned Happa or kept a change log. That's changing now. We've added a footer to happa that shows the current version as well as a link to this very page you are on now. 

### Changes
 - Added a footer with version number, a link to release notes, and a place to give feedback

-----------

.circleci/config.yml
```diff
@@ -55,6 +55,7 @@ jobs:
           fi
 
           docker run --name happa-branch -p 8000:8000 -d quay.io/$CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME:$CIRCLE_SHA1
+          sleep 2
```
oponder: The new startup script is "slow" enough (blaming `jq` here I think) that the curl command now failes with error code 35 (SSL connection error). 

2 seconds is enough for the sed substitutions happen and for the nginx server to definitely be up.

Dockerfile
```diff
@@ -1,6 +1,9 @@
 FROM nginx:1.14-alpine
 
+RUN apk add jq
```
marians:
```suggestion
RUN apk --no-cache add jq
```

For a smaller resulting image

marians:
Some detail notes and a bigger one:

- Let's use the middle dot ` · ` as we do between the cluster details instead of the dash
- I would remove the `View ` from `View Release Notes`

I am not sure a "Give Feedback" link pointing to GitHub issues is an improvement over how things work currently and will lead to more or better feedback. We currently get feedback every now and then via Slack, and to me this is absolutely fine. If I am not mistaken, all happa users are also in our Slack channels. So what I'd really like is encouraging more feedback via Slack. How could we reach that?

oponder:
> I am not sure a "Give Feedback" link pointing to GitHub issues is an improvement over how things work currently

True, it could open up a modal that then explains how to give feedback in slack? For now I'll remove it, since it is out of scope of what I intended to do with this PR.

I do somehow want to encourage more feedback / error reporting / feature requesting though.

oponder:
@marians  LGTY?

marians:
Content-wise it's fine now.

image

In small viewports I think there should be more space to the bottom viewport edge. Also ideally some more space to the upper content, in all viewport sizes.

.circleci/config.yml
```diff
@@ -22,7 +45,7 @@ jobs:
     - run:
         name: Install CI tools
         command: |
-          wget -q $(curl -sS -H "Authorization: token $RELEASE_TOKEN" https://api.github.com/repos/giantswarm/architect/releases/latest | grep browser_download_url | head -n 1 | cut -d '"' -f 4)
+          wget -q https://www.dropbox.com/s/b52zsxgfbr1qhc9/architect?dl=1
```
oponder: temporary way for me to test the new architect binary.
TODO: change it back to the normal way of downloading the latest release once the --group flag is tested.

marians:
The additional space between footer and viewport bottom should BTW be in all sizes, not only smaller viewports.

oponder:
@marians ok this should be ready now. The footer style has been updated as well, and the release process is just a matter of tagging happa. That can be done at the command line or by making a release in the github interface.

pipo02mix:
What if `deploy-to-all` it is the next step of the workflow (holding it)? so instead of add a tag, you just approve it and it spread to all clusters

oponder:
> What if `deploy-to-all` it is the next step of the workflow (holding it)? so instead of add a tag, you just approve it and it spread to all clusters

Could be, but what I wanted was to have proper semver versioned releases. Not that there really is a 'breaking change' for happa users. But from a professionalism point of view, I think that looks nice.

The hold step means we'd have to have a commit that changes a VERSION file. But with the tagging it all happens automatically, and the source of truth for the version is the tag.

.circleci/config.yml
```diff
@@ -1,4 +1,26 @@
...
+      - build
+      - deploy-to-all:
+          requires:
+            - build
+          filters:
+            tags:
+              only: /^[0-9]+\.[0-9]+\.[0-9]+$/
+            branches:
+              ignore: /.*/
+      - deploy-to-testing:
```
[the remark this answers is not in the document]
oponder: first test will be happening very soon. I'll announce the changes then as well

PR #505, opened 2019-02-25
Link commit hash to GitHub commit URL

marians:
Follow-up to #494


marians:
I actually got it working after roughly two hours of back and forth.

Learning: don't use a slash `/` as a seperator for `sed` when your replacement string contains slashes.

PR #508, opened 2019-02-27
Use noty for notifications

marians:
For https://github.com/giantswarm/giantswarm/issues/4533

- [x] Import Noty once on a global level
- [x] default position: topRight
- [x] animation to remove should be faster
- [x] Figure out how to replace `flashClearAll` (clear message queue)
- [x] Remove demo widget from home screen
- [x] Find out how much noty adds to the JS and CSS payload
- [ ] ~~Avoid certain messages stacking up, like `Unauthorized`, by giving queue name and max length = 1~~

For feedback on deletion, the `info` notification type is used, as deleting something should be acknowledged in a neutral way instead of be celebrated as a success.

Some messages are now shorter than they were.

Also removed the `Welcome` message on a successful login, as it felt as a distraction, with the entire screen changing and the user having to parse a lot of information potentially.


marians:
Tried out some of the Noty options.

image

The above shows the `bootstrap-v3` theme, adapted to use Roboto.

As for positioning, I tend to prefer `topLeft` over `topRight`.

image

marians:
State:

Most uses of the transitional `flashAdd` function are eliminated by `new FlashMessage()`. Some last ones are still to be replaced.

Debateable: does the term `action` make sense here? Flash messages are outside the redux event/action concept, so I am leaning towards renaming and moving elsewhere.

marians:
Things to test in particular:

- When a user is logged in, then this user is deleted, the user does not get logged out. However all API requests result in error messages.

marians:
Problem in login: login form not working at all.

image

marians:
Regarding payload: noty adds 9.1K for JS and 2.6K for CSS, both compressed.

marians:
Currently deployed on geckon for testing.

src/actions/flashMessageActions.js
```diff
@@ -1,23 +1,64 @@
...
+  MEDIUM: 3000,
+  LONG: 10000,
+  VERYLONG: 30000,
+};
 
-export function flashAdd(flashMessage) {
-  return function(dispatch) {
-    dispatch({ type: types.FLASH_ADD, flashMessage });
+export class FlashMessage {
+  constructor(text, type, ttl, subtext) {
```
pipo02mix: Why create a new class and not reuse the function signature but extend params?

Something like
```javascript
export function flashAdd(text, type, ttl, subtext) {

    this.noty = new Noty({
      type: type,
      text: this.text,
      timeout: this.timeout,
      theme: 'bootstrap-v3',
      layout: 'topRight',
      visibilityControl: true,
      animation: {
        close: 'flash_message_close',
      },
    });

    this.noty.show();
}
```

marians: Mainly because I find it very expressive to say "new Something" to "create" a something. Beyond that, the difference is small.

marians:
@oponder I'd still like your feedback here

src/actions/flashMessageActions.js
```diff
@@ -1,23 +1,64 @@
...
+    this.text = text;
+    if (subtext) {
+      this.text = '<p>' + text + '</p><p>' + subtext + '</p>';
```
oponder:
```suggestion
      this.text =  '<b>' + text + '</b>' + '<br/>' + subtext;
```

oponder: I noticed that the flash messages with subtext looked a bit different. Before:
screenshot 2019-03-06 at 14 54 07

after:
screenshot 2019-03-06 at 14 55 16

marians:
Good catch regarding the font weight. I deliberately did NOT make the first part of two part notifications bold, as I found it too much. But the type indeed is rendered with the wrong weight. I'll align this.

About XSS: here is what Noty has to say about this [in docs](https://ned.im/noty/#/options):

image

Are we safe if we run the escape function from https://stackoverflow.com/a/6234804/1228491 over all text parts coming from elsewhere (error messages, user input)? As the code is right now, this would have to be done before the texts are passed to `new FlashMessage()` though, which isn't ideal, as we have to think of it everywhere.

oponder:
> I'm pretty sure Noty thought of this

They did think of this, but chose to cover it with just a message in the docs:

screenshot 2019-03-06 at 15 28 13

see also: https://github.com/needim/noty/issues/318

I'm not a 1337 haxor and can't actually make scripts run, but I am able to get _some_ tags rendering, so it's a bit smelly to me. Easiest thing to do would be to add some kind of sanitisation to any text before we send it to noty.

oponder:
Heh great minds think alike :) I think that escaping should be ok for here.

> As the code is right now, this would have to be done before the texts are passed to new FlashMessage() though, which isn't ideal, as we have to think of it everywhere.

I think it can be done in the constructor, so doesn't need to be applied everywhere.

I'm thinking we santise `text` and `subtext` before calling `new Noty({....` in `flashMessageActions.js`

```
export class FlashMessage {
  constructor(text, type, ttl, subtext) {
     text = sanitize(text);
     ....
```

marians:
The thing is, I want to pass texts like `"blah <code>cluster-id</code> blah"` to `new FlashMessage(). I want to make use of HTML markup. If we escape at the very end, this wouldn't be rendered as expected.

oponder:
Oh I see... that's a pain. Maybe we can use something smarter that preserves some whitelisted elements? Found this while googlin': https://github.com/punkave/sanitize-html

marians:
Oh, that's too many dependencies for my taste. Then I'd rather do the whitelisting ourselves. SO far it's only `<code></code>` I am interested in, so that is simple enough.

marians:
The last commit fixes the escaping part.

PR #510, opened 2019-02-27
Remove or simplify constructors

marians:
This PR removes all constructors that

- do nothing
- only initialize state

and fixes a few possible bugs along the way


marians:
I'll test these changes today in a test installation

marians:
I had to revert the change I made for the password reset handling, as that didn't work. It was a rather complicated constructor. Now works fine again.

PR #511, opened 2019-02-28
Refactor - add more "routing" components

oponder:
This refactoring takes the `organizations` tree of views and refactors them into a pattern of routing components and rendering components.

The idea here is that we start to separate the rendering logic from the routing and data gathering logic. Eventually the actual views could possibly become pure functions, which are then easier to test and reason about.

New 'smells' are:

- If you have a set of routes that link deeper than one level, you are probably missing a routing component somewhere

- If the view component has a lot going on in the `mapStateToProps` function, then it can probably move up to the route component instead

- If you have a `Breadcrumb` in your `render`, then this should probably be a routing component.

The `routing` component should always be the `index.js` of a folder.

I'm still thinking about the naming of things.


oponder:
I'd be happy to walk through these changes and what motivated them with anyone that wants to review.

It gives a bit more room for structure and more refactoring to move out more logic from the 'view' components.

PR #512, opened 2019-02-28
Add dropdown button for AWS region selection in cluster creation

tuommaki:
There is now an upcoming support for inter-region clusters in aws-operator and
to use that one needs to be able to specify the region the cluster should be
created in. Therefore add a dropdown selection for this region.

### DIFF
--- a/src/components/create_cluster/index.js
+++ b/src/components/create_cluster/index.js
@@ -156,6 +156,7 @@ class CreateCluster extends React.Component {
           },
           name: this.state.clusterName,
           owner: this.props.selectedOrganization,
+          region: this.state.aws.region,
           release_version: this.state.releaseVersion,
           workers: workers,
         })
--- a/src/components/create_cluster/index.js
+++ b/src/components/create_cluster/index.js
@@ -455,13 +455,14 @@ class CreateCluster extends React.Component {
               switch (this.props.provider) {
                 case 'aws':
                   return (
-                    <div>
-                      <p>
-                        Select AWS Region for your cluster.
-                      </p>
+                  <div>
+                    <div className='row section'>
                       <div className='col-3'>
-                         <div className='textfield'>
-                            <label>Region:</label>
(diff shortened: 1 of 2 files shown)


src/components/create_cluster/index.js
```diff
@@ -197,6 +200,31 @@ class CreateCluster extends React.Component {
...
+          'eu-west-2',
+          'eu-west-3',
+          'eu-north-1',
+          'sa-east-1',
+          'us-gov-east-1',
+          'us-gov-west-1',
+      ];
+      return regions;
+  }
+
```
marians: Will this move to the info endpoint?

src/components/create_cluster/release_selector.js
```diff
@@ -250,8 +250,8 @@ function mapStateToProps(state) {
   var provider = state.app.info.general.provider;
   var releases = Object.assign({}, state.entities.releases.items);
 
-  var activeSortedReleases = _.filter(releases, release => release.active);
-  activeSortedReleases = activeSortedReleases.map(release => release.version);
+  //var activeSortedReleases = _.filter(releases, release => release.active);
+  var activeSortedReleases = _.map(releases, release => release.version); //activeSortedReleases.map(release => release.version);
```
tuommaki: :exclamation: REVERT!

src/components/create_cluster/index.js
tuommaki: Yes, I did couple shortcuts here and there in order to get this running E2E within two days :slightly_smiling_face:

tuommaki:
I guess this PR can act as a base, but proper version needs more work.

marians:
@tuommaki What should we do with this PR? AFAIK the companions in api etc. have been closed. Should we close this one, too? We can dig it up here once we want to implement this.

tuommaki:
Yes, let's close this one and redo when needed.

PR #513, opened 2019-03-05
Make input field selector not match radio, checkbox

marians:
Fixes https://github.com/giantswarm/giantswarm/issues/5379

### Preview

image

image

PR #514, opened 2019-03-06
Expose the InputField type property as HTML attribute

marians:
This PR fixes the problem that the AWS EC2 instance type input field and the Azure VM size input field did not carry a `type=text` HTML attribute.

As a consequence, they were rendered like this:

image

### DIFF
--- a/src/components/shared/input_field.js
+++ b/src/components/shared/input_field.js
@@ -118,6 +118,7 @@ class InputField extends React.Component {
           ref={i => {
             this.input = i;
           }}
+          type={this.props.type}
           id={this.props.name}
           value={this.state.value}
           onBlur={this.onBlur}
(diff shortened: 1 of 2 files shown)


marians:
The rendering bug was introduced by myself in https://github.com/giantswarm/happa/pull/513

src/components/shared/input_field.js
```diff
@@ -143,6 +144,7 @@ InputField.propTypes = {
   label: PropTypes.string,
   value: PropTypes.string,
   readOnly: PropTypes.bool,
+  type: PropTypes.stringm,
```
oponder: oh wait a minute...

oponder: `stringm` seems like a typo.

marians: Indeed. See https://github.com/giantswarm/happa/pull/515

PR #518, opened 2019-03-07
Wider layout to use more screen real estate

marians:
Fixes https://github.com/giantswarm/giantswarm/issues/4898


marians:
Deployed on `ginger` for your testing.

I recommend to test this out in action and compare current vs. PR and use different windows sizes and ideally laptop screen vs. external screen to get a feeling of the pros and cons.

oponder:
I think this works better for the smaller screens now, but worse for wider screens.

Perhaps we can apply some global max width, so that it doesn't get too wide?

I'm a fan of not letting UI's get as wide as the screen allows. It impacts readability and usability for me when related things aren't close together. Like how the "get started" button is now really quite far away from the cluster title, and for having a full overview of the UI it takes more eye saccades now.

marians:
> I'm a fan of not letting UI's get as wide as the screen allows.

Yes, in general I agree. But I think we should take care of that in the inner components, not on the global layout level. No text line should ever take the full width of a maximized browser window.

Are there cases we should treat in that sense, where we can limit the column width in the scope of this PR?

oponder:
Well I think everything now has the potential to be too wide.. so I wouldn't want to approach that on a component by component or page by page basis.

For a reference data point, you can see that this page on github doesn't allow itself to get wider than 980px, I'm a fan of this as a maximum width, but maybe that's too old fashioned.

marians:
> For a reference data point, you can see that this page on github doesn't allow itself to get wider than 980px

Except for when it does. See https://github.com/giantswarm/happa/pull/520/files?utf8=%E2%9C%93&diff=split for an example.

I don't want to limit us on a global basis. What I want is to allow some views like the key pairs table to use as much width as is available.

oponder:
ok, I hear ya. would be good to get some other opinions here too then.

pipo02mix:
I see the point of Marian. If we made it wider on the `main` div we can add another div as a child control for each section that let us chose wider or narrower layouts

oponder:
@marian is this one still relevant after the keypair table improvements?

marians:
The pain is reduced definitely after the work on the key pair table. Won't pursue it in this PR. CLosing.

PR #522, opened 2019-03-07
Sort imports

marians:
Fixes https://github.com/giantswarm/giantswarm/issues/5538

This adds an eslint plugin that checks whether Imports are sorted properly, plus it applies the right ordering.

There is a bit more to it than alphabetical sorting. Here is an example:

```js
import * as clusterActions from '../../actions/clusterActions';
import { bindActionCreators } from 'redux';
import { connect } from 'react-redux';
import { FlashMessage, messageTTL, messageType } from '../../lib/flash_message';
import _ from 'underscore';
import Button from '../shared/button';
import React from 'react';
```

The above example shows three groups of imports.

1. Imports that import all (`*`) exported symbols always come first.
2. Imports that use curly brackets to import multiple symbols from a module come second.
3. Imports for single symbols follow last. Note that if all lines were sorted strictly alphabetical, the `_` (underscore) would appear in the very first line, which would be wrong.

Note that the eslint plugin provides autofix capabilities. So in VS code, for example, there is an action called

    ESLint: Fix all auto-fixable Problems

That would bring the imports in the right order.

Also node that in **multiple**  symbol imports, the internal order is important, too. This is wrong:

```js
import { Z, A } from '../../lib/module';
```

while this is correct:

```js
import { A, Z } from '../../lib/module';
```

Finally, the ordering is **case insensitive**.


marians:
@oponder Good catch, thanks! And it just took me 3 commits to make index.html what it was again :o

PR #523, opened 2019-03-08
Remove increment/decrement buttons for disabled input

marians:
Fixes https://github.com/giantswarm/giantswarm/issues/5469

In Firefox, with the number picker being disabled, the input field would contain two tiny increment/decrement buttons.

image

This PR hides the buttons and aligns he appearance with Chrome.

### DIFF
--- a/src/styles/components/_number_picker.sass
+++ b/src/styles/components/_number_picker.sass
@@ -18,6 +18,11 @@
   position: relative
   margin-bottom: 5px
 
+  input:disabled
+    // to hide the increment/decrement buttons when disabled
+    -moz-appearance: textfield
+    background-color: #fff
+
 .number-picker--validation-error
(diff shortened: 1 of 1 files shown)


marians:
Re-ping @oponder

src/styles/components/_number_picker.sass
```diff
@@ -18,6 +18,11 @@
   position: relative
   margin-bottom: 5px
 
+  input:disabled
+    // to hide the increment/decrement buttons when disabled
+    -moz-appearance: textfield
+    background-color: #fff
```
oponder: is this background color right? Quickly tested the branch and it looks ok. But the `background-color: #fff` is being overruled, so it actually does nothing, and if it would do something, I think it would look odd.

PR #525, opened 2019-03-08
Add age condition for showing Get Started button

marians:
For https://github.com/giantswarm/giantswarm/issues/5410

This hides the `Get Started` button in the cluster overview if the cluster is older than 30 days.

Preview

image

### DIFF
--- a/src/components/home/cluster_dashboard_item.js
+++ b/src/components/home/cluster_dashboard_item.js
@@ -9,6 +9,7 @@ import { relativeDate } from '../../lib/helpers.js';
 import Button from '../shared/button';
 import ButtonGroup from 'react-bootstrap/lib/ButtonGroup';
 import ClusterIDLabel from '../shared/cluster_id_label';
+import moment from 'moment';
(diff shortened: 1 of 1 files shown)


marians:
Re-ping @oponder @pipo02mix

marians:
Very valid comment! Fix incoming...

src/components/home/cluster_dashboard_item.js
```diff
@@ -70,6 +71,19 @@ class ClusterDashboardItem extends React.Component {
...
+   * Returns true if the cluster is younger than 30 days
+   */
+  clusterYoungerThan30Days() {
+    var age = Math.abs(
+      moment(this.props.cluster.create_date)
+        .utc()
+        .diff(moment().utc()) / 1000
+    );
+
+    return age < 30 * 24 * 60 * 60;
```
pipo02mix: small perf improvement: calculate this in a class var as constant to avoid do the calculation every call :)

PR #526, opened 2019-03-11
Add catalog

oponder:
Because it's taking me a while, and didn't want to keep working in the dark, putting up the PR in draft state.

To finish:

- [x] Performance of the initial load of the catalog is bad. Needs pagination, or smarter delayed rendering of off screen apps
- [x] Finalize app detail page
  - [x] Add a disabled "install" button
  - ~Add a list of other versions~
  - [x] Add place for other metadata


oponder:
@marian Here are the things I worked on, and what's still todo, since we talked:

- [ ] lunar for improved clientside search
- [ ] back link go back to scroll position
- [ ] back link should get the catalog name in it too
- [ ] uppercase breadcrumbs
- [ ] name the repo 'helm-stable' instead of just 'helm'
- [ ] center the page loading spinner
- [ ] use same style for the catalog picker as used in the app list
- [ ] remove 'pick a catalog' label
- [ ] reduce size of the decription box below each app
- [ ] increase font size of the version
- [ ] remove hover effects, no drop shadows. experiment with a glow instead
- [ ] position the search bar further from the border, by making search bar a bit smaller
- [ ] center the no search results box
- [ ] include the search term in the no search results box
- [ ] detect images that fail to load on the app detail page and hide broken image
- [ ] different background color for tags vs the versions to differentiate them
- [ ] increase the font sizes of labels

marians:
Just noted that the repo name in breadcrumb and display are not matching. Should be MANAGED, is SAMPLE-CATALOG):

image

marians:
The error when YAML parsing doesn't work could be more expressive.

image

The exception is there: 
image

oponder:
> The error when YAML parsing doesn't work could be more expressive.

@marians I'm taking that on board and will do a follow up PR to address it. Right now I'd like to get these things integrated and do my first release of this feature.

PR #527, opened 2019-03-13
Auto-refresh for cluster details 

marians:
For https://github.com/giantswarm/giantswarm/issues/5556

This lets the cluster details refresh automatically. A note is added showing that this is a new feature and when the information has been last refreshed.

### Preview

image

### Details

- Loading of freshest data is done via `window.setInterval` every 30 seconds.
- the view component refreshes every 10 seconds, to update relative date strings like `a few seconds ago.`
- When the [page visibility API](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API) indicates that the tab is invisible, the refresh interval function is unregistered. When the page is revealed again the freshest data gets pulled immediately and the interval function is registered again.
- The current implementation doesn't simplify any re-use of the refresh mechanism in other components.
- Cluster deletion (by somebody else) is not handled. The page will continue to show the latest cluster details, plus an error messages, mostly as it used to be before. The error message re-appears every 30 seconds.

### TODO

- [x] fix refresh interval not being stopped when page is hidden
- [x] temporary visual highlighting of elements when they change, if simple enough
- [x] test if SSO JWT is refreshed before use
- [x] remove console debugging code
- ~~react sensibly in case of a 404 error to the getCluster request~~

### DIFF
--- a/src/reducers/clusterReducer.js
+++ b/src/reducers/clusterReducer.js
@@ -68,7 +68,7 @@ var ensureWorkersHaveAWSkey = function(clusterDetails) {
 };
 
 export default function clusterReducer(
-  state = { lastUpdated: 0, isFetching: false, items: {} },
+  state = { lastUpdated: null, isFetching: false, items: {} },
   action = undefined
 ) {
   var items;
@@ -84,6 +84,8 @@ export default function clusterReducer(
           ensureMetricKeysAreAvailable(cluster)
         );
 
+        items[cluster.id].lastUpdated = Date.now();
+
         // Guard against API that returns null for certain values when they are
         // empty.
         items[cluster.id].nodes = items[cluster.id].nodes || [];
@@ -92,14 +94,14 @@ export default function clusterReducer(
       });
 
       return {
-        lastUpdated: state.lastUpdated,
+        lastUpdated: Date.now(),
         isFetching: false,
         items: items,
       };
 
     case types.CLUSTERS_LOAD_ERROR:
       return {
-        lastUpdated: Date.now(),
+        lastUpdated: state.lastUpdated,
         isFetching: false,
         errorLoading: true,
         items: items,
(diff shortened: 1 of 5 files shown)


oponder:
Pretty cool! I think this is a good direction.

src/reducers/clusterReducer.js
```diff
@@ -92,14 +94,14 @@ export default function clusterReducer(
...
-        lastUpdated: state.lastUpdated,
+        lastUpdated: Date.now(),
         isFetching: false,
         items: items,
       };
 
     case types.CLUSTERS_LOAD_ERROR:
       return {
-        lastUpdated: Date.now(),
+        lastUpdated: state.lastUpdated,
```
marians: This is an unrelated fix. My thinking is that `lastUpdated` should only be set in the success case, not in the error case. It should carry the time when the data has been last fetched successfully.

src/reducers/clusterReducer.js
```diff
@@ -84,6 +84,8 @@ export default function clusterReducer(
           ensureMetricKeysAreAvailable(cluster)
         );
 
+        items[cluster.id].lastUpdated = Date.now();
+
```
marians: Here, every cluster item gets its own `lastUpdated` field.

src/reducers/clusterReducer.js
```diff
@@ -157,6 +159,8 @@ export default function clusterReducer(
         status: action.status,
       });
 
+      items[action.clusterId].status.lastUpdated = Date.now();
+
```
marians: Here every cluster status gets its own `lastUpdated` field.

marians:
I tried treating the 404 case (the cluster we are looking at has been deleted through some other way). Didn't manage to make it graceful.

- Simply redirecting to `/` yielded more errors, as that view also tries to load the status of the no-longer-existing cluster.
- My attempt to mitigate this by enforcing loading of all clusters before redirecting didn't fix that.

As it's not a very common case, I deecided not to pursue this now.

marians:
At least there is one reusable component now: `RefreshableLabel` is used to create a fading highlighting for some piece of content when it changes.

PR #531, opened 2019-03-22
Use react-timeout

marians:
A bit of refactoring, so we don't have to remember using `clearInterval` or `clearTimeout` before unmounting a component. react-timeout does that for us.

### DIFF
--- a/src/components/cluster/detail/cluster_detail_table.js
+++ b/src/components/cluster/detail/cluster_detail_table.js
@@ -7,6 +7,7 @@ import cmp from 'semver-compare';
 import moment from 'moment';
 import PropTypes from 'prop-types';
 import React from 'react';
+import ReactTimeout from 'react-timeout';
 import RefreshableLabel from '../../shared/refreshable_label';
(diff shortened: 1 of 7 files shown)


src/components/cluster/detail/cluster_detail_table.js
```diff
@@ -15,18 +16,14 @@ class ClusterDetailTable extends React.Component {
...
 
-  componentDidMount = () => {
+  componentDidMount() {
     this.registerRefreshInterval();
-  };
-
-  componentWillUnmount = () => {
-    window.clearInterval(this.reRenderInterval);
-  };
+  }
```
marians: I looked up a bit on the meaning of the arrow notation for a class method. It wasn't approprioate to apply that here, so I am reverting it. It's unrelated.

marians:
The good thing is, I didn't add the dependency for this. It was there already.

oponder:
> The good thing is, I didn't add the dependency for this. It was there already.

I wonder what it was once used for :P

PR #533, opened 2019-03-25
Publish semver versioning guidelines

marians:
As discussed on Slack https://gigantic.slack.com/archives/C13MG9F0W/p1553512913030900

### DIFF
--- a/docs/Release.md
+++ b/docs/Release.md
@@ -18,6 +18,14 @@ All you have to do is push a tag. CI will take care of it from there.
 
 Follow CircleCI's progress in https://circleci.com/gh/giantswarm/happa/.
 
+### Semantic versioning (semver) semantics
+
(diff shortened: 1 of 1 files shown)


docs/Release.md
```diff
@@ -18,6 +18,14 @@ All you have to do is push a tag. CI will take care of it from there.
 
 Follow CircleCI's progress in https://circleci.com/gh/giantswarm/happa/.
 
+### Semantic versioning (semver) semantics
+
+Releases are numbered after the semver system. As the notion of a breaking bs. non-breaking change isn't very meaningful for a GUI application, here is the system we apply:
```
oponder:
```suggestion
Releases are numbered after the semver system. As the notion of a breaking vs. non-breaking change isn't very meaningful for a GUI application, here is the system we apply:
```

PR #534, opened 2019-03-25
Fix access to undefined workers array on cluster details page

marians:
plus one additional error regarding RefreshableLabel supporting exactly one child.


marians:
Merging without approval

PR #536, opened 2019-03-26
Merge clusters into state to avoid flapping of cluster info

marians:
Fixes https://github.com/giantswarm/giantswarm/issues/5671 mostly

This seems to improve the situation.

Improvable: When values get first loaded, items highlight once, which they shouldn't ideally. But that's not the point of the fix.

Another thing I noticed, but couldn't explain when it happened:

image

Probably passed a bad value to `moment()`.

### DIFF
--- a/src/reducers/clusterReducer.js
+++ b/src/reducers/clusterReducer.js
@@ -4,52 +4,6 @@ import * as types from '../actions/actionTypes';
 import _ from 'underscore';
 import moment from 'moment';
 
-var metricKeys = [
-  'cpu_cores',
-  'cpu_used',
-  'ram_available',
-  'ram_used',
-  'pod_count',
-  'container_count',
-  'node_storage_limit',
-  'node_storage_used',
-  'network_traffic_incoming',
-  'network_traffic_outgoing',
-];
-
-// ensureMetricKeysAreAvailable
-// ----------------------------
-// Make sure that expected metrics keys are present on cluster and nodes
-// since Desmotes will omit them if they are not found in Prometheus
-var ensureMetricKeysAreAvailable = function(clusterDetails) {
-  clusterDetails.metrics = clusterDetails.metrics || {};
-  for (var metricKey of metricKeys) {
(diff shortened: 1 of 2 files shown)


src/reducers/clusterReducer.js
```diff
@@ -108,16 +73,9 @@ export default function clusterReducer(
...
 
     case types.CLUSTER_LOAD_DETAILS_SUCCESS:
-      items = Object.assign({}, state.items);
-
-      items[action.cluster.id] = Object.assign(
-        {},
-        items[action.cluster.id],
-        action.cluster
-      );
+      items = Object.assign(state.items, {});
```
oponder: My understanding is that when we do this we actually mutate the `state.items` object, instead of creating a new object. which is an anti pattern (mutating state). 

I think things are actually alright in this action (CLUSTER_LOAD_DETAILS_SUCCESS), did you find these changes necessary?

src/reducers/clusterReducer.js
```diff
@@ -75,14 +30,24 @@ export default function clusterReducer(
 
   switch (action.type) {
     case types.CLUSTERS_LOAD_SUCCESS:
-      items = {};
+      // if state was populated previously, let new data overwrite old data partially
+      var prevClusterIDs = Object.keys(state.items).sort();
+
+      // use existing state's items and update it
+      items = Object.assign(state.items, action.clusters);
```
oponder:
```suggestion
      items = Object.assign({}, state.items, action.clusters);
```

So that we get a new object, instead of mutating state.items.

I'm actually a bit surprised it didn't trigger a warning or error from redux. My understanding is a bit shaken. I'll check out the branch and do some poking tomorrow ok?

src/reducers/clusterReducer.js
marians: Good question. I didn't make sure. Will revert this part and see how that behaves.

src/reducers/clusterReducer.js
marians: sure

PR #537, opened 2019-03-27
Cluster details in tabs

marians:
This introduces the following tabs to the cluster details page:

- General (cluster details, also the Delete Cluster section)
- Key Pairs
- Managed Services

The loading behaviour hasn't been touched. This means that although we don't display key pairs by default, they are loaded immediately. This leaves room for improvements:

- don't load key pairs by default
- everytime when `Key Pairs` tab is revealed, load fresh data
- as long as the `General` tab is not active, don't refresh cluster details and status

### Preview

image

image

image


marians:
@oponder Wanna give this a test drive and let me know what you think?

marians:
At a second look, there should be more space between headline and tabs

oponder:
Looking good!

I think we can leave out that upper line that comes after the tab now, and possibly also the heading, now that the tab identifies what we're looking at.

So for example, instead of:

Screenshot 2019-03-27 at 19 18 44

I think something like this might make more sense?

Screenshot 2019-03-27 at 19 19 51

I'm not a fan of the slow transition fadeout / fadein. I bet its just a default thing but I'd vote to remove that transition, or make it a bit snappier.

marians:
Good points!

marians:
@oponder What do you think of this "transitional" bit?

image

oponder:
> What do you think of this "transitional" bit?

I'm ok with that! Maybe add the word "above" like: "We moved them into separate tabs above to clean up a bit"

oponder:
Any thoughts on this last bit from my first comment?

> I'm not a fan of the slow transition fadeout / fadein. I bet its just a default thing but I'd vote to remove that transition, or make it a bit snappier.

marians:
@oponder Sorry, I missed that one. I agree and I'll have to check how to adapt that.

marians:
@oponder I just gave my opinion :)

marians:
TODO: I tried updating two libraries in this PR, but the resulting yarn.lock keeps reverting to older versions. Will either have to figure this out or revert the changes in package.json.

PR #539, opened 2019-03-27
Fix cluster duplicates.

oponder:
Clusters were showing up twice in the organization list.

This is because of a mistake in the recently introduced merging code.

In the `clusters load success`  reducer, we were merging `action.clusters` which is an array of clusters into the `entites.items` object, which is supposed to be a dictionary of cluster id's to cluster objects:

`items = Object.assign({}, state.items, action.clusters);`


oponder:
Self merged because it was noticed by a couple of customers.

PR #540, opened 2019-03-28
Make parts of the keypair table copyable.

oponder:
This introduces a wrapper component called 'Copyable' and applies it to some elements in the keypair table.

It is a helper to allow us to make certain pieces of text quickly copyable to the clipboard.

See the behaviour in this gif:
copy

As you can see, you can click anywhere on the element, not just on the copyable indicator icon that appears on hover.

I made a change to the keypair table as well, combining the description and ID field in one column to allow for a bit more breathing room, and made most elements `small`er

Related: https://github.com/giantswarm/giantswarm/issues/2903

### DIFF
--- a/src/components/shared/copyable.js
+++ b/src/components/shared/copyable.js
@@ -11,6 +11,13 @@ class Copyable extends React.Component {
     copied: false,
   };
 
+  constructor(props) {
+    super(props);
+
+    this.copyToClipboard = this.copyToClipboard.bind(this);
+    this.mouseLeave = this.mouseLeave.bind(this);
+  }
+
   copyToClipboard(e) {
     e.preventDefault();
     e.stopPropagation();
@@ -31,8 +38,8 @@ class Copyable extends React.Component {
     return (
       <div
         className='copyable'
-        onClick={this.copyToClipboard.bind(this)}
-        onMouseLeave={this.mouseLeave.bind(this)}
+        onClick={this.copyToClipboard}
+        onMouseLeave={this.mouseLeave}
         style={{ cursor: 'pointer' }}
       >
         <div className='copyable-content'>{this.props.children}</div>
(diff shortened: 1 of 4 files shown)


marians:
- Copy interaction looks fine
- Placing both description and ID into the same column, and labelled "Descriptiion", is not nice
- I'm missing the truncation indicator (ellipses)

src/components/shared/copyable.js
```diff
@@ -11,6 +11,13 @@ class Copyable extends React.Component {
     copied: false,
   };
 
+  constructor(props) {
+    super(props);
+
+    this.copyToClipboard = this.copyToClipboard.bind(this);
+    this.mouseLeave = this.mouseLeave.bind(this);
```
marians: How about arrow notation for `copyToClipboard()`?

oponder: I did a bit more research (also spurred by @wind0r thx!), and this appears to be the most performant way. 

It is more verbose, but it avoids some issues that the arrow notation brings. 

It's a micro optimisation, but I think we could apply this everywhere and then not have to think about it anymore. 

This is probably a `fmt` repo discussion, and I realise I just introduced a different way of doing something again. So I'd be fine with going for the arrow notation now, since we agreed a while back to use that in cases where a bind was actually needed, and then we can discuss this a bit more internally, and then possibly change it everywhere in one fell swoop.

marians: Yeah, talking about it makes sense. Otherwise we'll end up reverting each other's standardization efforts.

wind0r: I only noticed in the initial version of this pr that in there was a .bind() call in the render function which should be avoided :)

I don't think/know if there is any performance difference between Arrow Function or Bind in Constructor (the result should be quite the same, but i never checked the generated code). The important part just should be to avoid bind/arrow in render functions and that the solution should be the standardized/same around the code base

There also is an eslint rule for this: [react/jsx-no-bind](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md)

PR #542, opened 2019-03-29
Make the keypair table sortable.

oponder:
This applies the `BootstrapTable` which we use in a few other places as well to the Key Pair table on the Cluster Detail screen, making the columns sortable.

keypairtablesorting


marians:
@oponder I agree that the cursor is weird. Not that you introduced it. But I ddn't notice it before. Can we set this?

```css
th.sortable {
  cursor: pointer;
}
```

oponder:
Looks like we had some styles from react-bootstrap-table-next to load that weren't actually being imported anywhere.

I moved the filed to `vendor` directory, since it's code from a different project, and made sure it's getting imported correctly.

PR #543, opened 2019-03-29
Reset scroll position when navigating.

oponder:
I noticed this while working on the app catalog, but it is also present in `master`. 

The scroll position is not reset to the top of the page when navigating. This is only matters if you navigate between two pages that actually have some content to scroll for like when moving between the pages of the getting started guide for example.

It's a bit weird, because I never noticed this before, so at first I thought this is a regression in a library somewhere.. but I checked and it happens still all the way to our first tagged release (0.1.31)

### DIFF
--- a/src/components/app.js
+++ b/src/components/app.js
@@ -32,6 +32,10 @@ var appContainer = document.getElementById('app');
 
 const store = configureStore({});
 
+history.listen(() => {
+  window.scrollTo(0, 0);
+});
+
 render(
   <Provider store={store}>
     <ConnectedRouter history={history}>


src/components/app.js
```diff
@@ -32,6 +32,10 @@ var appContainer = document.getElementById('app');
 
 const store = configureStore({});
 
+history.listen(() => {
+  window.scrollTo(0, 0);
+});
+
```
pipo02mix: But when you navigate back (previous page) it would be good to keep the position, dont you think?

oponder: Yeah indeed. The back button still does that though, for me at least. I think it's a standard browser behaviour somehow. Not sure how it is implemented to know the difference here, but it seems to work.

PR #546, opened 2019-04-08
add default sorted setting to reflect old table behaivor

wind0r:
befor #542 the tables was sorted by the create date in reverse order

`{_.sortBy(this.props.cluster.keyPairs, 'create_date').reverse()`

i quite liked it so i added this back to the new table :)


oponder:
Makes sense to me! Thx fo the contribution. And sorry for the regression there.

PR #547, opened 2019-04-09
Update minor and patch versions where possible.

oponder:


oponder:
fyi @sig-ux, self merging this minor dependency update.

PR #548, opened 2019-04-11
Improve responsiveness.

oponder:
This helps happa resize between its breakpoints in a smoother way.

This stops the effective viewport from getting smaller than it needs to before snapping into the 'medium' breakpoint.


oponder:
Deployed to `ginger` for review.

PR #549, opened 2019-04-12
Improve rendering of certificate orgs in key pairs table

marians:
- Splits multiple orgs, sorts them
- Highlights the cluster-admin org name `system:masters` (but not `systems:master` ;-))

### Before

image

### After

image

### DIFF
--- a/src/components/cluster/detail/certificate_orgs_label.js
+++ b/src/components/cluster/detail/certificate_orgs_label.js
@@ -1,0 +1,30 @@
+'use strict';
+
+import PropTypes from 'prop-types';
+import React from 'react';
+
+class CertificateOrgsLabel extends React.Component {
+  render() {
+    var orgs = this.props.value.split(',');
+    orgs.sort();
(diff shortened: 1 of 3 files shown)


marians:
@wind0r kindly suggested this.

oponder:
Nice!

I feel like the highlight `isadmin` state is a bit too harsh.. can we get away with something more subtle?

I think this is also a good opportunity to use a `React.PureComponent` instead of `React.Component`. https://reactjs.org/docs/react-api.html#reactpurecomponent

oponder:
you could try something like `border-left: 4px solid #6f0316` on those isadmin labels, so they get a bit of a smaller indication that they're special?

Screenshot 2019-04-12 at 21 06 52

(messed around quickly in the dev console for demonstration of what that looks like)

marians:
image

Toned down the red a little bit.

Also removed the nested `code` element to not have less DOM bloat.

PR #550, opened 2019-04-12
Structure general cluster info

marians:
This changes the cluster details page/tab in several ways:

- Remove SCALE button from top right, in a general attempt to get rid of those.
- Add vertical spacing to table rows to make table easier to read
- Remove top spacing and border form details table
- Regroup items to have the (unchangeable) VM size / instance type further in the top, above the dynamic details.

### Before

image

### After

image

### DIFF
--- a/src/components/cluster/detail/cluster_detail_table.js
+++ b/src/components/cluster/detail/cluster_detail_table.js
@@ -3,6 +3,7 @@
 import { relativeDate } from '../../../lib/helpers.js';
 import _ from 'underscore';
 import AWSAccountID from '../../shared/aws_account_id';
+import Button from '../../shared/button';
 import cmp from 'semver-compare';
 import moment from 'moment';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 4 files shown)


src/styles/_base.sass
```diff
@@ -143,3 +143,10 @@ div.tooltip-inner
 	padding: 0px 4px
 	line-height: 1.25em
 	border-radius: 2px
+
+.padding-left-10
```
oponder: I'm not in favour of these type of class names, as it means we're moving layout/presentation into the DOM, which we should try to avoid. 

 Looks like padding-left-20 and padding-left-30 are also not used in this case.

marians: You're right, changed.

src/components/cluster/detail/cluster_detail_table.js
```diff
@@ -315,7 +452,7 @@ class ClusterDetailTable extends React.Component {
                           {this.props.canClusterUpgrade ? (
                             <a
                               onClick={this.props.showUpgradeModal}
-                              className='upgrade-available'
+                              className='upgrade-available padding-left-10'
```
oponder: can be removed here as well

marians: Done, thanks!

PR #553, opened 2019-04-12
Improve scaling dialog warnings

marians:
Our warnings regarding scaling down needed improvements in several ways.

- Scenario: changing form pinned to 3 to min2/max3. The warnings indicated that worker nodes would be removed, which would in fact only happen if the autoscaler decides so. Accordingly, the submit button showed `Remove 1 node` when in fact only the scaling limits would be changed.
- We were lacking the info that a cluster with less than 3 nodes is not supported.
- We were showing some info regarding possible data loss that can be taken as general

This PR brings changes:

- Changed the logic regarding which warning appears in what situation, so that only warnings appear which apply to the situation
- change design of warnings, to prevent it from looking like an error message
- Simplify submit button working to make it shorter
- The label `To disable autoscaling, set both numbers to the same value.` changes to `To enable autoscaling, set minimum and maximum to different values.` when size is pinned.

### Before (AWS)

image

### Preview AWS autoscaling

x20VvfLWnx

### DIFF
--- a/src/components/cluster/detail/certificate_orgs_label.js
+++ b/src/components/cluster/detail/certificate_orgs_label.js
@@ -17,7 +17,11 @@ class CertificateOrgsLabel extends React.Component {
       if (element === 'system:masters') {
         classNames += ' isadmin';
       }
-      orgLabels.push(<span className={classNames}>{element}</span>);
+      orgLabels.push(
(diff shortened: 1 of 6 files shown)


marians:
Re-ping @oponder

PR #554, opened 2019-04-15
Add lunr to the app catalog.

oponder:


src/components/app_catalog/app_list/index.js
```diff
@@ -29,8 +45,10 @@ class AppList extends React.Component {
   componentDidMount() {
     // The hash value of the url is used by the app detail screen's back button
     // to indicate what app we should scroll to.
-    var scrolToApp = this.props.location.hash.substring(1);
-    window.scrollTo(0, this.appRefs[scrolToApp].offsetTop - 150);
+    var scrollToApp = this.props.location.hash.substring(1);
+    if (scrollToApp && scrollToApp !== '') {
+      window.scrollTo(0, this.appRefs[scrollToApp].offsetTop - 150);
+    }
```
oponder: Unrelated, but there was a bug here.

src/components/app_catalog/app_list/index.js
```diff
@@ -44,23 +62,19 @@ class AppList extends React.Component {
...
   // a filtered set of apps
   filterApps(allApps, filter) {
-    let apps = Array.from(Object.entries(allApps));
-    apps = apps.map(([, app]) => {
-      app.name = app[0].name;
-      app.version = app[0].version;
-      app.icon = app[0].icon;
-      return app;
-    });
+    console.log('f');
```
oponder: This is me getting a feel how often this method gets called, was already doing some checking on the performance

src/components/app_catalog/app_list/index.js
```diff
@@ -43,26 +59,23 @@ class AppList extends React.Component {
...
-    let apps = Array.from(Object.entries(allApps));
-    apps = apps.map(([, app]) => {
-      app.name = app[0].name;
-      app.version = app[0].version;
-      app.icon = app[0].icon;
-      return app;
-    });
+    // Lunr search
+    var lunrResults = this.index
+      .search(filter.searchQuery.trim() + ' ' + filter.searchQuery.trim() + '*')
```
pipo02mix: I dont get why it is repeated, can you explain to me?

oponder: Yeah, it has to do with how lunr handles the wildcard (`*`)

Without the double query like it is now, if we search for `prom` then the query will become `prom*`. And that will find `prometheus`.

However If we search for `prometheus` then the query will become `prometheus*` which will not match for `prometheus` since it is looking for something that has 1 or more characters after the string `prometheus`

So this is the workaround to allow the  'search as you type' behaviour we had before, without it  breaking as soon as you type the full matching word.

pipo02mix: ah ok so you can not use something like '+'

PR #555, opened 2019-04-15
Move dev instance types into dev cofiguration.

oponder:
Instead of having these devInstanceTypes in the source code, I move it to the dev configuration. This lets us avoid repeating checking wether or not instance types set and substituting in the dev values.

(Currently the cluster detail table is broken in dev, and this fixes it.)

Why I didn't do this in the first place? Not sure. I probably didn't think of the `.*` for the replacement in `start.sh`


oponder:
I also went ahead and changed everything to `.*`, that way we don't have to wonder 6 months down the road why one of them is special.

marians:
Just saw this now. It would be cool to set the azure VM sizes for dev, too. @oponder

PR #556, opened 2019-04-16
Remove concept of "current org"

marians:
This is a POC to demo how happa works if

- there is no such thing as a currently selected organization
- the front page lists all clusters

### TODO

- [ ] Check the getting started wizard
- [ ] Remove unused CSS
- [ ] Clean up `clusters` vs. `allClusters` duplication in `index.js`


oponder:
@marians is this something we still want to pursue?

ograu:
@marians can we close it?

PR #558, opened 2019-04-17
adding jest testrunner and some small inital tests

wind0r:
I created this quick unit testing draft since I had some time to kill and @marians  at giant summit said that happa should get some (unit) testing

i choosed jest as the test runner and enzeme as a utility to make it easier to test react components.
This is what nearly every React based project uses and also create-react-apps ships with it by default
```
"jest": "^24.7.1",
"babel-jest": "^24.7.1",
"enzyme": "^3.9.0",
"enzyme-adapter-react-16": "^1.12.1"
```

Jest detects following files as test cases.
* all files inside any `__test__` folder
* all files that end with `.test.js`
* all files that end with `.spec.js`

I decided to use `*.test.js` in this example draft

To remove duplicate code I moved the babel configuration to the babelrc file so it can be shared between webpack(dev/prod) and jest.

Personally I find most of the components are too big to test (hint ;) ) but since a few day ago the org labels component got added I had something to test the draft against.

While writing test for that component I noticed that a basic render test crashed, so I marked prop.value as isRequired and also a key was missing and throwing warnings.
```
// The component just checks for props.value === '' and does not check for undefined so it will crash without value
// props.value should get marked as isRequired
it('renders without crashing', () => {
  ReactDOM.render(<CertificateOrgsLabel />, div);
});
```

a different solution also would be to add default Props, but in this case it feels hacky
```
CertificateOrgsLabel.defaultProps = {
  value: "NoValidOrgLabelGiven",
}
```

The check for empty string happens inside the `key_pairs.js` and `certificate_orgs_label.js`
Also the code after the if clause technical also checks for empty string since the split will create an empty array, so the for each does not iterate and the function then return the empty orgLabels which is totally fine and valid.

I added a second commit which heavily opinionated refactored the `certificate_orgs_label.js` which of course should not get merged but could get used to get some inspiration.
my opinion in this case is heavily inspired by the airbnb rule set. The most obvious change is changing the forEach to map which removes the need for `var orgLabels = [];`

Personally I would like that the variable certificate_organizations gets modified(splitted) from a string to an array at the earliest place possible (api?) so all the converting and checking does not need to happen later.

Prettier does not really make the test files pretty. I was not bored enough to also change prettier settings or create extra settings for test files but that could be easier done because of the .test.js suffix.

Usage:
* npm test
* npm test -- --watch
* npm test -- --coverage

Because of cli auto completion I like to add those commands to the package.json instead of just using -- --x, so I added
* npm run test:watch

Sidenode:
* Instead of empty spans it is totally valid to return null or empty objects
* I may, would add the @babel/preset-react preset instead of adding the jsx plugin on its own (since the preset also adds extra plugins to dev builds which can sometimes help)
* Cypress(https://www.cypress.io/) is really nice For e2e-testing
* There is no need to add ''use strict';' to every file since the babel preset does it once per bundle
* To reduce duplicated configuration I prefer just having one webpack configuration with if classes for different environments. I can provide a PR if that is something you agree on.

PS: Sorry for the long pr, i did not had time to write a short one :) 

PSS: Ups, clicked create pr instead of draft. Sorry

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,4 @@
+{
+  "presets": ["@babel/preset-env"],
+  "plugins": ["transform-react-jsx", "transform-class-properties"]
+}
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,6 @@ Icon
 .tmp
 /secrets.env
 /dist
+
+# Test Coverage (Jest)
+/coverage
--- a/package.json
+++ b/package.json
@@ -7,7 +7,8 @@
   "src": "src",
   "test": "test",
   "scripts": {
-    "test": "grunt test",
+    "test": "jest",
+    "test:watch": "jest --watch",
     "start": "grunt serve",
     "prettier": "prettier --write \"src/**/*.js\"",
     "validate-prettier": "prettier --no-color --list-different \"src/**/*.js\""
@@ -16,6 +17,9 @@
     "singleQuote": true,
     "jsxSingleQuote": true,
     "trailingComma": "es5"
+  },
+  "jest": {
+    "setupFilesAfterEnv": ["./src/setupTests.js"]
   },
   "dist": "dist",
   "mainInput": "main",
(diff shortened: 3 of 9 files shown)


oponder:
Whoa @wind0r ! Thanks for this contribution! 

It's been on my forever todo list to add tests to Happa. Thx also for the sidetones and detailed info.

I'll take some time to review it later today, but it's looking good at first glance to me.

oponder:
Continued in #559

PR #559, opened 2019-04-17
Add jest

oponder:
See https://github.com/giantswarm/happa/pull/558 for @wind0r's introduction of jest. 

Made this PR so we can hook it up to CI, and do some small changes.

### DIFF
--- a/src/components/cluster/detail/certificate_orgs_label.test.js
+++ b/src/components/cluster/detail/certificate_orgs_label.test.js
@@ -1,0 +1,78 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { shallow } from 'enzyme';
+
+import CertificateOrgsLabel from './certificate_orgs_label.js';
+
+// Component only checks for props.value === '' and does not check undefined so it will crash without value
+/*
+it('renders without crashing', () => {
+  const div = document.createElement('div');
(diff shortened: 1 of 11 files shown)


oponder:
I decided to run install and run `yarn test` on the CI machine itself, its faster, and I don't really feel attached to the idea of everything _needing_ to be in a docker container anymore. 

Opinions welcome.

src/components/cluster/detail/certificate_orgs_label.test.js
```diff
@@ -0,0 +1,70 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { shallow } from 'enzyme';
+
+import CertificateOrgsLabel from './certificate_orgs_label.js';
+
+it('renders single label without crashing', () => {
```
pipo02mix: the message is a bit weird IMO, `without crashing` can be omit and it will be  clear enough

pipo02mix:
That's pretty great

thanks @wind0r

src/components/cluster/detail/certificate_orgs_label.test.js
wind0r: I do not know why but people started calling the first basic render test for react components 'renders without crashing'. Maybe because 'renders' is not a long enough.

a different and more fancy version would be 'renders without exploding' :)

PR #560, opened 2019-04-17
Auto login user after password reset

sohos:
This is a little contribution about Issue #143

PR #561, opened 2019-04-17
Improve jest

sohos:
Based on PR #558 I added full component testing with mocking of Redux, Promise and Superagent.
I also wrote tests for PR #560


marians:
Thanks for your contribution!

Could you please explain (or comment) a bit more what is happening here? That would make reviewing much easier.

sohos:
The idea here is to make full component tests. Enzyme + JSDom imitate react component mount (the same way as react mount component in the browser), with component lifecycle events (componentDidMount, componentWillUnmount and etc..) After that with enzyme you can assert, manipulate and traverse react component output. You can change the state, props and to test how the component respond to these changes. This PR contain simple test for <SetPassword /> component to test successful password change and automatic login, the same way as users do from the browser. Because the component depends on API requests we must mock these requests and return whatever result we want - success, error. You can add tests to other components, just wrap the component with <ComponentWrapper />. With all this, if you write more tests, your CI can run these tests and to be sure that the component does not break if you change some child component, library.

oponder:
Hey @sohos , I really appreciate the contribution here. There's a few too many things happening at once, I'll keep this for reference and possibly cherry pick some things from here if I find it needed as we continue down the path of adding tests. Hope that's alright!

PR #562, opened 2019-04-18
Auto login user after password reset

oponder:
Continuation of @sohos' contribution #560


oponder:
@sohos thanks for the contribution, before I continue can you acknowledge you've read and agree with our DCO? https://github.com/giantswarm/happa/blob/master/DCO

Replying with something like "I've read and agree with the DCO" is enough.

sohos:
I've read and agree with the DCO

Issue #234, opened 2017-10-11
Password reset jumps due to empty terminal during password change


marians:
After the user has entered the first password, the terminal element is emptied and changes its height. This makes the form jump upwards a bit, then downwards again once both passwords are entered.

To reproduce:

1. Initiate the "Forgot password workflow"
2. Follow the reset link in the email
3. Paste some text from the clipboard into the first password field
4. Hit TAB. The terminal is now empty.
5. Paste the same text again. The terminal is now filled.

As a simple solution, we should make sure that a little message like `Now please repeat that password` is shown in the terminal.

PR #563, opened 2019-04-18
Fix "empty password" is not valid anymore

sohos:
I think it's related to Issue #234 
In "Sign up" screen if you leave the password field empty and hit TAB, "Next" button becomes active.
In "Set your new password" screen after you reset your password, if you leave both password fields empty and hit TAB, the status text change to "Perfect match" and submit button becomes active.

### DIFF
--- a/src/lib/password_validation.js
+++ b/src/lib/password_validation.js
@@ -4,9 +4,7 @@ export function validatePassword(password) {
   var valid = false;
   var statusMessage = '';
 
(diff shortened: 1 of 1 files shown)


oponder:
Hey @sohos, thx again for your contribution!

I'll do the same as before. I'll close this PR and open a new one because of our CI limitations.

oponder:
Continued here: https://github.com/giantswarm/happa/pull/566

PR #564, opened 2019-04-24
Add label checks for labels used in k8s 1.14

MarcelMue:
Labels change with k8s 1.14 - this is basically the same PR as https://github.com/giantswarm/gsctl/pull/352 for gsctl. 

I tested this briefly on gauss and everything seems to be working fine.

### DIFF
--- a/src/components/cluster/detail/view.js
+++ b/src/components/cluster/detail/view.js
@@ -193,7 +193,7 @@ class ClusterDetailView extends React.Component {
       var workers = 0;
       nodes.forEach(node => {
         if (Object.keys(node).includes('labels')) {
-          if (node.labels['role'] != 'master') {
+          if (node.labels['role'] != 'master' || node.labels['kubernetes.io/role'] != 'master') {
             workers++;
           }
         }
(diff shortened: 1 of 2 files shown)


src/components/cluster/detail/view.js
```diff
@@ -193,7 +193,10 @@ class ClusterDetailView extends React.Component {
       var workers = 0;
       nodes.forEach(node => {
         if (Object.keys(node).includes('labels')) {
-          if (node.labels['role'] != 'master') {
+          if (
+            node.labels['role'] != 'master' &&
+            node.labels['kubernetes.io/role'] != 'master'
+          ) {
```
pipo02mix: Why do you need to check both labels? I dont see same logic in gsctl 🤔

MarcelMue: I added effectively the same logic in gsctl here:
https://github.com/giantswarm/gsctl/pull/352/files

The reason is that label name changed with k8s 1.14.

oponder: So basically it can be either `role` or `kubernetes.io/role` depending on the cluster version right?

MarcelMue: Yes, exactly.

PR #565, opened 2019-04-25
Extract and Refactor the Create Keypair modal using React Hooks.

oponder:
This PR extracts the Create Keypair modal from `key_pairs.js`. I also thought it would be a nice opportunity to try to refactor it using Hooks.

My main problem with using hooks was figuring out a way to make the button, which lived outside that component to be able to trigger the opening of the modal. Before I would use a ref and so be able to trigger a method (like `open` or `showModal`) on the instance of the modal class. Instead I moved the button into the modal class itself, so now it can take care of opening itself. 

An alternative would be to make the modal's open/close state be a prop, but what I don't really understand about Hooks, and that approach, is that then a lot of the 'business logic' that I would consider belongs to a component itself has to be managed by something outside of that component. 

I still need to level up more around the patterns here.


oponder:
Deployed to `gauss` so you can test if the keypair modal still works in the same way as before.

PR #566, opened 2019-04-25
Fix empty password is not valid anymore

oponder:
Continued from: https://github.com/giantswarm/happa/pull/563

related to: https://github.com/giantswarm/happa/issues/234


oponder:
Tested, works nicely! thx @sohos :)

Fixes #234 indeed.

PR #568, opened 2019-04-25
Fix member add notification

oponder:
reported by @jgsqware , we were using the `ERROR` type of flash message for a successful member adding.

Also noticed that only the first `<code> </code>` tag is whitelisted properly in our escapeHTML function. Fixed that as well.


oponder:
Before:
Screenshot 2019-04-25 at 5 07 36 PM

After:
Screenshot 2019-04-25 at 5 08 09 PM

marians:
Thanks!

PR #571, opened 2019-04-29
Don't show tab 2 as the first tab.

oponder:
I made a mistake and committed this, which I had set to `2` to more quickly test some changes I was making to the keypair table.

I'm gonna try to write a test so I don't make that same mistake again.


marians:
I would appreciate to approve this :)

oponder:
Luckily this isn't on customer installations yet, since the bug isn't in a release yet.

oponder:
I created our own `Tabs` component which wraps the Bootstrap's Tab component and has the props set to show the first tab. That made it easier to write a quick test.

PR #572, opened 2019-04-30
Wire up a keypair detail modal.

oponder:
@marians, this wires up a detail modal and adds a button to each row, placed hackishly in the `common name` column for now.

Feel free to take this PR as inspiration or continue from here.

I didn't manage to do what I wanted to do (which was to NOT have the visible state of the keypair tracked in the parent), so this is basically the same pattern we've done so far, except there is now this wrapped Bootstrap Modal (`generic_modal.js`) which makes it a bit less typing to get the structure of a basic modal up and running.

### DIFF
--- a/src/components/cluster/detail/key_pairs.js
+++ b/src/components/cluster/detail/key_pairs.js
@@ -7,6 +7,7 @@ import Button from '../../shared/button';
 import CertificateOrgsLabel from './certificate_orgs_label';
 import Copyable from '../../shared/copyable';
 import KeypairCreateModal from './key_pair_create_modal';
+import KeypairDetailModal from './generic_modal';
 import moment from 'moment';
 import OverlayTrigger from 'react-bootstrap/lib/OverlayTrigger';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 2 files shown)


src/components/cluster/detail/key_pairs.js
```diff
@@ -140,13 +147,17 @@ class ClusterKeyPairs extends React.Component {
...
     return (
-      <Copyable copyText={row.common_name}>
-        <small>{row.common_name}</small>
-      </Copyable>
+      <React.Fragment>
+        <Copyable copyText={row.common_name}>
+          <small>{row.common_name}</small>
+        </Copyable>
+        <br />
+        <Button onClick={this.showKeypairModal.bind(this, row)}>Details</Button>
```
oponder: I think we have to bite the bullet on this one, since we have to pass a param (row) to this function for every row of the table.

If anyone knows a different way, I'd learn so much if you could explain it.

src/components/cluster/detail/key_pairs.js
```diff
@@ -22,6 +23,12 @@ class ClusterKeyPairs extends React.Component {
     cn_prefix: '',
     cn_prefix_error: null,
     certificate_organizations: '',
+    keypairModal: {
+      visible: false,
```
oponder: This is what I wanted to avoid somehow.. but I couldn't get it working without code that I felt was just too much hassle. I wanted to fully encapsulate the concern of the modal's state, but maybe that is just not how to do things in React.

marians:
@oponder I think this PR is no longer needed

PR #573, opened 2019-05-02
Reduce key pairs table info, show details in modal

marians:
This changes the key pairs table, based on user feedback, to contain the most important info only, but in more readable fashion.

## Changes:

- Removed ID and description from table, to appear in a details modal, added a Details button to open the modal
- Column order. Put CN in front, as it is perceived as the identifier users care about.
- Removing common suffix (cluster API domain) from CNs in the table.
- Show full group information instead of truncated.
- Color hashing on group names

### Table preview

image

### Modal

image

image

image

### DIFF
--- a/src/components/cluster/detail/certificate_orgs_label.test.js
+++ b/src/components/cluster/detail/certificate_orgs_label.test.js
@@ -54,17 +54,3 @@ it('renders multiple labels in alphabetic order', () => {
       .props().children
   ).toEqual('Zepto');
 });
-
-it('gives system:masters a special isadmin class', () => {
-  const label = shallow(<CertificateOrgsLabel value={'system:masters'} />);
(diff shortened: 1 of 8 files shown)


src/components/cluster/detail/certificate_orgs_label.test.js
```diff
@@ -54,17 +54,3 @@ it('renders multiple labels in alphabetic order', () => {
       .props().children
   ).toEqual('Zepto');
 });
-
-it('gives system:masters a special isadmin class', () => {
```
oponder: Why did you remove this test? as far as I can see the isadmin code is still present.

marians: I forgot to remove it. Done now. The extra `isadmin` class makes no sense in combination with the hash colors. At least not with background color.

PR #574, opened 2019-05-02
Fetch installed apps for each cluster

oponder:
This adds fetching of apps installed on each cluster when the users hits the cluster detail page.


oponder:
Here's what it looks like for now:

Screenshot 2019-05-06 at 1 04 37 PM

The icons will be there in the left of the app, similar to how they are below for the preinstalled apps. 

Still discussing exactly how to achieve that, whether we store the icon url in the App CR or look it up in the appcatalog, but that would meaning loading the whole app catalog even for the cluster detail page, which I'd like to avoid.

Requesting feedback now, since I would like to keep moving. Design iteration and ideas are welcome, but keeping it at the bare basics for now.

src/components/cluster/detail/cluster_apps.js
```diff
@@ -142,32 +145,89 @@ class ClusterApps extends React.Component {
...
-            Soon you will be able to deploy managed apps like monitoring, log
-            storage, and more simply by selecting from our catalog.
-          </p>
-
-          <NavLink to={`/managed-apps/`}>
-            <Button>Browse Managed Apps</Button>
-          </NavLink>
+        <h3 className='table-label'>Installed Apps</h3>
+        <div className='row'>
+          {this.props.installedApps && this.props.installedApps.length === 0 && (
```
pipo02mix: Does it mean we only know if the customer can install apps if it has already app installed? 🤔

oponder: Nope, good question, here's what it looks like when no apps are on the tenant cluster yet:

Screenshot 2019-05-06 at 5 31 39 PM

src/reducers/clusterReducer.js
```diff
@@ -150,6 +150,50 @@ export default function clusterReducer(
...
+    case types.CLUSTER_LOAD_APPS:
+      items = Object.assign({}, state.items);
+
+      items[action.clusterId] = Object.assign({}, items[action.clusterId], {
+        isFetchingApps: true,
+      });
+
+      return {
+        lastUpdated: Date.now(),
+        isFetching: false,
```
pipo02mix: Is that right? Do we use two different flags for fetching?

src/components/cluster/detail/cluster_apps.js
pipo02mix: but how will it look like once we allow install apps but there is no app installed yet? If you have it under your radar and you will modify in the next PR I am fine

oponder: Yeah I will change the text as soon as it's possible in the next PR, and will be aware of that.

Just keeping it in a 'deployable' state.

src/reducers/clusterReducer.js
oponder: It's a bit confusing, but `isFetching` is on the collection of `clusters`. So that is true when we're doing `/v4/clusters/` and `isFetchingApps` is on a item in that collection (aka a cluster). So that is true/false depending on what cluster we're looking at, and is related to the `/v4/clusters/:clusterid/apps/` call.

PR #580, opened 2019-05-13
Install apps

oponder:
Work in progress, the main happy path is done.

Still need to:

- [x] Add clientside validation for name and namespace
- [x] Show loading state after clicking 'install app'
- [x] Handle errors if installing an app returns an error
- [x] Add lunr for searching through the clusters
- [x] Change the text in the managed apps tab, because now we _can_ install apps.
- [x] Preselect the clusterID if we arrived at the app catalog from a cluster detail screen.


oponder:
There's a few more features to do, but I'd like to get this reviewed and added to the codebase.

Here's the list, I'll make separate PR's for them probably instead of continuing to build on this one.

- [ ] Show app status on the cluster detail page. (pending, installed, error)
- [ ] Double check the max length for the name. 253 is maybe too long for helm.
- [ ] Preselect the Apps tab when going to the cluster detail page after installing an app.
- [ ] Show images for the apps on the cluster detail page.
- [ ] Show what clusters apps are installed on on the app detail page as well.
- [ ] Add a way to upload the values configmap during app installation.

src/components/cluster/detail/cluster_apps.js
```diff
@@ -158,6 +159,12 @@ class ClusterApps extends React.Component {
     });
   };
 
+  openAppCatalog = () => {
+    console.log(this.props.clusterId);
```
pipo02mix: is this needed anymore?

PR #582, opened 2019-05-14
Attempt fetching an index.yaml using cors anywhere if it fails.

oponder:
This is a lil workaround while we do not cache and serve the index.yamls ourself.

fyi @puja108

### DIFF
--- a/src/actions/catalogActions.js
+++ b/src/actions/catalogActions.js
@@ -5,8 +5,20 @@ import yaml from 'js-yaml';
 // loadCatalog takes a catalog object and tries to load further data.
 function loadCatalogIndex(catalog) {
   return fetch(catalog.spec.storage.URL + 'index.yaml', { mode: 'cors' })
+    .catch(() => {
+      console.log(`Fetch error for ${catalog.spec.storage.URL}, attempting with cors anywhere.`);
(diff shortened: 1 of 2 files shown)


src/components/app_catalog/catalogs/index.js
```diff
@@ -54,12 +54,6 @@ class Catalogs extends React.Component {
               </div>
             </React.Fragment>
           )}
-          <p className='well'>
```
oponder: This is not on topic for this PR sorry. It's something I forgot to remove before enabling app installation.

src/actions/catalogActions.js
```diff
@@ -6,10 +6,19 @@ import yaml from 'js-yaml';
 function loadCatalogIndex(catalog) {
   return fetch(catalog.spec.storage.URL + 'index.yaml', { mode: 'cors' })
     .catch(() => {
-      console.log(`Fetch error for ${catalog.spec.storage.URL}, attempting with cors anywhere.`);
-      return fetch('https://cors-anywhere.herokuapp.com/' + catalog.spec.storage.URL + 'index.yaml', { mode: 'cors' });
+      console.log(
+        `Fetch error for ${
```
pipo02mix: console log or console error?

oponder: At this point, its not an error yet. We are handling an error by trying an alternative method.

Just wanted to give some sign that we are in this path, since it won't be clear otherwise.

PR #589, opened 2019-06-04
Add a way to upload values.yaml during app installation.

oponder:
This adds a way for users to upload a `values.yaml` file during app creation which will get correctly associated with the app they are creation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "connected-react-router": "6.4.0",
     "copy-to-clipboard": "3.2.0",
     "deep-diff": "1.0.2",
-    "giantswarm-v4": "git+https://github.com/giantswarm/giantswarm-js-client.git#master",
+    "giantswarm-v4": "git+https://github.com/giantswarm/giantswarm-js-client.git#app-config",
     "history": "^4.7.2",
     "immutable": "3.8.2",
     "js-base64": "2.5.1",
--- a/src/actions/clusterActions.js
+++ b/src/actions/clusterActions.js
@@ -123,7 +123,7 @@ export function clusterInstallApp(app, clusterID) {
           .catch(error => {
             if (error.status === 409) {
               new FlashMessage(
-                `The user configuration configmap for <code>${app.name}</code> already exists on cluster <code>${clusterID}</code>`,
+                `The user configuration ConfigMap for <code>${app.name}</code> already exists on cluster <code>${clusterID}</code>`,
                 messageType.ERROR,
                 messageTTL.LONG
               );
--- a/src/actions/clusterActions.js
+++ b/src/actions/clusterActions.js
@@ -82,6 +82,11 @@ export function clusterLoadApps(clusterId) {
  * Takes an app and a cluster id and tries to install it. Dispatches CLUSTER_INSTALL_APP_SUCCESS
  * on success or CLUSTER_INSTALL_APP_ERROR on error.
  *
+ * If the app has valuesYAML set to a non empty object, then
+ * this will first try to create the user config configmap
+ * before creating the app, and it will make sure to fill in the
+ * user_config configmap fields correctly when making the app.
+ *
  * @param {Object} app App definition object.
  * @param {Object} clusterID Where to install the app.
  */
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -37,7 +37,7 @@
     "connected-react-router": "6.4.0",
     "copy-to-clipboard": "3.2.0",
     "deep-diff": "1.0.2",
-    "giantswarm-v4": "git+https://github.com/giantswarm/giantswarm-js-client.git#master",
+    "giantswarm-v4": "git+https://github.com/giantswarm/giantswarm-js-client.git#app-config",
```
oponder: TODO: change this back to master when api-spec and giantswarm-js-client are settled.

oponder:
Deployed to ginger. 

In the interface, here's what that looks like:

Screenshot 2019-06-17 at 10 29 15 PM

There is some error handling, so if what is uploaded doesn't parse as valid YAML the install button gets disabled and an error state is shown.

I think the description and label might need a bit more love though.

src/actions/clusterActions.js
```diff
@@ -98,16 +103,90 @@ export function clusterInstallApp(app, clusterID) {
...
+            // the app with the configmap.
+            resolve({
+              name: app.name + '-user-values',
+              namespace: clusterID,
+            });
+          })
+          .catch(error => {
+            if (error.status === 409) {
+              new FlashMessage(
+                `The user configuration configmap for <code>${app.name}</code> already exists on cluster <code>${clusterID}</code>`,
```
marians:
```suggestion
                `The user configuration ConfigMap for <code>${app.name}</code> already exists on cluster <code>${clusterID}</code>`,
```

src/actions/clusterActions.js
```diff
@@ -98,16 +103,90 @@ export function clusterInstallApp(app, clusterID) {
...
+                messageType.ERROR,
+                messageTTL.LONG
+              );
+            } else if (error.status === 400) {
+              new FlashMessage(
+                `Your user configuration configmap appears to be invalid. Please make sure all fields are filled in correctly.`,
```
marians:
```suggestion
                `Your user configuration ConfigMap appears to be invalid. Please make sure all fields are filled in correctly.`,
```

src/actions/clusterActions.js
```diff
@@ -98,16 +103,90 @@ export function clusterInstallApp(app, clusterID) {
...
+                messageType.ERROR,
+                messageTTL.LONG
+              );
+            } else {
+              new FlashMessage(
+                `Something went wrong while trying to create the user configuration configmap. Please try again later or contact support: support@giantswarm.io`,
```
marians:
```suggestion
                `Something went wrong while trying to create the user configuration ConfigMap. Please try again later or contact support: support@giantswarm.io`,
```

src/components/app_catalog/detail/install_app_form.js
```diff
@@ -36,6 +41,15 @@ const InstallAppForm = props => {
         hint={<React.Fragment>&nbsp;</React.Fragment>}
         value={props.namespace}
       />
+
+      <FileInput
+        label='User Configuration Configmap:'
```
marians:
```suggestion
        label='User Configuration ConfigMap:'
```

marians:
A question:

image

The text makes gives me the impression that the `values.yaml` is optional. How do I know that the app I am installing needs one? And what If I don't upload one right away, can I do it later, and how?

oponder:
> How do I know that the app I am installing needs one? And what If I don't upload one right away, can I do it later, and how?

@marians Good question, thats something we need to think about still. Editing apps or configuration is not something we wanted to build right away. Maybe something to discuss in sig-ux, let me also check the priorities with team-batman.

puja108:
> > How do I know that the app I am installing needs one? And what If I don't upload one right away, can I do it later, and how?
> 
> @marians Good question, thats something we need to think about still. Editing apps or configuration is not something we wanted to build right away. Maybe something to discuss in sig-ux, let me also check the priorities with team-batman.

There's 2 issues here and both are important to work on soon:
1. some apps (like Instana) require certain fields (but helm does not have that functionality so we need to have that for ourselves somehow)
2. Editing apps/config after install (this is even more important than 1 as it's kinda basic functionality)

pipo02mix:
> 1. some apps (like Instana) require certain fields (but helm does not have that functionality so we need to have that for ourselves somehow)

I guess this can temporarily fixed offering an installation guide for the managed app. Can we take it from the chart itself? like `info.md`?


> 2. Editing apps/config after install (this is even more important than 1 as it's kinda basic functionality)

This mean offer a way to interact with the existing `values.yaml` kind of editor, right?

PR #590, opened 2019-06-04
Show duration statistics on cluster creation page

marians:
Towards https://github.com/giantswarm/giantswarm/issues/6005

This adds some info regarding the actual cluster creation duration to the creation page. The info is coming from the statistics provided via the API's info endpoint.

### Preview

image

### TODO

- [x] revert JS client branch to master before merging

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "connected-react-router": "6.4.0",
     "copy-to-clipboard": "3.2.0",
     "deep-diff": "1.0.2",
-    "giantswarm-v4": "git+https://github.com/giantswarm/giantswarm-js-client.git#master",
+    "giantswarm-v4": "git+https://github.com/giantswarm/giantswarm-js-client.git#add-info-stats",
     "history": "^4.7.2",
     "immutable": "3.8.2",
     "js-base64": "2.5.1",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -37,7 +37,7 @@
     "connected-react-router": "6.4.0",
     "copy-to-clipboard": "3.2.0",
     "deep-diff": "1.0.2",
-    "giantswarm-v4": "git+https://github.com/giantswarm/giantswarm-js-client.git#master",
+    "giantswarm-v4": "git+https://github.com/giantswarm/giantswarm-js-client.git#add-info-stats",
```
pipo02mix: Is intended?

src/components/cluster/new/cluster_creation_duration.js
```diff
@@ -0,0 +1,29 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+class ClusterCreationDuration extends React.Component {
+  render() {
+    let minutes = 0;
+
+    if (this.props.stats.p75) {
```
pipo02mix: what does `p75` stands for?

marians: The `p75` attribute is the  75th percentile of a value distribution. See https://docs.giantswarm.io/api/#operation/getInfo

image

package.json
marians: This will be changed before moerging. See TODO list in the main comment.

PR #591, opened 2019-06-05
Prevent Happa from being shown in an iframe.

oponder:


oponder:
I also removed the nginx.conf as a configmap, which caused a bit of confusion, since we still had the old nginx.conf (since the plain old container that we run our tests against during the CI needed it).

The nginx.conf as a configmap is not needed anymore, since we don't need any helm templating in there as of: https://github.com/giantswarm/happa/pull/304/files

helm/happa-chart/templates/ingress.yaml
```diff
@@ -13,6 +13,8 @@ metadata:
     {{- if and (eq .Values.Installation.V1.Provider.Kind "aws") (.Values.Installation.V1.Security.RestrictAccess.GSAPI) }}
     nginx.ingress.kubernetes.io/whitelist-source-range: "{{ .Values.Installation.V1.Security.Subnet.VPN }},{{ .Values.Installation.V1.Security.Subnet.Customer }}"
     {{- end }}
+    nginx.ingress.kubernetes.io/configuration-snippet: |
+      add_header X-Frame-Options deny always;
```
oponder: This is the thing that really matters.

The rest is a bit of cleanup to hopefully make things less confusing next time around.

oponder:
Screenshot 2019-06-11 at 8 11 07 PM

marians:
A bit of a description would have been nice 🌷

marians:
Actually found the description in https://github.com/giantswarm/happa/pull/591#issuecomment-500426153

PR #593, opened 2019-06-11
Apply prettier 1.18

marians:
The changes in this PR are only made by executing

    prettier --write "src/**/*.js"

using prettier version 1.18.0


oponder:
Nice, I find this makes the code more readable personally.

PR #596, opened 2019-06-12
Simple grid to flexbox

ograu:
@oponder this is WIP

### DIFF
--- a/src/styles/vendor/_simple-grid.scss
+++ b/src/styles/vendor/_simple-grid.scss
@@ -100,7 +100,7 @@ li {
 // grid
 
 $width: 100%;
-$gutter: 4%;
+$half-gutter: 2%; // we are using space-between in the containers
 $breakpoint-small: 540px;
 $breakpoint-med: 825px;
 $breakpoint-large: 1200px;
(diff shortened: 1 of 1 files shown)


ograu:
@oponder I think this could possibly be the simpler, and most flexible, way of setting up a theme with emotion

oponder:
@ograu ok! interesting! One thing I don't like too much is that there has to be all these extra lines of "boilerplate". Is there some way we can do the 'withTheme' around only the ForgotPasswordStyles component?

ograu:
@oponder sure, done! :-)

ograu:
@oponder if the goal is to make reusable components to use later in our UI library, maybe would be better that ForgotPasswordStyles was ForgotPassword, .e.g a normal actual component with styles embedded. Just thinking

oponder:
Yeah the naming, i'd like to avoid calling it "SomethingSomethingStyles", so the idea woudl be to find a more semantic name. But for this refactoring, I kinda understand and it makes sense.

What I was thinking though was to move the `withTheme` further up to where it is really (potentially) being used.

So is there some way to get the 'styled' object, which we'd keep near the top of the file, wrapped with the `withTheme` instead of having to remember to do that at the end where we export it.

Also then we can keep the style props out of the propTypes.

Does that make sense?

oponder:
@ograu I like this direction regarding the theming! Thanks for your work here.

The PR itself is a bit unaligned now though. Totally fine, since this was about showing me your work and your progress there.

Since communication matters here, what changes might you make now to make sure that people reviewing it understand the intent of these changes?

Is the grid to flexbox a separate concern from the introduction of the theming?

Can we delete any css now that we have moved styles into the component?

ograu:
@oponder yes, I think I have mixed two related but different concerns here in this PR...

In order to clarify this, maybe I could change the name of the PR to something like "Simple grid to flexbox & Theming idea" or something like that. Or maybe you prefer to make two separate PRs, one with the grid replacement, and the other one with this theming proposal. ¿What do you think?

And yes, there are some styles to delete in the sass files.

ograu:
Most of the things we discussed here are in this PR: https://github.com/giantswarm/happa/pull/597

oponder:
Closing since we're migrating things in a way where we won't use these classes anymore anyways.

PR #597, opened 2019-06-14
Theming proposal

ograu:
@oponder I haven't deleted styles regarding this login box in the sass file because I've seen they are used in several components. I can do it as soon as we decide that the approach that here I am proposing makes sense to you.

Also, and regarding the UI library / reusability of components, I have made another change: it will be an interesting thing  - as we discussed two days ago - to have those styles embedded in a regular React component. So I went ahead and have created a _UI_ folder inside components for storing exclusively this style-related components (I've seen the _shared_ one, but I think we have different types of components there).

I'm using a props.children approach now, but we can also use render props if children is not enough for the functionality/es we want to achieve in each case.


oponder:
I like this approach a lot, and it gives us a clear path towards getting these components also into Framer X. 

So I see us taking each component and step by step extracting just the UI from them into these UI components. 

Can you update the branch to make sure it's up to date with master? Other than that it looks good to me.

One thing about this refactoring though is that the UI component is just a wrapper, that is not yet _that_ much better than having to go between styles and the react component to understand how things work. 

The end goal for me is to have these UI components contain more of the actual UI elements, and their styles, and their behaviour related to the different visual states they can be in all in one place. And this behaviour would be controlled by props in most cases. What do you think?

ograu:
I totally agree and I think that we could do it in two steps:

1. Extract just markup & styles, and once we have most of the UI elements as independent components
2. Add behaviour to each component

I just think this would be a nice way of avoiding premature abstractions on the behaviour side.

PR #598, opened 2019-06-25
Add serviceaccount, psp and supporting rbac

glitchcrab:
Towards giantswarm/giantswarm/issues/6013

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,6 +15,6 @@ ADD scripts/start.sh /
 # Test certifiates will be overwritten in production by configmap
 ADD test/certs /etc/nginx/certs
 
-RUN chown -R nginx.nginx /www
+RUN chown -R nginx:nginx /www
 
 CMD ["/start.sh"]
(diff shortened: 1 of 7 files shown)


glitchcrab:
Kinda expected that Happa would be more work:

```
--- Starting Happa nginx server ---
2019/06/25 08:03:50 [emerg] 1#1: mkdir() "/var/cache/nginx/client_temp" failed (13: Permission denied)
nginx: [emerg] mkdir() "/var/cache/nginx/client_temp" failed (13: Permission denied)
```

Dockerfile
```diff
@@ -15,4 +15,6 @@ ADD scripts/start.sh /
 # Test certifiates will be overwritten in production by configmap
 ADD test/certs /etc/nginx/certs
 
+RUN chown -R nginx.nginx /www
```
marians: @glitchcrab Is it really `nginx.nginx`, not `nginx:nginx`?

glitchcrab: Either way works, but I guess most people are used to the latter so I'll change it. This Dockerfile requires more work to run properly anyway, it's a bit of a WIP.

glitchcrab:
> Looks like it works!

Thanks for the help Oli!

PR #601, opened 2019-06-26
Ensure sorting of props of JSX components

marians:
This PR enforces alphabetic sorting of JSX props.

### DIFF
--- a/src/components/app.js
+++ b/src/components/app.js
@@ -39,15 +39,15 @@ render(
     <ConnectedRouter history={history}>
       <div>
         <Switch>
-          <Route path='/admin-login' component={AdminLogin} />
-          <Route path='/login' component={Login} />
-          <Route path='/logout' component={Logout} />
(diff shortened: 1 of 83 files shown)


marians:
Tested on ginger

src/components/app.js
```diff
@@ -39,15 +39,15 @@ render(
...
-          <Route path='/admin-login' component={AdminLogin} />
-          <Route path='/login' component={Login} />
-          <Route path='/logout' component={Logout} />
-          <Route path='/forgot_password/:token/' component={SetPassword} />
-          <Route path='/forgot_password' component={ForgotPassword} />
-          <Route path='/signup/:token' component={SignUp} />
-          <Route path='/oauth/callback' component={OAuthCallback} />
-
-          <Route path='/' component={Layout} />
+          <Route component={AdminLogin} path='/admin-login' />
```
oponder: Here's one case where I find the non alphabetical ordering more meaningful from a code reading point of view.

But I can live with it.

PR #602, opened 2019-06-26
Validate and run prettier in the same npm script

ograu:
So you don't have to worry about it

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "test:watch": "jest --watch",
     "start": "grunt serve",
     "prettier": "prettier --write \"src/**/*.js\"",
-    "validate-prettier": "prettier --version; prettier --no-color --check \"src/**/*.js\""
+    "validate-prettier": "npm run prettier && prettier --version; prettier --no-color --check \"src/**/*.js\""
   },
   "prettier": {
     "singleQuote": true,


package.json
```diff
@@ -11,7 +11,7 @@
     "test:watch": "jest --watch",
     "start": "grunt serve",
     "prettier": "prettier --write \"src/**/*.js\"",
-    "validate-prettier": "prettier --version; prettier --no-color --check \"src/**/*.js\""
+    "validate-prettier": "npm run prettier && prettier --version; prettier --no-color --check \"src/**/*.js\""
```
marians: Shouldn't that be consistent?

```suggestion
    "validate-prettier": "npm run prettier && npm run prettier --version; npm run prettier --no-color --check \"src/**/*.js\""
```

oponder:
Thx @ograu!  

However there is a `prettier` command already above there. So if you just want to run prettier you can do `npm run prettier`

So this `validate-prettier` is also for the CircleCI step. The idea is that I'd rather not have CircleCI automatically format code, I only want it to fail if the formatting hasn't been done.

So unless I misunderstood, I don't think this change is needed.

oponder:
@ograu I think we can close this one right?

ograu:
Yes, we can close it

marians:
@ograu It's your PR, so you are normally responsible for merging, closing, deleting the branch etc.

PR #606, opened 2019-06-27
App catalog list ui

ograu:


src/components/UI/app_container.js
```diff
@@ -0,0 +1,153 @@
...
+      iconErrors,
+      imgError,
+    } = this.props;
+    const { icon, name, repoName, version } = appVersions[0];
+    const to = `/apps/${catalog.metadata.name}/${appVersions[0].name}?q=${searchQuery}`;
+
+    return (
+      <Wrapper>
+        <StyledLink to={to}>
+          {repoName === 'managed' && <Badge>MANAGED</Badge> // This is always false. Is this ok?
```
oponder: for now it's based on the name of the catalog, but maybe in the near future the logic for determining this will change. for now its ok

oponder: I would like to avoid question comments such as this in the codebase though.

Although a comment explaining why something is the way it is for the future reader is fine.

src/components/UI/app_container.js
```diff
@@ -0,0 +1,153 @@
...
+  width: 'calc(25% - 15px)',
+  flexShrink: '0',
+  margin: '0 20px 20px 0',
+  '&:nth-of-type(4n + 0)': {
+    marginRight: '0px',
+  },
+
+  [`@media only screen and (max-width: ${props.theme.breakpoints.large})`]: {
+    width: 'calc(33.333% - 13.333px)',
+    // we are unsetting the 4n + 0 rule one above
```
oponder: It's a bit nit picky, and I used to not care about this, but I'd like join the Golang crew in the comment style, which is full sentence. So starting with a capital letter and ending with a dot.

```suggestion
    // We are unsetting the 4n + 0 rule one above.
```

src/components/UI/app_container.js
```diff
@@ -0,0 +1,153 @@
...
+  borderRadius: '4px',
+  height: '158px',
+  backgroundColor: '#2f556a',
+  position: 'relative',
+  cursor: 'pointer',
+  overflow: 'hidden',
+  '&:hover': {
+    zIndex: 10,
+    boxShadow: '0px 0px 5px #549ac3',
+    // [`${AppIcon}`]: {
```
oponder: I guess the latest changes aren't in yet?

ograu: I couldn't find the way, maybe I should write a todo and an issue?

oponder: If this way is like a workaround, and it works, not sure if we need a todo.

Maybe we'll figure out the composed form later, but if the PR works as intended, then its ok to leave out the comment and todo.

PR #609, opened 2019-06-27
Enforce button type

marians:
According to what I read, not having a type attribute on a button can be "bad".


ograu:
Submit is default type if you dont especify any type.

marians:
Right, that was it. And that's in general not what we want.

PR #610, opened 2019-06-27
AWSAccountID UI component

marians:


src/components/UI/theme.js
```diff
@@ -18,6 +18,9 @@ const theme = {
     med: '825px',
     large: '1200px',
   },
+  fontFamilies: {
+    console: 'Inconsolata, monospace',
+  },
```
marians: As I understand it, the naming here is up to us, right?

ograu: Yes, we can use the naming we want

marians:
@ograu I did that, didn't I? Have I forgotten anything? (See last file in https://github.com/giantswarm/happa/pull/610/files)

ograu:
You are so right! Sorry...

PR #612, opened 2019-06-27
ChangeEmailForm: remove constructor, rename state variables

marians:
Removes an unnecessary constructor, renames the state variables for better readability


ograu:
@marians One important thing I don't know if we have already dicussed: when we are not using the constructor, `this` is no longer binded in custom methods.

It doesn't apply in this specific case, because nothing was binded,  but we have to take it into account when doing this kind of refactors

marians:
Thanks! There is another PR coming where this might be of importance.

PR #614, opened 2019-06-28
Extract navigation into UI component

oponder:
It was a lot of sass to move, but it works. 

Making it a drat PR because I might want to further break this down into other subcomponents. 

@ograu @marians I'm thinking of making a folder in `UI` called `navigation`, and then have the subcomponents in there. Would that make sense?

There's some small color difference since I'm trying to stick to the color shades we're defining in the themes file, but that could be something we talk about together quickly.

This is also a nice example of removing redux and the connect from a component, and instead passing its behaviour in via a prop (see onselectorganisation). 

It used to be that Navigation knew how to dispatch actions, but UI components should not know how to do that.

At least, I think that makes most sense, curious to see if the added prop passing is more of a headache, or if it becomes a cleaner / clearer / less couple codebase overall.


marians:
We talked about sub-structure in the UI folder briefly yesterday. As soon as there are several components belonging together, that seems to make sense.

As this is still a draft, we can't actually review/approve.

ograu:
I think it makes a lot of sense, in the PR I am about to create I splitted one component into four, so I think it would be nice to create a subfolder in UI

PR #615, opened 2019-06-28
Deleted class .loader-button--status-indicator

ograu:


marians:
```js
const ProgressButtonStatusIndicator = styled.div({
  display: 'inline',
});
```

Can't we achieve the same behaviour by simply returning a `<span>`?

ograu:
@marians  yes, we could do it, but it feels weird to me, look what if found in [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/span): _The HTML <span> element is a generic inline container for phrasing content_

Do you want me to change it?

marians:
The explanation in https://developer.mozilla.org/en-US/docs/Web/HTML/Element/span represents exactly what I mean: `span` is a generic inline element (as `div` is a generic block element). So when we want something to render as `display: inline`, I am in favour of simply using a span over a div.

ograu:
@marian sorry, I am concerned about the semantics here, maybe I wasn't clear enough and should have stressed the expression **phrasing content**

We can definetily use `span` and it would probably work, and we would then have the benefit of not having to write this `display: inline`, but I think this is not a good practice. Please take a look at the w3 specs [here](https://www.w3.org/TR/html50/text-level-semantics.html#the-span-element)

I think that we should use html elements for what they are meant for because otherwise we can experience side effects: developers who build other technologies, such as screen readers for exemple, follow w3 recommendations when developing, so they might process `span`s assuming that they contain text and can lead our app to be misinterpreted by their software. 

This would be an edge case and I am being nit picky again... so, we can probaby put a span in place of this div and everything should be fine if you think that we should.

marians:
I read up on" phrasing content" now (learned something new!) and I think I get what you mean. Seems to make sense. So you want to use an element that is purely flow, correct? OK, then let's settle for `div`.

PR #616, opened 2019-06-28
App catalog detail

ograu:
I have splitted App catalog details index.js into four components in an attemp of simplify things - not sure if I really have simplified it. I have created a subfolder in UI with the components related to app_details.

Naming (variables and files) : not sure if I am doing a good job...

Semantic HTML: I have replaced `uls` for `dls` because I think that it is nice to have a more semantic markup, specially because of the a11y implications it has. ¿Am I complicating things too much? WDYT?

Emotion: also, ¿Do you fell comfortable using javascript objects for styles? I don't see the benefits it could have. I know there is one library for theming ([Facepaint](https://github.com/emotion-js/facepaint)?) that you can only use if you are using objects for styles. Drawbacks:

- We have to modifiy all styles when passing css to component
- I think that this is not going to be the "winning" syntax for obvious reasons, with its implications regarding documentation and community help


src/components/UI/app_details/index.js
```diff
@@ -0,0 +1,166 @@
...
+          </div>
+
+          <div className='version'>
+            <small>Chart&nbsp;Version</small>&nbsp;
+            <code>{version}</code> <small>App&nbsp;Version</small>&nbsp;
+            <code>{appVersion}</code>
+          </div>
+        </Title>
+
+        <Install>{children}</Install>
```
oponder: I think when we extract the install app modal, then we should place it here, instead of using this 'children'.

Or, it would be ok to use the {children} idea, but then not have styled `Install` component. Idea is that we don't want something outside the `InstallAppModal` affecting it, would make it harder to reason about.

But for now it makes sense.

oponder:
> Do you fell comfortable using javascript objects for styles?

I personally also prefer the css syntax.

src/components/UI/app_details/index.js
ograu: @oponder true, I was so focused in the details part that I kind of forgot to do it. `children` was meant just as a momentary solution. If we merge now, then we will have to do it anyway, so I will do it now (in a while) before merging, ok?

src/components/UI/app_details/body.js
```diff
@@ -0,0 +1,45 @@
...
+
+const DescriptionList = styled.dl({
+  margin: 0,
+
+  dd: {
+    margin: 0,
+    ul: {
+      marginBottom: 15,
+      padding: 0,
+      listStyleType: 'none',
```
marians: According to what we discussed on Slack, this would have to be `marginBottom: '15px'` etc., no?

marians:
> Semantic HTML: I have replaced uls for dls because I think that it is nice to have a more semantic markup, specially because of the a11y implications it has. ¿Am I complicating things too much? WDYT?

Totally fine with me

src/components/UI/app_details/body.js
ograu: Didn't know we agreed on that. Will change it

src/components/UI/app_details/index.js
ograu: @oponder  It involves more work than I thought, I'll merge now and continue working next week

PR #617, opened 2019-06-28
.eslintrc minor mods

ograu:
I have changed no-unused-vars error level from error to warning because when debugging I like to comment some parts of the code.

I also have sorted rules and favoured numbers instead of strings when defining levels of error.


ograu:
Oh, and also I see that my editor has formatted the code, this wasn't my intention I am changing it

PR #618, opened 2019-07-01
Use string notation for emotion/styled

marians:
Changing the notation from JS objects to CSS string.


marians:
@ograu @oponder This is not working currently. Can you spot the mistake I'm making?

marians:
Oh, I found an example here. https://emotion.sh/docs/keyframes#___gatsby

marians:
Solved it. Will merge into hackathon-ui right away.

PR #600, opened 2019-06-26
Hackathon Topic: UI Library

oponder:

PR #619, opened 2019-07-01
Add eslint emotion plugin

marians:
To merged into #600 

Adds checks regarding eslint (no CSS syntax checks though)

`emotion/syntax-preference` is about the string syntax being preferred and yields a warning of object syntax is used.

The other rules are checking imports.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -14,6 +14,11 @@
     "es6": true
   },
   "rules": {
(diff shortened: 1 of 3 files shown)


marians:
See https://github.com/emotion-js/emotion/tree/master/packages/eslint-plugin-emotion for details

PR #620, opened 2019-07-01
Tests setup

ograu:
Tests setup and converted all tests to React Testing Library.

### DIFF
--- a/src/components/cluster/detail/__tests__/tabs.js
+++ b/src/components/cluster/detail/__tests__/tabs.js
@@ -28,4 +28,10 @@ it('renders only the first tab', () => {
   );
 
   expect(container.querySelector('#tabs-pane-1')).toBeVisible();
+  expect(container.querySelector('#tabs-pane-2')).toHaveStyle(`
+    display: none;
+  `);
+  expect(container.querySelector('#tabs-pane-3')).toHaveStyle(`
+    display: none;
+  `);
 });
--- a/src/components/cluster/detail/__tests__/tabs.js
+++ b/src/components/cluster/detail/__tests__/tabs.js
@@ -12,7 +12,7 @@ it('renders without crashing', () => {
 
 it('renders only the first tab', () => {
   const div = document.createElement('div');
-  const { container } = render(
+  const { container, debug } = render(
     <Tabs>
       <Tab eventKey={1} title='first'>
         <h1>First Tab</h1>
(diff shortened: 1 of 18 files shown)


ograu:
@oponder CircleCI is crashing when running the tests :-(
Not sure why it is happening..., CircleCI is complaining about Node version:

`warning You are using Node "6.1.0" which is not supported and may encounter bugs or unexpected behavior. Yarn supports the following semver range: "^4.8.0 || ^5.7.0 || ^6.2.2 || >=8.0.0"`

src/components/cluster/detail/__tests__/tabs.js
```diff
@@ -22,12 +23,9 @@ it('renders only the first tab', () => {
...
         <h1>Third Tab</h1>
       </Tab>
-    </Tabs>
+    </Tabs>,
+    div
   );
 
-  expect(tabs.find('#tabs-pane-1').hasClass('active')).toEqual(true);
-
-  expect(tabs.find('#tabs-pane-2').hasClass('active')).toEqual(false);
```
oponder: I think part of it is also making sure the other panes are not visible. Not sure if now this test still ensures that.

src/components/cluster/detail/cluster_apps.js
```diff
@@ -168,20 +168,28 @@ class ClusterApps extends React.Component {
     return (
       <React.Fragment>
         {this.props.showInstalledAppsBlock && (
-          <div id='installed-apps-section'>
+          <div data-testid='installed-apps-section' id='installed-apps-section'>
```
oponder: Is there a way around having this `data-testid?` at first glance, not really liking that we have id and data-testid.

oponder:
I managed to get it running, see: https://github.com/giantswarm/happa/pull/621

If you approve that, I'll merge that into this branch and then I'll approve this one :) 

I just went with the latest version of Node, not sure what the advantage / disadvantage is, but if we can be as up to date as possible, I'd prefer that.

src/components/cluster/detail/__tests__/tabs.js
ograu: True, I assumed incorrectly that they were `position: absolute` and that if the first one was visible, the others will remain not visible. I am correcting it

ograu:
@marians  @oponder The problem in __tests__/tabs.js  is that we are not getting bootstrap styles into the tests. I think there must be a way to inject styles in the test render but for now I will not invest more time on finding how to test bootstrap styles because:

1. It is something that react doesn't control
2. We are moving to emotion

Please LMK if you think otherwise

oponder:
> So we will keep testing classes in those cases for now.
That's alright!

PR #621, opened 2019-07-02
Try to use node 10.16.0 in Circle

oponder:
This is me trying something based on ograu's test setup branch https://github.com/giantswarm/happa/pull/620

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,6 +32,19 @@ jobs:
     steps:
     - checkout
 
+    - run:
+        name: Install node@10.16.0
+        command: |
+          set +e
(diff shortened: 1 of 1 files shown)


oponder:
@ograu this seems to make the tests work.

I do notice there is an opportunity to streamline the CircleCI file. At some point I was not able to get tests running using the `happa-build` docker container myself and decided to run `yarn test` against the CircleCI machine executor itself.

But we still use `happa-build` to install dependencies and some other things. So it's a bit of a mishmash of things again.

I'm going to see if I can get everything running from `happa-build`, but that not with high priority now.

PR #622, opened 2019-07-02
Create component for availability zones label

marians:
This component should be re-usable to display which AZs a cluster or node pool uses.

## Preview

image

The full zone name, e. g. `eu-central-1b`, is shown on mouse over.


oponder:
Shouldn't this be in the UI folder :) ?

marians:
Yes, indeed. I forgot to move.

oponder:
I just noticed that `abbr` gets `cursor: help` from bootstrap applied to it. Which shows the cursor as a 'question mark' when hovering over these Availability Zones. Did you notice that? Is that something you want to keep?

marians:
@oponder I was aware and I didn't mind.

PR #625, opened 2019-07-02
Move NumberPicker to UI components, update styles

marians:
This converts the NumberPicker to use CSS in JS.

Also found and removed an unused component `NewKVMWorker` and a number of unused CSS classes.


marians:
Re-ping: Still needs a review

oponder:
Did you figure out what was causing the flashing?

PR #624, opened 2019-07-02
Change style notation to string

marians:
As we agreed on using the string notification for styles, this PR adapts some styles.

PR #626, opened 2019-07-02
Start adding stylelint

oponder:
It kinda works, you can see it run with `yarn run lint:css`. Not yet hooked up to circle.

There are a couple of weird things to figure out still.

1. According to the docs, it's not supposed to need `"processors": ["stylelint-processor-styled-components"],` in .stylelintrc. However I couldn't get it to actually lint anything without it.

2. I had to add `"no-descending-specificity": null`, otherwise it would raise some false positives for certain files that have more than one styled component in them.

3. Lastly, the styled component has to look like this:

```
const Wrapper = styled.span`
  font-family: ${props => props.theme.fontFamilies.console};
`
```

and not:

```
const Wrapper = styled.span(
  props => `
  font-family: ${props.theme.fontFamilies.console};
`
);
```

otherwise it won't pick it up for linting.


marians:
The style change mentioned in (3) is covered in this PR: https://github.com/giantswarm/happa/pull/624

PR #627, opened 2019-07-03
Fix handing of undefined/null prop in AvailabilityZonesLabel

marians:


oponder:
@oriol is there anything from your experience any best practices on how to deal better with these kind of `undefined` errors?

I myself am prone to making this kind of error quite often. 

I'm thinking typescript would help here, but I wonder if there is anything else you might know.

ograu:
@oponder @marians  `undefined` is a type in JS, and the way I deal with this type of comparisons is 
```
!zones || zones.length === 0
```
instead of
```
(typeof zones === 'undefined' || zones.length == 0)
```
By the way, it is usually considered a bad practice in JSland to use `==`, because with this: `'0' == 0` is `ture`

ograu:
But if we are using proptypes, we should not be using the `||` here...

oponder:
Let's see if there is some more linting we can turn on to help us with those things (like the `==`)

PR #629, opened 2019-07-04
Remove grunt

ograu:
@oponder @marians I was working yesterday on this PR. Stuff done:

- I have removed Grunt, this branch is using Webpack

- I have added this preset [@babel/preset-react](https://babeljs.io/docs/en/babel-preset-react) which is similar to what we had but with some additions (check the link for further details). Now we can do things like using the short syntax for `<React.Fragment>` : `<>` (not sure why we weren't having this, though, as the transform-react-jsx was plugin supposed to do this)

- new `webpack.NoEmitOnErrorsPlugin({ debug: true })`, removed because `hotOnly` is supposed to do the same job

- `use strict` removed as my ESLint says it is not required. (¿Do we really need this?)

- We now have this warning in the production build:

```bash
Warning

WARNING in asset size limit: The following asset(s) exceed the recommended size limit (244 KiB).
This can impact web performance.
Assets: 
  app.js (1.86 MiB)

WARNING in entrypoint size limit: The following entrypoint(s) combined asset size exceeds the recommended limit (244 KiB). This can impact web performance.
Entrypoints:
  main (2.01 MiB)
      main.css
      app.js


WARNING in webpack performance recommendations: 
You can limit the size of your bundles by using import() or require.ensure to lazy load some parts of your application.
For more info visit https://webpack.js.org/guides/code-splitting/
```

So there's a little bit of work still to do for improving performance: code splitting, dynamic imports, reducing bundle, etc.

I would like to test the production branch in any stagging/testing server we may have if that's possible


ograu:
Let's see if the modifications I did in the `make dist` are correct...

ograu:
@oponder I see that this is failing. Can it be because because webpack is not in the path??

ograu:
ping @oponder now it's failing again in CircleCI...

oponder:
@ograu it looks like the container that it builds does not show 'Happa' anywhere when trying to visit it.

This is a little smoke test I implemented just to make sure the production container actually works.

Have to see what is going on in that container and see if it is really able to start up correctly and serve the files.

I'd probably re-run the task with SSH on and then go into the circleci machine to then inspect the docker container (using something like `docker exec -ti /bin/sh`) to see what is going on in there.

Alternatively you can try to build the production docker container locally.

`make dist`
`make docker-build-prod`

After that you can run the docker container you just built:

`docker run -p 8000:8000 quay.io/giantswarm/happa`

After which you can try to visit localhost:8000 and see what might be going on, or otherwise `docker exec` and see if all the files are in the right places.

Happy to help with that, but for today I am going to try to finish some app catalog related things and then call it a day.

Let's meet tomorrow?

ograu:
@oponder @marians think I solved it, we have now a bit of a mixed solution with images: some of them are inlined by webpack, and the others are referenced using relative paths (and `images` folder is copied to `dist/images` as Grunt did). 

My idea is to use dynamic imports and get rid of this copy-pasting thing, the benefits being:

1. It's easier to manage images (see images/index.js), and
2. We let webpack to handle the images, so we can do anything we want with them: compression, inlining, etc.

ograu:
branch deployed in https://happa.g8s.ginger.eu-central-1.aws.gigantic.io/

oponder:
Cool! looks like it's working well.

One thing I noticed is that we lost the cache-busting functionality that we had before. 

The way we used to do the caching was to tell the browser it was to cache anything in the assets folder, and then to always append a unique hash to the end of app.js, based on the contents of the file. 

That would make cache invalidation a bit easier to think about, since whenever there was a new release, index.html would force the browser to fetch a file it never saw before.

I bet that's something that webpack has a plugin for too right?

ograu:
@ponder good point! In the process of solving this issue in the prod build (or IDK when TBH) I removed the hash appended to the filename. 

Added in https://github.com/giantswarm/happa/pull/629/commits/936f2b3681dfe58cf08ff020e4d57128bd820a52

webpack.common.js
```diff
@@ -8,7 +8,7 @@ module.exports = {
   output: {
     publicPath: '/',
     path: path.resolve(__dirname, 'dist'),
-    filename: 'app.js',
+    filename: 'app.[hash].js',
   },
```
marians: too easy :)

oponder:
@ograu, btw @ ponder is also a different guy :P

webpack.common.js
ograu: have to check config options for this...

ograu:
OMG true! Typos...

webpack.common.js
marians: I was trying to be funny. If this is all it takes, very cool!

ograu: yeah, what I wanted to say is that this is something that we would probably need to change when doing code splitting, and TBH I haven't spend much time finding out exactly the options we had here. Just went for the straightforward solution, now.

marians:
Lighthouse before (master):

image

This branch:

image

Good job! 🎉

marians:
@ograu I just noticed that `main.css` does not have a content hash in the file name for cache busting. Can you change that?

ograu:
@marians We are not referencing main.css but inlining instead, ¿or maybe we are referencing it somewhere and I am not seeing it for any reason? 

And about the 404, I see that it is loading fine now... not sure why we had a 404

ograu:
@oponder builds are failing also in this PR, not sure what's going on: I saw that very same error in my mac and read somewhere that it is related to the way homebrew installs yarn (I solved it in my computer), but what I can't understand is why it's also happening in CircleCI ... ¿could it be something related to this latest happa-built?

oponder:
@ograu I'll be taking a look at this today to see what I can do!

Since I think it has to do with the Snyk branch I merged, as a last resort, we can revert that entirely.

oponder:
The weird thing is that the Snyk branch on it's own worked fine. So somehow something since then is not playing nicely with it. Or something changed in Circle. Not sure yet.

ograu:
@oponder this error in CircleCI is because we don't have these dependencies in the image? :    
```
    "webpack": "4.32.2",
    "webpack-cli": "^3.3.5",
    "webpack-dev-server": "3.4.1",
    "webpack-merge": "^4.2.1"
```

oponder:
@ograu I think it's because we're back at happa-build:latest, and we haven't merged this yet: https://github.com/giantswarm/happa-build/pull/3

ograu:
hey @marians ¿are you ok with my comment about the changes you requested?

PR #633, opened 2019-07-05
Create UI component for a releases' component

marians:
This provides a UI component for displaying the component of a release and the version changes, if any. That should simplify further work on https://github.com/giantswarm/giantswarm/issues/6400

## Preview

image


marians:
Currently deployed on ginger

src/components/UI/release_component_label.js
```diff
@@ -0,0 +1,113 @@
+import { css } from '@emotion/core';
+import PropTypes from 'prop-types';
+import React from 'react';
+import styled from '@emotion/styled';
+
+const radius = '5px';
```
oponder: could this come from the theme instead? https://github.com/giantswarm/happa/blob/master/src/components/UI/theme.js

marians: As there is no 5px radius used by any other component, I think no. Possibly in the future.

src/components/cluster/detail/upgrade_cluster_modal.js
```diff
@@ -46,12 +47,12 @@ class UpgradeClusterModal extends React.Component {
     }
 
     var components = {};
-    _.each(currentComponents, component => {
+    currentComponents.forEach(component => {
```
ograu: @marians love it!

src/components/cluster/detail/upgrade_cluster_modal.js
```diff
@@ -76,57 +77,33 @@ class UpgradeClusterModal extends React.Component {
...
-                    {component.name}
-                  </span>
-                  <span className='release-selector-modal--component--version'>
-                    <span className='lhs'>
-                      <span>{component.version}</span> <span>(removed)</span>
-                    </span>
-                  </span>
-                </div>
+                <ReleaseComponentLabel
+                  isRemoved={true}
```
ograu: you can also do it as `isRemoved` there is no need for `={true}`

marians: Very nice, thanks!

PR #634, opened 2019-07-05
Remove react-addons-test-utils

ograu:


ograu:
@oponder @marians the CircleCI build is failing because of this: 

````
npm run snyk-protect
npm WARN lifecycle The node binary used for scripts is /tmp/yarn--1562657010829-0.6371811391889479/node but npm is using /usr/local/bin/node itself. Use the `--scripts-prepend-node-path` option to include the path for the node binary npm was executed with.
```

¿Any idea of what's going on?

oponder:
Can you try updating it against master, I merged something related to Snyk recently.

ograu:
I am receiving the same warning when doing `yarn install` (I have switched from `remove-grunt` branch)...

ograu:
It didn't work, actually I merged `master` it before here in github.

@oponder ¿Should I modify the Makefile to make it work? ?What do you suggest?

oponder:
I'm not sure at the moment! Will have to do some digging. But if you think there is something in the makefile that will fix it, feel free to try and add that to the PR. 

I can try to take a look a bit later today though otherwise!

ograu:
@oponder I couldn`t manage to solve this, in https://github.com/giantswarm/happa/pull/634/commits/6d1d727cf0a9e7d404d36d618388d5685c2eecf2 the warning about the node binary disappears but it fails in the same job

marians:
Related to https://github.com/giantswarm/giantswarm/issues/5677

PR #636, opened 2019-07-05
Adjust catalog overview to allow for more text in the descriptions.

oponder:
We'd like to add some more text to the catalog descriptions, currently that doesn't work so good in the layout we have.

These are some quick changes for now.

Screenshot 2019-07-05 at 11 19 01 PM

### DIFF
--- a/src/components/app_catalog/catalogs/index.js
+++ b/src/components/app_catalog/catalogs/index.js
@@ -8,11 +8,8 @@ class Catalogs extends React.Component {
     return (
       <DocumentTitle title={`Apps | Giant Swarm `}>
         <React.Fragment>
-          <h1>Apps</h1>
-          <p>
-            Browse and deploy apps from various sources, including managed apps
-            covered by SLAs from Giant Swarm.
-          </p>
+          <h1>App Catalogs</h1>
+          <p>Pick an App Catalog to browse all the Apps in it.</p>
           {Object.keys(this.props.catalogs.items).length === 0 ? (
             <p className='well'>
               <b>Could not find any appcatalogs:</b>
@@ -33,21 +30,23 @@ class Catalogs extends React.Component {
                       to={'/apps/' + catalogName + '/'}
                     >
                       <div className='app-catalog--card'>
-                        <h3>
-                          {this.props.catalogs.items[catalogName].spec.title}
-                        </h3>
                         <img
                           src={
                             this.props.catalogs.items[catalogName].spec.logoURL
                           }
                         />
                       </div>
+                      <div className='app-catalog--description'>
+                      <h3>
+                          {this.props.catalogs.items[catalogName].spec.title}
+                        </h3>
                       <p>
                         {
                           this.props.catalogs.items[catalogName].spec
                             .description
                         }
                       </p>
+                      </div>
                     </Link>
                   );
                 })}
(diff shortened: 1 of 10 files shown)


oponder:
I also noticed that the top level "Apps" isn't really totally correct. I'm ok with that in the navigation, but really the first level of this flow is the 'App Catalogs' so I made a quick change for that as well.

Screenshot 2019-07-05 at 10 39 11 PM

marians:
With https://github.com/giantswarm/happa/pull/637 we now have `react-markdown`. What do you think about using that for displaying of the catalog description? With the descriptions as they are, it would result in the URL being linked. Plus it would allow to add formatting to descriptions, if this is what we want.

Regarding security, I think we should check whether we can completely ensure the descriptions are coming from trustworthy sources. Of not, react-markdown has options to disable embedded HTML, to prevent XSS.

oponder:
> I think we should check whether we can completely ensure the descriptions are coming from trustworthy sources.

We're making the descriptions, it's not coming from the index.yaml of the app catalog repo or something like that. 

So I think it's ok. The description is a field on the App Catalog CR, which we create. Even the App Catalog CR's that point to partner or customer repo's will probably be created by us, or by the customer themselves.

Disabling embedded HTML regardless might be a nice preventative measure.

oponder:
@marians ok, I added react-markdown! It looks to me like the default options are 'safe'. i.e. it disabled embedded html by default: https://github.com/rexxars/react-markdown#options

marians:
Will test now on ginger

src/styles/components/_app_catalog.sass
```diff
@@ -75,43 +75,49 @@
...
   p
     margin-bottom: 0px
-    padding: 10px 20px
     border-bottom-left-radius: 5px
     border-bottom-right-radius: 5px
+    font-size: 14px
 
   &:hover
     text-decoration: none
     box-shadow: 0px 0px 8px lighten(#1e4156, 22%);
```
marians:
```suggestion
    box-shadow: 0px 0px 8px lighten(#1e4156, 22%)
```

Unrelated fix

marians: Please double-check the white space here. Not sure whether github got this right.

src/components/app_catalog/catalogs/index.js
```diff
@@ -30,24 +28,28 @@ class Catalogs extends React.Component {
...
+                          {this.props.catalogs.items[catalogName].spec.title}
+                        </h3>
+                        <p>
+                          <ReactMarkdown>
+                            {
+                              this.props.catalogs.items[catalogName].spec
+                                .description
+                            }
+                          </ReactMarkdown>
+                        </p>
```
marians: This is resulting in nested `<p>` tags, as ReactMardown is wrapping every paragraph in `<p></p>`, too. You can remove the outer one.

oponder:
Adjusted line length:

Screenshot 2019-07-08 at 5 15 29 PM

marians:
FYI: I suggested `max-width` with unit `em`, as it is useful to let the line length depend on the font size. That's what the typographs from the old days taught me at least.

oponder:
Would be useful if the font size was changing while it was resizing, or if we supported some kind of font-size change feature.

The way browsers zoom now is also not solely by increasing the font size, so I think we're ok.

But happy to rethink the typography and units usage (em / px) throughout happa at some point.

PR #638, opened 2019-07-10
Add a app detail modal and a way to update existing configs

oponder:
This adds an app detail modal where users can see a bit more detail on the app installed on their cluster.

This same modal also lets users update the apps configuration by uploading a yaml file in much the same way that is done during app creation.


oponder:
Work in progress, hope to discuss the UX details during our UX working hours.

My rough wireframe concept shows the two states that this modal can be in

On the left, when the app already has some user configuration applied to it
On the right, when the app doesn't have any configuration applied to it yet:

Screenshot 2019-07-10 at 2 42 16 PM

oponder:
After some feedback, we ended up with something like this:

Screenshot 2019-07-10 at 7 02 24 PM

The 'download configuration' link is not in this PR, as it is not essential for now.

src/actions/clusterActions.js
```diff
@@ -102,23 +102,19 @@ export function clusterInstallApp(app, clusterID) {
...
       if (Object.keys(app.valuesYAML).length !== 0) {
         // If we have user config that we want to create, then
         // fire off the call to create it.
-        appsApi
+        appConfigsApi
           .createClusterAppConfig(scheme + ' ' + token, clusterID, app.name, {
             body: app.valuesYAML,
           })
           .then(() => {
-            // The call succeeded, resolve with a configmap
```
oponder: the API does this linking of configmap to App CR automatically now, so that's why this step is gone.

src/components/UI/navigation/index.js
```diff
@@ -133,7 +133,7 @@ class Navigation extends React.Component {
             </NavLink>
             {this.props.showAppCatalog && (
               <NavLink activeClassName='active' to='/app-catalogs/'>
-                Apps
+                App Catalogs
```
oponder: I noticed there was a mental confusion when going between the Apps Tab on the cluster detail page and the Apps Link in the navigation. This resolves that and makes it clearer what you'll see when you click the link in the navigation.

oponder:
Deployed to `gauss` and `ginger`

marians:
Regarding consistency the installation and configuration dialogs could be aligned more, especially when it comes to the file input widget. I think the second variant is more elegant.

image

image

marians:
I'd like to see the app icon, if available, be used in the list of installed apps, too.

Catalog:

image

Apps panel now:

image

Mock:

image

marians:
The label of the `Browse Apps` button right under the list of installed apps makes me think: Which apps will I be browsing? Those I have in front of me?

image

I suggest we relabel the button to `Install App` and also give it the plus icon (`<i className="fa fa-add-circle"></i>`)

marians:
The distance between the `DANGER ZONE` label and the Delete button should be a bit larger, especially when compared to the spacing see right above under `USER CONFIGURATION`.

image

marians:
Regarding the defaul state `No user configuration yet.`: did we say we want to drop the `yet`, as it indicates that there should be a user config? In fact we don't know whether there should be one or not.

marians:
@oponder I checked from the UI/UX perspective. I'm done now. No blockers, just room for improvement.

oponder:
@marians thanks for the review! I'll try to get as much sorted in this branch still. Also thx for yesterday during the working hours.

oponder:
@marians I think I'll rename DANGER ZONE also.. we don't really do this anywhere else.

I'll follow the pattern we have for the cluster deletion.

PR #643, opened 2019-07-11
Version constraint for app installation features

oponder:
This ensures we don't show app installation related features for clusters that don't support it.


oponder:
This also introduces `cluster.capabilities`, an object that gets populated when the api call is made to get all the clusters.

This opens up a refactoring path to consolidate version comparison logic throughout Happa.

PR #644, opened 2019-07-11
Try to fix a weird behaviour of the loading spinner.

oponder:
I only see it in production, but the small loading spinners for the catalogs are sometimes spinning off axis and it looks very silly.

### DIFF
--- a/src/components/app_catalog/catalogs/index.js
+++ b/src/components/app_catalog/catalogs/index.js
@@ -46,15 +46,6 @@ class Catalogs extends React.Component {
                       >
                         <h3>
                           {this.props.catalogs.items[catalogName].spec.title}
-                          {!this.props.catalogs.items[catalogName]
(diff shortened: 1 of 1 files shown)


oponder:
Since I can't figure out on short notice why this is happening, I opt to remove these spinners from this location for now.

Clicking on an app catalog that hasn't loaded still brings you to a proper loading state for just that catalog. 

Once I figure out how to avoid this I can reintroduce these helpful little spinners again.

wacky

ograu:
@oponder I think it is the origin of the transform, taking a look at it now

ograu:
@oponder can't check it now cause after updating my testing api I lost again apps section in my local happa, ¿have you tried `transition-origin: center` in the image?

oponder:
@ograu Id like to fix it in another PR, but just remove it for now

PR #645, opened 2019-07-12
Also compute capabilities on cluster load details.

oponder:
When creating a cluster, a new cluster is added to the state without loading all clusters.

I was only computing the capabilities on the clusters load action, and not on the singular cluster load details action.


oponder:
I will merge this because I think it's important that customers do not run into this problem.

PR #648, opened 2019-07-15
Put the styleguide back.

oponder:
I believe it was removed accidentally in https://github.com/giantswarm/happa/pull/629/


marians:
Thank you!

PR #651, opened 2019-07-16
Node Pools - cluster detail

ograu:
# Cluster detail – Presentational PR

This PR is only for styles and markup.  Data is hardcoded. No tests yet.

I’ve been working on this with various goals in my mind:

- Reusability of styles (do we want a "styles library"?)
- Components as small as possible
- Render method:
    a) avoid cluttering render() with logic, conditionals, etc. if possible
    b) shorten number of lines for render
- Start using hooks

Wishful thinking: I would like to have components with a render method that could give us a general idea of what the component actually renders, **at a glance** 😛 (I know, this is not always possible)

I’ve been working in the ClusterDetailTable component basically. What I’ve done:

## Styles 

I have extracted reusable pieces of css – eg styles that we want to use in multiple places across Happa, and put them in a file inside the styles folder. I’ve divided css variables here into three types:

- Base styles: styles to be extended by others variables
- Layout styles: eg. rows
- Style wrappers and reusable elements: wrappers for labels or the dot we use to separate data

In the end this would be like our replacement for utility classes in regular css

Also, I started using grids for nodepools data: browser support with prefixes is at 93% and our target users are pretty technical, so it is relatively safe to think they are using modern machines or that they have installed latets versions of their browsers. No fallback at the moment. 🙀 

## Components

- Node Pools rows component
- I’ve added dropdown menu to the UI library (maybe we do have something like this? I haven’t found it)

## Other

- Naming: I prefer long explicit names than shorter-easy-to-write names

- Warning nth-childs in Emotion! This is a problem that I think most of CSS-in-JS libraries have. I am not taking this into account, now, cause this is a pseudo that I use a lot. I think we can address this problem when we decide to server render Happa, if we decide it at some point

Please, let's discuss it whenever fits you. I would love to know what do you think about what I'm doing.
(Maybe it's a too large PR, sorry)


src/actions/clusterActions.js
```diff
@@ -300,7 +300,17 @@ export function clusterLoadDetails(clusterId) {
     });
 
     var cluster;
-    var clustersApi = new GiantSwarmV4.ClustersApi();
+    var clustersApi = new GiantSwarm.ClustersApi();
+
+    clustersApi.getClusterV5(scheme + ' ' + token, 'm0ckd').then(
```
oponder: Is this some debug / local testing output?

src/actions/clusterActions.js
```diff
@@ -393,7 +403,7 @@ export function clusterCreate(cluster) {
       cluster,
     });
 
-    var clustersApi = new GiantSwarmV4.ClustersApi();
+    var clustersApi = new GiantSwarm.ClustersApi();
```
oponder: I think it would be good to separate this name change into its own PR.

src/actions/clusterActions.js
ograu: yes, it shouldn't be here...

src/actions/clusterActions.js
ograu: @marian is this something I can PR against master?

oponder: We'd have to get the same change in giantswarm-js-client as well, so if that can be done there too, seperate of the nodepools story, then that would be best.

PR #652, opened 2019-07-16
Extract organizations table into UI components.

oponder:
I got rid of a routing component layer which wasn't really needed.

Also the table items are now left aligned instead of centered, which is more similar to other tables we have in Happa.


oponder:
Screenshot 2019-07-16 at 4 57 03 PM

Screenshot 2019-07-16 at 4 57 18 PM

marians:
I think that for the CLUSTERS and MEMBERS column, centered would work better. Even better would be "centered right-aligned".

PR #653, opened 2019-07-16
Switch from js client giantswarm-v4 to giantswarm

ograu:
Not sure if this is something that we have to merge into master


ograu:
Is this failing because of any dependency in js-client? I assume it is so. Can't access details

marians:
Invited you to Snyk.

Currently I can't make anything out of the check. In Snyk's UI I don't see any information regarding this.

ograu:
I see a medium vulnerability in Synk for superagent. I see we are using version 3.8.3 while current version is 5.1.0: https://www.npmjs.com/package/superagent

Is there any reason why we are using 3.8.3? Can I update superagent?

But, still, I don't see how this is related to this PR, cause this dependency is in all PRs... I assume that this is just because `package.json` is in the commit and so it is verifying it.

ograu:
ok, I see that giantswarm client has it v3.5.2 as a dependency, so no way to change it from Happa's package.json

oponder:
I mentioned this version issue here by the way: https://github.com/giantswarm/giantswarm-js-client/pull/70#discussion_r301864069

package.json
```diff
@@ -39,7 +39,7 @@
     "copy-to-clipboard": "3.2.0",
     "deep-diff": "1.0.2",
     "emotion-theming": "^10.0.10",
-    "giantswarm-v4": "git+https://github.com/giantswarm/giantswarm-js-client.git#master",
+    "giantswarm": "git+https://github.com/giantswarm/giantswarm-js-client.git#remove-v4",
```
oponder: TODO: Change back to master when giantswarm-js-client is merged.

PR #656, opened 2019-07-16
Dropdown menu component to UI library

ograu:
To be 100% accessible we will have to implement arrow key interaction for the list items


src/components/UI/dropdown_menu.js
```diff
@@ -0,0 +1,73 @@
+import PropTypes from 'prop-types';
+import React, { useState } from 'react';
+import styled from '@emotion/styled';
+
+/*
+ * This is a reusable component. It is responsible for:
+ *
+ * 1. Rendering a styled popover menu
+ * 2. Provinding state and methods for handling the popover state in an accessible way
```
oponder:
```suggestion
 * 2. Providing state and methods for handling the popover state in an accessible way
```

PR #658, opened 2019-07-17
Added paths to webpack resolve moules & changed import paths

ograu:
We can add alias as well if we want: https://webpack.js.org/configuration/resolve/#resolvealias


ograu:
I have not changed `./` imports because I think those are nice and shorter than absolute imports. WDYT?

ograu:
thank you for being so supportive @oponder and @marians ! :D

PR #659, opened 2019-07-17
Cluster detail node pools

ograu:
Added cluster details node pools view and a route for it, so we now have:

-  the regular route `/organizations/{org}/clusters/{clusterID}/` with the current cluster details view, and
- the node pools view which has the same route with `np` appended to the end

This way node pools implementation development won't interfere with what we currently have and so we can make PRs directly to `master`.


ograu:
ping @marians

ograu:
Closed because I'm going to PR against a node pools master branch instead, until we add v5 endpoints to the API

PR #660, opened 2019-07-18
Two GS JS client libraries: master and node pools

ograu:
I have added the node pools branch of the JS client library, so I can use it for development.

@marians we talk about about creating a `node-pools` branch in Happa to issue all Node Pools related PRs against it. But this way we will have a long-lived branch and when we will want to merge it with `master` probably we will have a bunch of conflicts, so maybe it would be a better approach to commit Node Pools PRs directly to `master`. 

My two open PRs go in this direction. WDYT?

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
     "deep-diff": "1.0.2",
     "emotion-theming": "^10.0.10",
     "giantswarm": "git+https://github.com/giantswarm/giantswarm-js-client.git#master",
+    "giantswarm-v5": "git+https://github.com/giantswarm/giantswarm-js-client.git#node-pools",
     "history": "^4.7.2",
     "immutable": "3.8.2",
     "js-base64": "2.5.1",


marians:
If it works this way, fine with me. Out of curiosity, how much of an extra payload is this adding?

ograu:
Good point. I would say that too much: 1.7 MB is the weight I see it is adding to node_modules. Not sure how much would it add to the bundle.

Oddly enough, just before I've seen your comment I have added in the SIG-UX inbox: "Code splitting: check bundle size".

oponder:
I doubt the end result is 1.7MB... I tried deploying to ginger just now to see what the difference would be, but it looks like if we're not importing it anywhere in the code, then it is not added to the bundle.

ograu:
yes, it is not used now. I just made a test using the library and made a build locally: 

- Without v5: 1.85MB
- With v5: 1.94MB

Also tested a webpack compression plugin: 486KB

PR #661, opened 2019-07-22
Extract logic from AV label

ograu:
Extract logic from `AvailabilityZonesLabel` to make it a more reusable component and so we can use it in different layouts with different wrappers (markup and styles)

Is it `lib/utils.js` a good place to place reusable logic? WDYT?

### DIFF
--- a/src/components/UI/availability_zones_label.js
+++ b/src/components/UI/availability_zones_label.js
@@ -46,29 +46,17 @@ const Wrapper = styled.abbr`
   }
 `;
 
-const AvailabilityZonesLabel = props => {
-  const { zones } = props;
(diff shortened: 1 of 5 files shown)


oponder:
I'm mostly ok with this! but I also feel like it could be a component of its own, instead of a method in a utils file.

I think in general I want to discourage a utils file, since those kinds of files usually become a mixed bag of things. WDY? can we give this concept a name and a home in the UI folder instead?

ograu:
I agree with the comment regarding utils file...  I have made a component with it, as for the naming, maybe not super meaningful...  suggestions are welcome :-)

PR #662, opened 2019-07-22
Cluster detail node pools

ograu:
Just the view.
Modified styles:

- Row padding
- Font size
- Dropdown styles
- Added button

Availability zones still not working as they should because I need the PR I submitted this morning merged and the v5 client, which I am doing in a separate PR.


ograu:
ping @marians

oponder:
When there are visible changes or new UI components, I always appreciate a screenshot :)

src/components/UI/dropdown_menu.js
```diff
@@ -47,7 +66,7 @@ function DropdownMenu(props) {
 
   const onBlurHandler = () => {
     timeOutId = setTimeout(() => {
-      setIsOpen(false);
+      // setIsOpen(false);
```
oponder: As a codestyle thing, I think we should avoid commented code in the codebase .

Is the surrounding code and blurhandler still relevant?

src/components/cluster/detail/node_pool_dropdown_menu.js
```diff
@@ -0,0 +1,36 @@
+import DropdownMenu from '../../UI/dropdown_menu';
```
oponder: Let's use the absolute path now that we can? Also in some other places.

marians:
Looks good mostly. Some notes:

### Node pool menu:

- the menu should be closed when hitting ESC or when clicking outside the overlay.
- menu items should have a mouse-over state
- no separeator lines needed between the three items

### Node count indicator

- In the case where the current number equals the desired number (as in the first NP row), we don't need the special background for the number `3`.
- For the case where the current number derives from the desired number:
  - the background shape should have a corner radius
  - let's try a different color. We don't want to indicate that something is wrong. The message is more: pay attention, there is something going on here. Can we try the gold-ish color we use for the "Upgrade available" label? Maybe we can adapt it so that it works for backgrounds.

### Code labels

- The labels like node pool ID, instance type name, API endpoint URL should have more horizontal padding, so that it appears optically balanced with the vertical padding.
- 

### Availability Zones column

(just saying this, you are probably aware)

- We wanted to spell out the label as `AVAILABILITY ZONES`
- The letter alignment is not yet as planned. The upper row should have a gap for the `C`. The `D`s should appear in the same horizontal position.

image

src/components/UI/dropdown_menu.js
ograu: Good point. Yes, I should have uncommented it

src/components/cluster/detail/node_pool_dropdown_menu.js
ograu: 🤦‍♂true

ograu:
@marians thanks for the comments, ping you when I fix everything

marians:
Just tested the view in a browser window that's slightly more narrow than my 13" laptop permits. I see room for improvement in the behaviour here.

The region label should not wrap at all in my opinion.

image

The next two items should ideally wrap in a way that keeps the items together.

image

image

Example for `14 nodes in 2 node pools · 105 GB RAM · 30 CPUs`:

```
14 nodes in 2 node pools
105 GB RAM · 30 CPUs
```

ograu:
@marians this is what I've done:

### Node pool menu

- [x] the menu should be closed when hitting ESC or when clicking outside the overlay.
- [x] menu items should have a mouse-over state
- [x] no separeator lines needed between the three items

### Node count indicator


- [x] In the case where the current number equals the desired number (as in the first NP row), we don't need the special background for the number 3.
- [x] For the case where the current number derives from the desired number:

  - [x] the background shape should have a corner radius let's try a different color. We don't want to indicate that something is wrong. The message is more: pay attention, there is something going on here. Can we try the gold-ish color we use for the "Upgrade available" label? Maybe we can adapt it so that it works for backgrounds.

### Code labels

- [x] The labels like node pool ID, instance type name, API endpoint URL should have more horizontal padding, so that it appears optically balanced with the vertical padding.

### Availability Zones column

- [x] We wanted to spell out the label as AVAILABILITY ZONES
- [ ] The letter alignment is not yet as planned. The upper row should have a gap for the C. The Ds should appear in the same horizontal position.

### Small screens

- [x] The region label should not wrap at all in my opinion.
- [x] The next two items should ideally wrap in a way that keeps the items together. (I also had aligned contend to right)

ograu:
The only thing that is not yet done is the alignment of availability zones, I was kind of waiting for the NP data fetching in redux, but I have changed my mind and will do the styles now. So sorry for this pinging 🙏  will add it to this PR and pinged you again

ograu:
@marians all fixed now

ograu:
Captura de pantalla 2019-07-26 a las 12 44 15

ograu:
ping @marians waiting for your review to merge. Just want to make sure that you recieved my request.

PR #664, opened 2019-07-23
WIP Consider data in the index.yaml for getting first app

pipo02mix:
We are returning the first app of the index.yaml but sometimes it is not served order by date. We need to order first to pick first up


oponder:
It's failing because of code style. If you do `yarn run prettier` and commit the result, it should bring things in order.

pipo02mix:
tested on ginger and works

PR #666, opened 2019-07-23
V5 endpoints and Node pools cluster in Happa

ograu:
In this PR I have modified the action creators that are responsible of dispatching the data needed in the clusters list view and the cluster details view. The goal is make `m0ckd` cluster available in Happa so I can keep working on the node pools views. Also:

-  I have extracted instatiation of the clusters API in the functions I've been working with. Maybe I am missing something... ¿Is there any reason why we reinstantiate it inside all the action creators?

- I have changed promise `then/catch` workflows for `async/await` workflows in the methods I've been working. I have nothing against promise workflow. Just prefer `async/await` for the sake of readibility.

I am PRing against a node pools master branch. Maybe should have sumitted it against `master`?  WDYT?


oponder:
Hey @ograu thanks again for your efforts here! I'm ok with async await, and I think the instantiation of the clusters API is just some bad copy pasting.

As for this approach, I think intertwining the v5 with the v4 things is going to be quite hard to keep track off.

I'd think going for something where we have a `clustersLoad` and a `clustersLoadV5` action would be a bit easier to manage right?

Also: 

> window.config.environment === 'development'

I think these kind of conditionals inside the actions will lead to some pain down the road.

Instead I'd like to pull that conditional 'higher'. So wherever we do the initial fetching of clusters, we can have that conditional there once, and then do a `clustersLoadV5` if `config.environment === development`. 

We can then do the adding of clusters to state in the reducer.

by the way, I am not as deep in the topic and code for this as you are, so what you have right now might be the best thing considering the tradeoffs you had to deal with. But would like to consider something that results in an easier mental model.

ograu:
@oponder thanks for the comments!

I'm still not being able to take like informed decisions because my mental model of the redux store is still partial and I am not as deep as you may think in redux topic, so any suggestion from your side is appreciated.

I was kind of making momentary solutions to just make it work, but after reading your comments I now think it makes sense to spend more time in the Redux store now. So, yes, I am going to separate regular clusters from NP clusters when we fetch them and add additional actions when needed. Ping you then for review.

ograu:
hey @oponder In today commits:

- I am adding specific actions for v4 and v5 clusters fetching so we can have separate logic in reducers. If at some point down the road we find that it is too much boilerplate, we can remove v4/v5 actions and use generic actions if the code is the same for the two types of clusters.

- Regarding promises instead async/away: if we separate the fetching of v4 and v5 clusters, there are too many sequencial await statements. This is an antipattern because it takes longer than with promises. And also no callback nesting needed. So I'm switching back to promises

- Conditional `window.config.environment === 'development'`, moved it as high as I can. (or at least this is what I think 🤔 )

- I have added a `nodePoolsClusters` key in `clusters` object in store where I want to store NP clusters ids, so we can use it when we have to check whether a cluster is a NPs one or not. Not sure if it is the best solution, yet.

Finally, please keep in mind that this is not the code that will we merged master, though we could merge it if we want. Instead it is something that should allow us (me) continue working with the NP Happy implementation. It has to be something that goes in the right direction, but we have some limitations with the mocked API... that's why I am PRing against `np-master` instead of `master`.

Looking forward your comments. Thanks in advance.

ograu:
Still iterating on this

ograu:
Done, waiting for review.

src/actions/clusterActions.js
```diff
@@ -6,6 +6,9 @@ import APIClusterStatusClient from 'lib/api_status_client';
 import cmp from 'semver-compare';
 import GiantSwarm from 'giantswarm';
 
+// API instantiations
+const clustersApi = new GiantSwarm.ClustersApi();
```
oponder: This change would be a good candidate for a seperate small PR next time

ograu:
You mean open a PR `master`<- `np-master`? Yeah, haven't done it yet. I will

This error you have: have you deleted `node_modules/giantswarm` before running `yarn install`?

ograu:
@oponder merging and opening a PR `master` <- `np-master`

PR #668, opened 2019-07-24
Cluster reducer with spread operator

ograu:
Minor refactor: I saw that inside every `case` we were overwriting state. I decided to refactor in a way that the state is modified, but never overwritten.

I also had simplified this:

`{ items: items }`

To this:

`{ items }`

But no real change here, it is the exact same thing with ES6 syntax.

Also I have removed state `isFetching: false` from returns because I haven't found a single `isFetching: true`, then we can remove them safely (?). I am wondering if we really need this property here...

If you are ok with this I will do the same in the other reducers


ograu:
And it works 🙂  because you are generating a new state object in every `return`, that's fine. But this way will be easier to manage and to read. Next step is enforce inmutability in nested objects like `items`. Thinking about it already

PR #670, opened 2019-07-25
Fix proptypes on AvailabilityZonesLabels

ograu:
`key` is not a prop, it's for React's internal use.

### DIFF
--- a/src/components/UI/availability_zones_label.js
+++ b/src/components/UI/availability_zones_label.js
@@ -46,14 +46,13 @@ const Wrapper = styled.abbr`
   }
 `;
 
-const AvailabilityZonesLabel = ({ key, label, letter, title }) => (
-  <Wrapper className={letter} key={key} title={title}>
+const AvailabilityZonesLabel = ({ label, letter, title }) => (
+  <Wrapper className={letter} title={title}>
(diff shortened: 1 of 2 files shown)


src/components/UI/availability_zones_label.js
```diff
@@ -46,14 +46,13 @@ const Wrapper = styled.abbr`
   }
 `;
 
-const AvailabilityZonesLabel = ({ key, label, letter, title }) => (
-  <Wrapper className={letter} key={key} title={title}>
```
oponder: isn't the key attribute still needed though for react to do the iterating properly? is there no warning in the console about that now?

ograu: It is needed, and in fact it is passed to the label, _but_ it is not a prop and therefore there's no need to make any reference to it in the label, React justs passes it down in the return statement of `AvailabilityZonesLabel` component. 

This is what I see in my `clusters-detail-node-pools`after merging this `fix-availability-zones-proptypes`branch:

Captura de pantalla 2019-07-26 a las 8 24 02

oponder: Nice! thanks :)

PR #675, opened 2019-07-30
Number of NPs in Clusters list

ograu:
Added node pools number to clusters list


ograu:
Closing and crating againg because commits are duplicated in both branches

PR #679, opened 2019-07-31
NPs Clusters detail view w/ store data

ograu:
- No data on K8s version because can't found any version for 8.0.0 release version
- Nodes are the sum of nodes in all NPs
- Adding RAM and CPUs data in another PR


ograu:
wrong branch

ograu:
Actually right branch...

PR #683, opened 2019-08-01
Simplified props in dashboard items and added NPs

ograu:


src/components/home/cluster_dashboard_item.js
```diff
@@ -130,31 +130,25 @@ class ClusterDashboardItem extends React.Component {
     var cpus = this.getCpusTotal();
     var numNodes = this.getNumberOfNodes();
 
+    const np =
+      window.config.environment === 'development' &&
+      this.props.cluster.id === 'm0ckd'
```
oponder: I'm a bit concerned about `m0ckd` spreading throughout the code base. 

I'd like to maybe have a refactoring session where we can discuss other strategies here.

In this case it's not a problem, but in another situation, it makes developing happa now only possible on the `v5-dev` branch.

ograu: also a bit concerned, happy to hear suggestions regarding this :-)

PR #685, opened 2019-08-02
Add a label that does some expectation management depending on the catalog type.

oponder:
This adds a label to the catalog listing, which on hover, gives a bit of expectation management for the `incubator`, `test` and `community` catalog types.

The `test` type of catalog should by the way never appear on customer installations, but I decided to cover that case anyways.

Screenshot 2019-08-02 at 4 47 47 PM

I will also add a warning on app installation when someone is about to install a community app, but that will be in a different PR.

### DIFF
--- a/src/components/UI/catalog_type_label.js
+++ b/src/components/UI/catalog_type_label.js
@@ -1,0 +1,74 @@
+import OverlayTrigger from 'react-bootstrap/lib/OverlayTrigger';
+import PropTypes from 'prop-types';
+import React from 'react';
+import styled from '@emotion/styled';
+import Tooltip from 'react-bootstrap/lib/Tooltip';
+
+const Wrapper = styled.div`
+  font-size: 12px;
+  display: inline-block;
+  margin-bottom: 15px;
+
+  span {
+    background-color: #1e4156;
+    padding: 4px 6px;
+    border-radius: 4px;
+    margin-right: 5px;
+  }
+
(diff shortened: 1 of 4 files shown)


src/components/UI/catalog_type_label.js
```diff
@@ -0,0 +1,74 @@
...
+  }
+`;
+
+/**
+ * CatalogTypeLabel shows some information about a catalog depending on its type.
+ */
+const CatalogTypeLabel = props => {
+  var icon;
+  var message;
+
```
ograu: @oponder I think it is best practice to try to avoid `var`and use `let` or `const` depending on the case.

oponder: My bad habits. Thanks ograu!

oponder:
@ograu could you take a look again, would love some feedback on my little tests.

src/components/UI/catalog_type_label.js
ograu: If you don't want to rebind the variable it's best to use `const` because you will get a `TypeError`if you try to do so.

So my rule of thumb is: use always `const`, except when you want to rebind the variable, then use `let`. 
Keep in mind that you can have a `const` variable and mutate their keys, array items though.

ograu: And never use `var` cause it can be redeclared or leaked to the global object

PR #687, opened 2019-08-02
Nodepools with store data

ograu:
Replace NPs hardcoded objects with NPs data in store

### DIFF
--- a/src/components/cluster/detail/cluster_detail_node_pools_table.js
+++ b/src/components/cluster/detail/cluster_detail_node_pools_table.js
@@ -9,12 +9,6 @@ import ReactTimeout from 'react-timeout';
 import RefreshableLabel from 'UI/refreshable_label';
 import styled from '@emotion/styled';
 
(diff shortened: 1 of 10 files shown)


src/components/cluster/detail/cluster_detail_node_pools_table.js
```diff
@@ -239,7 +216,7 @@ class ClusterDetailNodePoolsTable extends React.Component {
           </GridRowNodePoolsNodes>
           <GridRowNodePoolsHeaders>
             <span>ID</span>
-            <span>NAME</span>
+            <span style={{ paddingLeft: '8px' }}>NAME</span>
```
oponder: In general, I'd like to avoid "inline" style, but I think in this case, it's a fine compromise. I think sometimes the benefit of having things 'closer together' can outweigh the rule.

ograu: yeah, in this case this is a very specific thing for only this column. In such situations  I think that this is the best thing to do

PR #688, opened 2019-08-02
Some small style adjustments to app catalog index

oponder:
I noticed I had messed up the font size. This sets it back to what it was, and adjust the space between the catalog type label and the text a bit.


oponder:
Before:
Screenshot 2019-08-02 at 9 35 19 PM

After:
Screenshot 2019-08-02 at 9 35 27 PM

PR #689, opened 2019-08-05
Add info on KVM node storage use and size

marians:
As a follow-up from support questions, this adds some info regarding storage usage for KVM.

### Preview

image

### DIFF
--- a/src/components/cluster/new/view.js
+++ b/src/components/cluster/new/view.js
@@ -447,9 +447,9 @@ class CreateCluster extends React.Component {
                       <div className='col-9'>
                         <p>
                           Configure the amount of CPU, RAM and Storage for your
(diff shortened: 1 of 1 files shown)


oponder:
It might be interesting to have a label appear next to the storage picker like:

[!] you will need 40GB available, since this setting applies to two volumes

Alternatively , we could have two pickers there (one for Kubelet and one for docker) which are always linked, to really drive home the point?

Not a prop for me though.

marians:
> Alternatively , we could have two pickers there (one for Kubelet and one for docker) which are always linked, to really drive home the point?

Yes, I would prefer that, but the API doesn't support it yet. This PR here is about adapting to the way it currently works behind the API.

PR #692, opened 2019-08-05
Modified config with data from gauss

ograu:
I am using data from `gauss` because I need it for node pools cluster.
No idea how this affects production Happa exactly


marians:
What about the idea to mover this JSON structure into a separate file where it can be read? Having thise huge minified string in the code base will create unreadable/unreviewable diffs whenever we want to change something.

src/index.html
```diff
@@ -21,7 +21,8 @@
     passageEndpoint: 'http://localhost:5001',
     environment: 'development',
     ingressBaseDomain: 'k8s.sample.io',
-    awsCapabilitiesJSON: '{"m3.large":{"description":"M3 General Purpose Large","memory_size_gb":7.5,"cpu_cores":2,"storage_size_gb":32},"m3.xlarge":{"description":"M3 General Purpose Extra Large","memory_size_gb":15,"cpu_cores":4,"storage_size_gb":80},"m3.2xlarge":{"description":"M3 General Purpose Double Extra Large","memory_size_gb":30,"cpu_cores":8,"storage_size_gb":160}}',
+    awsCapabilitiesJSON:
```
oponder: This would need to stay on a single line. We're doing some find/replace in the index.html to set production values for this during a startup script: https://github.com/giantswarm/happa/blob/master/scripts/start.sh

marians: @oponder I have discussed this with Oriol last week. The question was: Why do we need to inject this at build time? We can maintain a list of all known instance types in happa. The info endpoint delivers the list of allowed types (only names) and the default type. That would simplify some things.

I don't see this reflected in the PR yet.

oponder: We want to have a single source of truth for it (the installations repo), thats why in dev it is actually 'less' than what we have in prod. To really indicate that this is not the real list.

When we add a new instance type to the installations repo, I would like to avoid having to make changes in any other repos.

oponder: Oh and the reason this info is 'extracted' to the installations repo is because `cluster-service` also makes use of it.

ograu:
@marians yes, but as we discussed on Friday, this script, as far as I understand, runs after the bundle, and therefore is not that easy a refactor, and as I mentioned in planning, I am avoiding refactors becase otherwise I can make progress on NPs. So skipping it now.

That being said, @oponder I totally think that we should have config as a js file imported as a regular js module, and I think that these replacements should be done inside an environment file where we could set up some values in a more canonical way, because this is kind of hidden and it is happening outside of our control, like out of the app, and hence it is hard to view/change

WDYT?

src/index.html
ograu: We can fetch it in build time, can't  we?

ograu:
@oponder Is it ok now?

oponder:
> because this is kind of hidden and it is happening outside of our control, like out of the app, and hence it is hard to view/change

Sure, that would be a better place. But the point is that this information shouldn't be changed often, and it is not the place to change values for production. It's just a placeholder that holds enough information for us to be able to test various cases with.

oponder:
> @oponder Is it ok now?

@ograu what is the data that we actually need from gauss to be able to test / visualise something node pool related? I'd rather add just that, instead of adding much more info than needed.

I think this line of data that you're editing should be considered as like dev placeholder data, and by putting everything in there we can possibly confuse the idea now.

ograu:
Ok, so then what I will do is take what we had before and just add the line I need.

ograu:
@oponder I've checked out file from master and added just the object I need

oponder:
@ograu nice! Now i see a bit clearer what you needed: The "r3.xlarge" instance.

So now I'm wondering, why do we need this instance type specifically in development of node pools? Why is this one important to add?

ograu:
Because is the one I get when I hit the mocked endpoints

oponder:
Ah ok.. does it make sense to change happa's dev placeholder data to fit the mocking? or do you think it might be better to change the mocked response instead? 

Sorry if I'm being too difficult about a small thing! Not my intention to block you guys here.

ograu:
IDK, I am ok with either the solution. If you think it's best to change the mocked response, I can ask Toby to do so.

ograu:
> Sorry if I'm being too difficult about a small thing! Not my intention to block you guys here.

No worries, thanks to you questions and also to @marians comments I now understand better what's reaaly going on!

ograu:
> Yeah I think it makes more sense to change it in the API. but don't want to block here.

Asking Toby to do so.

ograu:
Closing as we have changed mocked API response

PR #697, opened 2019-08-06
Dynamic thunk function in clusterName

ograu:
Little change to make `clusterName` reusable with different endpoints. 

Naming sucks... any suggestion is welcome

### DIFF
--- a/src/actions/clusterActions.js
+++ b/src/actions/clusterActions.js
@@ -658,107 +658,79 @@ export function clusterLoadKeyPairs(clusterId) {
   };
 }
 
-export function clusterLoadDetailsSuccess(cluster) {
-  return {
-    type: types.CLUSTER_LOAD_DETAILS_SUCCESS,
-    cluster,
-  };
-}
+export const clusterLoadDetailsSuccess = cluster => ({
+  type: types.CLUSTER_LOAD_DETAILS_SUCCESS,
+  cluster,
+});
 
-export function clusterLoadDetailsError(error) {
-  return {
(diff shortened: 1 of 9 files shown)


ograu:
@oponder how about this: https://github.com/giantswarm/happa/pull/697/commits/8beea2de45f1ee18a8fa126e7cacb9f1fe3f8b3d ?

oponder:
Hey ograu, my idea actually was to keep the details of dispatching away from the UI elements. 

So what I was aiming for actually is having just one function that gets passed in, and that is all that needs to get called, but without mapping `dispatch` also to props in the UI component.

So the function that is passed in from the parent would already contain the dispatch inside of it... does that make sense?

So onSubmit could be a function that looks like

```
function() {
 return dispatch(...);
}
```

ograu:
hey @oponder I understand your idea and found it an interesting approach, but being `dispatch` a function that this component will always need I think that my approach makes more sense, because this way we inject it here and then any component that uses this `ViewAndEdit`component doesn't have to know about dispatch and so we won't have to map `dispatch` to props in every component that uses it.

IMO if what we want is to have superclean UI components then maybe we should extract all the business logic (validation and dispathing) and split this component into two. But again, this does not make a lot of sense to me since dispatching the action is something that this component will always have to deal with.

ograu:
Merging this as I need it, @oponder let's keep the discussion open, decide something and modify this component, if needed

oponder:
All good! Thanks ograu!

PR #700, opened 2019-08-07
Updates in Redux store in an optimistic fashion

ograu:
I found that we were not updating the store when patching cluster names. We are now in a situation where the app shows what's in `ViewAndEditName` state instead of what we have in the store. When the app refreshes the store is synchronized and so this problem is solved. But this means we have more than one source of truth during 30 secs and we don't wan't this because this can result in weird side-effects.

A quick note on _optimistic updates_: it is a way to perform updates on an database - in order to improve the user experience and give a feel of speediness - that consists in updating the local store first and then the database. If there is an error, we catch it and revert the situation in the store. This should be done only in operations with a high probability of success.

To solve this situation I've added the following:

- `case`s for cluster patch call in reducer
- A new argument to patch `thunk` function: the whole object that we want to update, so this way we can revert the situation
- I had to modify the arguments we pass to `ViewAndEditName` and also simplified it a bit

By the way, I think that what I added in the reducer can be highly simplified with [Immer](https://github.com/immerjs/immer). In the meanwhile I will do it this way, but even doing this way this is not 100% inmutable because nested properties in `cluster` object are pointing to same references.


ograu:
ping @oponder

PR #701, opened 2019-08-08
Node pool patch

ograu:
- Add nodePoolPatch function
- Inform `NodePool` component of the state of `ViewEditName` so we can modify the view accordingly
- Let node pool name take all the room in the grid

Note: endopoint for patching node pools is not working, Toby is working on it


ograu:
Added the cases for the reducer, now that the endpoint is working in local testing API.

Please @oponder @marians take a look at the reducer and see how much code we have to write just for patching a object. This is because we don't have a library for updateding state in an inmutable fashion (we knew that), but also because of this: https://github.com/giantswarm/giantswarm/issues/6711

src/actions/clusterActions.js
```diff
@@ -808,3 +807,56 @@ export function clusterCreateKeyPair(clusterId, keypair) {
...
+      clusterId,
+      payload,
+    });
+
+    return nodePoolsApi
+      .modifyNodePool(scheme + ' ' + token, clusterId, nodePool.id, payload)
+      .catch(error => {
+        // Undo update to store if the API call fails.
+        dispatch({
+          type: types.NODEPOOL_PATCH_ERROR,
```
oponder: Does the user get any feedback when things go wrong? Or does the name just jump back? Was trying to find a flash error message or something like that, but not seeing it yet.

ograu: It was in the `ViewEditName` component, I think it's cleaner to put this notifications here, so this is what I did

ograu:
Styles will be modified in next PR, I still have more interactions to work with in this view, so this is not final:

Captura de pantalla 2019-08-09 a las 18 07 40

ograu:
For some reason synk check didn't finish and it was like _pending checks_, now it has finished.

ograu:
ping @oponder

PR #706, opened 2019-08-12
Node pools reducer

ograu:
Hey, first of all, sorry for such a big PR. 

I actually wanted to make smaller PRs but when I tried to do so, everything broke, so I ended up with this PR, but keep in mind that most of it is just moving the code from `clusterReducer` to `nodePoolsReducer` and changes consequence of having moved state from `entities.clusters.items[id]nodePools` to `entities.nodePools`, so it seems bigger than it actually is, there is virtually no real addition:

Changes this PR introduces to the codebase:

- Created `nodePoolsReducer.js` and `nodePoolsActions.js` and moved relevant code there
- Added `nodePools` as independent piece of state in store
- We now just keep an array with ids in `entities.clusters.items[id]nodePools`
- Added a method in `clusterUtils.js` that get cluster node pools and return an array with them

The idea behind this change is basically 

- easier updates to store and 
- performance: avoid re-renderings

If we like what I've done here with node pools and think it is worth it, we can normalize all the store in the future. It is something we can think about.

I know I told you I was going to add `immer` to Happa, and this is what I'm doing, but I thought this change goes in the same direction and now is the moment to make it before the app keeps growing on the node pools side, so I submit this first.

Thanks for being so patient with my changes.


ograu:
Thanks for reviewing so quickly! 😸

PR #707, opened 2019-08-13
Node pools reducer with immer

ograu:


ograu:
The idea here is that we can totally forget about using immutable methods in reducers and we can mutate this `draft` object with any JS method we want to use, Immer will take care of updating the pieces of state that have changed in an immutable manner.

PR #712, opened 2019-08-13
Immutable Invitation reducer

ograu:
⚠️⚠️ ⚠️ Can't test because it never worked in my local Happa  ⚠️ ⚠️ ⚠️

### DIFF
--- a/src/reducers/invitationReducer.js
+++ b/src/reducers/invitationReducer.js
@@ -1,32 +1,24 @@
 import * as types from 'actions/actionTypes';
+import produce from 'immer';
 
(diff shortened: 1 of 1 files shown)


oponder:
To test invitations, from the happa project root folder do: `docker-compose up -d passage redis mailcatcher`

That'll bring up passage, which handles the invitations for us.

It also starts mailcatcher, so you can see the emails it sends at: `localhost:1080`

No real mails are sent, mailcatcher is the fake smtp server that passage is configured to talk to in this case.

Looks like it's working!

PR #714, opened 2019-08-13
Immutable Organization reducer

ograu:
⚠️ ⚠️ ⚠️ Can't test deletion of organization, never worked on my local Happa ⚠️ ⚠️ ⚠️


oponder:
Hmm.. org deletion should work, can't think of a reason that it shouldn't work for you locally.

Let's take a look together at that!

PR #716, opened 2019-08-14
Update getting started

oponder:
Things have changed a bit since we wrote the getting started part of the README.

This updates that.

It also removes passage and mailcatcher from the dockerfile, I will move it to the dockerfile in api instead, so we don't have to think about starting another docker-compose before testing anything related to invitations. 

FYI @ograu. I gave you some instructions on how to start `passage`, but now I am removing that from `happa`. So what I said here will no longer work after this PR is merged. Instead I will add it to the docker-compose in `api` so that it is always running, and invitations are always working.

https://github.com/giantswarm/happa/pull/712#issuecomment-521104734

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ Currently only Giant Swarm employees are able to see that repo. We're considerin
 open sourcing it.
 
 Start the `api` and dependencies first by going to the `api` [repo](https://github.com/giantswarm/api) and running
-the dockercompose file there:
+the dockercompose file in the testing folder there:
 
 ```nohighlight
 cd $GOPATH/src/github.com/giantswarm/api/testing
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -42,16 +45,6 @@ make up
 As part of `make up`, `./fixtures.sh` will run and create the initial user and
 organization you can log in with.
```
marians: I am still missing two pieces of info:

1. The docker-compose stuff happens in the `testing` sub folder of giantswarm/api
2. You have to run `make aws|azure|kvm` to create a docke-compose.yaml file.

marians:
Very good to clean this up before the next people are hit! ❤️

oponder:
> Very good to clean this up before the next people are hit! ❤️

Yeah I'm kinda in a "lets make dev smoother" mode.. which conflicts a bit with other prios, but I think it might be one the best places to put effort, since more people are coming on board.

README.md
oponder: Ok! adding that.

We do have a `cd` command as part of the example. but i'll add the `make aws` part there.

PR #717, opened 2019-08-14
Remove some flashing of unstyled content.

oponder:
This makes the initial load of Happa more calm.

The footer content won't show up on a blank page anymore, and the page should start off blue.

I've also fixed a unreadable portion of the warning that shows up when users do not have javascript enabled. The `strong` tag rendered the text in white.


marians:
Release please! :)

PR #718, opened 2019-08-14
0.0.0.0 to localhost

ograu:
0.0.0.0 triggers an error in auth0.com


oponder:
Oops. yeah.

that means though we can't get it to work easily for the Docker set up. I'm fine putting this pack to localhost for now. But have to think of a way to let webpack know if it is in docker or not, so that it can run the server on an interface that is actually reachable.

ograu:
> Oops. yeah.
> 
> that means though we can't get it to work easily for the Docker set up. I'm fine putting this back to localhost for now. But have to think of a way to let webpack know if it is in docker or not, so that it can run the server on an interface that is actually reachable.

¿Is there any way of letting auth0 know that 0.0.0.0 is localhost?

PR #719, opened 2019-08-14
Fix number of Nodes, CPU and RAM

ograu:
Now it is working, but I wouldn't merge this to master. 
Not proud at all of this code 😢 

- Fix conditionals for checking status in `clusterUtils`
- Added hardcoded `status` object for developing
- Commented refresh function: `Home` component has to be rewritten as now we have an immutable store and what made sense before (force re-renderings, deep comparisions) now not only doesn't make sense but also kind of clashes with what we have done with Immer.

Also I think that that actions should be redefined, simplified and some of them removed.

### DIFF
--- a/src/components/home/index.js
+++ b/src/components/home/index.js
@@ -53,7 +53,7 @@ class Home extends React.Component {
   };
 
   refreshClustersList = () => {
-    this.props.actions.clustersLoad();
+    // this.props.actions.clustersLoad();
   };
 
   handleVisibilityChange = () => {
--- a/src/components/home/index.js
+++ b/src/components/home/index.js
@@ -39,6 +39,7 @@ class Home extends React.Component {
 
   componentWillUnmount() {
     this.visibilityTracker.removeEventListener(this.handleVisibilityChange);
+    this.props.clearInterval(this.refreshInterval);
   }
 
   /**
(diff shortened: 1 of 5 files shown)


oponder:
I think there were some other things going on that confused things. 

Mostly that fetching the list of clusters again kind of cleared the workers or nodes field of the cluster object.

Also that refreshing continued even when away from the cluster list.

marians:
Had some trouble testing, but it looks fine on geckon.

src/components/home/index.js
```diff
@@ -39,6 +39,7 @@ class Home extends React.Component {
 
   componentWillUnmount() {
     this.visibilityTracker.removeEventListener(this.handleVisibilityChange);
+    this.props.clearInterval(this.refreshInterval);
```
marians: @ograu This is needed to stop the client from pulling data in the background, even when the user is no longer on the page.

ograu: thanks @marians , I know what a `clearInterval()` do, I saw that we didn't cleared intervals here and actually noticed weeks ago that in Happa we have a lot of `setIntervals`/`setTimeouts` not cleared and, yeah, this is clearly something that has to be done.

I decided (maybe it was a wrong decision) to comment this refreshing because I was kind of thinking in redoing all of.

marians: OK. Just wanted to point it out since it must have gotten lost in some of the recent refactoring.

PR #721, opened 2019-08-16
Make view and edit name know a bit less about our tech stack.

oponder:
@ograu, this would be my approach to make it so that the view and edit name component doesn't know it has to dispatch anything. 

Instead all it expects is a Promise for the onsubmit call, so it stays concerned with calling that function and dealing with a success or a failure.

Side effect, no more optimistic updates. Which I question a bit to be honest. I'd rather show a spinner, to let the user know something is happening, and then an error, instead of doing a optimistic update, and then throwing an error after the fact.

This PR is for discussion so you can see the approach I would take here, and see what you think of it.

Is this promise stuff too old school or weird or bad practice?

### DIFF
--- a/src/actions/nodePoolsActions.js
+++ b/src/actions/nodePoolsActions.js
@@ -90,13 +90,6 @@ export function nodePoolPatch(nodePool, payload) {
 
     return nodePoolsApi
       .modifyNodePool(scheme + ' ' + token, clusterId, nodePool, payload)
-      .then(() => {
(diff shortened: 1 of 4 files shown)


oponder:
I will go ahead and merge this for now, we can continue refactoring and thinking of different approaches.

The reason for the merge is because right now not all 3 affected actions will give a success flash message, so I'd like to bring that back.

PR #722, opened 2019-08-16
Fix flapping 0s

oponder:


oponder:
We went with a different approach for now, see https://github.com/giantswarm/happa/pull/723

PR #723, opened 2019-08-16
Fetch everything at once

oponder:
Instead of trying to merge things as they come in, this makes the clustersLoad action fetch everything about all clusters all at once.

The disadvantages here are that this will be fine for small installations, but not so great for installations with many clusters.

What is happening now, if we have 100 clusters across 10 orgs, is that each time we will do ~300~ 201 requests, even if we are only looking at a org that has only 10 clusters.

So that's probably one of the reason the code evolved the way it did. Lets keep thinking, and submitting this now so you can see what it looks like.

### DIFF
--- a/src/actions/clusterActions.js
+++ b/src/actions/clusterActions.js
@@ -87,15 +87,67 @@ function clustersLoadV4(token, scheme, dispatch, getState) {
   return clustersApi
     .getClusters(scheme + ' ' + token)
     .then(clusters => {
-      const lastUpdated = Date.now();
       const enhancedClusters = enhanceWithCapabilities(
         clusters,
         getState().app.info.general.provider
       );
 
       // Clusters array to object, because we are storing an object in the store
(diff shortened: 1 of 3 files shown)


oponder:
Im not 100% convinced the trade off of more requests is worth it... let me think a bit. will have to continue this tomorrow

oponder:
or maybe do this is a temporary fix for now... thinking

ograu:
Hey @oponder nice! I modified some things here, no big deal, it's basically the same but with `await`, maybe more readable: https://github.com/giantswarm/happa/pull/724

> What is happening now, if we have 100 clusters across 10 orgs, is that each time we will do ~300~ 201 requests, even if we are only looking at a org that has only 10 clusters.

yeah, I understand that as a temporary fix kinda of sucks... I assume that we have to continue refactoring and limiting the clusters to fetching to those of the org we are looking at...

PR #724, opened 2019-08-16
Fetch everything at once with await

ograu:
This is my take on what we are PRing in `fetch-everything-at-once`: it is the same you did but using await and because of that a little more readable in my opinion.

### DIFF
--- a/src/actions/clusterActions.js
+++ b/src/actions/clusterActions.js
@@ -80,28 +80,84 @@ export function clustersLoad() {
   };
 }
 
-function clustersLoadV4(token, scheme, dispatch, getState) {
+async function clustersLoadV4(token, scheme, dispatch, getState) {
   // TODO this will be in getClusters() here in this function we just want to
   // dispatch specific actions for v4 clusters.
   // Or maybe we won't need this method at all.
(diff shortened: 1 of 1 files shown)


oponder:
Thx oriol! I like this approach more. I will merge it later, want to do some exploratory testing to see how it handles certain error edge cases. (mostly because I need to understand better (await / async) and will probably learn a lot

ograu:
ping @oponder 

Do you think we can merge it into master? I have to work with this `clustersLoadV4()` function now and find it a little bit difficult to follow.

oponder:
There was a problem with this version, which is fixed here: https://github.com/giantswarm/happa/pull/732

Also the client has changed slightly, it doesn't need to take `sheme + token` as a parameter anymore:

https://github.com/giantswarm/happa/pull/733

So this is a bit outdated now. Merging it as is will introduce some regressions.

If you want I can try to change what we have now into async await, using this as an example.

ograu:
I am doing it, thanks!

ograu:
continued in https://github.com/giantswarm/happa/pull/744

PR #725, opened 2019-08-16
Immutable App reducer

ograu:
- Cleanup and Immer in `appReducer`
- Local storage utils in its own file


src/actions/userActions.js
```diff
@@ -138,6 +141,7 @@ export function giantswarmLogin(email, password) {
         return userData;
       })
       .then(userData => {
+        localStorage.setItem('user', JSON.stringify(userData));
```
oponder: would it make sense to use the new setUser here? (and in a few other cases too)

src/utils/localStorageUtils.js
```diff
@@ -0,0 +1,39 @@
...
+    user.auth = {
+      scheme: 'giantswarm',
+      token: user.authToken,
+    };
+  }
+
+  return user;
+};
+
+export const setUser = userData => {
```
oponder: I'm worried that maybe we dont' always have `userData.auth.token` and `userData.auth.scheme`, so we might 'unset' the authentication details of the api client... need to think a bit about this

oponder: Also I think this could maybe be better described as `saveUser` or `persistUser`

especially if we also add the organization setting method here, then it is nice to have some kind of verb that implies persistence, since that's what this file is about.

src/actions/organizationActions.js
```diff
@@ -14,6 +14,8 @@ export function organizationSelect(orgId) {
   return function(dispatch) {
     dispatch(push('/'));
 
+    localStorage.setItem('app.selectedOrganization', orgId);
```
oponder: perhaps we can make a method in localStorageUtils for this as well.

src/actions/userActions.js
ograu: TBH I don't know, I wasn't sure why `setUser()` declare and assign `defaultClient`and `defaultClientAuth` so I left it as I found it.

src/utils/localStorageUtils.js
ograu: IMO using __set__ and __localStorage__ toghether is fine, I mean I think it's clear what it does, but we can change it if you think it's clearer!

src/actions/organizationActions.js
ograu: mmm... Why should we do that?
Are you thinking in adding any check in the wrapper method for setItem?

src/utils/localStorageUtils.js
oponder: Yeah the naming isn't a big deal, but I just kinda wish in the other files, it would be clearer. So in this file it's ok, but where it's being implemented, because of the way we import it, it's hard to know if this is some local function or something imported, and then what package it's coming from. 

But I'm ok with the name for now to be honest, not a big deal.

src/actions/organizationActions.js
oponder: Just to keep all the localStorage related things together. 

Another bonus is that then the `app.selectedOrganization` string gets defined in only one place, because there is another 

` localStorage.setItem('app.selectedOrganization', selectedOrganization);`

further down in this file.

src/utils/localStorageUtils.js
ograu: Aah, ok I didn't understand you... you're right... we can rename it when importing it also

ograu: Changed method names and added `To/FromStorage`suffix to all of them, following what we already have in the `fetch...` methods

PR #727, opened 2019-08-19
Network policy

glitchcrab:
towards giantswarm/giantswarm/issues/4383

(blocks giantswarm/g8s-oauth2-proxy/pull/11)

### DIFF
--- a/helm/happa-chart/templates/network-policy.yaml
+++ b/helm/happa-chart/templates/network-policy.yaml
@@ -1,0 +1,24 @@
+kind: NetworkPolicy
+apiVersion: networking.k8s.io/v1
+metadata:
+  namespace: {{ .Values.namespace }}
+  name: {{ .Values.name }}-network-policy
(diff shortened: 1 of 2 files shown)


glitchcrab:
This isn't going to work as Happa is exposed using nodeports.

glitchcrab:
tested - happa is still accessible.

helm/happa-chart/templates/network-policy.yaml
```diff
@@ -0,0 +1,20 @@
...
+spec:
+  podSelector:
+    matchLabels:
+      app: {{ .Values.name }}
+  ingress:
+  - ports:
+    - port: {{ .Values.port }}
+      protocol: {{ .Values.protocol }}
+  egress:
+  - to:
```
corest: Let's keep one format everywhere. afaik we dropped just `- {}` previously to allow all traffic

PR #728, opened 2019-08-20
Show an error if anything goes wrong while loading clusters.

oponder:


oponder:
Will self accept this one now. I want to do a release and I want this in the release.

PR #729, opened 2019-08-20
Fix errors getting cluster status

oponder:
This fixes something new that I don't have a good explanation for.

Suddenly the `getclusterstatus` call of the `giantswarm-js-client` started returning `{}` even though I could see in the network tab that a successful request had been made, and it contained real data in the response.

It's very weird, and couldn't be due to any changes or experiments in `giantswarm-js-client`, I haven't merged any of my changes there to master.

This PR also fixes some console warnings and output I noticed while debugging.

### DIFF
--- a/src/actions/clusterActions.js
+++ b/src/actions/clusterActions.js
@@ -153,7 +153,6 @@ function clustersLoadV4(token, scheme, dispatch, getState) {
       });
     })
     .then(clustersObject => {
-      console.log(clustersObject);
       const lastUpdated = Date.now();
       dispatch(clustersLoadSuccessV4(clustersObject, lastUpdated));
     })
--- a/src/actions/clusterActions.js
+++ b/src/actions/clusterActions.js
@@ -128,9 +128,16 @@ function clustersLoadV4(token, scheme, dispatch, getState) {
             return { id: clusterId, ...mockedStatus };
           } else {
             return clustersApi
-              .getClusterStatus(scheme + ' ' + token, clusterId)
-              .then(clusterStatus => {
-                return { id: clusterId, ...clusterStatus };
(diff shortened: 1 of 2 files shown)


src/actions/clusterActions.js
```diff
@@ -128,9 +128,16 @@ function clustersLoadV4(token, scheme, dispatch, getState) {
             return { id: clusterId, ...mockedStatus };
           } else {
             return clustersApi
-              .getClusterStatus(scheme + ' ' + token, clusterId)
-              .then(clusterStatus => {
-                return { id: clusterId, ...clusterStatus };
+              .getClusterStatusWithHttpInfo(scheme + ' ' + token, clusterId)
```
pipo02mix: Why this works if you have changed the function name but I dont see in the commit the change in any function definition? 🤔

oponder: getClusterStatusWithHttpInfo is also available to use from the `giantswarm-js-client`, which is a generated client.

It returns the raw response, as well as the json parsed object.

So in the case i need to get the raw response and parse it myself. For a reason that I don't understand yet.

src/actions/clusterActions.js
```diff
@@ -128,9 +128,16 @@ function clustersLoadV4(token, scheme, dispatch, getState) {
...
+              .then(data => {
+                // For some reason we're getting an empty object back.
+                // The Giantswarm JS client is not parsing the returned JSON
+                // and giving us a object in the normal way anymore.
+                // Very stumped, since nothing has changed.
+                // So we need to access the raw response and parse the json
+                // ourselves.
```
pipo02mix: It is weird, but dont have any better idea, maybe a TODO prefix for clean in the future

PR #730, opened 2019-08-21
Fix params for cluster patch

oponder:
Fixes a bug when trying to scale clusters.


oponder:
A quick fix, merging without review.

PR #733, opened 2019-08-28
Simplify auth state

oponder:
This removes the need to constantly set scheme and token when doing an api request.

The api wrapper handles fetching a new token when a call fails, however with the way we are doing it in some chained requests, when this new token is acquired, we're still using the old token that we took from the state, instead of being able to rely on the instance of the API having its authentication set correctly.

By removing the authentication header we manually added to each call in the api-spec, we can use the javascript client in this more 'natural' way.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "copy-to-clipboard": "3.2.0",
     "deep-diff": "1.0.2",
     "emotion-theming": "^10.0.10",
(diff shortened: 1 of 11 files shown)


oponder:
see: https://github.com/giantswarm/api-spec/pull/165
and: https://github.com/giantswarm/giantswarm-js-client/pull/77

PR #735, opened 2019-09-02
Node pool name edit

ograu:
- Conditional styles for node pools in `ViewAndEditName`
- Trigger edit mode from dropdown

### DIFF
--- a/src/components/UI/view_edit_name.js
+++ b/src/components/UI/view_edit_name.js
@@ -192,5 +192,7 @@ function mapDispatchToProps(dispatch) {
 
 export default connect(
   undefined,
-  mapDispatchToProps
+  mapDispatchToProps,
+  undefined,
+  { forwardRef: true }
 )(ViewAndEditName);
--- a/src/components/UI/view_edit_name.js
+++ b/src/components/UI/view_edit_name.js
@@ -17,7 +17,10 @@ const FormWrapper = styled.div`
     padding: 0px 5px;
     width: 320px;
     margin-right: 5px;
-    font-size: 85%;
+    font-size: ${props => (props.entityType === 'node pool' ? '15px' : '85%')};
+    line-height: ${props =>
+      props.entityType === 'node pool' ? '1.8em' : null};
+    margin-bottom: ${props => (props.entityType === 'node pool' ? 0 : null)};
   }
   .btn[type='submit'] {
     display: inline;
@@ -25,7 +28,11 @@ const FormWrapper = styled.div`
   .btn-group {
     float: none;
     margin-left: 4px;
-    top: -2px;
+    top: ${props => (props.entityType === 'node pool' ? 0 : '-2px')};
+  }
+  button {
+    font-size: ${props => (props.entityType === 'node pool' ? '13px' : null)};
+    padding: ${props => (props.entityType === 'node pool' ? '4px 10px' : null)};
   }
 `;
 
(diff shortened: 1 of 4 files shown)


src/components/UI/view_edit_name.js
```diff
@@ -17,15 +17,22 @@ const FormWrapper = styled.div`
     padding: 0px 5px;
     width: 320px;
     margin-right: 5px;
-    font-size: 85%;
+    font-size: ${props => (props.entityType === 'node pool' ? '15px' : '85%')};
```
oponder: This seems a bit like a code smell to me.

Is there some way we can achieve the same by giving the element a className, and then attach these css differences to that className? (But still keeping that class related css in a css-in-js way?)

Alternatively, is there some way we can move the css that is important in a specific context out of this component, and apply it to this component from the place where it gets used?

I actually think `entityType` is a problem, and we should be careful depending on it.

ograu: yeah, I don't like it either... I am changing it with a `className` approach.

As about extracting css, we can pass an styles object as a prop and then apply it in `ViewAndEditName` but it makes things worse IMO basically because you will have the styles applied to an element, eg an `input`, splitted into different components, and so it will be more difficult to manage them.

ograu:
ping @oponder WDYT?

src/components/UI/view_edit_name.js
oponder: I was thinking of wrapping the component with another styled component that would then select on the child. But then that styled wrapping component would live in the nodepools view, where it is actually used, instead of teaching viewandeditname about specific things like what a nodepool is and what to do depending on what it is.

The latest edit is good, but I'm wondering if we can take it a step further. I'm concerned about the boundaries between components, and `ViewAndEditName` shouldn't know anything about node pools. 

So in my point of view, the 'implementor' of a component _can_ know details of the thing it is implementing.

But it is a bad pattern to teach the dependency about things outside of its domain.

So I would almost prefer to pass a styles object as a prop, because that is a more generic way of teaching this component that it should be able to change its style based on something that the implementor wants.

src/components/UI/view_edit_name.js
```diff
@@ -185,5 +201,7 @@ function mapDispatchToProps(dispatch) {
 
 export default connect(
   undefined,
-  mapDispatchToProps
```
oponder: I don't think this component dispatches anything actually anymore right?

src/components/UI/view_edit_name.js
ograu: I understand what you mean, but I am not sure if we have to think of components as dependencies...

> I was thinking of wrapping the component with another styled component that would then select on the child. But then that styled wrapping component would live in the nodepools view, where it is actually used, instead of teaching viewandeditname about specific things like what a nodepool is and what to do depending on what it is.

If we do what you propose, then we should do the exact same thing with all the other use cases, otherwise we will have a css specificity issue... And we will have an extra wrapper for this with specific styles, so we were splitting styles in different places. 

>where it is actually used

Keep in mind that these styles are used in `ViewAndEditName` when it is embeded in a NPS cluster view. And this is important, specially because with emotion we are using scoped styles.

>So I would almost prefer to pass a styles object as a prop, because that is a more generic way of teaching this component that it should be able to change its style based on something that the implementor wants.

If we do this, we would have to split this object and embed each chunk into the different elements, also to avoid specificity issues.

IMO when the _correct_ approach/pattern implies more headaches that the _simpler_ approach maybe we should stick with the simpler one, keep an eye on it and change it when we feel we have to do so.

WDYT?

PR #736, opened 2019-09-03
Remove nodeport from service

glitchcrab:
- remove unecessary nodeport from service.
- migrate to using values instead of hardcoding.

PR #737, opened 2019-09-03
Add migration for service to clusterIP

glitchcrab:
Required by #736 - the service must be removed as it cannot be amended in place.


glitchcrab:
Finally - success!

```
$ kubectl get svc -l app=happa
NAME    TYPE        CLUSTER-IP       EXTERNAL-IP   PORT(S)    AGE
happa   ClusterIP   172.31.129.115   <none>        8000/TCP   52s
```

PR #738, opened 2019-09-03
Fix cluster status call by using gs clusters api instead of separate client

tfussell:
I noticed that Happa was calling `GET:/v4/clusters/:id/` and `GET:/v4/clusters/:id/status/` every 30 seconds on the detail view. The status call was succeeding on initial page load and then failing with a 401 afterwards. It seems there are two different methods for getting the cluster status, one using `giantswarm-js-client` and the other using an internal client in Happa itself. That Happa-internal client was being called with one argument instead of two leading to the 401. Rather than fix it, I thought it made sense to get rid of it completely and use `giantswarm-js-client` for this.

I don't have the context behind this structure so let me know if this seems wrong. I also haven't tested the code as I haven't taken the time to set up a dev environment for Happa. Happy to test it tomorrow if the idea looks correct.


ograu:
Hey Thomas! Thanks for the PR :-)

TBH I don't know exactly why are we using this client in `lib/`folder. Probably we are using it wrong as you mentioned. But I think that the problem with this endpoint is related to the local API itself not returning anything for this call. If you `curl` it it returns a `404`.

That's why we are using (momentarily) a mocked response in the local environment for this first call and this is the reason why It looks that it is succeeding while actually there's no call at all.

tfussell:
Interesting. I see the mocked status for the development environment. It seems that the status call is working fine on `gauss` and `ginger` though. Are you running the API locally or is this running against one of the test clusters? What `curl` request returns 404?

Just FYI, I'm digging into cluster status for https://github.com/giantswarm/giantswarm/issues/6805

oponder:
Interesting. I'm not sure what the reason is for this client either anymore. Maybe introduced to get something working before the changes were ready in api-spec and giantswarm-js-client, and then forgotten.

I certainly didn't notice it.

I think this change makes sense though.

ograu:
> Interesting. I see the mocked status for the development environment. It seems that the status call is working fine on `gauss` and `ginger` though. Are you running the API locally or is this running against one of the test clusters? 

Yes, I am running the testing API locally

>What `curl` request returns 404?

I'm logging in using `gsctl` and then copy/pasting its auth token:

 `curl -sS -H "Authorization: giantswarm d597e9e4-19ab-46a2-b739-21050e1bc63b" http://localhost:8000/v4/clusters/3zqsq/status/ | jq . `

tfussell:
I think I see the issue now. The [cluster status endpoint](https://github.com/giantswarm/api/blob/master/server/endpoint/v4/cluster/status/endpoint.go#L86) returns 404 if statusSearcher returns a `not found` error. And [the cluster service config field `StatusPathFormat`](https://github.com/giantswarm/api/blob/master/service/service.go#L430) is set to a path like `"apis/provider.giantswarm.io/v1alpha1/namespaces/default/awsconfigs/%s/status"`. Since the latest `aws-operator` doesn't use `AwsConfig`, it makes sense that we would get a 404 here. The solution is can be any of the following (in order of increasing complexity and elegance): 
1. Create a legacy (i.e. v4) cluster
2. Only call `status` endpoint for v4 clusters
3. For v5 clusters, get cluster status from the cluster CR status directly

These are probably outside the scope of this story, just wanted to close the loop on the current state of things.

oponder:
@tfussell would you do the honors and merge this PR :)

tfussell:
@oponder Yes, sorry, wanted to take some time to actually run Happa and the API locally so I can test this before merging. It's probably fine as is, but those are famous last words... I'll get it merged today.

oponder:
For what it's worth, I tried it out :) and it works for me. But always nice to get a second check from the author too

tfussell:
Sorry for the delay. I finally took the time to learn how to do a branch deployment on a test cluster and revert it. Merging now.

PR #740, opened 2019-09-05
Extract UI component: Cluster dashboard item

oponder:


src/components/home/index.js
```diff
@@ -1,14 +1,13 @@
-import * as clusterActions from 'actions/clusterActions';
```
oponder: I actually prefer just a good ol dispatch of an action instead of doing the bindActionCreators thing. I think I used `bindActionCreators` when I started because I thought it was needed, but unless anyone disagrees, I'd like to avoid that from now on.

oponder: see also related change on line 43

src/components/home/index.js
```diff
@@ -84,7 +83,10 @@ class Home extends React.Component {
       <DocumentTitle title={this.title()}>
         {
           <div>
-            {this.props.selectedOrganization ? (
+            <EmptyStateDisplay
```
oponder: I'm not sure about this EmptyStateDisplay component refactoring. 

It seems to run all the code in the child component whereas the ternary would skip executing it.

src/components/UI/__tests__/cluster_dashboard_item.js
```diff
@@ -0,0 +1,45 @@
...
+import { render } from '@testing-library/react';
+import React from 'react';
+import ClusterDashboardItem from 'UI/cluster_dashboard_item';
+
+it('renders without crashing', () => {
+  const div = document.createElement('div');
+  render(<ClusterDashboardItem />, div);
+});
+
+it('has links to the cluster detail page', () => {
```
oponder: @ograu can we pair and implement these tests together? I'm not sure where to start.

src/components/UI/cluster_dashboard_item.js
```diff
@@ -0,0 +1,260 @@
...
+        .diff(moment().utc()) / 1000
+    );
+
+    return age < 30 * 24 * 60 * 60;
+  }
+
+  /**
+   * Returns the amount of memory in this cluster.
+   */
+  memory() {
```
oponder: this and other function like `cpus` `storage` and `numNodes` used to be procedural code at the start of the `render()` method.

I think we should try to keep the render method as clean as possible, so i moved it out into individual functions.

It does mean we call these functions about 2 more times (since they are used in two places in the render method) but if that really becomes a performance problem, it would be easy to memoize this function.

But for now let's not premature optimise?

ograu: Instead memoization or as a way of doing it, let's use state. That's what it is for. We can execute this only when `cluster` prop change, no need to check for deep nested properties thanks to Immer.

src/components/UI/__tests__/cluster_dashboard_item.js
ograu: Yes, just ping me on slack please

src/components/home/index.js
ograu: Never used it, I see this from the docs: 

_The only use case for bindActionCreators is when you want to pass some action creators down to a component that isn't aware of Redux, and you don't want to pass dispatch or the Redux store to it._

And I think it's not the case here, so 👍

oponder:
> Sorry but I don't see the point of storing components like this one in UI folder.

Maybe we can break this one down further later, but it's a starting point. I do agree it's a bit of a big component though.

What's important for me here is the removal of dispatch and redux and other stack specific things. As well as the move to emotion.

This also makes it easier to test in "isolation", and to continue refactoring it into smaller pieces if we want to.

And also, one big thing I am trying to go towards is to integrate these components into Framer X, so that we can rapidly prototype and view different UI states. 

So I want the "interface" to UI components to be just purely props and nothing that fetches something directly from the redux store. (So no mapStateToProps)

ograu:
Ok if FramerX is the reason and we are getting a lot of value from it

src/components/UI/__tests__/cluster_dashboard_item.js
```diff
@@ -0,0 +1,60 @@
...
+      </MemoryRouter>
+    </ThemeProvider>
+  );
+
+  const links = container.querySelectorAll('a');
+
+  expect(links[0]).toHaveAttribute('href', href);
+  expect(links[1]).toHaveAttribute('href', href);
+});
+
```
ograu: @oponder nitpick: I think that one step further to this test would be: `links lead the user to the cluster details page`  we would fire a click event and see if the router leads us to the `href` it is essentially the same test but with the benefit of also testing the router.

oponder: I did it, but I think maybe we're testing implementation detail at this point? 

This test doesn't wanna make sure that the memory router works, or that click events on `a` elements behave as expected. (Since that is a browser feature which I'm pretty sure we can rely on, and don't need to test)

ograu: Yes, you are now testing implementation details. You can easily avoid it by testing if any of the elements in the cluster details page is rendered.

ograu: > This test doesn't wanna make sure that the memory router works, or that click events on `a` elements behave as expected. (Since that is a browser feature which I'm pretty sure we can rely on, and don't need to test)

Maybe the router is broken for whatever reason and `Link` components are not rendering an `<a>`.

oponder: > Yes, you are now testing implementation details. You can easily avoid it by testing if any of the elements in the cluster details page is rendered.

but then I have to render the full app... i think like this is ok for now right?

ograu: And there's another thing to keep in mind: our users are using our application __with__ the router, the store, etc. They don't care about it, but what I mean is that if we want to test Happa in a way that ressembles how our users are using it we should go for integration tests in most cases.

oponder:
closing this since it got stale. will try to make it again from master at a later date.

marians:
Do you want to delete the branch, too?

BTW, I am currently working on that component.

oponder:
Yeah that's why I closed this PR. 

A lot has happened in the meanwhile and it doesn't make sense to try and merge it anymore I think.

PR #741, opened 2019-09-05
NPs clusters nodes scaling modal

ograu:
This PR is just for:

- Triggering the modal, and
- Updating values in the store.

Also:

- Simplified nodePolsPatch() thunk to accept clusterId as a parameter instead of fetching it

Logic in ScaleClusterModal component has to be adapted to node pools after this is merged.
Tests are coming after this PR.


src/components/cluster/detail/scale_cluster_modal.js
```diff
@@ -88,38 +93,68 @@ class ScaleClusterModal extends React.Component {
...
-              messageTTL.SHORT
-            );
-
-            this.props.clusterActions.clusterLoadDetailsSuccess(patchedCluster);
-          })
-          .catch(error => {
-            this.setState({
-              loading: false,
-              error: error,
+        if (this.state.nodePool) {
```
oponder: I'm worried about these kinds of conditionals going throughout the UI.

I'd rather have an action called like 'nodePoolorClusterPatch' which takes three parameters (clusterId, Nodepool, scaling) and where nodepool can be null or undefined or something, and then the conditional to decide which api method to call would live in there?

Or perhaps we should make a dedicated `scale_nodepool_modal` instead?

src/components/cluster/detail/scale_cluster_modal.js
```diff
@@ -88,38 +93,68 @@ class ScaleClusterModal extends React.Component {
...
+          // Node Pools
+          const scaling = {
+            Min: this.state.scaling.min,
```
oponder: Is this supposed to be capitalized?

ograu: I can store `Min` property as `min` but the ideal solution is to have `min` fetched as `min` instead as `Min`,  and `max intead of `Max`.

ping @marians

ograu:
> Hey @ograu, I think in the spirit of avoiding patterns and things that would be bad in the long run, this review mentions something about the nodepool conditional.
> 
> Do you agree we should avoid that in the UI?

Totally, thank you for not letting me do this kind of things. Sometimes I want to go faster than I should... 🙇 

> The second alternative I propose is to make a second modal just for scaling node pools. I think sometimes duplication is the lesser evil, especially when we might have some change in the way the nodepools scaling interaction needs to be from the cluster scaling interaction.
> 
> Also at somepoint all clusters will be nodepool clusters, (though that might take longer than we think). And then it becomes easy to just delete the cluster_scaling_modal, instead of trying to untangle the conditonals.

src/components/cluster/detail/scale_cluster_modal.js
ograu: As I said thank you for not letting me do this kind of things, but I am rethinking it: one of my main concerns now is get things done here, don't wanna be a blocker in the NPs roadmap, and taking into account that no matter the solution we choose now, we will have to refactor it in the future.... ¿What if we leave this conditional here and refactor it when all clusters are NPs clusters?

Despite everything we discussed, one thing I like about this solution is that is very clear what's going on and shouldn't be too difficult to modify it when needed.

WDYT?

src/components/cluster/detail/scale_cluster_modal.js
oponder: Not sure I understand what's going on here.. so nodepools store their min/max as Min and Max?

Oh! I see it now. These are the parameters for the scaling action. Hmm I think there might be something wrong in the api spec then, or api implementation?

src/components/cluster/detail/scale_cluster_modal.js
oponder: What about making a second modal just for scaling node pools?

Then possibly extracting what they have in common if you can find the time?

I'm just concerned that it becomes easy to add another thing, and another thing to this modal, and soon it becomes difficult to make sure cluster scaling works AND node pool scaling works.

Also means we have to test cluster scaling each time we make a change related to node pool scaling.

src/components/cluster/detail/scale_cluster_modal.js
marians: Is this still not fixed in the API? The JSON attributes have to be lowercase of course. Let's get this fixed ASAP.

src/components/cluster/detail/scale_cluster_modal.js
ograu: ok, I'm going to try to do it

src/components/cluster/detail/scale_cluster_modal.js
marians: See https://github.com/giantswarm/api/pull/842

ograu: I can approve it if you want... but can't judge the code

ograu: Thanks for the quickness! :-)

marians: https://github.com/giantswarm/api/pull/842 is merged

src/components/cluster/detail/scale_cluster_modal.js
ograu: done ping @oponder

src/components/cluster/detail/scale_node_pool_modal.js
```diff
@@ -0,0 +1,439 @@
...
+        warnings.push(
+          <CSSTransition
+            classNames='rollup'
+            enter={true}
+            exit={true}
+            key={1}
+            timeout={this.rollupAnimationDuration}
+          >
+            <p key='node-removal'>
+              <i className='fa fa-warning' /> The cluster currently has{' '}
```
oponder: Would instances of 'cluster' have to be changed to 'nodepool' in the text here?

src/components/cluster/detail/scale_node_pool_modal.js
```diff
@@ -0,0 +1,439 @@
...
+        warnings.push(
+          <CSSTransition
+            classNames='rollup'
+            key={2}
+            timeout={this.rollupAnimationDuration}
+          >
+            <p key='node-removal'>
+              <i className='fa fa-warning' /> You are about to enforce the
+              removal of {diff === 1 ? 'one node' : diff + ' nodes'}. Please
+              make sure the cluster has enough capacity to schedule all
```
oponder: in this we're scaling a nodepool, does this warning still make sense? ping @marians as well

src/components/cluster/detail/scale_node_pool_modal.js
ograu: Changes committed and pushed

Yeah there are more changes to do here, I just wanted to split them into various PRs in order to make PRs more reviewable. I stated that in my first comment, sorry if I wasn't clear enough

> This PR is just for:
> 
> * Triggering the modal, and
> * Updating values in the store.
>

PR #742, opened 2019-09-05
Node Pools tests

ograu:
This is just a draft PR for discussion. My intention was to actually submit tests, but here's what I've done instead:

- Added conf in .babelrc in order for jest to support async in tests
- Added global config object in jest.config.js because otherwise tests won't find config
- Added tests

Well, actually just one test. The others are failing or not yet implemented because I don't find the way to test connected components. I think that we should do this (but don't want to invest more time in it now) :

1. Learn how to mock the redux store (with thunks) and/or API calls
2. Extract providers and mocked data into reusable wrapping functions so we can keep tests as simple and short as possible

Hopefully once we have solved this two points tests are going to be easier to write.

PS: alternative approach: use Cypress for these kind of testing where different components are involved.


ograu:
ping @oponder

ograu:
I actually was doing it right with the store the problem here are others: 

- [x] localStorage in Node (solved)
- [ ] auth0: how can we test/bypass this?

oponder:
There are also normal password logins, so we could skip talking to auth0.

jest.config.js
```diff
@@ -15,4 +16,17 @@ module.exports = {
     ),
   },
   testPathIgnorePatterns: ['/node_modules/', 'node_modules_linux'],
+  globals: {
+    // window.config object will now be available in all tests
+    config: {
```
oponder: So I guess this is now duplicated between index and this config file. Which means we have to maintain it in two places. Is there any idea you have for making a single source of truth for this config info?

src/reducers/appReducer.js
```diff
@@ -34,6 +34,7 @@ const appReducer = produce((draft, action) => {
       return;
 
     case types.LOGIN_SUCCESS:
+      console.log(action.userData);
```
oponder: todo: remove debug output when you're ready

jest.config.js
ograu: Not right now.
I tried to extract it to its own module in `lib` folder, but then I think that `scripts/start.sh` wasn't able to find the `config` object, so I reverted it.
We definitely have to find a way to extract this data outside of `index.html`

ograu:
I think this is ready for tomorrow, I've tried different approaches and end up with the following:

- I've mocked the store( in `test_utils/initialState.js`), we can modify this object to fit our needs
- I've mocked some of the action creators/thunks (in `actions/__mocks__`), probably we will have to mock all or most of them, because we don't want to hit API endpoints
- I'm using a custom render function `renderRouteWithStore()` that does exactly this: render whatever the browser render in a specific route with the benefits of including the store provider and the theme provider

We can, of course, test non-conected components in isolation or even we maybe could test connected components in isolation exporting a non-connected version. 

My idea (and this setup) is to test just components, then we would have to add some stuff if we want to test also redux (actions and reducers).

ograu:
We are not making those calls, in theory... I'm afk now. This is in circle CI? Cause I think I' ve not seen this in my machine

ograu:
I forgot to mock `catalogsLoad`, I've added it and now these errors are gone. Weird that it just popped up in CircleCI

PR #745, opened 2019-09-06
NPs clusters in cluster list

ograu:
Marian has added the mocked NP cluster to the cluster list response in the `add-mock-cluster-to-list` API branch so now we can have a `clustersLoad` method that gets all clusters (v4 and v5)

In this PR I have:

- Modified the `clustersLoad()` function to fetch everything
- Removed logic for introducing `m0ckd` v5 cluster manually in the store
- Adapt `ClusterDashboardItem` to the modifications made

⚠️ ⚠️  From now on (once this is merged) in the local testing api we must use the `add-mock-cluster-to-list` branch. ⚠️ ⚠️

### DIFF
--- a/src/actions/nodePoolsActions.js
+++ b/src/actions/nodePoolsActions.js
@@ -6,20 +6,21 @@ import GiantSwarm from 'giantswarm';
 const nodePoolsApi = new GiantSwarm.NodepoolsApi();
 
 //Loads all node pools for all node pools clusters.
-export function nodePoolsLoad(clusters) {
+export function nodePoolsLoad() {
   return async function(dispatch, getState) {
-    // const clusters = getState().entities.clusters.nodePoolsClusters || [];
+    const nodePoolsClustersId =
+      getState().entities.clusters.nodePoolsClusters || [];
 
     return Promise.all(
-      clusters.map(async cluster => {
-        const nodePools = await nodePoolsApi.getNodePools(cluster.id);
+      nodePoolsClustersId.map(async clusterId => {
+        const nodePools = await nodePoolsApi.getNodePools(clusterId);
 
         // Receiving an array-like with weird prototype from API call,
         // so converting it to an array.
         let nodePoolsArray = (Array.from(nodePools) || []).map(np => np.id);
 
         // Dispatch action for populating nodePools key inside cluster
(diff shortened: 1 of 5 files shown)


src/actions/nodePoolsActions.js
```diff
@@ -8,10 +8,11 @@ const nodePoolsApi = new GiantSwarm.NodepoolsApi();
 //Loads all node pools for all node pools clusters.
```
marians: nitpick :)

```suggestion
// Loads all node pools for all node pools clusters.
```

src/actions/nodePoolsActions.js
```diff
@@ -8,10 +8,11 @@ const nodePoolsApi = new GiantSwarm.NodepoolsApi();
...
 export function nodePoolsLoad() {
   return async function(dispatch, getState) {
-    const clusters = getState().entities.clusters.nodePoolsClusters || [];
+    const nodePoolsClustersId =
```
marians: Do we really do `nodePoolsClustersId`, not `nodePoolsClustersID` ?

ograu: What do you mean? Do we have any format or styling rule for this? I think I always have written `id`...

ograu:
ping @oponder 
Are you ok with me merging this?

src/actions/nodePoolsActions.js
marians: Not really a rule yet. Worth discussion I think. govet reminds devs to uppercase abbreviations like "ID", "API" in variable names, so e. g. `myAPIEndpoint` instead of `myApiEndpoint`. If you want `nodePoolsClustersId`, ignore me.

ograu:
Thanks to this code we will be able to develop receiving a mix of v4 and v5 clusters when we call `clustersLoad()`. This is exactly how it is going to work when the API is ready, so this way we are able to work in the right direction instead of having to manually inject v5 mocked cluster

When not in development it will just load v4 clusters.

When the v5 API endpoints are ready, we just have to remove the conditional in line 179 in `clustersActions`:

```javascript
if (window.config.environment === 'development') {
      const clusters = await Promise.all(
        v5Clusters.map(cluster => clusterDetailsV5(dispatch, cluster))
      );
      ...
```

ograu:
And all that depends on the development API branch is inside this conditional also.

PR #746, opened 2019-09-16
Set the cpu/ram default to 4/4 and min to 2/3

oponder:
This is a quick fix to make sure that Happa can't make clusters that are invalid. 

Once https://github.com/giantswarm/api/pull/855 is merged , kvm clusters with less than 2CPU and 3GB RAM are invalid, so this makes sure people can't even try to create smaller clusters.

Once https://github.com/giantswarm/api-spec/pull/167 is _implemented_ I will make this value dynamic, populated by what the info endpoint returns.

But for now this is a quick fix to stop people from making invalid clusters.

See also: https://gigantic.slack.com/archives/C76JX6YLQ/p1567516000045900?thread_ts=1567509332.042700&cid=C76JX6YLQ for the 4/4 default.

### DIFF
--- a/src/components/cluster/new/view.js
+++ b/src/components/cluster/new/view.js
@@ -456,7 +456,7 @@ class CreateCluster extends React.Component {
                         <NumberPicker
                           label='CPU Cores'
                           max={999}
-                          min={1}
+                          min={2}
                           onChange={this.updateCPUCores}
                           stepSize={1}
                           value={this.state.kvm.cpuCores.value}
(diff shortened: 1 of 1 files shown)


oponder:
> I think it would be nice to have this configuration variables in a config file instead of here

Yeah it will come from the info endpoint (`/v4/info`), and be controllable per installation: https://github.com/giantswarm/installations/pull/887

but that needs to be implemented in the API still.

PR #747, opened 2019-09-16
Node Pool deletion

ograu:
In this PR I added the logic for deleting a Node Pool. 

The endpoint is not ready, so if you try to click on `delete` you will get a `405` ...
Do you think it makes sense to merge it?

I think we can merge it since this view isn't rendered in production and we will test it before releasing it.


ograu:
Yes, I totally would prefer using local state instead redux store, it's simpler. Using redux store is overkill for modals.

Now I am focusing on getting things done with what we actually have in Happa, but after node pools, this modal refactor is one of the things I have in mind.

PR #748, opened 2019-09-16
Fix on refresh nps clusters

ograu:
This is a continuation of https://github.com/giantswarm/happa/pull/745 

I have deleted actions regarding v4 and v5 clusters and added a generic one that will be responsible for:

- Adding v4 and v5 clusters to store, and
- Notifying the app when both types if clusters are correctly added to store

Without this changes, if you refresh manually a NPs Cluster details view Happa will break. This is because `CLUSTERS_LOAD_SUCCESS_V4` action is dispatched, so Happa thinks everything is ready for rendering, but it is perfectly possible that v5 clusters are still being fetched and hence not actually ready.


src/actions/actionTypes.js
```diff
@@ -58,11 +58,8 @@ export const CLUSTER_LOAD_STATUS_SUCCESS = 'CLUSTER_LOAD_STATUS_SUCCESS';
 export const CLUSTER_LOAD_STATUS_NOT_FOUND = 'CLUSTER_LOAD_STATUS_NOT_FOUND';
 export const CLUSTER_LOAD_STATUS_ERROR = 'CLUSTER_LOAD_STATUS_ERROR';
 
-export const CLUSTERS_LOAD_SUCCESS_V4 = 'CLUSTERS_LOAD_SUCCESS_V4';
-export const CLUSTERS_LOAD_ERROR_V4 = 'CLUSTERS_LOAD_ERROR_V4';
-
-export const CLUSTERS_LOAD_SUCCESS_V5 = 'CLUSTERS_LOAD_SUCCESS_V5';
-export const CLUSTERS_LOAD_ERROR_V5 = 'CLUSTERS_LOAD_ERROR_V5';
+export const CLUSTERS_LOAD_SUCCESS = 'CLUSTERS_LOAD_SUCCESS';
```
oponder: Nice that we're folding things back into one path. Awesome

src/actions/clusterActions.js
```diff
@@ -166,7 +167,7 @@ export function clustersLoad() {
     });
 
     const lastUpdated = Date.now();
-    dispatch(clustersLoadSuccessV4(v4ClustersObject, lastUpdated));
+    // dispatch(clustersLoadSuccessV4(v4ClustersObject, lastUpdated));
```
oponder: Would like to avoid commented out code in the codebase.

ograu: sorry... I will end up writing a git hook for this because I am doing it so frequently!

PR #749, opened 2019-09-16
Removed setState from constructor in ClusterDetailView

ograu:
removed `setState()` calls in constructor cause it is an antipattern (see https://reactjs.org/docs/react-component.html#constructor) and removed constructor itself.


oponder:
Thanks @ograu !

PR #750, opened 2019-09-16
NPs text with link in NPs Clusters

ograu:
ping @marians 

- Which is the link for _Read more_ ?
- If I stick with the font size we are using for this kind of texts and the button dimensions we are using the result is not very pleasant to my eyes:

Captura de pantalla 2019-09-16 a las 17 25 22

Maybe we could downsize the font to 14px. WDYT?


marians:
The URL for "Read more" is https://docs.giantswarm.io/basics/nodepools/ - but the page isn't there yet.

marians:
I agree that font size and button don't fit well. Also the line height is too small for the font size. Feel free to try 14.

ograu:
I like with 14px more:

Captura de pantalla 2019-09-23 a las 11 34 54

ograu:
@marians Are you ok with me merging this?

marians:
As long as it it not appearing anywhere for users, yes.

ograu:
V5 clusters won't be in production until we decide so. In production no node pool clusters will be dispatched to the store, so no node pools clusters in Happa:

```javascript
// clusterActions.j
if (window.config.environment === 'development') {
    
   // ...
   dispatch(
     clustersLoadSuccess(
        v4ClustersObject,
        v5ClustersObject,
        nodePoolsClusters,
        lastUpdated
      )
    );

 } else {
   // Dispatch with an empty object for v5Clusters and an empty array for node pools clusters.
   dispatch(clustersLoadSuccess(v4ClustersObject, {}, [], lastUpdated));
 }

PR #751, opened 2019-09-16
Removed sorting props alfabetically

ograu:
I've tried to get used to this linting rule, I've never liked it but wanted to give it a try.
Now, after using it for a while, I can say without hesitance that I don't like because:

- it forces you to use a complete arbitrary order: props starting with a can be the less important ones, and

- it collides with Emmett, so every time I use it I have to reorder all attributes

Do you think we can delete it? Or on the contrary you think it is useful?


oponder:
I’m ok to remove it

ograu:
@marians are you ok with this?

oponder:
Lets get this merged in as well already?

PR #752, opened 2019-09-17
Fix menu button in AWSInstanceTypeSelector

ograu:
Removed the space under the menu button and centered hamburger. What we had:

Captura de pantalla 2019-09-17 a las 5 57 47

Now:

Captura de pantalla 2019-09-17 a las 6 01 35


oponder:
This seems to happen when you have Roboto installed in your system instead of taking the one that is downloaded via the browser.

What happens now for me in this branch is the opposite somehow:

Screenshot 2019-09-17 at 9 28 29 AM

I think we need a better solution than I made with these hardcoded pixel values.

ograu:
is it fixed now for you?

marians:
❤️

PR #754, opened 2019-09-17
Add node pool

ograu:
Hey @marians I need you to take a look at this screenshot:

- Are you ok with this background color? I am not super excited about it but I dont like any of the shades for background we have in `theme.js` either. Which color you would like to have here?

- Where does _Select distinct availability zones_ link point to?

Captura de pantalla 2019-09-17 a las 14 00 17

thanks!


src/components/cluster/detail/add_node_pool.js
```diff
@@ -0,0 +1,220 @@
...
+    return (
+      <WrapperDiv>
+        <h3 className='table-label'>Add Node Pool</h3>
+        <FlexColumnForm>
+          <label htmlFor='name'>
+            <span className='label-span'>Name</span>
+            <input id='name' type='text'></input>
+            <p>
+              Pick a name that helps team mates to understand what these nodes
+              are here for. You can change this later. Each pool also gets a
```
marians:
```suggestion
              are here for. You can change this later. Each node pool also gets a
```

src/components/cluster/detail/add_node_pool.js
```diff
@@ -0,0 +1,220 @@
...
+            <FlexWrapperDiv>
+              <AWSInstanceTypeSelector
+                // allowedInstanceTypes={this.props.allowedInstanceTypes}
+                allowedInstanceTypes={['m3.large', 'm5.large']}
+                // onChange={this.updateAWSInstanceType}
+                readOnly={false}
+                // value={this.state.aws.instanceType.value}
+                value='m3.large'
+              />
+              <p>2 CPU cores, 8GB RAM each</p>
```
marians: What is this hard coded paragraph here for?

marians: I just noticed that this is a draft and WIP. Never mind!

src/components/cluster/detail/add_node_pool.js
```diff
@@ -0,0 +1,220 @@
...
+              style={{ background: 'red' }}
+              type='button'
+            >
+              Cancel
+            </Button>
+          </FlexWrapperDiv>
+          <p style={{ fontSize: '16px' }}>
+            Note that it takes around 15 minutes on average until a new node
+            pool is fully available.
+          </p>
```
marians: FYI: Initially we wont have that data yet, so we should remove the entire paragraph `Note that it takes around`...

I'm wondering about the `<p style={{ fontSize: '16px' }}>`. Why is that necessary? 16px should be the standard font size.

ograu: Because I'm overwriting it in the other `<p>`s

src/components/cluster/detail/AddNodePool.js
```diff
@@ -0,0 +1,488 @@
...
+        color: ${props => props.theme.colors.white1};
+        font-weight: 400;
+      }
+      & > div:nth-of-type(2) {
+        display: none;
+      }
+    }
+  }
+`;
+
```
ograu: This component is basically copy/pasted from `CreateCluster` component and adapted

src/components/cluster/detail/AddNodePool.js
```diff
@@ -0,0 +1,488 @@
...
+        value: this.props.defaultMemorySize,
+        valid: true,
+      },
+      diskSize: {
+        value: this.props.defaultDiskSize,
+        valid: true,
+      },
+    },
+  };
+
```
ograu: Do we need kvm and azure? I mean, AFAIK we don't need them now, but we will need them in the future. Then ¿It makes sense to keep them here?

ograu:
Continued in https://github.com/giantswarm/happa/pull/754

PR #755, opened 2019-09-17
Prepare for secrets. App detail modal refactor

oponder:
These are some changes to the app detail modal to prepare it for the secrets file uploading.

- Extract the YAML uploading to its own component
- Fix a bug where clicking outside of the modal to close it would not return the modal to its initial pane
- Fix some rendering bugs with the modal (div cannot be inside p)
- Make sure we don't try to change the state when the modal is not on the screen


oponder:
I started this by adding the secrets stuff, but then realised I needed to refactor some things in the modal.

So the commit history mentions secrets and stuff, but I removed that from this PR towards the end.

oponder:
@ograu the underscore in function naming is addressed in the follow up PR: https://github.com/giantswarm/happa/pull/757

PR #757, opened 2019-09-18
Add Secret management to AppDetailModal

oponder:
This adds a control similar to the one for uploading `ConfigMaps` to the `AppDetailModal`, but for uploading a `Secrets` file instead.

Screenshot 2019-09-18 at 2 03 14 AM

Ping @giantswarm/sig-ux for interface feedback as well. Now that we have two files to upload, does extending what we had before make sense? Should there some totally different approach? Personally, for me it still makes sense. 

Though I'd like to add some question mark bubbles perhaps next to the labels for ConfigMap and Secret with a bit more explanation about the files, how they should be formatted, and possibly a way to download a sample file. Though that is out of scope of this PR.

This PR also changes the term from `user configuration configmap` to just `ConfigMap` and tries to standardise on that capitalisation as well. 

That's up for discussion by the way, should it be `configmap` or `ConfigMap` ? I'm not super attached to any but would like to have it consistent throughout Happa.


package.json
```diff
@@ -39,7 +39,7 @@
     "copy-to-clipboard": "3.2.0",
     "deep-diff": "1.0.2",
     "emotion-theming": "^10.0.10",
-    "giantswarm": "git+https://github.com/giantswarm/giantswarm-js-client.git",
+    "giantswarm": "git+https://github.com/giantswarm/giantswarm-js-client.git#app-secrets",
```
oponder: TODO: Merge https://github.com/giantswarm/giantswarm-js-client/pull/78 and change this back to master

PR #761, opened 2019-09-19
TESTS - Cluster management

ograu:


src/components/cluster/__tests__/cluster_managament.js
```diff
@@ -0,0 +1,106 @@
...
+  const payload = {
+    availability_zones,
+    scaling,
+    name,
+    owner,
+    release_version,
+    workers,
+  };
+
+  expect(mockClusterCreate).toHaveBeenCalledTimes(1);
```
oponder: I'd rather see here an expectation that we navigated to the page we expect, and that we see the title of the cluster as provided by the user and other expected details.

So that would be a matter of adding some details to the mocked `clusterCreate` call up there, taking the params from the payload, and dispatching a valid `CLUSTER_CREATE_SUCCESS` action with the parameters that the reducer expects.

Ideally though, if we could mock at the API response level, and not at the action level, that would be even better for this type of test.

Otherwise, what are we testing exactly here? That this button calls ends up calling an action with a certain payload? I think that is heavily in the realm of implementation detail, and not the right thing to be testing given the way that this test is set up.

ograu: Mmmmh... you are right!  😄 

It's just that with the previous router setup I wasn't able to click and push a route. Now we maybe can do that... I will give it a try tomorrow.

Thanks!

ograu: Not merging for now

ograu: I have added a `statusState` object to inject when running the scaling test. I'm adding it to the `initialState` object.

And also I modified `node_count_selector`, now `<label>`s wrap `<input>`s

ograu: ping @oponder You approved this, but as I did some modifications, asking you again for approval

src/components/cluster/__tests__/cluster_managament.js
```diff
@@ -0,0 +1,176 @@
...
+
+  await wait(() => {
+    const button = getByText(/launch new cluster/i);
+    fireEvent.click(button);
+  });
+
+  expect(getByTestId('cluster-creation-view')).toBeInTheDocument();
+});
+
+it(`renders the form when in new cluster route with default values and calls
```
oponder: this is kinda testing two things, and we're checking on things which aren't really under the setup of this test.

Like:  `expect(inputs[0]).toHaveValue('My cluster');`

It's ok to check if the form loads well and shows up right, in my tests I also check for output that isn't really under test, as a way to ensure I am on the page I expect. 

But I try not to lock down too much detail.

If there was a way for us to set the default new cluster name as part of the config, and you can manipulate that config at the start of this test ,and then verify that the form comes out as expected, then it would make sense to check on all of those things.. but in this case, I think we're doing a few too many expectations on things that could change, and wouldn't really mean this test should fail for that.

src/components/cluster/__tests__/cluster_managament.js
```diff
@@ -0,0 +1,176 @@
...
+  expect(mockClusterDelete).toHaveBeenCalledTimes(1);
+  expect(mockClusterDelete).toHaveBeenCalledWith(
+    initialState().entities.clusters.items[clusterId]
+  );
+});
+
+// The modal is opened calling a function that lives in the parent component of
+// <NodePoolDropdownMenu>, so we can't test it in isolation, we need to render
+// the full tree.
+it(`shows the scaling settings modal when the button is clicked with default values and calls
```
oponder: this test is pretty cool, I think it's testing more real functionality than the other one.

src/components/cluster/__tests__/cluster_managament.js
ograu: Yeah I agree that this test has to be simplified and maybe splitted into two. Rethinking it now.

I don't think we are going to have problems identifying what's going wrong in a test when it fails. Even if it is long.

ograu: I've  changed it a bit, anyway I will iterate on them because I have to look for rendered elements instead of checking the action creator is called now that we have `nock` and what you've done with it in the mix. Not merging for now.

PR #762, opened 2019-09-19
Fix history and initialState sharing.

oponder:
The way I made a weird singleton out of `history` and `store` was confusing things in the tests.

Also the way `initialState` was implemented means that it is shared between test cases, which confused things for me a bit as well.

These changes make it possible to really test the full 'stack' of the app, make sure that things navigate and render the right components.

### DIFF
--- a/src/components/cluster/detail/__tests__/cluster_details_node_pools_table.js
+++ b/src/components/cluster/detail/__tests__/cluster_details_node_pools_table.js
@@ -1,6 +1,7 @@
 import 'jest-dom/extend-expect';
 import { fireEvent, render, wait } from '@testing-library/react';
 import { nodePoolPatch as mockedNodePoolPatch } from 'actions/nodePoolActions';
+import { push } from 'connected-react-router';
 import { renderRouteWithStore } from 'test_utils/renderRouteWithStore';
 import { ThemeProvider } from 'emotion-theming';
 import initialState from 'test_utils/initialState';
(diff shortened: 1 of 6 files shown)


test_utils/initialState.js
```diff
@@ -1,4 +1,5 @@
-const initialState = {
+const initialState = function() {
```
oponder: this was my way of making sure we get a new initialState each time we call it.

src/stores/configureStore.js
```diff
@@ -1,23 +1,18 @@
 import { applyMiddleware, compose, createStore } from 'redux';
 import { routerMiddleware } from 'connected-react-router';
-import history from './history';
```
oponder: Because I needed to access the store also in the giantswarm client wrapper, I made a weird singleton out of history.

I found a better way to solve that now.

src/lib/giantswarm_client_wrapper.js
```diff
@@ -3,12 +3,11 @@
 
 import { auth0Login } from 'actions/userActions';
 import { isJwtExpired } from 'lib/helpers';
+import { store } from '../components/app';
```
oponder: importing the store that we initialise in app, instead of loading the singleton store

src/components/cluster/detail/__tests__/cluster_details_node_pools_table.js
```diff
@@ -57,15 +56,13 @@ it('shows the dropdown when the three dots button is clicked', () => {
 // calling the right method with the right arguments when clicking the submit button
 it('patches node pool name correctly', async () => {
   const div = document.createElement('div');
-  const { getByText, container } = renderRouteWithStore(
-    '/organizations/acme/clusters/m0ckd/np',
-    div
-  );
+  const { store, getByText, container } = renderRouteWithStore(div);
+  store.dispatch(push('/organizations/acme/clusters/m0ckd/np'));
```
oponder: This is how we travel to some URL as part of the setup of the test.

I realise now that we could put it back as a params of `renderRouteWithStore` and do the dispatch there instead if you'd like.

PR #763, opened 2019-09-19
Fix mocked status response.

oponder:
I noticed in development we occasionally got the 0 node, 0 CPU bug again.

That's because of a change in the status action which wasn't carried through in the mocked status response part.

This feels like a regression though, not sure how this came back in the dev env, which is a bit scary.

### DIFF
--- a/src/actions/clusterActions.js
+++ b/src/actions/clusterActions.js
@@ -129,7 +129,7 @@ export function clustersLoad() {
     const status = await Promise.all(
       Object.keys(v4ClustersObject).map(clusterId => {
         if (window.config.environment === 'development') {
-          return { id: clusterId, ...mockedStatus };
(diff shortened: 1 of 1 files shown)


oponder:
It's also possible I noticed this and fixed this in a different branch that I havent pushed yet, and that's why I think it's a regression.. not sure.

ograu:
I think that this is because you didn't merge the PR where you fixed this. It's this one: https://github.com/giantswarm/happa/pull/740

I saw it yesterday and though that you wanted to checked anything else before merging...

oponder:
Ah yeah, I think that one I wanted to make flatter, so I have a continuation where I do that. But then for now it's ok to get this small PR in and fix it on it's own, without making it depend on that other PR.

PR #766, opened 2019-09-20
Add a login test.

oponder:
This adds an integration test for logging in.

The boundary are the actual network requests to the API, which are mocked using `nock`. 
Nock works by overriding Node's http.request function. 

https://github.com/nock/nock

The stuff under `GIVEN` could be extracted somehow into a 'Happy API setup' function somehow.

Since I think we will be repeating it often.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,13 +33,13 @@ jobs:
     - checkout
 
     - run:
-        name: Install node@10.16.0
+        name: Install node@12.10.0
         command: |
           set +e
           export NVM_DIR="/opt/circleci/.nvm"
           [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
-          nvm install v10.16.0
-          nvm alias default v10.16.0
+          nvm install v12.10.0
+          nvm alias default v12.10.0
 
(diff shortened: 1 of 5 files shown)


oponder:
Fails currently in circleci, since I need to get it running a later version of node. (Array.flat is not a function)

ograu:
> The stuff under `GIVEN` could be extracted somehow into a 'Happy API setup' function somehow.
> 
> Since I think we will be repeating it often.

Cool! Love the approach ❤️ 

So if I understand it well `nock` will intercept all requests that match with the route passed as an argument, righ?

This way we could define diferent responses in each case, which is great.
Thanks!

oponder:
> So if I understand it well nock will intercept all requests that match with the route passed as an argument, righ?

Correct!

> This way we could define diferent responses in each case, which is great.

Yup! And it even complains if the same request happens more than once, so for each request (even to the same endpoint) you have to make a nock like that. 

So you could even test (if we increase the jest timeout for that page) the pages that do like refreshing requests to `/v4/clusters` and pretend that one of the requests fails or something like that.

It will also reveal if we are making requests we don't expect for a certain path.. the errors are pretty helpful so far.

oponder:
FYI @marians , this might be a usable test example you've been waiting for.

PR #767, opened 2019-09-20
Finish up the login tests.

oponder:
I adjusted a little bit the way I do the given/when/then commenting, no need for the capitalization or the indentation.

### DIFF
--- a/__tests__/login.js
+++ b/__tests__/login.js
@@ -20,7 +20,7 @@ it('renders the login page at /login', async () => {
 });
 
 it('redirects to / and shows the layout after a succesful login', async () => {
-  // GIVEN I have a Giant Swarm API with no clusters, organizations, appcatalogs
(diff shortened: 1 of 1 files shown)


__tests__/login.js
```diff
@@ -14,82 +14,140 @@ it('renders the login page at /login', async () => {
...
 
-it.skip('shows an error if the user logs in with invalid credentials', async () => {
-  // To be implemented
+it('shows an error if the user logs in with invalid credentials', async () => {
+  // Given I have a Giant Swarm API that does not accept my login attempt
+
+  // The failed 401 response to the login call
+  const authTokensRequest = nock('http://localhost:8000')
+  .post('/v4/auth-tokens/')
+  .reply(401);
```
oponder: this causes the test to log the error, (since I have these `console.error` calls in the catch blocks of most of the actions. But that is expected. It might be confusing though in the test output.

PR #768, opened 2019-09-20
Added nock utils

ograu:
To reuse nock function and responses. WDYT?


ograu:
BTW, is there any reason why you put login tests in a __tests__ folder at root? Just asking

oponder:
> BTW, is there any reason why you put login tests in a tests folder at root? 

yeah these are the 'feature' tests... they aren't testing a single component, theyre testing everything at once, so I don't think they belong in a component folder

PR #769, opened 2019-09-20
Logout tests

oponder:
Currently breaking due to localstorage not working.

This is my WIP attempt at getting it to work, so you can take a look when you have some time, or otherwise we'll pair on it on monday.

You can run the tests that fail due to localstorage with: `yarn test __tests__/logout.js`

### DIFF
--- a/__tests__/login.js
+++ b/__tests__/login.js
@@ -20,7 +20,7 @@ it('renders the login page at /login', async () => {
 });
 
 it('redirects to / and shows the layout after a succesful login', async () => {
-  // GIVEN I have a Giant Swarm API with no clusters, organizations, appcatalogs
+  // Given I have a Giant Swarm API with no clusters, organizations, appcatalogs
   // that I can log in to.
 
   // The response to the login call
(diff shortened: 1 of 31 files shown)


oponder:
ping @ograu

src/actions/userActions.js
```diff
@@ -40,6 +40,7 @@ export function logoutError(errorMessage) {
 // refreshUserInfo performs the /v4/user/ call and updates what Happa knows
 // about the user based on the response.
 export function refreshUserInfo() {
```
oponder: it's the localstorage call in here that fails and eventually causes `Cannot read property 'origin' of undefined` to bubble up.

ograu:
ping @oponder 

I think the issue was that you have to tell Jest to mock `localStorage`. 

Let me know if it doesn't work for you, and will take a look later or tomorrow.

__tests__/login.js
```diff
@@ -14,82 +14,140 @@ it('renders the login page at /login', async () => {
...
-    clustersRequest.done();
-    appcatalogsRequest.done();
+  // The response to the login call
+  const authTokensRequest = nock('http://localhost:8000')
+  .post('/v4/auth-tokens/')
+  .reply(200, {"auth_token":"efeb3f94-8deb-41a0-84cc-713801ff165e"});
+
+  // The response to the user info call
+  const userInfoRequest = nock('http://localhost:8000')
+  .get('/v4/user/')
```
ograu: I think it would be nicer to have our global `endpoints` object and use something like axios. Now way we have to:

1. look at redux dev tools to find  which actions are triggered, then
2. look at actions files and find which is the API method we are using when/before/after this action is dispatched, and
3. find which endpoint we are actually hitting

If we got rid of the gs js client, at least there will be no need to do point 3 and we would have enpoints written just once in our codebase and not these strings repeated everywhere.

@oponder Does this make sense to you?

oponder: Not sure i understand! lets talk during our meeting!

oponder:
Got it all to work, could you take a look again @ograu ?

test_utils/browserMocks.js
```diff
@@ -0,0 +1,22 @@
+var localStorageMock = (function() {
+  var store = {};
+
+  return {
+    getItem: function(key) {
+      return store[key] || null;
+    },
+    setItem: function(key, value) {
+      store[key] = value;
```
oponder: there was a `toString` here which caused some trouble, I removed that. Not sure if that changes heavily how a real browser would deal with it.. 

It was breaking when trying to persist `undefined` for the selected organisation for when there are no `orgs` yet in Happa.

Since there is no toString method on undefined.

But in a real browser, that worked fine, so I felt ok about changing this mock to make it work.

__tests__/logout.js
```diff
@@ -0,0 +1,97 @@
...
+  });
+
+  // Assert that the mocked responses got called, tell them to stop waiting for
+  // a request.
+  userInfoRequest.done();
+  infoRequest.done();
+  orgRequest.done();
+  clustersRequest.done();
+  appcatalogsRequest.done();
+});
```
ograu: I like this test a lot :-)

PR #772, opened 2019-09-24
Unified validation messages

ograu:
For consistency. Added a new UI component that takes a string and renders a span.

Before: 

Captura de pantalla 2019-09-24 a las 10 52 21

After: 

Captura de pantalla 2019-09-24 a las 12 49 46

### DIFF
--- a/src/components/UI/ValidationErrorMessage.js
+++ b/src/components/UI/ValidationErrorMessage.js
@@ -1,0 +1,22 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import styled from '@emotion/styled';
+
+const MessageSpan = styled.span`
+  color: ${props => props.theme.colors.error};
+  margin-right: 10px;
+  display: block;
+  font-size: 14px;
+  font-weight: 400;
+  white-space: nowrap;
+`;
+
+const ValidationErrorMessage = ({ message }) => (
+  <MessageSpan>{message}</MessageSpan>
+);
+
+ValidationErrorMessage.propTypes = {
+  message: PropTypes.string,
+};
+
+export default ValidationErrorMessage;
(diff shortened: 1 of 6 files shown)


ograu:
Also instance type selector is now read only, what means that users have to choose a value from the popup and can't write in the field.

ograu:
Removind readonly as now the value can't be modified by the dropdown...

ograu:
@ping marians

ograu:
> I think the typing into the instance type files was a “feature” originally. But I’m ok with disabling that.

It has its drawbacks, cause now it's not reachable through keyboard, but anyway I think you couldn't display the dropdown with the keyboard either. Maybe things to fix in the future...

marians:
Editing the number via direct input is required, yes. We introduced this after an explicit complaint by a user. They found it pretty stupid to have to click the increment button 20 or more times when they were trying to make their cluster fit a traffic spike.

marians:
Just tested. Good to see that direct input is still possible. :)

I found it's not possible to delete the entire value of the input field. Try this:

1. place the cursor behind the number, or select the entire number
2. hit backspace

The number won't be deleted. Instead the error message `Field must not be empty.` appears. However, that's been the case in master, too. So don't worry.

Another nitpick: I think we should remove the full stop from all these short error messages. E. g. `Field must not be empty.` -> `Field must not be empty`

ograu:
> The number won't be deleted. Instead the error message `Field must not be empty.` appears. However, that's been the case in master, too. So don't worry.

Fixed
 
> Another nitpick: I think we should remove the full stop from all these short error messages. E. g. `Field must not be empty.` -> `Field must not be empty`

Fixed

ograu:
@marians Could you please take a look whenever you can and tell me what you think? Thanks 🙂

PR #774, opened 2019-09-25
Add node pool

ograu:
`AddNodePool` component is basically copy/pasted from CreateCluster component and adapted to its needs: I removed everything related to Azure and KVM because AFAIK node pools aren't supported in Azure and KVM by now. Correct me if I'm wrong, please.

Also in this PR: 

- I've started to use Immer to update nested properties on local state object
- I've started to add the HTML element name as a suffix in emotion styled variables. I think Marian does it as well (?). It makes JSX in render easier to read IMO.

Form view:

Captura de pantalla 2019-09-25 a las 13 10 36

With validation errors/messages (some adjustments to be made on margins):

Captura de pantalla 2019-09-25 a las 13 11 30


ograu:
@marian "Add Node Pool" button excludes the form and the other way around, I mean when the form pops up, the button is hidden. This is not what I see in the wireframes but if I follow what we have there, the user could have more than one form in her/his screen and... I don't like it. Let's discuss it if you think otherwise.

Captura de pantalla 2019-09-25 a las 15 11 33

src/components/cluster/new/aws_instance_type_selector.js
```diff
@@ -103,7 +103,7 @@ class AWSInstanceTypeSelector extends React.Component {
 
   render() {
     return (
-      <div className='col-4'>
+      <div style={{ maxWidth: '200px' }}>
```
oponder: There might be a good reason in this case, but generally, would like to avoid inline styles.

src/reducers/clusterReducer.js
```diff
@@ -134,6 +134,10 @@ const clusterReducer = produce((draft, action) => {
     case types.CLUSTER_PATCH_ERROR:
       draft.items[action.cluster.id] = action.cluster;
       return;
+
+    case types.NODEPOOL_CREATE_SUCCESS:
+      draft.items[action.clusterId].push(action.nodePool.id);
```
oponder: not sure about the API.. does it make sense to push the nodepool into the clusters list?

ograu: It is actually an error, it has to be pushed inside the `nodePools`array of a cluster. 
Didn't noticed it because endpoint is not working and so this code has never been reached.
Thanks!

ograu:
ping @marians 

Now the form can switch from availability zones picker:

Captura de pantalla 2019-10-01 a las 10 22 26

To availability zones labels:

Captura de pantalla 2019-10-01 a las 10 23 03



In order to have _clickable_ labels for availability zones I have:

- Adapted `AvailabilityZonesLabel` and `AvailabilityZonesLabels` components to accept checked/unchecked state
- Created a wrapper for `AvailabilityZonesLabels` responsible for passing the state in the correct shape to the form

This is how labels looks like when they are checked (A, B, C) and unchecked (D, E):

Captura de pantalla 2019-10-01 a las 10 19 39

ograu:
- Changed default values for scaling, 
- Maximum availability zones is 4 now
- Improved validation and messages to the user for labels


Captura de pantalla 2019-10-01 a las 12 53 44

Captura de pantalla 2019-10-01 a las 13 02 01

ograu:
ping @marians Just realized that you weren't added here as a reviewer...

ograu:
In these lasts commits I made the modifications we've been discussing this morning:

- _Region:_ before region
- `0 nodes` when no the cluster has no node pools
- Kubernetes endpoint copyable
- Upgrade available underlined when hover
- Added text _To enable autoscaling, set minimum and maximum to different values._  below picker
- Changed background (too dark in my opinion)
- Disable text selection highlighting in all AZ labels

Things I couldn't change:

- Can't use the region instead AZ because in testing API we don't have the `datacenter` property on the object retrieved...

And finally, that weird `span`is something that `ReactCSSTransitionGroup` is adding 🤷‍♂ 

I will merge it and when we have the region will update it.

PR #775, opened 2019-09-25
Add globals file

ograu:
I think we could have a globals (or whatever name you like) file where we could store harcoded values.

WDYT?  Would you prefer another name? Inside a folder?

### DIFF
--- a/src/globals.js
+++ b/src/globals.js
@@ -1,0 +1,1 @@
+export const nodePoolsRelease = '8.2.0';


ograu:
BYW this value is not correct, I am just usign it for dev purposes.

oponder:
I'd rather this constant be somewhere closer to where it gets used.. thats maybe more a personal preference thing... but yeah.

And if it's configuration, then it could go into the window.config 'global'.

ograu:
I wouldn't put this in the window object... I'd rather have this controlled somewhere, actually this is what we want to do with this config object in the future, ¿Did we agree on that? or maybe I am wrong...

And having this variables close to where they are used it is something that I alwys do when I start an app, but IMO is not the best choice in an app that is getting bigger as Happa.

What if I rename it as `config.js` and place it in `src` folder?

oponder:
I think while we don't have a final solution, we could still use the window.config object.. instead of creating two places now to store config... that would be more confusing personally. 

Like when would we use one or the other?

Not super picky about it.. but Happa is in a state of flux already, and I don't see `config.js` or the globals as a better final solution, so I'd rather not create _another_ thing we'll replace with something else once we figure out the 'best' solution....

hope that makes some sense.

ograu:
Good point, agree... let's keep the config object we have now until we figure out what to do 👍

PR #776, opened 2019-09-25
Select multiple organizations when inviting users.

oponder:
Towards https://github.com/giantswarm/giantswarm/issues/6940

Screenshot 2019-09-25 at 4 01 59 PM

Screenshot 2019-09-25 at 4 02 10 PM

ping @T-Kukawka


src/styles/components/_form.sass
```diff
@@ -71,3 +71,18 @@ form
       i
         margin-right: 8px
         color: #23c451
+
+.multi-select
+  input
+    border-color: #3a5f7b !important
```
oponder: the library I used for this multiselect dropdown seems to add styles to the elements, so with the highest specificity possible... abusing `!important` seemed like the only way to override this.

PR #778, opened 2019-09-26
Trim NP name when is too long and test it

ograu:
Also I increased the output for debugging tests and added a utils file where we can write little useful funtions.

I know that you are not a big fan of this @oponder, but it also makes easier testing... this is what I wrote in the top of `utils.js`:

```javascript
// This file is not meant to be a nonsense hodgepodge of functions that we don't know
// where to store, instead it is a file where we store little utility functions that
// can be used and *reused* in broad contexts, eg string or date transformers, methods
// for calculations, etc.
```

WDYT?


oponder:
Hmm, I think we have a place like this already with such a function sorry 😓 

See `truncate` in `lib/helpers.js`

ograu:
> Hmm, I think we have a place like this already with such a function sorry 😓
> 
> See `truncate` in `lib/helpers.js`

really? 🤦‍♂ 😂

ograu:
@oponder can I merge this?

PR #779, opened 2019-09-26
Disabled number picker buttons

ograu:
Buttons " + " and " - " will be disabled once the value in the input equals maximum and minimum value respectively.

@marians I've used same styling that we have for disabled buttons, not thrilled though

### DIFF
--- a/__tests__/password_forgot.js
+++ b/__tests__/password_forgot.js
@@ -1,0 +1,68 @@
+import '@testing-library/jest-dom/extend-expect';
+import { fireEvent, wait } from '@testing-library/react';
+import { renderRouteWithStore } from 'test_utils/renderRouteWithStore';
+import initialState from 'test_utils/initialState';
+import nock from 'nock';
+
+it('takes us to the forgot password form when clicking on "Forgot your password?" from the login form', async () => {
+  // Given I am not logged in and visit the app.
+  const div = document.createElement('div');
+  const { getByText, debug, getByLabelText } = renderRouteWithStore(
+    '/',
+    div,
+    {}
+  );
+
+  // Wait till the app is ready and we're on the login page.
+  await wait(() => {
+    expect(getByText(/Log in to Giant Swarm/i)).toBeInTheDocument();
+  });
+
+  // When I click "Forgot your password?".
+  const forgotPasswordLink = getByText('Forgot your password?');
+  fireEvent.click(forgotPasswordLink);
+
+  // Then I should get redirected to the Forgot Password Form.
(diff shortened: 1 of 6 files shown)


ograu:
And now it is possible to delete the input with backspace key, it will mark the form as invalid and so it cannot be submitted

oponder:
I think the state is still kinda going through, and being validated against the invalid state... so like when I click beyond the allowed range i get the validation error, even though the number didnt change and the button was disabled.

To be a bit clearer, I can click on it in the disabled state, which leads to the validation error showing, but the number not changing.

oponder:
Maybe add a bit of blue to the gray so it doesn't stand out that much?

556677
Screenshot 2019-09-26 at 5 01 49 PM

Instead of 888888
Screenshot 2019-09-26 at 5 02 28 PM

ograu:
Sorry I forgot the image, this is how it looks now:

Captura de pantalla 2019-09-26 a las 16 45 51

> To be a bit clearer, I can click on it in the disabled state, which leads to the validation error showing, but the number not changing.

I know, I have followed @marians instructions, if I undestood them well... IMO it feels better or more natural to me now.

oponder:
But the thing is I can put the form into an invalid state, even though the number picker hasn't changed to an invalid value. Then I have to click to subtract one (and the picker still shows 3) to enable to 'submit' button again.

oponder:
Oh! I'm looking at the number picker on the create cluster page, maybe it's behaving differently there

ograu:
Same behaviour in both places

> But the thing is I can put the form into an invalid state, even though the number picker hasn't changed to an invalid value. Then I have to click to subtract one (and the picker still shows 3) to enable to 'submit' button again.

Oh, I get you now! fixing it

ograu:
> Maybe add a bit of blue to the gray so it doesn't stand out that much?
> 
> 556677
> Instead of 888888

I don't know, let's see what Marian says... If it doesn't stand out that much (I don't like it either, at least with these styles) it's not evident that has changed its state... plus I would like to be consistent with the colors

ograu:
@oponder @marians Can we merge it and decide the colors later?

marians:
I'll check

marians:
Yeah, the gray for a disabled up/down button is pretty much highlighting the button right now, instead of dimming it. I suggest `rgb(60, 90, 113)` or `#3c5a71`.

marians:
After I looked for a color I saw that @oponder already has suggested one. `556677` works for me, too.

ograu:
Changed

PR #791, opened 2019-10-07
Es lint changes

ograu:
- Unreacheable code to warning, because I like to do early returns when debugging without needing to comment x lines of code below the `return`statement.
- Removed sort imports rule because I prefer to sort imports in custom ways (node modules, components, custom hooks, UI, styles)  or not ordering them at all

WDYT Marian?

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -30,11 +30,8 @@
     "react/prefer-es6-class": [1, "always"],
     "react/no-find-dom-node": 0,
     "strict": 2,
-    "sort-imports-es6-autofix/sort-imports-es6": [2, {
-      "ignoreCase": true,
-      "ignoreMemberSort": false,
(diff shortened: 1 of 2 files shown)


marians:
I think removing the sorting of imports will force us (or rather: you and Oliver) to think about them again. But if you prefer it that way, fine with me.

marians:
Please also remove the dependency from package.json

ograu:
> I think removing the sorting of imports will force us (or rather: you and Oliver) to think about them again. But if you prefer it that way, fine with me.

Yeah... actually as this is something that affects all of us who are contributing here, so we should decide it as a team.

Changing error to warning for now and will discuss it with Oliver when he's back.

marians:
Cool!

Issue #790, opened 2019-10-07
Fix IP value rendering


paurosello:
Related to https://github.com/giantswarm/giantswarm/issues/7141

This is an example of the values that the helm chart will be rendered:
```
    ...
    Security:
      RestrictAccess:
        Enabled: true
        GSAPI: true
        GuestAPI:
          Private: false
          Public: true
      Subnet:
        CI: 127.0.0.1/32
        Customer:
          Private: ""
          Public: 127.0.0.1/32,127.0.0.2/32
        VPN: 127.0.0.1/32,127.0.0.2/32
    ...
```

https://github.com/giantswarm/happa/blob/4880ff4571be050fcdcf96ecfe56bcb40d9171e4/helm/happa-chart/templates/ingress.yaml#L14

Will not render correctly as Customer is an object and UI will render 503 error:
    `nginx.ingress.kubernetes.io/whitelist-source-range: '127.0.0.1/32,127.0.0.2/32,map[Private:Public:127.0.0.1/32,127.0.0.2/32]'
`

PR #792, opened 2019-10-07
Join CIDRs ranges using helper function

paurosello:
Fixes https://github.com/giantswarm/happa/issues/790


helm/happa-chart/templates/_helpers.tpl
```diff
@@ -0,0 +1,15 @@
...
+    {{- range $i, $e := (split "," .) -}}
+        {{- if and $e (not (has $e $CIDRs.whitelist)) -}}
+            {{- $noop := append $CIDRs.whitelist $e | set $CIDRs "whitelist" -}}
+        {{- end -}}
+    {{- end -}}
+{{- end -}}
+
+{{- join "," $CIDRs.whitelist -}}
+
+{{- end -}}
```
pipo02mix:
```suggestion
{{- end -}}

```

PR #793, opened 2019-10-07
Add np cluster

ograu:
This is the view for v5 cluster creation.

Release selector can't change forms yet, because `ReleaseSelector` component currently selects the higher active release on render, so even when the user chooses a release that should send her/him to v4 form, when the new form is rendered, `ReleaseSelector` component autoselects the higher release available and so redirects the user to v5 cluster creation form.

When we release NPs happa we will have to change it so the value is passed from parent component instead from `ReleaseSector`.

localhost_7000_organizations_acme_clusters_new_ (1)


ograu:
Form validation and sending data to the endpoint to be implemented

ograu:
@marians  this is ready for review, we can take a look at it tomorrow if you want.

The only thing that I haven't yet implemented is the call to the create cluster endpoint.

ograu:
ping @marians 

- Changed texts in AZ selectors (both NPs form and v5 Cluster form) and pointer cursor added.
- AZ labels in v5 Cluster form with radio buttons behaviour. 
- Dynamic `<hr>` above _Add node pools_ button
- Name of the node pool not truncated but wrapped
- _Unnamed Node Pool_ and _Unamed Cluster_ as placeholders
- _Region_ in v5 cluster view as a tooltip
- Little animation added in the AZ selector in th node pool form

PENDINGS

- v5 Cluster detection (through `v4/info` endpoint)
- update local testing API, update region with `datacenter` value

ograu:
I have changed labels in AZ selectors and now the user can't select more than 4 AZs in the node pools form:

Captura de pantalla 2019-10-15 a las 12 59 31

Captura de pantalla 2019-10-15 a las 13 00 12

@marians this PR is becoming monstrous, meaning huge, so I would like to merge this soon and so avoid conflicts, etc. Do you think it is possible? Thanks

PR #795, opened 2019-10-10
Store nodepols inside items object

ograu:
Solved a problem with  a cluster not having node pools by moving node pools object inside `nodePools.items`object, following same shape as all other entities in store.


ograu:
ok... then I suppose I will have to merge it without approval

PR #798, opened 2019-10-16
Added message in v5 cluster view when there are no node pools

ograu:
Captura de pantalla 2019-10-16 a las 13 07 20

### DIFF
--- a/src/components/cluster/detail/cluster_detail_node_pools_table.js
+++ b/src/components/cluster/detail/cluster_detail_node_pools_table.js
@@ -577,7 +577,7 @@ class ClusterDetailNodePoolsTable extends React.Component {
           <FlexWrapperDiv className={noNodePools && 'no-nodepools'}>
             {noNodePools && (
               <p>
-                Add at least one node pools to this cluster so that you can
+                Add at least one node pool to this cluster so that you can
                 actually run workloads.
               </p>
             )}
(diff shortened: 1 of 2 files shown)


src/components/cluster/detail/cluster_detail_node_pools_table.js
```diff
@@ -222,6 +222,18 @@ const FlexWrapperDiv = styled.div`
   button {
     margin-right: 16px;
   }
+  &.no-nodepools {
+    flex-direction: column;
+    justify-content: space-between;
+    height: 147px;
```
axbarsan: This is really specific haha

ograu: You mean the height? Yeah, this is what I found in wireframes, and BTW I forgot to add an image for this: 

Captura de pantalla 2019-10-16 a las 13 07 20

src/components/cluster/detail/cluster_detail_node_pools_table.js
```diff
@@ -561,7 +574,13 @@ class ClusterDetailNodePoolsTable extends React.Component {
             </AddNodePoolWrapperDiv>
           </ReactCSSTransitionGroup>
         ) : (
-          <FlexWrapperDiv>
+          <FlexWrapperDiv className={noNodePools && 'no-nodepools'}>
+            {noNodePools && (
+              <p>
+                Add at least one node pools to this cluster so that you can
```
axbarsan: I think this should say "add at least one node pool", (without the "s" at the end)

ograu: good catch, thanks!

PR #801, opened 2019-10-17
Fix AZ selector in cluster creation form

ograu:


ograu:
Merging this as without this fix no clusters can be created using happa in testing installations

PR #802, opened 2019-10-18
New V4 cluster form

ograu:
- Restyled form
- Added flash notification when `loadReleases()` returns an empty array instead of breaking the app (is this what we want? should we disable the submit button? Which validation message do we want? see Azure pic)
- Added default Azure instance type

localhost_7000_organizations_giantswarm_clusters_new_

localhost_7000_organizations_giantswarm_clusters_new_ (1)


src/components/cluster/new/cluster_creation_duration.js
```diff
@@ -11,14 +11,18 @@ class ClusterCreationDuration extends React.Component {
 
     if (minutes > 0) {
       return (
-        <p>
+        <p style={{ marginTop: '23px' }}>
```
axbarsan: If you're repeating this, why not make another styled component with this property and use it in both cases?

ograu: Good question: I'm totally fine with inlined styles, maybe it is something we could discuss in SIG Frontend because I think Oliver is not.

The reason why I didn't styled it is because I find it overkill to import `styled` and create a variable just for one css property... but yeah, we can talk about it if you think otherwise.

BTW, I refactored the component, this is how I would have written it. Happy to hear your opinions about it :-)

ograu:
Tested this in testing AWS and Azure installations and it is working fine, couldn't test in `gorgoth` cause there is an issue with credentials, there are still some thinks to be done:

- Unnamed cluster as a placeholder, when no name is provided, and
- RAM and cores of instance type in form

But I would like to merge it, and complete it in another PR.

marians:
When there is no release, creating a cluster will likely fail. Not sure whether cluster creation should be possible (button enabled) in that state.

ograu:
> When there is no release, creating a cluster will likely fail. Not sure whether cluster creation should be possible (button enabled) in that state.

Yeah, I think we should disable the button.
Will do this in the next PR, where I am adding also  "Unnamed cluster" as a placeholder and the "X RAM and X cores per each" message next to the instance type selected.

This PR was just for reestyling.

marians:
Style LGTM!

PR #805, opened 2019-10-21
Refactor App Catalog list components

axbarsan:
This addresses giantswarm/giantswarm#6358

- Refactor App Catalog List inner component for better readilibity and to follow React best practices


axbarsan:
> Hey @axbarsan, I've noted that all your new components are class based ones. We decided to use functional components as a first option. Please take a look at it: https://github.com/giantswarm/fmt/blob/master/javascript/general_style_guide.md#functional-or-stateless-components-vs-class-components
> 
> We use functional components when no local state is needed. We are still experimenting with hooks, so feel free to use class or functional components with hooks for stateful components, but please lets use funtional component for stateless components.
> 
> Please ping me when you have transformed class components into functional ones, or lets have a discussion about it if you disagree with this. Thanks :-)

Ahhh yes, I'm just used to only using class components 😁 Not really fond of hooks right now, they're too young to be used in a large-scale app, and I don't usually jump on the latest trends.

Will make the changes 👍

src/components/app_catalog/app_list/list_inner.js
```diff
@@ -0,0 +1,178 @@
...
+    const currentSearchQuery = searchParams.get('q') || '';
+
+    if (prevState.searchQuery === null) {
+      return {
+        searchQuery: currentSearchQuery,
+      };
+    }
+
+    return null;
+  }
```
ograu: This look like a code smell for me, ¿Could we do this in `ComponentDidUpdate` instead?

src/components/app_catalog/app_list/list_items.js
```diff
@@ -0,0 +1,42 @@
+import AppContainer from 'UI/app_container';
+import PropTypes from 'prop-types';
+import React from 'react';
+
+const AppListItems = props => {
+  const { apps, searchQuery } = props;
```
ograu: If you want to write less you could do as well:

```javascript
const AppListItems = ({ apps, searchQuery }) => {
    // ...
```

src/components/app_catalog/app_list/list_search.js
```diff
@@ -0,0 +1,28 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+const AppListSearch = props => {
+  const { value } = props;
```
ograu: Could be shortened as well:

```javascript
const AppListSearch = {value} => {
    // ...
```

src/components/shared/lazy_loaded_image.js
```diff
@@ -0,0 +1,25 @@
...
+
+    if (current && src) {
+      current.src = src;
+    }
+  }
+
+  render() {
+    return <img {...this.props} src={null} ref={this.imageRef} />;
+  }
+}
```
ograu: Not sure what this component is doing...

axbarsan: > Not sure what this component is doing...

This delays image loading to after the component has already loaded, to make the first page interaction quicker

src/components/app_catalog/app_list/list_inner.js
axbarsan: No, this is the recommended way of setting the state based on a props change. Here I'm using it to set the initial state based on the props.

ograu: Well, `componentDidUpdate` won't rerender anything if you compare props inside it and just triggers the ones you want.

As for derived state, sorry but I don't get the point of `getDerivedStateFromProps` here... 
Now that I am rereading it, with the code you have inside `getDerivedStateFromProps` and initializing `this.state.searchQuery` with `null` what this would do is just set an empty string in `searchQuery` when component mounts, right? Why don't initialize `searchQuery`with `''` instead? 

TBH I never use `getDerivedStateFromProps` so I might be missing something.

Thanks!

ograu:
> As discussed in today's meeting. I've split my work in multiple pull-requests. This is just a refactor, the actual performance optimization will come in a later pull-request

Ah, ok, I just read the title of the PR and thought that this was the actual improvement.

src/components/app_catalog/app_list/list_inner.js
axbarsan: `getDerivedStateFromProps` merges the update triggered by changing props and the update triggered by changing state. `componentDidUpdate` doesn't re-render anything, unless you change the state in it. (which just means you're triggering another update after an update)

As said before, here I'm using it to set the initial state based on the props. Setting the initial state based on props in the constructor is an anti-pattern.

I'm setting `searchQuery`'s initial value to `null`, so I could diferentiate the initial value of it, to the value that it would have if the search input would be empty. 

Having `if (prevState.searchQuery === '') {` would mean that a state change would be triggered every time the props change.

ograu: > `getDerivedStateFromProps` merges the update triggered by changing props and the update triggered by changing state. `componentDidUpdate` doesn't re-render anything, unless you change the state in it. (which just means you're triggering another update after an update)

Ok if this is what you need here.

 
> As said before, here I'm using it to set the initial state based on the props. Setting the initial state based on props in the constructor is an anti-pattern.

Yeah, that's the mantra, but it's an oversimplification, basically this is not true. We do it all the time. It is an antipattern as long as you expect this to trigger rerenders. Otherwise it is perfectly fine.

> I'm setting `searchQuery`'s initial value to `null`, so I could diferentiate the initial value of it, to the value that it would have if the search input would be empty.
> 
> Having `if (prevState.searchQuery === '') {` would mean that a state change would be triggered every time the props change.

Don't agree, but anyway, that's not the point I was willing to raise. I am just asking what problem does this code actually solve, because as far as I understand this just runs when component is mounted, and when this happens it will set `searchQuery` always to `' '` because it is initialized as `null`. So the question is ¿Does this code run in other situations? 

Just asking because I want understand it, that's all.

axbarsan: > > As said before, here I'm using it to set the initial state based on the props. Setting the initial state based on props in the constructor is an anti-pattern.
> 
> Yeah, that's the mantra, but it's an oversimplification, basically this is not true. We do it all the time. It is an antipattern as long as you expect this to trigger rerenders. Otherwise it is perfectly fine.

It's actually always an anti-pattern. The reason is because of React Fiber, the component might be instantiated a while after the component has been mounted, and if that happens, the props might not be there yet. That will become an even bigger problem later when React makes rendering asynchronous.

> > I'm setting `searchQuery`'s initial value to `null`, so I could diferentiate the initial value of it, to the value that it would have if the search input would be empty.
> > Having `if (prevState.searchQuery === '') {` would mean that a state change would be triggered every time the props change.
> 
> Don't agree, but anyway, that's not the point I was willing to raise. I am just asking what problem does this code actually solve, because as far as I understand this just runs when component is mounted, and when this happens it will set `searchQuery` always to `' '` because it is initialized as `null`. So the question is ¿Does this code run in other situations?

This does not set `searchQuery` to always be `''`. If there is a `q` parameter in the URL. This sets the `searchQuery` value to the one in the URL. So for example if you send the URL (`some_URL?q=someApp`) to someone, `searchQuery` would be set to `someApp` for the initial render, and the browser would scroll to that app.

ograu: We discussed about that, don't wanna make this longer than what it actually is, just could you please take a look at here? https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops and here: https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#what-about-memoization and decide if this fits in some of the cases where we don't need `getDerivedStateFromProps`.

I'm ok with whatever you decide.

ograu:
> > As discussed in today's meeting. I've split my work in multiple pull-requests. This is just a refactor, the actual performance optimization will come in a later pull-request
> 
> Ah, ok, I just read the title of the PR and thought that this was the actual improvement.

So, could we please change the title, it confuses me a little bit.  Thanks!

src/components/app_catalog/app_list/list_inner.js
```diff
@@ -0,0 +1,178 @@
...
+
+      const apps = Object.values(props.catalog.apps).map(appVersions => {
+        return appVersions.sort((a, b) => {
+          return new Date(b.created) - new Date(a.created);
+        })[0];
+      });
+
+      for (const app of apps) this.add(app);
+    });
+  }
```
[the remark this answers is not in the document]
axbarsan: I actually forgot about this, sorry.
What I was trying to propose is dropping `lunr` from this altogether, since it's not providing an actual performance benefit. What do you think about that?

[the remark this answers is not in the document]
marians: Lunr is providing a UX benefit. What are you going to replace it with?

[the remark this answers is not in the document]
axbarsan: I've replaced lunr with a simpler algorithm. Now search times are comparable to before, but with a much lower CPU usage

PR #808, opened 2019-10-28
Unhide nodepools

ograu:
First of all apologies, this PR is way too big. Unhidding and implementing NPs has resulted in a lot of changes that have to be merged at once, otherwise we will break Happa.

In this PR:

- Removed conditionals that hide everything NPs related when `environment !== development`
- Removed other hardcoded data
- Changed the data flow in v5 cluster creation so the action creator is in control of handling responses/errors and setting state in store accordingly. And components then are reacting to whatever state is set. I think probably this should be the data flow we want.
- Added action types for getting control of the rendering from store `isLoading` flags
- Fixed errors, typos and changes in `api-spec` regarding the shape of the data we send/receive when hitting API endpoints
- Changed `nodePoolsCreate` action creator so it can be used for creating either a single or various node pools
- Skipped some tests because as the UI changes, some test have to be rewritten

PENDING: node pools scaling modal


marians:
I noticed this in the release selector in cluster creation: When hitting "Select" in the modal, sometimes the modal closes and sometimes it doesn't.

To reproduce:

1. load https://happa.g8s.gauss.eu-central-1.aws.gigantic.io/organizations/giantswarm/clusters/new/
2. click "Details and Alternatives"
3. select "Version 7.3.1". The modal closes.
4. click "Details and Alternatives".
5. select "Version 8.5.0". The model stays open.

I'd prefer the modal to stay open in all cases.

marians:
The number of availability zones offered for master placement and for worker nodes differs.

image

Region `eu-central-1` where gauss is placed has only 3 AZs. This is also contained in the info endpoint response in `general.availability_zones.max`.

For the master node placement, this information seems to be used. We also have to apply it to the worker nodes. However it can never be larger than 4.

marians:
Typo:

> Covering one availability zone, [...] same availability zones as the [...]

Should be

`same availability zone as`

marians:
When hitting the submit button in the Create cluster form, I got two spinners next to the buttons instead of one.

marians:
**This is severe**:

From the cluster details page we are hammering the API with https://api.g8s.gauss.eu-central-1.aws.gigantic.io/v5/clusters/h2z66/nodepools/

This seems to be related to the fact that my test cluster had no node pools.

marians:
Master node availability zone name must be prefixed with the region name.

This is what is currently submitted when we select a specific master AZ:

image

Instead of the letter `"b"` this should be `"eu-central-1b"`. The region name must be prepended to the letter.

As a result, currently the master is placed in A instead of B. The value seems to be ignored by the API.

marians:
Severe: Key pairs are not appearing

I created a cluster without node pools. After some time on the cluster details page I clicked the Key Pairs tab and only the spinner appeared. There were no signs of a request to  the API.

marians:
Improvement:

image

Instead of `0 node pools, 0 nodes, .. CPU cores .. RAM` I'd like to simply show `0 nodes`.

marians:
Let's close the gap above `Scaling range`, even if it means to move the content when toggling between Automatic and Manual AZ placement.

image

Also there could be more space before the submit buttons.

marians:
Deleting a node pool creates an exception:

image

marians:
In NP creation:

image

Here we show 3 zones, but the text says "up to 4". It should be "up to 3" in this case.

If the region has 5 zones, the text must be "up to 4" however, as 4 is the absolute limit.

marians:
The success message for creating a node pool should disappear after some time. It stays forever until clicked.

image

ograu:
> Edit: here i would expect the "Number of availability zones to use" selector to offer the numbers from 1 to 3 only.

Should I pick the minimum from the info endpoint response in `general.availability_zones.default`? Or should we hardcode `1`?

ograu:
> The success message for creating a node pool should disappear after some time. It stays forever until clicked.
> 
> image

Marked this as fixed

ograu:
> I noticed this in the release selector in cluster creation: When hitting "Select" in the modal, sometimes the modal closes and sometimes it doesn't.

> I'd prefer the modal to stay open in all cases.

This requires a little bit of work because currently the modal is a child component of the form. We switch forms when a no NPs release is selected from de NPs cluster creation form (and viceversa) so the modal is automatically unmounted with its parent when this happens.

It is something totally doable, of course, but it will require some time: 

- Create a wrapper for the two modals (or maybe reuse tho one we already have)
- Extract the modal from the `ReleaseComponent` (the button with _Details and Alternatives_ text inside)
- Place the modal in the wrapper and write the logic to make the modal and the `ReleaseComponent` communicate throught the wrapper

ograu:
@marians Finally had some time to add NP scaling modal. I don't understand exactly what this scaling modal does (how does it compute the values, etc.) so please tomorrow let's take a look toghether and see if it does what it shold do. Thanks!

marians:
image

Let's make that `1 node pool` instead of `1 node pools`

ograu:
- `0 nodes` and nothing else when there are no nodes
- Kubernetes version in cluster details view
- AZ minimum is now `1`, harcoded in Happa. Default value for the input is now `general.scaling.default` from info endpoint response
- Max value for scaling is 999

src/actions/clusterActions.js
```diff
@@ -178,32 +175,27 @@ export function clustersLoad() {
     );
 
     // Get the details for v5 clusters.
-    if (window.config.environment === 'development') {
-      const clusters = await Promise.all(
-        v5Clusters.map(cluster => clusterDetailsV5(dispatch, getState, cluster))
-      );
+    const v5ClustersDetails = await Promise.all(
+      v5Clusters.map(cluster => clusterDetailsV5(dispatch, getState, cluster))
```
axbarsan: Sweet! 😁

src/actions/clusterActions.js
```diff
@@ -553,12 +547,12 @@ function clusterLoadStatusV4(dispatch, clusterId) {
  *
  * @param {Object} cluster Cluster definition object
  * @param {Boolean} isV5Cluster
+ * @param {Array} nodePools array of node pool objects
```
axbarsan: I think this is missing from the function 😬

src/actions/clusterActions.js
```diff
@@ -553,12 +547,12 @@ function clusterLoadStatusV4(dispatch, clusterId) {
...
  */
 export function clusterCreate(cluster, isV5Cluster) {
   return function(dispatch) {
```
axbarsan: We could make this an arrow function

src/actions/nodePoolActions.js
```diff
@@ -9,6 +9,11 @@ const nodePoolsApi = new GiantSwarm.NodePoolsApi();
 //Loads all node pools for all node pools clusters.
 export function nodePoolsLoad() {
   return async function(dispatch, getState) {
```
axbarsan: This could be an arrow function

src/components/cluster/detail/cluster_detail_node_pools_table.js
```diff
@@ -476,21 +467,29 @@ class ClusterDetailNodePoolsTable extends React.Component {
               </GridRowNodePoolsHeaders>
               {nodePools &&
                 nodePools
-                  .sort((a, b) => (a.name > b.name ? 1 : -1))
+                  .sort((a, b) =>
+                    a.name > b.name
```
axbarsan: Can you please refactor this to use regular `if` blocks? I know it might be verbose, but it would be much easier to understand

src/components/cluster/detail/node_pool_dropdown_menu.js
```diff
@@ -2,71 +2,73 @@ import DropdownMenu from 'UI/dropdown_menu';
...
+            type='button'
+          >
+            •••
+          </button>
+          {isOpen && (
+            <ul aria-labelledby='node_pools_dropdown' role='menu'>
+              <li>
+                <a
+                  href='#'
+                  onClick={e => {
```
axbarsan: What do you think about extracting these lambdas outside the `NodePoolDropdownMenu` component?

src/components/cluster/new/CreateNodePoolsCluster.js
```diff
@@ -574,10 +584,19 @@ CreateNodePoolsCluster.propTypes = {
...
+  minAZ: PropTypes.number,
+  defaultAZ: PropTypes.number,
 };
 
 function mapStateToProps(state) {
-  let availabilityZones = state.app.info.general.availability_zones.zones;
+  const { availability_zones: AZ } = state.app.info.general;
+  const availabilityZones = AZ.zones;
+  // More than 4 AZs is not allowed by now.
+  const maxAZ = AZ.max > 4 ? 4 : AZ.max;
```
axbarsan: Can we extract the `4` number into a constant somewhere, so it'll be easier to find in the future?

src/components/shared/node_count_selector.js
```diff
@@ -114,7 +114,7 @@ class NodeCountSelector extends React.Component {
                 </SpanWrapper>
                 <NumberPicker
                   label=''
-                  max={99}
+                  max={999}
```
axbarsan: Same with this, can we extract it to a constant somewhere? (for example, at the beginning of the file?)

src/actions/clusterActions.js
ograu: Totally. 
I won't do it because this is the pattern I found and we have followed in all thunks.
I think it is interesting to follow the same patterns for thunks, just for readibility sake.

src/components/cluster/detail/cluster_detail_node_pools_table.js
ograu: Agree

src/components/cluster/detail/node_pool_dropdown_menu.js
ograu: Yeah, we can do it. But why?

src/components/cluster/new/CreateNodePoolsCluster.js
ograu: I really think that this very much needed. Tried to do it, unsuccessfully: https://github.com/giantswarm/happa/pull/775

Feel free to bring this up on the next SIG Frontend sync. I will support you, cause I don't feel comfortable at all doing this

src/components/cluster/detail/node_pool_dropdown_menu.js
axbarsan: Just for better readibility and for easier editing

ograu: When there is so little code I like to inline these functions. If the function has to grow for whatever reason, then I would write a named function.

Please, feel free to open a PR in `fmt` as a way to start a discussion for this if you think it's worth it.

src/components/cluster/new/CreateNodePoolsCluster.js
axbarsan: I would just extract it at the top of the file, it doesn't need to be global (at least for now)

marians: How about `Math.min(4, AZ.max)`?

ograu: You make me think about it! 
I really love the conciseness of `Math.min` but have chosen instead a more resilient approach instead

src/components/home/index.js
```diff
@@ -33,7 +33,7 @@ class Home extends React.Component {
    * Load clusters list periodically
    */
   registerRefreshInterval = () => {
-    var refreshIntervalDuration = 3 * 1000; // 30 seconds
+    var refreshIntervalDuration = 30 * 1000; // 30 seconds
```
axbarsan: This could be a `const`

src/index.html
```diff
@@ -17,15 +17,14 @@
 
   <script>
   var config = {
-    apiEndpoint: 'http://localhost:8000',
+    // apiEndpoint: 'http://localhost:8000',
+    apiEndpoint: "https://api.g8s.gauss.eu-central-1.aws.gigantic.io",
```
axbarsan: You forgot this here 😬

ograu:
@marians this is solved: https://gigantic.slack.com/archives/C0251MXL5/p1572520669431400?thread_ts=1572509325.407700&cid=C0251MXL5

And I have added the notice about refreshing:

image

image

Pending:  

1. Add 'RefreshableLabel' as a wrapper for every piece of unit (in v4 and v5 clusters). I assume that we still want this highlighting when data is updated. Correct me if I am wrong.
2. Change the way v5 clusters update information: now we do a full rerender of the view, which I think is not acceptable. I will fix it, but before I will fix this: https://github.com/giantswarm/giantswarm/issues/7521 unless we decide otherwise.

marians:
From the surface, this looks mostly good.

Just this: For 1 node, it should be `node`, not `nodes`.

image

src/index.html
marians: Fixed meanwhile in https://github.com/giantswarm/happa/pull/808/commits/3a930e188771973aa50dc7fcdfd3fc3965b6cfe8

PR #811, opened 2019-10-30
Fix organization navigation

axbarsan:
Fixes giantswarm/giantswarm#6632 and giantswarm/giantswarm#6875

### DIFF
--- a/src/actions/__mocks__/organizationActions.js
+++ b/src/actions/__mocks__/organizationActions.js
@@ -1,2 +1,3 @@
 export const organizationsLoad = () => () => Promise.resolve();
+export const organizationSelect = () => () => { };
 export const organizationCredentialsLoad = () => () => Promise.resolve();
--- a/src/actions/__mocks__/organizationActions.js
+++ b/src/actions/__mocks__/organizationActions.js
@@ -1,3 +1,3 @@
 export const organizationsLoad = () => () => Promise.resolve();
-export const organizationSelect = () => () => { };
+export const organizationSelect = () => () => {};
 export const organizationCredentialsLoad = () => () => Promise.resolve();
--- a/src/components/layout.js
+++ b/src/components/layout.js
@@ -68,37 +68,40 @@ class Layout extends React.Component {
   }
 
   selectOrganization = orgId => {
-    this.props.dispatch(organizationSelect(orgId));
+    const { dispatch } = this.props;
+
+    dispatch(organizationSelect(orgId));
+    dispatch(push('/'));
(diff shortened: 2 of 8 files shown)


src/components/layout.js
```diff
@@ -68,37 +68,43 @@ class Layout extends React.Component {
   }
 
   selectOrganization = orgId => {
-    this.props.dispatch(organizationSelect(orgId));
+    const { dispatch } = this.props;
+
+    dispatch(organizationSelect(orgId));
+    dispatch(push('/'));
```
ograu: Is there any reason why you are pushing after dispatching instead of adding the `'/'` in the return`getOrganizationURL()`?

src/components/organizations/index.js
```diff
@@ -6,24 +6,19 @@ import List from './list/';
...
+  return (
+    <Breadcrumb data={{ title: 'ORGANIZATIONS', pathname: url }}>
+      <Switch>
+        <Route component={List} exact path={`${path}`} />
+        <Route component={Detail} path={`${path}/:orgId`} />
+        <Redirect path={`${path}*`} to={`${url}`} />
+      </Switch>
+    </Breadcrumb>
+  );
+};
```
ograu: Nice work, so much cleaner!

src/actions/__mocks__/organizationActions.js
```diff
@@ -1,2 +1,3 @@
 export const organizationsLoad = () => () => Promise.resolve();
+export const organizationSelect = () => () => {};
```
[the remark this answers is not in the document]
axbarsan: I know, but since it's used in the `Layout` component, it throws an error if I don't overwrite it here :(

src/components/layout.js
axbarsan: In the previous implementation, the `push` was in the `organizationSelect` action creator. It kind of makes sense, since when selecting an organization from the dropdown in the header, you could be on a page that might not be available for another organization, so it just directs you to the homepage, to avoid any unwanted errors.

The reason for not leaving it in `organizationSelect` is that it's only needed when changing organizations from the header dropdown, otherwise not. Hope that makes sense 😁

ograu: Totally, thanks!

src/actions/__mocks__/organizationActions.js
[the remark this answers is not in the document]
ograu: Ah, ok, yeah makes sense. Anyway we are going to get rid of this mocks probably.

PR #813, opened 2019-11-04
Virtualize scrolling in app catalog

axbarsan:
Fixes giantswarm/giantswarm#6358

- Improve scrolling performance in app catalog apps view 
  - Virtualized scrolling with the help of the `react-window` library
  - Improved the way we handle app icon errors (it would expose a memory leak in isolated cases before)
- Included `WindowScroller` from `react-virtualized` to not add the whole library as a dependency just for this functionality, since most of its functionality is covered by `react-window`
  - This allows us to have virtualized scrolling by scrolling on the browser window (just like facebook / instagram), instead of scrolling inside of a `div`
  - I didn't really refactor this code, to not break anything (you can see it relies heavily on hoisting `var`s and using function prototypes in certain files)
- Refactor and rename existing components (improve methods and convert to functional components -> where possible)
- Fixes a bug which made it unable to go back from the Detail Mode of an app, and also access a certain app catalog with its direct URL

Everything should work and behave the same as before, but with lower memory usage and much lower CPU usage


axbarsan:
Re-assigning since the old SIG got deleted

marians:
Deploying to ginger for testing...

oponder:
Hey Adrian, thanks for the PR

Nice work improving the performance of the app catalog pages, while fixing some bugs along the way!

In general though I'd like to see smaller PR's if possible. Sometimes a large PR is inevitable, in that case a commit history that tells the story is good too. 

Check out: https://github.com/giantswarm/giantswarm/blob/130dae32606aab0ae6b7dfc4c37f07fba45a396d/processes/pull-requests.md for a distillation of a discussion we had at Giant Swarm on this topic.

So it's a bit hard to go into feedback right now on the code, because it is a bit much at once :P I don't have any significant nitpicks at the moment, but I am not very picky about code style, usually.

However, one thing I do have feedback on is that it seems we're taking a library (https://github.com/sdecima/javascript-detect-element-resize), making some modifications, and adding it to our `components` folder. 

I'd prefer not to include library code like this. 

Ideally we can make the modifications and contribute it to the upstream. Less ideally we make a fork of the repo with our changes in it and point the `package.json` to that. And as a last resort, if it is necessary, I'd like to make sure we put it in a folder that identifies it as vendored / external code. Perhaps a `vendor` folder in the root of the project. `src/lib` might seem like a candidate at first, but I'd like to consider anything in the `src` as something that really belongs to the `happa` codebase.

Other than that, I think it's really awesome you've identified ways to improve the performance here, and thanks for doing the heavy lifting to improve the overall situation.

axbarsan:
> Hey Adrian, thanks for the PR
> 
> Nice work improving the performance of the app catalog pages, while fixing some bugs along the way!
> 
> In general though I'd like to see smaller PR's if possible. Sometimes a large PR is inevitable, in that case a commit history that tells the story is good too.
> 
> Check out: https://github.com/giantswarm/giantswarm/blob/130dae32606aab0ae6b7dfc4c37f07fba45a396d/processes/pull-requests.md for a distillation of a discussion we had at Giant Swarm on this topic.
> 
> So it's a bit hard to go into feedback right now on the code, because it is a bit much at once :P I don't have any significant nitpicks at the moment, but I am not very picky about code style, usually.
> 
> However, one thing I do have feedback on is that it seems we're taking a library (https://github.com/sdecima/javascript-detect-element-resize), making some modifications, and adding it to our `components` folder.
> 
> I'd prefer not to include library code like this.
> 
> Ideally we can make the modifications and contribute it to the upstream. Less ideally we make a fork of the repo with our changes in it and point the `package.json` to that. And as a last resort, if it is necessary, I'd like to make sure we put it in a folder that identifies it as vendored / external code. Perhaps a `vendor` folder in the root of the project. `src/lib` might seem like a candidate at first, but I'd like to consider anything in the `src` as something that really belongs to the `happa` codebase.
> 
> Other than that, I think it's really awesome you've identified ways to improve the performance here, and thanks for doing the heavy lifting to improve the overall situation.

Hey Oliver. Thank you for pointing it out. I actually extracted [WindowScroller from react-virtualized](https://github.com/bvaughn/react-virtualized/tree/master/source/WindowScroller). I now realize that `javascript-detect-element-resize` was actually included as a vendor dependency [in the original library] (https://github.com/bvaughn/react-virtualized/blob/master/source/vendor/detectElementResize.js). I had no idea where it came from. The reason for not extracting it as a vendor dependency is because I doubt we'll ever need it again (but I might be completely wrong here), and also, now I also see that it has been modified, compared to the one on github. 

Also, about the pull request size, this is actually the second one for this issue, and I was unsure how to actually split all of it in pieces, since the changes are kind of glued together. I will pay more attention to this in the future.

I would like an advice for the `detectElementResize` dependency, not sure what to do with it, since it has been modified. Should I extract it to a `vendor` folder, even though we might not use it again? (also while `ResizeObserver` exists and might be a better tool these kinds of issues)

oponder:
Yeah I think extracting to a `vendor` folder would be ok for now.

What do you mean with `we might not use it again?`? Like it won't be used in more than one place?

I think that's fine for it to still live in a `vendor` folder to identify it as 'outside' code.

And if we end up not using it _any_ places, then we can just delete the file entirely :P

axbarsan:
> Yeah I think extracting to a `vendor` folder would be ok for now.
> 
> What do you mean with `we might not use it again?`? Like it won't be used in more than one place?
> 
> I think that's fine for it to still live in a `vendor` folder to identify it as 'outside' code.
> 
> And if we end up not using it _any_ places, then we can just delete the file entirely :P

I will extract it to a vendor folder, although I'm not confortable in creating a vendor folder outside the `src` folder, since that is know to cause issues with `webpack` - it might bundle the file multiple times in our built static assets.

ograu:
> I will extract it to a vendor folder, although I'm not confortable in creating a vendor folder outside the `src` folder, since that is know to cause issues with `webpack` - it might bundle the file multiple times in our built static assets.

Hey Adrian, one thing I don't get is ¿what is the motivation for extracting it? Bundle size? Customizing it?

axbarsan:
> > I will extract it to a vendor folder, although I'm not confortable in creating a vendor folder outside the `src` folder, since that is know to cause issues with `webpack` - it might bundle the file multiple times in our built static assets.
> 
> Hey Adrian, one thing I don't get is ¿what is the motivation for extracting it? Bundle size? Customizing it?

Hey Oriol. Maybe `extracting` is a strong word. I just moved it from the `utils` folder of the `WindowScroller` component to our `vendor` folder in `src`. 😬

ograu:
> Hey Oliver. Thank you for pointing it out. I actually extracted [WindowScroller from react-virtualized](https://github.com/bvaughn/react-virtualized/tree/master/source/WindowScroller). I now realize that `javascript-detect-element-resize` was actually included as a vendor dependency [in the original library] (https://github.com/bvaughn/react-virtualized/blob/master/source/vendor/detectElementResize.js). I had no idea where it came from. The reason for not extracting it as a vendor dependency is because I doubt we'll ever need it again (but I might be completely wrong here), and also, now I also see that it has been modified, compared to the one on github.

Sorry for not being clear enough. AFAIU you are extracting something from `react-virtualized`. What I was wondering is what do we want to accomplish doing this. We want to reduce bundle size? Customize it a little bit?

ograu:
> The reason for using `react-window` over `react-virtualized` is that `react-window` is something like `react-virtualized` v2. It's made by the same developer, but he made it much faster and lighter. The only issue is that it doesn't include something like `WindowScroller`, so the solution I've found is to use the one from `react-virtualized`. But that library is pretty big, and it will increase our bundle size a lot, so I just included a part of it (which I also refactored a little).

I’ve been reading your PR more carefully today, I appreciate your effort in writing clean code and your commitment in not inflating too much the bundle size. Thanks :-)

About `WindowScroller`, I’ve taken a look at `react-window` library and I do share your concern about the size of the library, it is quite big. But at the same time I am not a huge fan of having ‘external’ code in our `src` folder, either. Also my (I think our) main concern in Happa is to make complex things simple when possible, especially simple to mantain.

So what if instead of having all these code in `src`, we add `react-window` and use a named import for `WindowScroller`? (that would reduce the size of the library considerably). And this increase in the bundle size should not be a huge problem, especially having code splitting in our list.

WDYT? Would this work for you?

src/components/shared/VirtualizedScrollableGrid/index.js
```diff
@@ -0,0 +1,222 @@
...
+  }
+
+  render() {
+    const { data } = this.props;
+    const { columnCount, width } = this.state;
+    const columnWidth = width / columnCount;
+    const rowCount = Math.ceil(data.length / columnCount);
+
+    // Weird fix, together with setting `List`'s direction to 'rtl', to render elements in the correct order
+    const newStyle = Object.assign({}, this.props.style, { direction: 'ltr' });
```
oponder: Does it make sense to set this object each time during the render call?

src/components/app_catalog/AppList/AppListInner.js
```diff
@@ -1,5 +1,5 @@
+import { memoize } from 'lodash';
```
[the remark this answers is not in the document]
axbarsan: `React.memo` is used to optimize the number of updates on React functional components (similar to how `PureComponent` does it for class components).

The reason for memoizing this is to prevent having to filter and parse the whole list every time we re-render the component. This way we avoid a whole lot of CPU usage.

src/components/shared/VirtualizedScrollableGrid/index.js
axbarsan: Kind of, because with each re-render, there might be some new `style` props coming, so I'm appending some new stuff to it.

Also, since `Object.assign` is pretty quick, memoizing this might actually be slower than it is right now.

axbarsan:
I'll import `react-virtualized` and use `WindowScroller` from that 👍

src/components/app_catalog/AppList/AppListInner.js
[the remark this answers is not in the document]
ograu: Ok, then if you have to use it, I think it would be best to use `Underscore` cause this is the library we are using for these kinf of things.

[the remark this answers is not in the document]
axbarsan: Done 👍

oponder:
Screenshot 2019-11-06 at 2 37 42 PM
It might be a good idea to invest a bit of time in making your dev environment run prettier on save.

I keep putting this off also, so I'll follow my own advice now :P

src/components/shared/VirtualizedScrollableGrid.js
```diff
@@ -0,0 +1,222 @@
+import { debounce } from 'underscore';
+import { FixedSizeGrid as List } from 'react-window';
+import { WindowScroller } from 'react-virtualized';
```
ograu: I'd say that you have to import `WindowScroller` specifying the path. Otherwise the full library is bundled.

axbarsan:
> Screenshot 2019-11-06 at 2 37 42 PM
> 
> It might be a good idea to invest a bit of time in making your dev environment run prettier on save.
> I keep putting this off also, so I'll follow my own advice now :P

Maybe we can get [husky](https://github.com/typicode/husky) to do it for us 😬

src/components/shared/VirtualizedScrollableGrid.js
axbarsan: Yes, you were right! 😬 Updated it now

ograu:
LGTM as well! Thanks! :-)
Just out of curiosity: Is there any reason why you are importing `WindowScroller` from `commonjs` folder instead of from `es` folder?

axbarsan:
> LGTM as well! Thanks! :-)
> Just out of curiosity: Is there any reason why you are importing `WindowScroller` from `commonjs` folder instead of from `es` folder?

Because importing the `esmodule` variant of it has problems in the jest runtime (aka no tests actually run) 😬

ograu:
Makes sense. Probably we will have to fix it when doing code splitting.

PR #816, opened 2019-11-06
Fix broken cluster creation form in Azure and KVM installations caused by NPs features merged

ograu:
- Provide a fallback when no `features.nodepools.release_version_minimum` in info endpoint
- Hide `region` in cluster view when there is no region in the API info enpoint response (KVM)


ograu:
Moving region inside `Versions` component to avoid code duplication

PR #818, opened 2019-11-06
Make version links have an underline to make them actually look like links

axbarsan:
Fixes giantswarm/giantswarm#5772

**Relaxed**
Screenshot 2019-11-06 at 17 43 53

**Full flex hover mode - Kubernetes version 💪**
Screenshot 2019-11-06 at 17 48 37


marians:
Great! Have you checked `:hovber` styles in other places? Let's strive for some consistency here.

axbarsan:
> Great! Have you checked `:hover` styles in other places? Let's strive for some consistency here.

Hello! As far as I've seen, the way it was before was pretty much how our links look. 
What I just realized that what I've done with opacity is kind of the opposite way that I should've done. We usually have things with lower opacity, and we increase the opacity on hover.

What do you think about inversing their states? (having the hover look for regular state, and the regular look for hovering)

src/components/cluster/detail/Versions.js
```diff
@@ -25,17 +34,17 @@ function Versions({
...
+                <i className='fa fa-version-tag' />
                 {releaseVersion ? releaseVersion : 'n/a'}
-              </span>
+              </ReleaseDetail>
             </>
           </RefreshableLabel>
         </span>
-        <span
+        <ReleaseDetail
           className='pointer'
```
ograu: We can remove this class, this was what we were actually using for the underline + pointer styles on hover. Think I lost the css when I crated this component

axbarsan:
***Regular***
Screenshot 2019-11-07 at 12 01 52

***Hover***
Screenshot 2019-11-07 at 12 02 01

This is how it looks now

src/components/cluster/detail/Versions.js
```diff
@@ -5,6 +5,23 @@ import PropTypes from 'prop-types';
...
 import ReleaseDetailsModal from 'modals/release_details_modal';
+import styled from '@emotion/styled';
+
+const ReleaseDetail = styled.span`
+  text-decoration: underline;
+  cursor: pointer;
+  color: ${({ theme }) => theme.colors.white2};
+
+  &:hover {
+    color: ${({ theme }) => theme.colors.white1};
```
ograu: LGTM 

I think it would be interesting to have it extracted in the theme as something more semantic as  eg `textColor` and `hoveredTextColor` or something like that, so we can be easily consistent when applying styles.

oponder: I'm in favour of textColor and hoveredTextColor, that's a great idea.

by the way, I think themeing and how we go forward with that could use a discussion between us 3 to make sure we're on the same page. Right now i haven't really thought much more about it yet, and as you can see we have lots of these colors now, and giving them some more meaning would be cool.

For this, it would be nice though if textColor references `white2` instead of defining the same hex value again.

ograu: I think we had an issue opened than we can use as a basis for further discussion, and Marian is willing to work a little bit on theming as well: https://github.com/giantswarm/giantswarm/issues/7346

marians:
Checking...

Issue #820, opened 2019-11-07
VSCode reference & definition tools not working


axbarsan:
Options such as `Find all occurences` or `Go to definition` which are really useful for debugging and understanding other people's code don't work because of our path resolvers

PR #821, opened 2019-11-07
Add config file for code editor TS, so it will know where to check for modules

axbarsan:
Fixes giantswarm/happa#820

This makes VSCode's (and other TS-based code editor's) reference tools work again with our codebase.

Because of the way we have path resolvers setup, this will ramp up the CPU a bit more than usual while resolving paths. The solution for that would be to setup separate path resolvers for everything (such as `@UI` coresponding to `/src/components/UI`, instead of having the resolver look through all of `/src` and `/src/components` for the right folder.


ograu:
also LGTM if it works for you :-)
Could we add `"exclude": ["node_modules"]` also? Intellisense is slow in my computer
Thanks!

axbarsan:
> also LGTM if it works for you :-)
> Could we add `"exclude": ["node_modules"]` also? Intellisense is slow in my computer
> Thanks!

Yes! Good idea 🥇

axbarsan:
> also LGTM if it works for you :-)
> Could we add `"exclude": ["node_modules"]` also? Intellisense is slow in my computer
> Thanks!

Hey Oriol. There's no need to do that, since we're using `include`, TS will only look in the paths in `include`.

The reason for it consuming more resources than it should (as stated in the description) is because we have the whole `src` and `src/components` as path resolvers, so it has to look through a lot of files to find the ones we need. For it to run faster we'd have to setup resolvers separately (for example, `@UI` for `/src/components/UI`)

ograu:
Ok, then I would say go ahead with that refactor if you feel like it :-)
thanks!

PR #824, opened 2019-11-07
Replace react addons css transition group with react transition group

axbarsan:
Fixes giantswarm/giantswarm#5676

- Replace the obsolete `react-addons-css-transition-group` module with `react-transition-group`
- Create a generic `BaseTransition` component that we can use as a base for all transitions
- Create a specific component for slide transitions
- Make the transition used for the login form generic, since it was used in multiple places (I included it in the slide transitions)


src/components/getting-started/fileblock.js
```diff
@@ -110,18 +110,13 @@ class FileBlock extends React.Component {
...
+            classNames='checkmark'
           >
-            {this.state.clicked ? (
-              <i
-                aria-hidden='true'
-                className='fa fa-done codeblock--checkmark'
-              />
-            ) : null}
-          </ReactCSSTransitionGroup>
+            <i aria-hidden='true' className='fa fa-done codeblock--checkmark' />
```
oponder: This one seems to not work anymore, at least for me.

They show up on the getting started pages of a cluster, when clicking the clipboard icon: http://localhost:7000/organizations/giantswarm/clusters/bs04o/getting-started/configure/

Screenshot 2019-11-13 at 10 02 35 AM

oponder: actually the checkmark became super tiny

src/components/getting-started/codeblock.js
```diff
@@ -116,18 +116,13 @@ export class CodeBlock extends React.Component {
               <i aria-hidden='true' className='fa fa-content-copy' />
             </a>
           </div>
-          <ReactCSSTransitionGroup
-            transitionEnterTimeout={1000}
-            transitionLeaveTimeout={1000}
-            transitionName={'checkmark'}
+          <BaseTransition
```
oponder: possibly same here as in the other comment.

ograu:
@axbarsan v5 cluster view is broken. Have you tried to test all this with `gauss` endpoint?

ograu:
wait, forgot to `yarn install` 🤦‍♂

ograu:
No, still broken. v5 cluster form and details view broken for me.

axbarsan:
The problem in `codeblock` and `fileblock` was that the state change that triggered the update happened so quick, that the animation didn't have time to execute. I've added a way to delay prop changes (similar to debouncing, but there are no changes lost)

The rest were just problems with using the animation wrong (for arrays they needed a `TransitionGroup` parent component)

ograu:
Still broken...

ograu:
It's working fine now 🎉  just two things:

1. There was more room below the labels before:

image

2. The animation in the node pools Availability Zones section now isn't as smooth as it was. There is some clipping between elements. 

I can take over if you prefer me to do this.

axbarsan:
> It's working fine now 🎉 just two things:
> 
> 1. There was more room below the labels before:
> 
> image
> 
> 1. The animation in the node pools Availability Zones section now isn't as smooth as it was. There is some clipping between elements.
> 
> I can take over if you prefer me to do this.

Sure, that would be great, since you already know how they should look like. Thank you!

ograu:
Taking over, please don't merge this until I've fixed this.

src/styles/transitions/BaseTransition.js
```diff
@@ -0,0 +1,39 @@
...
+    <CSSTransition in={delayedInProp} mountOnEnter unmountOnExit {...props}>
+      {children}
+    </CSSTransition>
+  );
+};
+
+BaseTransition.defaultProps = {
+  in: false,
+  timeout: 200,
+};
```
ograu: hey @axbarsan  ¿Is there any reason why you wrote this instead of using `transition-delay?

I am reading this and don't understand the purpose of this...

src/styles/transitions/SlideTransition.js
```diff
@@ -0,0 +1,22 @@
...
+SlideTransition.defaultProps = {
+  direction: 'right',
+};
+
+SlideTransition.propTypes = {
+  ...BaseTransitionPropTypes,
+  direction: PropTypes.oneOf(['up', 'down', 'right', 'left']),
+};
+
+export default SlideTransition;
```
ograu: @axbarsan  Maybe I'm missing something here, but id this is just for adding a class to `<BaseTransition>` I really would prefer to keep things simple and just add a css class name.

Is there any other reason to have this?

src/styles/transitions/BaseTransition.js
axbarsan: Since in most places we do our animations something like "set this state variable to `true` and then set it back to `false`". Because that is so quick, React merges the state changes and only does the state change to `false`, so the `enter` animation never happens, just the `exit` one.

What I've done is I've pretty much load balanced the animation `in` property change. This is similar to the concept of debouncing, but this one just delays the `in` property changes, where debouncing merges them.
After each change of the `in` property, it waits `500ms`, and if there is another change happening, it will postpone it to when `500ms` have passed from the last change. If the `500ms` have passed without any change, the next change will execute immediately, without any delay.

axbarsan: This pretty much makes us not need to use `setTimeout` in every component we want to do these types of animations

src/styles/transitions/SlideTransition.js
axbarsan: I'd rather leave it like this, in case we ever need to change the implementation of the animation (for example: switching to a full JS animation technique, without any CSS classNames)

ograu:
After taking a look more carefully to this PR, I don't think it's a good idea to merge this to master.

I don't know exactly how this `react-transition-library` works, but IMO if it is not capable of managing a fade in/fade out after/before the mounting/unmounting and we have to add a `setTimeout` I'd rather don't want to use it. 

I mean, we were using the old one because it added classes / removed classes before state flags are toggled, it did this for us. If we are going to do setTimeouts, we are adding complexity to our codebase while doing something that this library should be doing for us. Even adding a little bit more to this `useDelayedChange` hook, eg the capacity of producing css class names, we could be writing our own implementation of an animation library on top of `react-transition-library`.

What I mean is that I think this is not the direction to go, it has to be possible to do this simple animations we do using the `api react-transition-group` provides, specially taking into account that this library is inspired in the Angular one, which I do have used and provide this functionallity. Otherwise it wouldn't exist.

axbarsan:
I don't see any problem with having the `useDelayedChange` hook, since you don't have to worry about it while writing your own animations.

As for the problem we face, it is mainly because we're using the library for animations, when it was actually designed for transitions (hence the name `react-transition-group`). And the library may have worked for this before (when it was called `react-addons-css-transition-group`), since it had a different implementation, but right now we have to either use methods like the `useDelayedChange` hook, or write the animations as CSS animations, with keyframes.

I'd love to also hear @oponder 's take on this

ograu:
That's not true, at least for the animations I've done. Also everything else we have where we are using the old `react-addons-...` are transitions from a css point of view. 

So, ¿you saying that something that can be done with the old library can't be done with the new one without extra abstractions?

src/styles/transitions/BaseTransition.js
oponder: does that effectively limit our time between "in and out" animations to being 500ms 

So like a checkmark can appear quickly, hold for a while, and then disappear quickly.

But that 'hold for a while', cant be any shorter than 500ms right?

I'm generally ok with that. Just getting an understanding.

oponder:
So my general sentiment here is : I wouldn't block it because of the delayedChange hook, and I think we can eventually take it further from here, to a point where we don't need it (like refactoring to CSS animations) or figuring out that the library has a better way of using it.

I leave it up to @axbarsan to decide if he wants to continue and see if he can find a way to trigger the animations without needing that delay hook, or to revisit it later, since the PR is getting big and perhaps tiring.

oponder:
Only thing is I don't have a 100% good feeling that we've seen and tested every animation in Happa to make sure everything still looks right. 

I've been trying to get a nodepools release in my local dev environment so I can also click through and do some exploratory testing on the animation changes there, but it's been a bit of a struggle so far.

This might be a moment to add some dummy versions of the components that are changing here to the `styleguide` component so that we can test their animations more easily: https://happa.g8s.ginger.eu-central-1.aws.gigantic.io/styleguide

ograu:
Nothing to add to what I already said, just about this:

> So my general sentiment here is : I wouldn't block it because of the delayedChange hook, and I think we can eventually take it further from here, to a point where we don't need it (like refactoring to CSS animations) or figuring out that the library has a better way of using it.

Taking further from here is just the problem I see. 
But if you guys think this is ok, I will just disagree and commit

ograu:
I'm ok with you merging this before I smooth the animation.
Will do it later.

PR #826, opened 2019-11-08
Changed file names in cluster folder

ograu:
Towards https://github.com/giantswarm/giantswarm/issues/6695

Renamed everything inside `cluster` folder following https://github.com/giantswarm/fmt/blob/master/javascript/general_style_guide.md#naming

`new/VMSizeSelector.js` component is not used anywhere. Can we delete it @oponder ?


oponder:
> new/VMSizeSelector.js component is not used anywhere. Can we delete it @oponder ?

Yup. If it's not being used, its dead to me!

PR #828, opened 2019-11-12
Respect worker node limit when creating cluster

axbarsan:
Fixes giantswarm/giantswarm#6572 and giantswarm/giantswarm#6573

- Make the `NodeCountSelector` component accept minimum and maximum values
- Make the number of maximum worker nodes be dictated by the API
- Make the `NodeCountSelector` component use the `readOnly` props
- Clean up the `mapStateToProps` function that injects the store props to `CreateRegularCluster` & move the default props declared there to the component's default props


src/components/cluster/detail/ScaleClusterModal.js
```diff
@@ -331,6 +331,9 @@ class ScaleClusterModal extends React.Component {
               this.props.provider,
               this.props.cluster.release_version
             )}
+            valueConstraints={{
+              max: this.props.maxWorkersPerClusters,
+            }}
             onChange={this.updateScaling}
```
marians: I think naming could be improved here. On AWS, cluster sizes are configured by the users in a range between `min` and `max`. On Azure and KVM, for now, instead there is only one value for the desired number of worker nodes.

If we set constraints for these, we should make sure that it is clear to what they apply. For example, there could be a lower and an upper bound for the `min` value and likewise for the `max` value. Or there could be an absolute minimum and maximum of worker nodes to permit.

If necessary, please clarify in comments what this constraint really applies to.

src/components/cluster/detail/ScaleClusterModal.js
```diff
@@ -415,18 +418,31 @@ ScaleClusterModal.propTypes = {
...
 };
 
+function mapStateToProps(state) {
+  let maxWorkersPerClusters = 0;
+
+  if (state.app.info.workers.count_per_cluster.max) {
+    maxWorkersPerClusters = state.app.info.workers.count_per_cluster.max;
+  } else {
+    maxWorkersPerClusters = state.app.info.workers.count_per_cluster.default;
+  }
```
marians: I don't think it makes sense to use the `workers.count_per_cluster.default` value (normally `3`) as the maximum.

src/components/shared/node_count_selector.js
```diff
@@ -146,10 +161,10 @@ class NodeCountSelector extends React.Component {
             >
               <NumberPicker
                 label=''
-                max={999}
-                min={1}
+                min={valueConstraints.min}
+                max={valueConstraints.max}
                 onChange={this.updateNodeCount}
-                readOnly={false} // TODO
+                readOnly={readOnly} // TODO
```
marians: Why is this TODO?

marians:
Not sure this is ready from a logical standpoint. Added comments. Hoping for the SIG front end folks to chime in.

src/components/cluster/detail/ScaleClusterModal.js
axbarsan: For AWS, the value constraints here (minimum and maximum) apply to the overall values. (the minimum constraint value applies as a minimum value for the `MIN` number selector, and the maximum constraint value applies as a maximum value for the `MAX` number selector.

For KVM and Azure, they act as constraints for the single number selector.

As for clarification, there is no need to set individual limits for each number selector.
With this, the `MIN` number selector has the minimum value of the `min` constraint, and the maximum value of `MAX`'s current value; the `MAX` number selector has a minimum value of `MIN`'s current value and the maximum value of the `min constraint`

Let me know if you need any more information

src/components/shared/node_count_selector.js
axbarsan: Was there before, not sure what it means, since there's no additional info

src/components/cluster/detail/ScaleClusterModal.js
marians: All good!

src/components/shared/node_count_selector.js
marians: Then I think you can remove it. It will only raise questions again otherwise.

src/components/shared/node_count_selector.js
```diff
@@ -28,6 +33,21 @@ class NodeCountSelector extends React.Component {
     scaling: this.props.scaling,
   };
 
+  static mergeConstraints(partialConstraints, defaultConstraints) {
```
oponder: I feel like this is a bit more complicated than it needs to be?

Can it be just like `Object.assign({}, defaultConstraints, partialConstraints)` ?

Or am I missing a bit what this function does?

src/components/shared/node_count_selector.js
```diff
@@ -17,6 +17,11 @@ import styled from '@emotion/styled';
 // scaling - The current value of the input field[s].
 //
 
+const DEFAULT_VALUE_CONSTRAINTS = {
+  min: 1,
+  max: 999,
+};
+
```
ograu: I really think (and I think Adrian will agree) that this should be in a constants file.
We are spreading constants in a lot of components...

oponder: Ok lets talk in happa2020.. I like having related things close to each other. If possible in the same file, if not in the same folder. Somehow I don't think a big constants file with constants related to all possible components in happa will help us, because then you have to go back and forth between things and it's another context switch. 

But if you guys are strongly in favour of centralising things like that in one place, then I am of course happy to disagree and commit!

ograu: ok, let's talk about it quickly tomorrow, and decide what is best for us

src/components/shared/node_count_selector.js
axbarsan: Well it is too complicated, but it makes it easier to handle undefined input.

The actual usage example:
I only inject a `maxValue` prop in the parent component, if the value is bigger than 0 in the store. But if it's not, in reality I would just inject an object like
```
{
  max: undefined
}
```
that messes up the `NodeCountSelector` component.

The check can be done outside the `NodeCountSelector` component, and inject a `valueConstraints` prop equal to `undefined` if there is no `maxValue` in the store.
However, that check will be done multiple times, in different places. So, for less code reusability, I decided to include a safety net in the actual `NodeCountSelector` component.

So if I inject an object like 
```
{
  min: 3,
  max: undefined
}
```
The `min` value will be used, but the `max` value will be replaced by the default value.

We could handle this less painfully and have separate props for the minimul value and maximum value.

Please let me know what you think

oponder: Separate props sounds good and would be ok with me.

ograu:
And also because tests setup it's still kinda WIP, or at least it was until yesterday XD

> We have to still officially agree on the definition of done for Happa PRs, and I dont' wanna block this because it is an improvement, on something that already existed, and so it's a bit mean to asks for tests in this case :P

But in the other hand we will have to write tests when improving/refactoring, otherwise we will never have a good test coverage. We can do it as an aftermath of improvement PRs in order to not endind up with big fat PRs (shame on me BTW!). We have to talk about it...

axbarsan:
I'll add some tests 👍

ograu:
In order to not having to write them twice, maybe it would be wise to wait for me to merge this tests PRs.

axbarsan:
> In order to not having to write them twice, maybe it would be wise to wait for me to merge this tests PRs.

Aiiiight

src/components/shared/__tests__/NodeCountSelector.js
```diff
@@ -0,0 +1,307 @@
...
+
+    const input = getByTestId(labelTestID).querySelector('input');
+
+    fireEvent.change(input, {
+      target: {
+        value: `${targetValue}`,
+      },
+    });
+
+    expect(onChangeCallback).toBeCalledWith({
```
oponder: cool!

src/components/shared/__tests__/NodeCountSelector.js
```diff
@@ -0,0 +1,307 @@
...
+
+import NodeCountSelector from 'shared/NodeCountSelector';
+
+function getComponent(props) {
+  return (
+    <ThemeProvider theme={theme}>
+      <NodeCountSelector {...props} />
+    </ThemeProvider>
+  );
+}
```
ograu: What about something like `renderComponentWithTheme` instead `getComponent`? 

We could reuse this nice and handy function by storing it in  the `renderRouteWithStore.js` file that we could be renamed as `testUtils.js` or something like that

src/components/shared/__tests__/NodeCountSelector.js
```diff
@@ -0,0 +1,307 @@
...
+
+function renderWithProps(props) {
+  return render(getComponent(props));
+}
+
+const labelTestID = 'node-count-selector-picker';
+
+afterEach(cleanup);
```
ograu: I would say this is not needed, now it is a default setting.

src/components/shared/__tests__/NodeCountSelector.js
ograu: you rock @axbarsan ! 🚀

src/components/shared/__tests__/NodeCountSelector.js
```diff
@@ -0,0 +1,307 @@
...
+          maxValid: true,
+        },
+      })
+    );
+
+    expect(getByTestId(labelTestID).textContent).toBe(
+      'To disable autoscaling, set both numbers to the same value.'
+    );
+  });
+});
```
ograu: you rock @axbarsan ! 🚀

ograu: Good work, thanks! :-)

PR #829, opened 2019-11-12
Test fixes localstorage wip

oponder:
ping @ograu


test_utils/renderRouteWithStore.js
```diff
@@ -20,8 +20,11 @@ export function renderRouteWithStore(
   initialRoute = '/',
   container,
   state = initialState(),
-  history = createMemoryHistory()
+  storage = {},
```
ograu: What if we set the user object here as the default and use and empty object in login tests?
We will be using the authenticated version most of the time.

oponder: Yeah that makes sense.

I think actually we can drop also `initialState`.

I see some refactoring to do. I think better in a new PR.

PR #831, opened 2019-11-13
Remove console warnings in jest

ograu:


ograu:
Yeah, I have the same concern.

It's in my list of priorities and I always work with the browser console opened, it's so annoying that it is impossible to forget about it. Won't stop until we get rid of all warnings. I promise!

Issue #830, opened 2019-11-12
Deep linking to App Catalog does not work. Some part of the required app state is not initialised but the layout is still trying to reach it.


axbarsan:

axbarsan:
Initially opened by @oponder .

Should be fixed already

PR #834, opened 2019-11-14
App catalog deep linking

axbarsan:
Fixes giantswarm/giantswarm#7646 and giantswarm/happa#830


axbarsan:
Not resolved yet, investigating further

axbarsan:
Just found out I can't test on `gollum`.

Can't replicate the issue anymore on local app connected to `gollum`, neither or on the deployed `gollum` installation.

I'd consider it fixed

oponder:
The changes make sense to me, but sometimes it loads and sometimes it doesn't. It's a bit weird.

It might have something to do with: https://github.com/giantswarm/giantswarm/issues/7220

I'll try to fix that and combine it with this branch to see if that solves it totally.

axbarsan:
> The changes make sense to me, but sometimes it loads and sometimes it doesn't. It's a bit weird.
> 
> It might have something to do with: [giantswarm/giantswarm#7220](https://github.com/giantswarm/giantswarm/issues/7220)
> 
> I'll try to fix that and combine it with this branch to see if that solves it totally.

I suspect that might be some kind of load balancing from heroku

oponder:
What's weird is that it's more reliable from localhost than it is from gollum, but that shouldn't be a difference.. since it's the same browser making the request.

oponder:
I think there is some race condition actually, or the order that things are happening the reducer isn't what I expect it to be.

I added a log like this to the `appList` index and refreshed a bunch of times and I got some output like this at some point:

```
catalog undefined
catalog {metadata: {…}, spec: {…}, isFetchingIndex: true}
catalog {metadata: {…}, spec: {…}, isFetchingIndex: false, apps: {…}}
catalog {metadata: {…}, spec: {…}, isFetchingIndex: true}
catalog {metadata: {…}, spec: {…}, isFetchingIndex: true}
```

So for a split second it showed the app list and then it went back to isFetching and never recovered.

Actually, i think i found it.

For some reason the `CATALOGS_LOAD_SUCCESS` action is firing twice. And sometimes it lands after a `CATALOG_LOAD_INDEX_SUCCESS`

I think this is overwriting the isFetchingIndex, and possibly clearing the apps again for catalogs that have been loaded up till that point.

Screenshot 2019-11-14 at 8 06 51 PM

Investigating...

oponder:
Oh I see we're trying to load catalogs in `layout.js` and in `app_catalog/index.js`
Screenshot 2019-11-14 at 8 09 55 PM

oponder:
By the way the things I am finding are not directly the cause of deep linking problems. I think you fixed that now. I am just uncovering some more underlying bugs and problems.

oponder:
notes for myself for tomorrow:

1. We wanted to preload the catalogs so that they would be ready faster if people arrived on the home page and navigated manually to them

2. I think we can take that back though. Loading all catalogs on every visit to happa is not needed. And the helm-stable catalog is I think not used visited that often, and the apps in it don't even work so lets not optimize for that use case.

3. we need more than just isFetching. We also need isFetched. the `apps` key on a catalog kind of represents that it has been fetched at some point, but maybe that isn't explicit enough. Should think of a better way to represent entities loading and the various states that can be in.

4. We should not set `isFetching` for individual entities from the collection loader. i.e. we shouldn't set `isFetching` to true for an individual catalog from the action that loads the list of all catalogs. Instead it should be set to true only when we actually start the `catalog load` action, as opposed to the plural `catalogs load`.

I tried to have some kind of system in the way these entities are loaded and the way we set the loading flags, but it needs revision and some better thought.

ograu:
> I tried to have some kind of system in the way these entities are loaded and the way we set the loading flags, but it needs revision and some better thought.

This is actually what I wanted to do in the data flow refactor. As I mentioned in our last meet I think we should have one place, eg one function for each view, responsible:

- batch all the calls, 
- chaining them, waiting for the resolved promises, and
- setting loading flags accordingly

My thinking is that the calls should be performed by independent, reusable functions that just do the API call and modifies the shape of the data received so it fits in our store. 

Then we would be able to use this functions as building blocks of our _batch_ functions, that will somehow document what exactly is going on in each view.

ograu:
BTW this is something that I wanted to do because I think that we really need this: for easy debugging and because I believe we do too many API calls in confusing ways. But also happy if any of you wants to take over it.

axbarsan:
Huge 👍 for @ograu

ograu:
Thanks @axbarsan :-)
Let's do it!

PR #838, opened 2019-11-15
Fix apps loading in reducer

ograu:
We were spreading, and hence iterating over an array-like value. It works, apparently, but tests throw an error that makes sense to me, so fixing it.

To be clear, we are removing `Array(..arrayLike)` and using `Array.from(arrayLike)` instead.

The conditional is not actually needed, `Array.from` would be enough, but I prefer it like this because:
- In tests, where we are mocking the response with a proper array, we don't need `Array.from`
- In case js client is finally fixed we won't need it.


src/actions/clusterActions.js
```diff
@@ -245,10 +245,14 @@ export function clusterLoadApps(clusterId) {
     return appsApi
       .getClusterApps(clusterId)
       .then(apps => {
+        // For some reason the array that we get back
+        // from the generated js client does not have
+        // .map on it. So I make a new one here.
+        const appsArray = apps.map ? apps : Array.from(apps);
```
axbarsan: The actual explanation would be that the value returned is an [*Iterable* (or an *Array-Like* data structure)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols) which does not have methods specific to arrays (such as `forEach` or `map`). Hence the need for creating an array out of it.

Since `Array.from()` also works on arrays, can we just keep `Array.from(apps)`? (to decrease complexity and for easier understanding later on)

oponder: I'd be happy with just Array.from() instead of the ternary.

ograu:
I decided to write it cause I thought it was clearer and kinda tells what is going on... And also because I don't like to run code that is not needed. Yeah, this is insignificant, but as a general principle I think it makes sense. Don't wanna be dogmatic, though, so ... removing the conditional :-)

ograu:
ping @axbarsan

PR #842, opened 2019-11-19
Add a component for showing port mappings.

oponder:
Brings back the ingress port mappings on KVM clusters:

Screenshot 2019-11-19 at 11 37 27 AM


src/components/cluster/detail/PortMappingsRow.js
```diff
@@ -0,0 +1,65 @@
...
+    cluster &&
+    cluster.kvm &&
+    cluster.kvm.port_mappings &&
+    cluster.kvm.port_mappings.length !== 0
+  ) {
+    return (
+      <PortMappings>
+        <FlexRowWithTwoBlocksOnEdges>
+          <div>
+            Ingress Ports:
```
marians:
```suggestion
            Ingress ports:
```

src/components/cluster/detail/PortMappingsRow.js
```diff
@@ -0,0 +1,65 @@
...
+            <dl>
+              {cluster.kvm.port_mappings.map(mapping => {
+                return (
+                  <div key={mapping.protocol}>
+                    <dt>{mapping.protocol}</dt>
```
marians:
```suggestion
                    <dt>{mapping.protocol.toUpperCase()}</dt>
```

src/components/cluster/detail/__tests__/PortMappingsRow.js
```diff
@@ -0,0 +1,49 @@
...
+it('doesnt show port mappings when given an invalid cluster object', () => {
+  const { queryByText } = render(
+    <Router>
+      <ThemeProvider theme={theme}>
+        <PortMappingsRow cluster={{ kvm: { yolo: [{ foo: 'bar' }] } }} />
+      </ThemeProvider>
+    </Router>
+  );
+
+  expect(queryByText(/Ingress Ports/i)).not.toBeInTheDocument();
```
marians:
```suggestion
  expect(queryByText(/Ingress ports/i)).not.toBeInTheDocument();
```

src/components/cluster/detail/__tests__/PortMappingsRow.js
```diff
@@ -0,0 +1,49 @@
...
+        <PortMappingsRow
+          cluster={{
+            kvm: { port_mappings: [{ protocol: 'http', port: '12345' }] },
+          }}
+        />
+      </ThemeProvider>
+    </Router>
+  );
+
+  expect(getByText(/http/i)).toBeInTheDocument();
```
marians:
```suggestion
  expect(getByText(/HTTP/)).toBeInTheDocument();
```

src/components/cluster/detail/__tests__/PortMappingsRow.js
```diff
@@ -0,0 +1,49 @@
...
+
+it('renders without crashing', () => {
+  const div = document.createElement('div');
+  render(
+    <Router>
+      <ThemeProvider theme={theme}>
+        <PortMappingsRow />
+      </ThemeProvider>
+    </Router>,
+    div
```
axbarsan: According to [the docs](https://testing-library.com/docs/react-testing-library/api#render),  you'd have to provide the container in a different way, but also remember you'd still have to append the newly created `div` to the DOM for it to actually get rendered (there's an example on the docs link).

ograu: Yeah this is how you would use the regular react `render`.
It works because `react-testing-library``render` uses react render under the hood I think, so if you don't wanna make queries, I would say you can totally use it like Oliver does here.

src/components/cluster/detail/__tests__/PortMappingsRow.js
ograu: Oliver uses the regex version, which is suposed to be more resilient (though I am experiencing some weordness using it)

src/components/cluster/detail/__tests__/PortMappingsRow.js
axbarsan: The way it's used here is the same as just using `render(<Component />)` (which also works), the second argument of the function should be an options object (you can see that in the docs). If you really want to render in a container you'd provide a second argument similar to:
```js
  render(<Component />, {
    container: containerElement
  }
```

But again, to actually render in a container element, that container element should be in the DOM. Otherwise, nothing happens.

axbarsan: Citing the `testing-library` docs "By default, React Testing Library will create a div and append that div to the document.body and this is where your React component will be rendered. If you provide your own HTMLElement container via this option, it will not be appended to the document.body automatically."

oponder: Not sure what the consequence / impact is of how it's done right now.

If we should change it, then lets change it everywhere, because I think I copied it from another test.

Will do some more reading later, and consider it out of scope for this PR, but certainly something to align if it's wrong.

oponder: Im surprised it works then...

Will consider it out of scope of this PR. If we're doing it wrong then lets fix it everywhere at once. Have to do a bit of reading to understand this a bit beter.

PR #844, opened 2019-11-19
Replace global modules with local ones

axbarsan:
Fixes giantswarm/happa#840 and giantswarm/happa#841

- Use local modules instead of using the global ones (to have control over versions)
- Add `husky` for easily configuring pre-commit hooks on a project level
- Add `pretty-quick` to only run prettier on the changed files, not on the whole project
- Add a pre-commit hook to run prettier on all the staged files

**How it looks in the terminal**
Screenshot 2019-11-19 at 14 10 29


oponder:
Nice, I think this needs maybe a bit of a docs update in `README.md`. 

Is there some command to initialise husky? The precommit hook has to be 'installed' individually right?

Or does it install some kind of global precommit hook when you install husky?

axbarsan:
> Nice, I think this needs maybe a bit of a docs update in `README.md`.
> 
> Is there some command to initialise husky? The precommit hook has to be 'installed' individually right?
> 
> Or does it install some kind of global precommit hook when you install husky?

That's the beauty of it, you don't have to do anything, just `yarn install` when you clone the project, and you're done.

What it does behind the scenes, is that it creates git hooks in your hidden `.git` project folder.

PR #845, opened 2019-11-19
V4 cluster tests

ograu:
Before this PR: https://github.com/giantswarm/happa/pull/843

V4 (mainly) cluster views tests. Tests included: 

- Deleting
- Patching name
- Scaling
- Renders data correctly
- Creates a v4 cluster
- Creates a v5 cluster
- Switches form when a new release is selected

Pending: in these tests we are just testing AWS. Scaling and data rendering in KVM and Azure are not tested.


__tests__/ClusterManagement.js
```diff
@@ -0,0 +1,204 @@
...
+    v5ClusterResponse,
+  ]);
+
+  // Cluster GET request
+  const clusterRequest = getPersistedMockCall(
+    `/v5/clusters/${v5ClusterResponse.id}/`,
+    v5ClusterResponse
+  );
+
+  const div = document.createElement('div');
```
axbarsan: As we talked on the previous PR, this is not actually needed, since Testing Library creates its own element to render in

ograu: :+1: agree,  I was used to add the container before when using `ReactDOm` for tests, but this is not needed here, let's get rid of this.

ograu: But I suggest it would be best to merge this first, and then I will add a PR modifying the function and all the calls.

ograu:
@oponder  Are you ok with me merging this?

PR #848, opened 2019-11-20
Display deleted clusters differently in clusters table

marians:
Towards https://github.com/giantswarm/giantswarm/issues/7516

Changes:

- Add DELETED column to clusters table showing delete_date
- Remove Details button for deleted clusters

### Before

image

### After

image


marians:
Thanks @ograu ! Now that the code has a LGTM, pinging SIG UX for the UX part.

oponder:
@marians does the deleted cluster still light up on hover? 

I think we should disable that. 

My thinking is the row is 'dead' and we shouldn't give any hint that it might be interactive still in some way.

marians:
@oponder Yes, it still gets highlighted.

Somehow I think that's fine, as it is a visual aid to focus on a specific table row and is not necessary related semantically to the row's status.

PR #849, opened 2019-11-20
Modify cluster overview for deleted clusters

marians:
Towards https://github.com/giantswarm/giantswarm/issues/7516

Also make a step towards https://github.com/giantswarm/giantswarm/issues/7346 by introducing variables for shades of dark blue.

Related: https://github.com/giantswarm/happa/pull/848

### Before

image

### After

image


marians:
Asking SIG front end for review first, to get the internals straight. Of course, feel free to comment on everything else.

src/styles/components/_cluster_dashboard.sass
```diff
@@ -30,8 +30,19 @@
...
   b
     font-weight: 400
+  
+  
+
+.cluster-dashboard-item--name
+  font-weight: 700
 
 .cluster-dashboard-item--content
   flex: 1;
```
ograu: This semicolon didn't break anything? weird...

src/components/home/cluster_dashboard_item.js
```diff
@@ -108,6 +108,27 @@ class ClusterDashboardItem extends React.Component {
...
+              </span>
+            </div>
+            <div className='cluster-dashboard-item--delete-date'>
+              Deleted {relativeDate(cluster.delete_date)}
+            </div>
+          </div>
+        </div>
+      );
+    }
+
```
ograu: In order to not duplicate code, I suggest you store the `deleted` string in a variable conditionally. Then you can interpollate it wherever you want it to be printed:

```javascript
const deletedClass = typeof cluster.delete_date !== 'undefined' && cluster.delete_date ? 'deleted' : null;

// And  then use it like: 

return (
  <div className={`cluster-dashboard-item well ${deletedClass}`}>
  // ...
)

File: src/styles/components/_cluster_dashboard.sass [hunk 3]
```diff
@@ -30,8 +30,19 @@
...
+    .cluster-dashboard-item--delete-date, .cluster-dashboard-item--content
+      color: $lighterblue5
+
   b
     font-weight: 400
+  
+  
+
+.cluster-dashboard-item--name
+  font-weight: 700
```
ograu: We agreed in moving styles to components when we had to add/change  styles

src/styles/components/_cluster_dashboard.sass
marians: Removing it

src/styles/components/_cluster_dashboard.sass
marians: ACK, will do

src/components/home/cluster_dashboard_item.js
axbarsan: > In order to not duplicate code, I suggest you store the `deleted` string in a variable conditionally. Then you can interpollate it wherever you want it to be printed:
> 
> ```js
> const deletedClass = typeof cluster.delete_date !== 'undefined' && cluster.delete_date ? 'deleted' : null;
> 
> // And  then use it like: 
> 
> return (
>   <div className={`cluster-dashboard-item well ${deletedClass}`}>
>   // ...
> )
> ```

That will actually add the `null` classname if `typeof cluster.delete_date !== 'undefined' && cluster.delete_date` is not true, since you're converting the `deletedClass` variable to a string, and it would have a null value

ograu: Sorry, true @axbarsan is right, an empty string would do the trick

marians: Changed everything to use styled now. Does this comment still apply?

src/styles/components/_cluster_dashboard.sass
marians: Done

marians:
PTAL @ograu

src/components/app_catalog/detail/cluster_picker.js
```diff
@@ -18,28 +18,28 @@ const ClusterList = styled.div`
 
 const Cluster = styled.div`
   align-items: center;
-  border-radius: ${theme.BORDER_RADIUS};
+  border-radius: ${theme.border_radius};
```
axbarsan: The theme doesn't actually need to be imported, it is in the context of all styled components (so all of them receive it as a prop).

Usage is like:
```css
border-radius: ${({ theme }) => theme.border_radius};
```

Importing it separately will make the components not respond to dynamic theme updates

marians: Interesting, thanks!

src/components/home/ClusterDashboardItem.js
```diff
@@ -108,27 +160,41 @@ class ClusterDashboardItem extends React.Component {
 
     const hasNodePools = cluster.nodePools && cluster.nodePools.length !== 0;
 
+    if (typeof cluster.delete_date !== 'undefined' && cluster.delete_date) {
```
oponder:
```suggestion
    if (cluster.delete_date) {
```

src/components/app_catalog/detail/cluster_picker.js
marians: Changed.

marians:
I'm seeing this error after changing the theme usage:

image

marians:
Instead of 

```
const Cluster = styled.div`
  align-items: center;
  border-radius: ${({ theme }) => theme.border_radius};
  display: flex;
  padding: 10px 15px;
```

should I do this?

```
const Cluster = styled.div`
  align-items: center;
  ${({ theme }) => css`border-radius: ${theme.border_radius}`};
  display: flex;
  padding: 10px 15px;
```

axbarsan:
> Instead of
> 
> ```
> const Cluster = styled.div`
>   align-items: center;
>   border-radius: ${({ theme }) => theme.border_radius};
>   display: flex;
>   padding: 10px 15px;
> ```
> 
> should I do this?
> 
> ```
> const Cluster = styled.div`
>   align-items: center;
>   ${({ theme }) => css`border-radius: ${theme.border_radius}`};
>   display: flex;
>   padding: 10px 15px;
> ```

We should keep using the way recommended [in the docs](https://emotion.sh/docs/theming), so that is the first way you suggested.

The error is probably coming from somewhere else, I'll look for it

src/styles/theme.js
```diff
@@ -1,5 +1,32 @@
 const theme = {
   colors: {
+    // dark blue as defined in Giant Swarm's style guide
+    darkBlue: '#234a61',
+
+    // Darker shades of above darkBlue.
+    // The higher the number, the darker the tone.
+    darkBlueDarker1: '#204358',
```
axbarsan: What we need to have in mind right here, is what happens if we ever chang happa's colors? So what if we start using gray instead of blue?

I would much rather categorization based on color importance (primary / secondary / tertiary), and then splitting those by shade (100 - 900, 100 being the lightest, and 900 being the darkest)

Example
```js
{
  primary:  {
    100: '#fff',
    200: '#eee',
    300: '#ddd',
    400: '#ccc',
    500: '#bbb',
    600: '#aaa',
    700: '#999',
    800: '#888',
    900: '#777'
  }
}
```

ograu: I'd rather use `rgba` or a helper function that outputs some `rgba` instead of harcoding the colors.

axbarsan: You mean `hsl`, right?

axbarsan: Having a specific set of shades will mean we'd have consistent colors across happa. Generating these colors might be done in a different way other than hardcoding them, but that shouldn't be done in the actual component styles, but in the top-level of the app.

oponder: good point about not naming colors though!

ograu: ok, yeah we can discuss this wednesday: not `hsl`we need alpha, so `hsla`or `rgba`

axbarsan: Not sure about needing opacity for anything, since between those shades we're pretty much changing the lightness of the color, which is easier to do with `hsl`, rather than `rgb`/`rgba`

But I agree, we should talk about this in our meeting

axbarsan:
I'm happy about this now, I wouldn't hold you back for the colors on the theme, since that's something we'll discuss and refactor later

src/styles/theme.js
marians: As interesting as that topic is, can we have it separate from this PR?

src/components/app_catalog/detail/cluster_picker.js
```diff
@@ -16,30 +16,42 @@ const ClusterList = styled.div`
...
+`;
+
+const themeBackgroundShade4 = ({ theme }) => css`
+  background-color: ${theme.colors.shade4};
+`;
+
+const themeColorWhite2 = ({ theme }) => css`
+  color: ${theme.colors.white2};
+`;
+
```
marians: @axbarsan I have my doubts here. Changing the syntax to this style everywhere bloats up the code a lot. And I am confused by the fact that I find exactly the style you proposed to me initially, like here:

https://github.com/giantswarm/happa/blob/master/src/components/cluster/detail/RegionAndVersions.js#L12-L20

or here

https://github.com/giantswarm/happa/blob/master/src/components/UI/AppContainer.js#L21-L22

axbarsan: Well this is the standard way of using styled components. We could turn to another styling solution, but if we want to keep using styled, we kind of have to use the library the way it was meant to be used

src/components/home/ClusterDashboardItem.js
```diff
@@ -22,6 +23,56 @@ import moment from 'moment';
...
+
+const WrapperStyles = css`
+  display: flex;
+  background-color: ${props => props.theme.colors.darkBlueLighter1};
+  border-radius: 5px;
+  border: 0px;
+  min-height: 20px;
+  padding: 19px;
+  margin-bottom: 20px;
+  box-shadow: inset 0 1px 1px rgba(0, 0, 0, 0.05);
```
ograu: I would say that, reading the error, whihc is quite explicit, this is what is causing the warning. 
AFAIK we use this without the function:

```javascript
background-color: ${props.theme.colors.darkBlueLighter1};
````

the function is for `styled`

marians:
So, after we spoke about this, I as able to fix the error and reverted one change where my alignment just wasn't useful. Would be happy about refreshed approvals.

PR #850, opened 2019-11-20
Show an icon if the organization has credentials setup or not

axbarsan:
Fixes giantswarm/giantswarm#6630

- Add getting credentials to getting organizations action
- Show icons in the table view if the organization has credentials setup or not
- Clean up code

**What it looks like**

Screenshot 2019-11-20 at 19 28 11


src/components/UI/organization_list/index.js
```diff
@@ -2,16 +2,22 @@ import { clustersForOrg } from 'lib/helpers';
...
       <thead>
         <tr>
-          <th>Name</th>
-          <th style={{ textAlign: 'center' }}>Clusters</th>
-          <th style={{ textAlign: 'center' }}>Members</th>
-          <th />
+          <StyledTableHeader>Name</StyledTableHeader>
+          <StyledTableHeader centered={true}>Clusters</StyledTableHeader>
+          <StyledTableHeader centered={true}>Members</StyledTableHeader>
+          <StyledTableHeader centered={true}>Has Credentials</StyledTableHeader>
```
marians:
```suggestion
          <StyledTableHeader centered={true}>Provider Credentials</StyledTableHeader>
```

marians:
This new column would only make sense on AWS and Azure, not on KVM. I don't see any condition for this so far.

While I like that you are using an icon for this column, I don't find the ✖️ icon ideal. We use it for closing modals. Can we find an alternative? What do others think?

axbarsan:
> This new column would only make sense on AWS and Azure, not on KVM. I don't see any condition for this so far.
> 
> While I like that you are using an icon for this column, I don't find the ✖️ icon ideal. We use it for closing modals. Can we find an alternative? What do others think?

I actually didn't know that, I'll make the change now.
I also agree about the icon, I'll look for another one.

Thank you 👍

oponder:
How about only using the CHECK icon and just leaving it blank for when there are no credentials?

axbarsan:
> How about only using the CHECK icon and just leaving it blank for when there are no credentials?

Screenshot 2019-11-21 at 09 55 07

Looks pretty good, what do you think, @marians ?

marians:
Like it!

src/actions/organizationActions.js
```diff
@@ -59,39 +59,46 @@ const determineSelectedOrganization = (organizations, selectedOrganization) => {
...
+                return organizationsApi
+                  .getCredentials(organization.id)
+                  .then(credentials => {
+                    const organizationWithCredentials = Object.assign(
+                      {},
+                      organization,
+                      { credentials }
+                    );
+
+                    return organizationWithCredentials;
```
ograu: Maybe, probably, you have a reason for this, IDK, but there are two things that catch my attention here:

1.  Nesting. If we really want to have a clear data flow as discussed, we should avoid nesting as much as possible. Nesting makes it difficult to read. BTW No blaming here: I am guilty of this as well, especially in actions files. Just let's stop doing this.

Best approach, as I we have it described here: https://github.com/giantswarm/giantswarm/issues/7521 IMO is to have a separate method that does this, for reusability and readability. 

2. I think we already have a method for this, maybe we should refactor it? Or maybe this is another thing we will hav to have in mind when doing the data flow refactor

axbarsan: I just added to what was already there, since I know there's a refactor coming.

To avoid nesting, which is indeed difficult to read, we can use async await, which I'll also do here to illustrate my point.

The function we currently have loads credentials for a given id and then sets the credentials for the current organization in the state

ograu: > just added to what was already there, since I know there's a refactor coming.

If that's not a lot of hassle, I think we shouldn't continue writing stuff the old way, cause in the end it will cause more work during the refactor

oponder:
Let's merge this?

PR #851, opened 2019-11-21
Fix direct visit to cluster detail view on clusters that don't exist anymore.

oponder:
Fixes attempts to reach`cluster.id` on the cluster detail page when a cluster isn't there.


marians:
Thank you! Can't vouch for the code

PR #853, opened 2019-11-21
Guard against setting on undefined cluster.

oponder:
There are cases where the cluster object doesn't exist anymore, but parts of happa are still trying to refresh status or doing other calls that trigger the reducer to do something on them.

This guards against such errors.


src/actions/clusterActions.js
```diff
@@ -174,11 +174,16 @@ export function clustersLoad() {
     );
 
     // Get the details for v5 clusters.
-    const v5ClustersDetails = await Promise.all(
+    let v5ClustersDetails = await Promise.all(
```
oponder: fyi @axbarsan , I pushed this commit after you approved.

Can you take a look again sorry. I just deployed the branch and found an underlying related error.

oponder: Will merge, assuming its ok.

PR #855, opened 2019-11-22
Show byoc credentials correctly for V4 and V5 clusters

oponder:
Fixes a typo in the code.


oponder:
Also @marians @ograu I think the V5 cluster detail table doesn't show credentials at all?

oponder:
Extracted component and added also to v5 detail table.

src/components/cluster/detail/CredentialInfoRow.js
```diff
@@ -0,0 +1,77 @@
+import { FlexRowWithTwoBlocksOnEdges } from 'styles';
+import AWSAccountID from 'UI/aws_account_id';
+import PropTypes from 'prop-types';
+import React from 'react';
+// import styled from '@emotion/styled';
+
+function CredentialInfoRow({ cluster, credentials, provider }) {
+  const { credential_id } = cluster;
+
+  let credentialInfoRows = [];
```
axbarsan: This should be a `const`, since we're not changing its value

PR #856, opened 2019-11-22
Remove double alert on NewCluster screen when no releases available.

oponder:


oponder:
I think I will remove the alert entirely since the preceding `loadReleases()` also throws an alert already. So then we have two very similar alerts on the screen.

oponder:
Requested again cuz the nature of the PR changed from a typo fix to a full removal.

oponder:
Fixes this double alert:

Screenshot 2019-11-22 at 1 32 09 PM

PR #861, opened 2019-11-25
Cluster form keeps data when swithching releases

ograu:
Fix for https://github.com/giantswarm/giantswarm/issues/7626


ograu:
Sorry, don't understand.

¿Do you think that `localStorage` is not the best way to persist this? 
Or are you saying that you don't think that this is useful/good from a UX point of view?

oponder:
I thought the 'bug' we want to fix is that the name is reset when we change the release version. (I can't recreate this bug by the way, so maybe I don't understand it.)

I figured there was just something reseting the internal state when we changed a release, and the fix I expected was not something involving `localStorage`.

So yeah I'm wondering why we want to persist it in this way.

And then from a UX point of view, if we're only persisting the changes to the `clusterName`, it's a bit odd. Shouldn't we then also want to persist the entire state of the form?

ograu:
> I thought the 'bug' we want to fix is that the name is reset when we change the release version. (I can't recreate this bug by the way, so maybe I don't understand it.)

It only happens when swithcing from v4 cluster creation form to v5 cluster creation form and viceversa.

> I figured there was just something reseting the internal state when we changed a release, and the fix I expected was not something involving `localStorage`.

Form data is local state, so when one form is unmounted, data form vanishes. 

> So yeah I'm wondering why we want to persist it in this way.

For simplicity, we could lift up state, use store, etc. I just went for `localStorage` because I thought it was the simpler way of get the job done.

> And then from a UX point of view, if we're only persisting the changes to the `clusterName`, it's a bit odd. Shouldn't we then also want to persist the entire state of the form?

No opinion regarding this, the only thing I can think of is that there is not a lot more identical inputs shared between the different forms. We can decide to persist more data or no data at all.

@marians Any other reason?

ograu:
In the end it turned out that using state in parent wasn't that much work ... :-P

ograu:
Waiting for gauss to be free to test it.

PR #862, opened 2019-11-25
Apply as text in button when no nodes available yet

ograu:
Fixes https://github.com/giantswarm/giantswarm/issues/7561


__tests__/V5ClusterManagement.js
```diff
@@ -256,7 +256,7 @@ scales node pools correctly`, async () => {
   // Change the values and modify the scaling settings.
   fireEvent.change(inputMin, { target: { value: newScaling.min } });
   fireEvent.change(inputMax, { target: { value: newScaling.max } });
-  const textButton = `Increase minimum number of nodes by ${newScaling.min}`;
+  const textButton = `Apply`;
```
axbarsan: This shouldn't be a template string

ograu: Why please?

axbarsan: Because you're not using it to integrate any variables, you're just using it as a regular string, so why not make it a regular string?

ograu: It's perfectly fine like this, even more performant I believe, if that matters.
If you are suggesting to use template strings _only_ for interpolations, I will probably suggest to use it _always_ instead :-)

And I would say that it is good to have rules and so make easier to anyone to read/write code in Happa, but I also do think that it is good to be flexible to some extent and let everyone to write code in the way it is more comfortable to him, unless there is a clear benefit, of course. 

In this case I don't care a lot, TBH.

We can discuss it if you think it is important.

oponder: Kinda agree with Adrian. I wouldn't expect to see a template string if it's not going to be used for something. It looks a little weird to my eyes personally. 

> good to be flexible to some extent and let everyone to write code in the way it is more comfortable to him

Self expression is important for sure! 

And template string or not doesn't really matter much to me, but I can imagine for other people contributing and wondering how we do things, it would be nice to have a consistent approach throughout the codebase.

I know I'm a hypocrite because I haven't been consistent, but this does stand out as a new approach, so nice to have a moment and agree on it or not.

And I agree it is perfectly fine though, performance wise and code wise. No argument really to be made besides consistency, which I think is a good argument.

ograu: Consistency is what we want, so lets agree with using regular strings when no interpolation or multiline is needed.

PR #863, opened 2019-11-25
Fix desired nodes rendered value

ograu:
Tiny PR, just wanted to make sure we all agree on this cause I've seen this several times.

I think it's better to return a number for functions that compute values, instead of returning `null`, so we don't have to recheck down the component tree, just render. Obviously there may be scenarios when this might  be useful/needed, but I think it makes sense as a general rule.


src/components/cluster/detail/ClusterDetailView.js
```diff
@@ -194,7 +194,7 @@ class ClusterDetailView extends React.Component {
     ) {
       return this.props.cluster.status.cluster.scaling.desiredCapacity;
     }
-    return null;
+    return 0; // if we return null no value is rendered in AWS v4 cluster view
```
axbarsan: Can you also return `0` in the first `if` statement of the method, please?

Also, please replace all `!=` with `!==` 👍

ograu: > Can you also return `0` in the first `if` statement of the method, please?

I am working wih AWS views now, and realize of this, so no, I won't cause I haven't check all use cases and maybe we will run into some issue.  Maybe later when I'm working with other providers. 

> Also, please replace all `!=` with `!==` 👍

👍

src/components/cluster/detail/ClusterDetailView.js
```diff
@@ -181,7 +181,7 @@ class ClusterDetailView extends React.Component {
     // Desired number of nodes only makes sense with auto-scaling and that is
     // only available on AWS starting from release 6.3.0 onwards.
     if (
-      this.props.provider != 'aws' ||
+      this.props.provider !== 'aws' ||
       cmp(this.props.cluster.release_version, '6.2.99') != 1
```
axbarsan: This one also, please

oponder:
```suggestion
      cmp(this.props.cluster.release_version, '6.2.99') !== 1
```

ograu:
> There's one left that adrian spotted, I made a suggested change so it can be applied easily :)

thanks!

PR #864, opened 2019-11-26
Refactored refreshable label

ograu:
Changes in `RefreshableLabel` Component:
- Simplified and its name changed
- Removed `underscore` and `ReactTimeout`
- Now accepts a primitive, string or number, instead an array because this way it is much simpler. Not sure this is the best solution, but it's the simplest one I could think of.

Changes in cluster views:
- Added `RefreshableLabel` to items that might change.
- Extracted nodes, RAM and CPUs info into a new `NodesRunning` component:

image


src/components/UI/RefreshableLabel.js
```diff
@@ -0,0 +1,73 @@
...
+
+  return (
+    <Wrapper className={hasDataChanged && 'changed'} style={style}>
+      {children}
+    </Wrapper>
+  );
+}
+
+RefreshableLabel.propTypes = {
+  children: PropTypes.any,
```
axbarsan: This should be `PropTypes.node`, I believe

src/components/UI/RefreshableLabel.js
```diff
@@ -0,0 +1,73 @@
+import { keyframes } from '@emotion/core';
+import PropTypes from 'prop-types';
+import React, { useEffect, useState } from 'react';
+import styled from '@emotion/styled';
+
+const yellowfade = keyframes`
```
axbarsan: Please provide a more suggestive camelcased name, such as `yellowFadeAnimation`

src/components/UI/RefreshableLabel.js
```diff
@@ -0,0 +1,73 @@
...
+ * which is an array of arbitrary values. When this
+ * array changes, visual highlighting is triggered.
+ */
+
+// As this hook has styles, we are going to pass a styles prop to overwrite/add any styles
+export function RefreshableLabel({ children, value, style }) {
+  // used for outputting 'changed' css class
+  const [hasDataChanged, setHasDataChanged] = useState(false);
+  // used for storing dataItems and so be able to compare with new props
+  const [prevValue, setPrevValue] = useState(value);
```
axbarsan: You should use a ref for storing previous values. You could create a hook that we can re-use later 👍 

https://reactjs.org/docs/hooks-faq.html#how-to-get-the-previous-props-or-state

src/components/cluster/detail/NodesRunning.js
```diff
@@ -0,0 +1,41 @@
...
+    <div>
+      {!workerNodesRunning ? (
+        <span>0 nodes</span>
+      ) : (
+        <>
+          <span>
+            {workerNodesRunning}
+            {workerNodesRunning === 1 ? ' node' : ' nodes'}
+            {nodePools &&
+              ` in ${nodePools.length}${
```
axbarsan: Can you please move this to a separate function? It's hard to read

src/components/cluster/detail/WorkerNodesAWS.js
```diff
@@ -15,6 +16,23 @@ function WorkerNodesAWS({
   workerNodesDesired,
   workerNodesRunning,
 }) {
+  const instanceTypeText = instanceType
+    ? `${instanceType.cpu_cores} CPUs, ${instanceType.memory_size_gb.toFixed(
```
axbarsan: Can you please move this to a separate function? It's hard to read

src/components/cluster/detail/WorkerNodesAzure.js
```diff
@@ -21,21 +22,30 @@ export const LineDiv = styled.div`
 `;
 
 function WorkerNodesAzure({ instanceType, nodes, showScalingModal }) {
+  const instanceTypeText = instanceType
+    ? `${instanceType.cpu_cores} CPUs, ${instanceType.memory_size_gb.toFixed(
```
axbarsan: Same here

src/components/UI/RefreshableLabel.js
ograu: Thanks, I actually think that it doesn't make any sense to propType children here when we are not doing anywhere else. So I think I am going to remove it from `propTypes`.

src/components/UI/RefreshableLabel.js
ograu: Oh, I see, this is the way to do it with hooks, thanks :-)

src/components/cluster/detail/WorkerNodesAWS.js
ograu: Sorry, I don't find it hard to read, I guess if I make a function of it it will look quite similar... maybe it's because this `0` in its own line?
Could you please provide an example of how you would like it to be so I can understand it better?

src/components/UI/RefreshableLabel.js
```diff
@@ -59,13 +59,12 @@ export function RefreshableLabel({ children, value, style }) {
 
   return (
     <Wrapper className={hasDataChanged && 'changed'} style={style}>
-      {children}
+      {this.props.children}
```
axbarsan: This is a functional component, so `this` is not the same as in a class component, it doesn't have the `props` property. If you don't want to destructure it, you can use the rest operator in the compoonent props

Example:
```js
export function RefreshableLabel({ children, value, style, ...props })
```

and then use `props.children` as you would normally.

ograu: Thanks for the explanation, I guess it was a lapse :-P

src/components/cluster/detail/WorkerNodesAWS.js
axbarsan:
```js
const getInstanceTypeLabel = (instanceType) => {
  const cpuCores = 0;
  const memorySize = 0;

  if (instanceType) {
    cpuCores = instanceType.cpu_cores;
    memorySize = instanceType.memory_size_gb.toFixed(0);
  }

  const label = `${cpuCores} CPUs, ${memorySize} GB RAM`;

  return label;
}

const instanceTypeLabel = getInstanceTypelabel(instanceType);
```

ograu: Ok, then we don't agree here... 😢 I find this too verbose 😛 

@oponder help us please!

oponder: I think prettier is making it hard here, because the line indeed got long and then it splits it in a nasty way.

Like this I can follow it fine and get the intent quickly:

```
  const instanceTypeText = instanceType
    ? `${instanceType.cpu_cores} CPUs, ${instanceType.memory_size_gb.toFixed(0)} GB RAM`
    : '0 CPUs, 0 GB RAM';
```

Not sure what the best approach here is. Adrian's is more verbose for sure, but I do find it quicker to get the intent when compared to:

```
  const instanceTypeText = instanceType
    ? `${instanceType.cpu_cores} CPUs, ${instanceType.memory_size_gb.toFixed(
        0
      )} GB RAM`
    : '0 CPUs, 0 GB RAM';
```

Just because of the wacky line break, makes me spend a few more cpu cycles trying to figure out what is going on there.

ograu: And what if we use `// prettier-ignore` so the line isn't splitted?

oponder: I leave it up to you @ograu! 

I would also be ok with `// prettier-ignore` . Just tested that out and it leaves it alone.

```
  // prettier-ignore
  const instanceTypeText = instanceType
    ? `${instanceType.cpu_cores} CPUs, ${instanceType.memory_size_gb.toFixed(0)} GB RAM`
    : '0 CPUs, 0 GB RAM';
```

And then maybe we have a discussion later about increasing the prettier line length. But that has it's own issues

ograu: are you ok with this @axbarsan ?

```javascript
  // prettier-ignore
  const instanceTypeText = instanceType
    ? `${instanceType.cpu_cores} CPUs, ${instanceType.memory_size_gb.toFixed(0)} GB RAM`
    : '0 CPUs, 0 GB RAM';
```

src/components/cluster/detail/WorkerNodesAzure.js
oponder: I just realised azure's instance type structure looks different to this, so we get a error on `toFixed`

In azure it's like `instanceType.memoryInMb`, which you then have to divide by 1000.

And `instanceType.numberOfCores` for the CPU core count.

oponder: So it would look like this

```
  // prettier-ignore
  const instanceTypeText = instanceType
    ? `${instanceType.numberOfCores} CPUs, ${(instanceType.memoryInMb/1000).toFixed(0)} GB RAM`
    : '0 CPUs, 0 GB RAM';
```

src/components/cluster/detail/WorkerNodesAWS.js
axbarsan: For me readibility is way more important, compared to not being verbose.

But you can go ahead and do it your way 👍

src/components/cluster/detail/WorkerNodesAzure.js
ograu: Didn't notice, thanks!
( this one and kvm views are the only views without tests.. )

src/components/cluster/detail/WorkerNodesAWS.js
ograu: > For me readibility is way more important, compared to not being verbose.

Readability is paramount for me as well. I think it's just that we have different ideas of what's readable and what's not. And that's natural, we just started working together. That's why I asked for an example, because I want to understand you better so maybe I can make my code more readable to you. Hope you can appreciate my effort in trying to understand your point.

> But you can go ahead and do it your way 👍

And yeah, "my way", if there is no consensus on how to do/write something I think it makes sense to leave it up to the author of the PR.

BTW I suggest that if you have strong opinions and clear ideas about readability, interpolations, etc. please feel free to open a PR in our coding guidelines repo and let's keep this kind of discussions there. This way we could agree in things and there will be no need to have discussions this long in PRs.

Meanwhile, and especially when there is no consensus, I propose let's be a little bit flexible with the work of our peers.

src/components/cluster/detail/WorkerNodesAzure.js
```diff
@@ -21,21 +22,29 @@ export const LineDiv = styled.div`
 `;
 
 function WorkerNodesAzure({ instanceType, nodes, showScalingModal }) {
+  const instanceTypeText = instanceType
+    ? // prettier-ignore
+      `${instanceType.cpu_cores} CPUs, ${(instanceType.memoryInMb / 1000.0).toFixed(1)} GB RAM`
```
oponder: @ograu: oops, for the CPU, this one needs to be `instanceType. numberOfCores`

ograu: 🤦‍♂ opening a pr...

PR #865, opened 2019-11-26
Replace less than three nodes warning with recommendation in scaling dialog

marians:
Towards https://github.com/giantswarm/giantswarm/issues/5770

This replace a warning mentioning that we do not support or monitor clusters below three worker nodes (which is not the case) with a recommendation to have at least three worker nodes.

### Before

image

### DIFF
--- a/src/components/cluster/detail/ScaleClusterModal.js
+++ b/src/components/cluster/detail/ScaleClusterModal.js
@@ -306,8 +306,8 @@ class ScaleClusterModal extends React.Component {
           timeout={this.rollupAnimationDuration}
         >
           <p key='unsupported'>
-            <i className='fa fa-warning' /> We recommend that you run clusters with at
-            least three worker nodes.
+            <i className='fa fa-warning' /> We recommend that you run clusters
+            with at least three worker nodes.
           </p>
         </CSSTransition>
       );
--- a/src/components/cluster/detail/ScaleClusterModal.js
+++ b/src/components/cluster/detail/ScaleClusterModal.js
@@ -306,7 +306,7 @@ class ScaleClusterModal extends React.Component {
           timeout={this.rollupAnimationDuration}
         >
           <p key='unsupported'>
-            <i className='fa fa-warning' /> We recommend to run clusters with at
+            <i className='fa fa-warning' /> We recommend that you run clusters with at
             least three worker nodes.
           </p>
         </CSSTransition>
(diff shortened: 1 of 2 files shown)


src/components/cluster/detail/ScaleClusterModal.js
```diff
@@ -305,11 +305,8 @@ class ScaleClusterModal extends React.Component {
           timeout={this.rollupAnimationDuration}
         >
           <p key='unsupported'>
-            <i className='fa fa-warning' /> With less than 3 worker nodes, the
-            cluster does not fall under the Giant Swarm{' '}
-            <abbr title='Service Level Agreement'>SLA</abbr>. Giant Swarm staff
-            will not be alerted in case of problems and will not provide
-            proactive support.
+            <i className='fa fa-warning' /> We recommend to run clusters with at
```
oponder:
```suggestion
            <i className='fa fa-warning' /> We recommend that you run clusters with at
```

grammar suggestion

src/components/cluster/detail/ScaleNodePoolModal.js
```diff
@@ -266,11 +266,8 @@ class ScaleNodePoolModal extends React.Component {
           timeout={this.rollupAnimationDuration}
         >
           <p key='unsupported'>
-            <i className='fa fa-warning' /> With less than 3 worker nodes, the
-            cluster does not fall under the Giant Swarm{' '}
-            <abbr title='Service Level Agreement'>SLA</abbr>. Giant Swarm staff
-            will not be alerted in case of problems and will not provide
-            proactive support.
+            <i className='fa fa-warning' /> We recommend to run clusters with at
```
oponder:
```suggestion
            <i className='fa fa-warning' /> We recommend that you run clusters with at
```

grammar suggestion

J-K-C: remove 'that' -

oponder: I think that wouldn't be right either. `Recommend you to run clusters` is like we're recommending that person as someone who can run clusters.

Like: "We recommend you to paint the walls" -> You're a painter and we recommend you as the person to do the job.

"We recommend that you paint the walls" -> You're some guy with unpainted walls and we're wall specialists and we think you should paint your walls.

Being a bit nitpicky. Not a big deal, but I think I'm right on this one?

J-K-C: yes, you're right ;) I used to over use the word 'that' and now i'm a bit aware of it ;)

PR #866, opened 2019-11-26
Move providers to enums

axbarsan:
Fixes giantswarm/happa#859

- Create shared folder where we can place constants / other cool stuff
- Create constants for providers
- Create constants for authorization types
- Replace usage of hard-coded providers/authorization types with the enum-like object values

Please go on and suggest what else to write in this manner. (apart from the URLs, that will be separate)

### DIFF
--- a/src/lib/giantswarm_client_patcher.js
+++ b/src/lib/giantswarm_client_patcher.js
@@ -1,4 +1,5 @@
 import { auth0Login } from 'actions/userActions';
+import { AuthorizationTypes } from 'shared/constants';
 import { isJwtExpired } from 'lib/helpers';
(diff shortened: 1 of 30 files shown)


src/lib/giantswarm_client_patcher.js
```diff
@@ -42,7 +43,9 @@ function monkeyPatchGiantSwarmClient(store) {
             store.dispatch(auth0Login(result));
 
             // Ensure the second attempt uses the new token.
-            headerParams['Authorization'] = 'Bearer ' + result.accessToken;
+            headerParams[
+              'Authorization'
```
oponder: regrettable what prettier does here :P

PR #867, opened 2019-11-26
Enable Multi-AZ for Azure

axbarsan:
References giantswarm/giantswarm#7738

- Show az counter on the create new regular cluster view

The only thing left to do in order for it to work properly is to make the API change to send a maximum AZ count bigger than 1


axbarsan:
Closed because there was a different implementation desired.
Easier to remake than to rework

PR #869, opened 2019-11-27
Show cluster's app tab after creating an app.

oponder:
This lets us show a specific tab of the cluster detail view.

It then uses that when creating an app, to show the app tab of a cluster directly after creating an app. Instead of showing the general tab, which feels like a sudden context drop.

### DIFF
--- a/src/components/app_catalog/detail/install_app_modal.js
+++ b/src/components/app_catalog/detail/install_app_modal.js
@@ -169,7 +169,7 @@ const InstallAppModal = props => {
           push(
             `/organizations/${
               props.clusters.find(c => c.id === clusterID).owner
-            }/clusters/${clusterID}/`
+            }/clusters/${clusterID}/apps/`
           )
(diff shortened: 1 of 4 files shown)


oponder:
Hey guys, I made it possible to go for example to `http://localhost:7000/organizations/byoc/clusters/39yxv/apps/` and it will show the apps tab of that cluster directly.

And `http://localhost:7000/organizations/byoc/clusters/39yxv/keypairs/` shows the keypairs tab.

Now when navigating between tabs by clicking on them, that URL doesn't change yet.

Would you expect it to change? I think so yes.

But the real question I have is: would you expect it to add to your history? 

I.E. after clicking between the tabs, when you press 'back' would you expect to go back through the tabs you clicked?

oponder:
I'm not very happy with this kinda hacked way of making the tab show up.

I'd like to use the router for that more properly.

PR #871, opened 2019-11-28
Filter out apps that are pre installed from the user installed apps list

oponder:
This filters out apps like `chart-operator`, `kube-state-metrics`, `metrics-server`, and `node-exporter`

from the user installed apps list. They appear there now because we have `App CRs` for them in the meanwhile and Happa assumed every App was user installed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "copy-to-clipboard": "3.2.0",
     "deep-diff": "1.0.2",
     "emotion-theming": "^10.0.10",
-    "giantswarm": "git+https://github.com/giantswarm/giantswarm-js-client.git#add-app-labels",
+    "giantswarm": "git+https://github.com/giantswarm/giantswarm-js-client.git",
     "history": "^4.7.2",
     "immer": "^3.2.0",
     "immutable": "3.8.2",
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "copy-to-clipboard": "3.2.0",
     "deep-diff": "1.0.2",
     "emotion-theming": "^10.0.10",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -44,7 +44,7 @@
     "copy-to-clipboard": "3.2.0",
     "deep-diff": "1.0.2",
     "emotion-theming": "^10.0.10",
-    "giantswarm": "git+https://github.com/giantswarm/giantswarm-js-client.git",
+    "giantswarm": "git+https://github.com/giantswarm/giantswarm-js-client.git#add-app-labels",
```
marians: Reminder to revert this

src/components/cluster/detail/ClusterApps.js
```diff
@@ -190,15 +199,29 @@ class ClusterApps extends React.Component {
     });
   };
 
+  // userInstallApps returns a list of just the apps that the user installed
+  // since the list of apps in a cluster also includes apps that were installed
+  // automatically.
+  userInstalledApps = () => {
```
axbarsan: Since this is a function, can you please add a verb at the beginning of its name (such as `getUserInstalledApps`)?

src/components/cluster/detail/ClusterApps.js
```diff
@@ -190,15 +199,29 @@ class ClusterApps extends React.Component {
...
+    if (!this.props.installedApps) {
+      return [];
+    }
+
+    return this.props.installedApps.filter(
+      x => x.metadata.labels['giantswarm.io/managed-by'] !== 'cluster-operator'
```
axbarsan: Can you please give `x` a more meaningful name?

package.json
oponder: thanks!

PR #875, opened 2019-11-29
Add stricter eslint config

axbarsan:
Fixes giantswarm/giantswarm#7804

- Add stricter rules, for keeping a consistent code style and prevent errors.
- Temporary disable eslint during webpack (until we fix the errors)
- Add lint, and lint + auto-fix commands

**Current state**
Screenshot 2019-11-29 at 20 03 42

Checking with the team if the rules are fine for you. If it is, I'll run the auto-fixer.


webpack.common.js
```diff
@@ -13,12 +13,12 @@ module.exports = {
   },
   module: {
     rules: [
-      {
-        test: /\.(js|jsx)$/,
-        enforce: 'pre',
-        exclude: /node_modules/,
-        loader: 'eslint-loader',
-      },
+      // {
```
oponder: If we won't use these rules, I'm ok removing them entirely instead of commenting them out.

axbarsan: These lines of code make `eslint` run as a step of the `webpack` compilation. This was temporary disabled, so if we merge this to `master` as it is, compilation won't work (because of our 300+ errors), so temporary disabling the linting step will give us time to fix our errors.

ograu:
I am ok with most of the rules. Maybe will suggest some changes after a time having been working with this config.

One thing that I am not comfortable, though, is having too many `2`s  because I break some rules when I am working or debugging and it's quite annoying IMO. So I will change at least some ot them to `1`s.

axbarsan:
> I am ok with most of the rules. Maybe will suggest some changes after a time having been working with this config.
> 
> One thing that I am not comfortable, though, is having too many `2`s because I break some rules when I am working or debugging and it's quite annoying IMO. So I would change at least some ot them to `1`s.

Please suggest some changes and we can all discuss about them

.eslintrc
```diff
@@ -20,36 +20,175 @@
...
+    "no-unused-vars": 2,
     "no-useless-escape": 0,
-    "quotes": [1, "single", {"allowTemplateLiterals": true}],
-    "semi": [1, "always"],
+    "quotes": [2, "single", { "allowTemplateLiterals": true }],
+    "semi": [2, "always"],
+    "strict": 2,
+    "no-cond-assign": 2,
+    "no-constant-condition": 2,
+    "no-unreachable": 2,
```
ograu: this is one I don't like as I sometimes use early returns when I am debugging, as a quick and dirty way of doing breakpoints

.eslintrc
```diff
@@ -20,36 +20,175 @@
...
-    "no-unused-vars": 1,
+    "no-unused-vars": 2,
     "no-useless-escape": 0,
-    "quotes": [1, "single", {"allowTemplateLiterals": true}],
-    "semi": [1, "always"],
+    "quotes": [2, "single", { "allowTemplateLiterals": true }],
+    "semi": [2, "always"],
+    "strict": 2,
+    "no-cond-assign": 2,
+    "no-constant-condition": 2,
```
ograu: This is another one I don't like, as I actuallyt do some nasty things as `if(true)...` when I am working or debugging

oponder:
Maybe we can set it up in the CI so that it fails on warnings. So we can be a bit more relaxed in development, while still getting fully ok code into codebase.

axbarsan:
For debugging we can use [eslint-disabling comments](https://eslint.org/docs/user-guide/configuring#disabling-rules-with-inline-comments), as I think those should be errors, and not warnings, but @oponder has a good suggestion on this.

As for @oponder's suggestion. I'm fine with that. We can take it out of the compilation step and add it as a pre-commit hook, and also during CI.

But keep in mind that the existing code base will make CI fail right now, unless we fix all the errors.

oponder:
> But keep in mind that the existing code base will make CI fail right now, unless we fix all the errors.

As discussed in Slack, let's not enable the CI failing step then for now and fix the issues in a separate PR.

And then when we feel ready we can enable the CI check again.

Goal is to get to smaller easier to review PR's so we can improvements and changes like this out already to everyone.

ograu:
> For debugging we can use [eslint-disabling comments](https://eslint.org/docs/user-guide/configuring#disabling-rules-with-inline-comments), as I think those should be errors, and not warnings, but @oponder has a good suggestion on this.

I prefer warnings as I said ao I don't have to do this. 
But fine with that, don't wanna block.

PR #876, opened 2019-12-01
Name app catalog files according to conventions.

oponder:
This renames the app catalog related components so they follow conventions.

Also changed the hyphen in the `getting-started` folder to an underscore.


ograu:
sorry @oponder but this was one of the things I wanted to change here: https://github.com/giantswarm/fmt/pull/83/files 

Couldn't merge yet. Need another approval cc/ @axbarsan

src/components/getting_started/index.js
```diff
@@ -1,10 +1,10 @@
 import { Breadcrumb } from 'react-breadcrumbs';
 import { Redirect, Route, Switch } from 'react-router-dom';
 import DocumentTitle from 'react-document-title';
-import Page0_Overview from './0_overview.js';
-import Page1_ConfigureKubeCTL from './1_configure_kubectl.js';
-import Page2_SimpleExample from './2_simple_example.js';
-import Page3_NextSteps from './3_next_steps.js';
+import Page0_Overview from './0_overview';
```
axbarsan: This doesn't really match our naming rules (because of the numbers and underscores)

oponder: yeah I only changed the folder, not the components in here.

The PR is originally just for app catalog components, but noticed the hyphen in `getting-started` folder and decided to fix that quickly too.

I'll make other PR's for other components. But actually have been thinking of saving that as a starting task for @ubergesundheit so he can get a bit familiar with the layout of Happa's files.

oponder:
So for now I'd like to roll with this, and when we finish the discussion I'm happy to apply the result.

Though I think this is in line with what we wanted right (for the app catalog components)?

src/components/getting_started/index.js
ubergesundheit: The [file name guide](https://github.com/giantswarm/fmt/blob/master/javascript/general_style_guide.md#naming) does not mention numbers being not allowed or did I miss something in the style document?

Which one of those should be the new file name? I think `Overview.js` is suited best.

- `PageZeroOverview.js`
- `Overview.js`
- `Page0Overview.js`
- `Page0.js`

oponder: `Overview.js` and `NextSteps.js` is fine, I guess we don't need the page numbers in the component file names.

PR #877, opened 2019-12-01
Allow setting secrets during app creation

oponder:
This allows you to upload a YAML file for secrets during app creation. Much like the Config Map.

### DIFF
--- a/__tests__/AppCatalog.js
+++ b/__tests__/AppCatalog.js
@@ -1,0 +1,84 @@
+import '@testing-library/jest-dom/extend-expect';
+import {
+  appCatalogsResponse,
+  appsResponse,
+  AWSInfoResponse,
+  getPersistedMockCall,
+  ORGANIZATION,
+  orgResponse,
+  orgsResponse,
+  releasesResponse,
+  userResponse,
+  V4_CLUSTER,
+  v4AWSClusterResponse,
+  v4AWSClusterStatusResponse,
+  v4ClustersResponse,
+} from 'testUtils/mockHttpCalls';
+import { fireEvent, wait } from '@testing-library/react';
+import { renderRouteWithStore } from 'testUtils/renderUtils';
+import nock from 'nock';
+
+// Cluster and route we are testing with.
+const ROUTE = `/app-catalogs/`;
+
+// Tests setup
+const requests = {};
+
+// Responses to requests
+beforeAll(() => {
+  // prettier-ignore
+  requests.userInfo = getPersistedMockCall('/v4/user/', userResponse);
+  requests.info = getPersistedMockCall('/v4/info/', AWSInfoResponse);
+  requests.organizations = getPersistedMockCall(
+    '/v4/organizations/',
+    orgsResponse
+  );
+  requests.organization = getPersistedMockCall(
+    `/v4/organizations/${ORGANIZATION}/`,
+    orgResponse
+  );
+  requests.clusters = getPersistedMockCall('/v4/clusters/', v4ClustersResponse);
+  requests.cluster = getPersistedMockCall(
+    `/v4/clusters/${V4_CLUSTER.id}/`,
+    v4AWSClusterResponse
+  );
+  requests.status = getPersistedMockCall(
+    `/v4/clusters/${V4_CLUSTER.id}/status/`,
+    v4AWSClusterStatusResponse
+  );
+  requests.apps = getPersistedMockCall(
+    `/v4/clusters/${V4_CLUSTER.id}/apps/`,
+    appsResponse
+  );
+  requests.credentials = getPersistedMockCall(
+    `/v4/organizations/${ORGANIZATION}/credentials/`
+  );
+  requests.releases = getPersistedMockCall('/v4/releases/', releasesResponse);
+  requests.appcatalogs = getPersistedMockCall(
+    '/v4/appcatalogs/',
+    appCatalogsResponse
+  );
+});
+
+// Stop persisting responses
+afterAll(() => {
+  Object.keys(requests).forEach(req => {
+    requests[req].persist(false);
+  });
+});
+
+/************ TESTS ************/
+
+it('renders all the app catalogs in the app catalogs overview', async () => {
+  const { getByText, debug, getAllByText, getByTestId } = renderRouteWithStore(
+    ROUTE
+  );
+
+  await wait(() => {
+    expect(
+      getByText('Pick an App Catalog to browse all the Apps in it.')
+    ).toBeInTheDocument();
+  });
+
+  debug();
+});
--- a/__tests__/AppCatalog.js
+++ b/__tests__/AppCatalog.js
@@ -103,11 +103,39 @@ it('renders all apps in the app list for a given catalog', async () => {
   nock('https://catalogshost').get('/giantswarm-test-catalog/index.yaml').reply(200, catalogIndexResponse);
   nock('https://catalogshost').get('/helmstable/index.yaml').reply(200, catalogIndexResponse);
 
-  const { debug, findByText } = renderRouteWithStore(
+  const { findByText } = renderRouteWithStore(
     ROUTE + "giantswarm-incubator/"
   );
 
   const catalogTitle = await findByText('Giant Swarm Incubator')
   expect(catalogTitle).toBeInTheDocument();
+});
+
+it('renders all apps in the app list for a given catalog', async () => {
+  nock('https://catalogshost').get('/giantswarm-incubator-catalog/index.yaml').reply(200, catalogIndexResponse);
+  nock('https://catalogshost').get('/giantswarm-test-catalog/index.yaml').reply(200, catalogIndexResponse);
+  nock('https://catalogshost').get('/helmstable/index.yaml').reply(200, catalogIndexResponse);
+
+  const { findByText } = renderRouteWithStore(
+    ROUTE + "giantswarm-incubator/"
+  );
+
+  const catalogTitle = await findByText('Giant Swarm Incubator')
+  expect(catalogTitle).toBeInTheDocument();
+});
+
+it('renders the app detail page for a given app', async () => {
+  nock('https://catalogshost').get('/giantswarm-incubator-catalog/index.yaml').reply(200, catalogIndexResponse);
+  nock('https://catalogshost').get('/giantswarm-test-catalog/index.yaml').reply(200, catalogIndexResponse);
+  nock('https://catalogshost').get('/helmstable/index.yaml').reply(200, catalogIndexResponse);
+
+  const { findByText } = renderRouteWithStore(
+    ROUTE + "giantswarm-incubator/nginx-ingress-controller-app/"
+  );
+
+  // The app's description should be there.
+  // This comes from parsing the index.yaml, which is mocked in catalogIndexResponse.
+  const appDescription = await findByText('A Helm chart for the nginx ingress-controller')
+  expect(appDescription).toBeInTheDocument();
 });
 
--- a/__tests__/AppCatalog.js
+++ b/__tests__/AppCatalog.js
@@ -70,7 +70,7 @@ afterAll(() => {
 /************ TESTS ************/
 
 it('renders all non internal app catalogs in the app catalogs overview', async () => {
-  const { findByText, debug, getAllByText, getByTestId } = renderRouteWithStore(
+  const { findByText } = renderRouteWithStore(
     ROUTE
   );
 
--- a/__tests__/AppCatalog.js
+++ b/__tests__/AppCatalog.js
@@ -3,7 +3,9 @@ import {
   appCatalogsResponse,
   appsResponse,
   AWSInfoResponse,
+  catalogIndexResponse,
   getPersistedMockCall,
+  getMockCall,
   ORGANIZATION,
   orgResponse,
   orgsResponse,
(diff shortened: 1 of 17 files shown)


src/actions/clusterActions.js
```diff
@@ -348,7 +349,49 @@ export function clusterInstallApp(app, clusterID) {
...
+      if (Object.keys(app.secretsYAML).length !== 0) {
+        // If we have user config that we want to create, then
+        // fire off the call to create it.
+        appSecretsApi
+          .createClusterAppSecret(clusterID, app.name, {
+            body: app.secretsYAML,
+          })
+          .then(() => {
+            // The call succeeded, resolve the promise
+            resolve();
```
axbarsan: Why do you resolve here with no parameters, but on the other `resolve` call, you resolve it with an empty object?

src/components/app_catalog/app_detail/InstallAppForm.js
```diff
@@ -50,6 +54,15 @@ const InstallAppForm = props => {
         validationError={props.valuesYAMLError}
         value={props.valuesYAML}
       />
+
+      <FileInput
+        description='Sensitive configuration can be uploaded separately as a secret.'
+        hint={<React.Fragment>&nbsp;</React.Fragment>}
```
axbarsan: I think we should discuss if we'll use `<React.Fragment />` or `< />` from now on, we kind of have a mix of both right now

src/components/app_catalog/app_detail/InstallAppModal.js
```diff
@@ -124,6 +127,24 @@ const InstallAppModal = props => {
     reader.readAsText(files[0]);
   };
 
+  const updateSecretsYAML = files => {
+    var reader = new FileReader();
+
+    reader.onload = (function() {
```
axbarsan: Can you please refactor this to not be an IIFE?

src/components/app_catalog/app_detail/InstallAppModal.js
```diff
@@ -124,6 +127,24 @@ const InstallAppModal = props => {
...
+      return function(e) {
+        try {
+          let parsedYAML = yaml.safeLoad(e.target.result);
+          setSecretsYAML(parsedYAML);
+          setSecretsYAMLError('');
+        } catch (err) {
```
axbarsan: Also, [you can exclude the parameter, since it's not used](https://v8.dev/features/optional-catch-binding)

src/actions/clusterActions.js
ograu: If we are only interested in errors, `then()` could be completely removed.

src/actions/clusterActions.js
```diff
@@ -348,7 +349,49 @@ export function clusterInstallApp(app, clusterID) {
...
+              );
+            }
+
+            reject(error);
+          });
+      } else {
+        // Otherwise, no user config, so don't do anything.
+        resolve({});
+      }
+    });
```
ograu: This is too much for me! :-)
I am ok with not refactoring some of the things we have, but we could use `async/await` for new code and as a way to avoid too much nesting.
It's hard when something fails and you have to debug things like this.

If you need to return a promise, as it seems to be the case, I would write this probably this way:

```javascript
const optionalCreateAppSecret = Object.keys(app.valuesYAML).length === 0
    ? Promise.resolve()
    : await appSecretsApi
          .createClusterAppSecret(clusterID, app.name, {
               body: app.secretsYAML,
           })
          .catch(functionThatHandlesTheseErrrors(error);
```

ograu: `functionThatHandlesTheseErrrors` makes sense here since I think you are using the same errors twice. 

This code above is something that I find easy to read (I know, this is subjective...)

src/components/app_catalog/app_detail/InstallAppForm.js
oponder: Let's use `< />` only. 

I'll fix that in a new PR.

ping @ograu @ubergesundheit

src/components/app_catalog/app_detail/InstallAppModal.js
oponder: Done. I'm really not sure what that was there like that.  It's not like I am trying to use that parameter. Kinda confused myself.

src/components/app_catalog/app_detail/InstallAppForm.js
ubergesundheit: `<></>` is generally nicer than `<React.Fragmen></React.Fragment>`! I like this decision. There one exception I can think of when returning fragments from `.map`. Here using `<React.Fragment>` seems the way to go to add the required `key=` attribute for iteration.

src/actions/clusterActions.js
```diff
@@ -300,128 +300,157 @@ export function clusterLoadApps(clusterId) {
...
-    var optionalCreateAppConfiguration = new Promise((resolve, reject) => {
-      if (Object.keys(app.valuesYAML).length !== 0) {
-        // If we have user config that we want to create, then
-        // fire off the call to create it.
-        appConfigsApi
+    try {
+      // If we have user config that we want to create, then
+      // fire off the call to create it.
+      Object.keys(app.valuesYAML).length !== 0 &&
+        (await appConfigsApi
```
axbarsan: Please move this to a separate variable, as it is really difficult to read

src/actions/clusterActions.js
```diff
@@ -300,128 +300,157 @@ export function clusterLoadApps(clusterId) {
...
+              clusterID,
+              error
+            );
+            throw error;
+          }));
+
+      // If we have an app secret that we want to create, then
+      // fire off the call to create it.
+      Object.keys(app.secretsYAML).length !== 0 &&
+        (await appSecretsApi
```
axbarsan: Same here

src/actions/clusterActions.js
```diff
@@ -300,128 +300,157 @@ export function clusterLoadApps(clusterId) {
...
+          .createClusterAppSecret(clusterID, app.name, {
+            body: app.secretsYAML,
           })
           .catch(error => {
```
axbarsan: I would suggest keeping everything consistent and always using `try/catch` (since we'll need it for stacking multiple promises together in most places)

src/actions/clusterActions.js
```diff
@@ -300,128 +300,157 @@ export function clusterLoadApps(clusterId) {
...
 }
 
+/**
+ * appConfigErrorFlashMessages provides flash messages when something went wrong
+ * when creating either the ConfigMap or Secret during app installation.
+ *
+ * @param {string} resource Name of the resource we were trying to create.
+ * @param {string} appName Name of the app.
+ * @param {string} clusterID Where we tried to install the app on.
+ *  @param {object} error The error that occured.
```
axbarsan: Formatting is wrong here

src/actions/clusterActions.js
```diff
@@ -300,128 +300,157 @@ export function clusterLoadApps(clusterId) {
...
+    );
+  }
+}
+
+/**
+ * appInstallationErrorFlashMessage provides flash messages for failed app creation.
+ *
+ * @param {string} appName Name of the app.
+ * @param {string} clusterID Where we tried to install the app on.
+ *  @param {object} error The error that occured.
```
axbarsan: Same here

src/components/app_catalog/app_detail/InstallAppModal.js
```diff
@@ -109,17 +112,31 @@ const InstallAppModal = props => {
...
-          setValuesYAML(parsedYAML);
-          setValuesYAMLError('');
-        } catch (err) {
-          setValuesYAMLError('Unable to parse valid YAML from this file.');
-        }
-      };
-    })(files[0]);
+    reader.onload = function(e) {
+      try {
+        let parsedYAML = yaml.safeLoad(e.target.result);
```
axbarsan: This should be a `const`

src/components/app_catalog/app_detail/InstallAppModal.js
```diff
@@ -109,17 +112,31 @@ const InstallAppModal = props => {
...
-        try {
-          let parsedYAML = yaml.safeLoad(e.target.result);
-          setValuesYAML(parsedYAML);
-          setValuesYAMLError('');
-        } catch (err) {
-          setValuesYAMLError('Unable to parse valid YAML from this file.');
-        }
-      };
-    })(files[0]);
+    reader.onload = function(e) {
```
axbarsan: Please use an arrow function here

src/components/app_catalog/app_detail/InstallAppModal.js
```diff
@@ -109,17 +112,31 @@ const InstallAppModal = props => {
   const updateValuesYAML = files => {
     var reader = new FileReader();
```
axbarsan: This should be a `const`

src/components/app_catalog/app_detail/InstallAppModal.js
```diff
@@ -109,17 +112,31 @@ const InstallAppModal = props => {
...
+        setValuesYAMLError('');
+      } catch {
+        setValuesYAMLError('Unable to parse valid YAML from this file.');
+      }
+    };
+
+    reader.readAsText(files[0]);
+  };
+
+  const updateSecretsYAML = files => {
```
axbarsan: Same for this function as the one before

src/actions/clusterActions.js
```diff
@@ -300,128 +300,157 @@ export function clusterLoadApps(clusterId) {
...
-                messageTTL.LONG
-              );
-            } else {
-              new FlashMessage(
-                `Something went wrong while trying to install your app. Please try again later or contact support: support@giantswarm.io`,
-                messageType.ERROR,
-                messageTTL.LONG
-              );
-            }
+            appConfigErrorFlashMessages('Secret', app.name, clusterID, error);
```
axbarsan: I'd like to propose always using a verb as the first word of a function name

src/components/app_catalog/app_detail/InstallAppForm.js
ograu: I like both. Happy with whatever you decide.

oponder:
Thanks for the reviews so far! I'll also add some tests before considering this done.

src/actions/clusterActions.js
oponder: so a nested try catch in this case?

I actually moved this now out to it's own function, still using `.catch` but let me know what you think. Will push in a sec.

src/actions/clusterActions.js
oponder: Thx for the help here! 

It's been broken down further now.

src/actions/clusterActions.js
```diff
@@ -300,128 +300,177 @@ export function clusterLoadApps(clusterId) {
...
+ * createAppConfig takes an app and a clusterID and tries to create the config map
+ * of that app. The app object must have a valuesYAML field representing the config
+ * map to be created.
+ * @param {object} app
+ * @param {string} clusterID
+ */
+async function createAppConfig(app, clusterID) {
+  const appConfigsApi = new GiantSwarm.AppConfigsApi();
+
+  return await appConfigsApi
```
axbarsan: This would be bad practice, also impossible to debug.

[Please read up on this](https://eslint.org/docs/rules/no-return-await)

src/components/app_catalog/app_detail/InstallAppModal.js
```diff
@@ -107,19 +110,33 @@ const InstallAppModal = props => {
...
-        } catch (err) {
-          setValuesYAMLError('Unable to parse valid YAML from this file.');
-        }
-      };
-    })(files[0]);
+    const reader = new FileReader();
+
+    reader.onload = e => {
+      try {
+        let parsedYAML = yaml.safeLoad(e.target.result);
```
axbarsan: I think you forgot about these 2

src/actions/clusterActions.js
oponder: Ok, removed the return. thx for the link

__tests__/AppCatalog.js
```diff
@@ -0,0 +1,141 @@
...
+    `/v4/organizations/${ORGANIZATION}/credentials/`
+  );
+  requests.releases = getPersistedMockCall('/v4/releases/', releasesResponse);
+
+  requests.appcatalogs = getPersistedMockCall(
+    '/v4/appcatalogs/',
+    appCatalogsResponse
+  );
+
+  nock('https://cors-anywhere.herokuapp.com').get(/.*$/).reply(500, 'Not supposed to get here in these tests.').persist();
```
axbarsan: You can use `getPersistedMockCall` for this

__tests__/AppCatalog.js
```diff
@@ -0,0 +1,141 @@
...
+afterAll(() => {
+  Object.keys(requests).forEach(req => {
+    requests[req].persist(false);
+  });
+});
+
+/************ TESTS ************/
+
+it('renders all non internal app catalogs in the app catalogs overview', async () => {
+  nock('https://catalogshost').get('/giantswarm-incubator-catalog/index.yaml').reply(200, catalogIndexResponse);
```
axbarsan: For these 3, as well

__tests__/AppCatalog.js
oponder: I kind of like the story this tells, that these 3 requests happen once in this test.

oponder: I think for now I don't want to sweep under the rug that all catalog indexes are loaded, even if we just go straight to a specific catalog.

axbarsan: What I meant is using `getPersistedMockCall` basically does the same thing, but in a shorter form.
Although I now realize that this returns status code `500`, which is not supported by `getPersistedMockCall`. So it will only work for the next 3 ones in `renders all non internal app catalogs in the app catalogs overview`

axbarsan: Nevermind, I didn't realize `getPersistendMockCall` does not support custom domains and custom status codes.

PR #879, opened 2019-12-02
Azure kvm cluster views tests

ograu:
In this PR:

- Added integration tests for Azure and KVM views
- Added custom mock responses, and 
- Added files, so more integration tests can be added (scaling nodes for example)


test_utils/mockHttpCalls.js
```diff
@@ -53,42 +54,39 @@ export const infoResponse = {
...
     },
   },
 };
 
+export const azureInfoResponse = {
+  general: {
+    availability_zones: { default: 1, max: 1 },
+    datacenter: 'westeurope',
+    installation_name: 'ghost',
+    provider: Providers.AZURE,
```
oponder: Since this is technically out from the API, I don't know if we need the Providers.AZURE thing here.

That said, I don't mind.

But I would like to keep it the same between the types of calls. So in `KVMINfoResponse` we use just `kvm` instead of `Providers.KVM`. Can we align this here so it's the same?

Also noticed further down `Providers.azure` a few times (no uppercase).

https://github.com/giantswarm/happa/pull/879/files#diff-f5234f59c0885943656b6bba42845e98R403

__tests__/V4AWSClusterManagement.js
```diff
@@ -79,7 +79,7 @@ afterAll(() => {
 // and hence the warning because we are transforming an array into an array
 it('renders all the v4 AWS cluster data correctly without nodes ready', async () => {
   const div = document.createElement('div');
-  const { getByText, getAllByText, getByTestId } = renderRouteWithStore(
+  const { getByText, getAllByText, getByTestId, debug } = renderRouteWithStore(
```
oponder: debug not used here.
```suggestion
  const { getByText, getAllByText, getByTestId } = renderRouteWithStore(
```

__tests__/V4KVMClusterManagement.js
```diff
@@ -0,0 +1,108 @@
...
+// Stop persisting responses
+afterAll(() => {
+  Object.keys(requests).forEach(req => {
+    requests[req].persist(false);
+  });
+});
+
+it('renders all the v4 KVM cluster data correctly without nodes ready', async () => {
+  const div = document.createElement('div');
+  const { getByText, getAllByText, debug } = renderRouteWithStore(
```
oponder:
```suggestion
  const { getByText, getAllByText } = renderRouteWithStore(
```

test_utils/mockHttpCalls.js
ograu: Changed, thanks!

PR #880, opened 2019-12-03
Mock console.error in a test that triggers it when it's green

ograu:


oponder:
Any specific reason? I think we should keep it, but instead figure out if we can surpress or remove a `console.error` somewhere in the app where it is causing confusion instead.

We're also tending to have no `console` whatsoever in the app at all, with the new stricter linting. So maybe that would solve this.

ograu:
Nevermind, you are right: there's no reason to silence `console.error` in tests, we actually want to see them.

I changed this and instead of mocking it globally, now I am mocking `console.error` only inside the test that is polluting the terminal.

> We're also tending to have no `console` whatsoever in the app at all, with the new stricter linting. So maybe that would solve this.

really? 🤦‍♂  didn't saw that one

PR #882, opened 2019-12-03
Data flow refactor POC

ograu:
Towards https://github.com/giantswarm/giantswarm/issues/7521

This is a simplified POC, no loading flags and still action creators dispatching thunks.
The general idea is simplify components, avoid performing chained api calls in there and document better the data flow.

What do you think about this? Do you like it? What would you change?
Do you think it's worth it?


ograu:
> It doesn't seem like a real change in the dataflow though, but it's a nice grouping / refactoring to make the layout component less complex.

Yeah, sorry, too soon. 

Iterating on it and trying to add some value, so we can decide if it pays off.

ograu:
Added a batch function for `organizationSelect` and modified `clustersLoad` so now it only fetches clusters that belong to the selected organization.

ograu:
Merging if everyone is ok with this. cc/ @axbarsan  @ubergesundheit

PR #884, opened 2019-12-03
Get rid of componentWillMount and componentWillReceiveProps

axbarsan:
Fixes giantswarm/giantswarm#7059

- Further configure eslint rules to prevent false alarms
- Update 'react-bootstrap' to the latest version
- Update our components to use the right 'react-bootstrap' components and props


axbarsan:
Closed because updating `bootstrap` broke pretty much all of our styling, and fixing it would've taken too much time, taking into consideration that we're planning of getting rid of bootstrap in the future

PR #885, opened 2019-12-04
Use short form of React.Fragment

oponder:
Replaces `<React.Fragment></React.Fragment>` with `<></>` everywhere.

### DIFF
--- a/src/components/app_catalog/catalog_list/CatalogList.js
+++ b/src/components/app_catalog/catalog_list/CatalogList.js
@@ -23,50 +23,46 @@ class CatalogList extends React.Component {
               resources yet for your installation. Please come back later!
             </p>
           ) : (
-            <>
-              <div className='app-catalog--repos'>
-                {Object.keys(this.props.catalogs.items).map(catalogName => {
-                  return (
-                    <div
-                      className='app-catalog--repo'
-                      key={this.props.catalogs.items[catalogName].metadata.name}
-                    >
-                      <img
-                        height='100px'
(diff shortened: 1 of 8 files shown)


src/components/app_catalog/catalog_list/CatalogList.js
```diff
@@ -23,7 +23,7 @@ class CatalogList extends React.Component {
               resources yet for your installation. Please come back later!
             </p>
           ) : (
-            <React.Fragment>
+            <>
```
ubergesundheit: Is this one really needed? I see only one child

oponder: Yeah doesn't look like it needs to be there. Removing.

src/components/cluster/detail/ClusterApps.js
```diff
@@ -215,7 +215,7 @@ class ClusterApps extends React.Component {
 
   render() {
     return (
-      <React.Fragment>
+      <>
```
ubergesundheit: Could be refactored to return early like

```
if (!this.props.showInstalledAppsBlock) {
  return null;
}
```

but this is fine I guess

oponder: :+1: I'm not a big fan of the `&&` conditional rendering pattern but at one point I accepted it into the codebase as part of another review. This was way before the other front end engineers joined too. 

We can always discuss if we want to avoid that pattern, for now let's keep it.

PR #887, opened 2019-12-05
Tests: Add appInstallation placeholder. Rename files according to fmt.

oponder:
Hey guys, 

Started work on the app installation tests and had a question about the filenames for tests.

I wouldn't consider them components, so the filename should start with a lowercase right?

Thought I would start the discussion right away with a PR.

Also we have to make sure to set: `git config core.ignorecase false` in our git configs.

Otherwise it will not notice these types of renames. (fyi @ubergesundheit)


oponder:
I think I always had a voodoo feeling towards caps in filenames because of this. But now I understand why and can deal with it in my config.

axbarsan:
I would use PascalCase for everything tbh, unless it's a file with an utility function (or a React hook), then it would be camelCase

oponder:
I find PascalCase for filenames really weird. 

But I'm ok when it indicates this is a renderable react `Component`, because that's something really specific and means something in our context.

So I'd like to keep it for that specific case, instead of making it the general file name default.

ubergesundheit:
> Also we have to make sure to set: git config core.ignorecase false in our git configs.
> 
> Otherwise it will not notice these types of renames. (fyi @ubergesundheit)

thanks. Maybe this is a macOS thing (since I think the whole underlying filesystem is not case aware?)? Here on linux git noticed all renames I did already

For the files in the `__tests__` I'll just won't rename them in my upcoming PR

oponder:
Not needed anymore, already fixed by @ubergesundheit in a different commit. 👍

PR #888, opened 2019-12-05
File name convention 

ubergesundheit:
This is the first bunch of file/folder renames related to giantswarm/giantswarm#7915 according to the rules layed out in fmt.

This should be squashed I think


oponder:
> This should be squashed I think

It will be squashed on merge, as configured in github.

src/components/app_catalog/AppList/AppList.js
```diff
@@ -19,7 +19,7 @@ const AppList = ({ catalog, ...props }) => {
       }}
     >
       <DocumentTitle title={'Apps | Giant Swarm '}>
-        <>
+        <React.Fragment>
```
oponder: I guess things got a bit confused here, sorry for doing large changes while you're also doing the filename change.

ubergesundheit: I'm going to change this one to the short form in here

ubergesundheit:
I'll try to fix the conflicts now and start another round of renamings ;)

src/components/app_catalog/AppList/AppList.js
ubergesundheit: But I guess the Fragment isn't required at all.. Will remove it

PR #889, opened 2019-12-06
Don't show empty state when switching orgs.

oponder:
Fixes the empty state that pops up when switching orgs.

### DIFF
--- a/src/components/Home/Home.js
+++ b/src/components/Home/Home.js
@@ -93,7 +93,7 @@ class Home extends React.Component {
                     <i className='fa fa-add-circle' /> Launch New Cluster
                   </Button>
                 </Link>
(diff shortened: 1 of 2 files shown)


oponder:
This doesn't add a spinner, that requires a bit of thought, adding it on the top makes it push the list down, which looks odd with the transition that is also happening.

I'd vote to remove the transitions on the overview. Then we can instantly snap to the spinner in the middle of the screen and back to a full list when things have loaded. ping @marians

marians:
Related to https://github.com/giantswarm/giantswarm/issues/7979

oponder:
Closing. We're rolling back with: https://github.com/giantswarm/happa/pull/890

PR #894, opened 2019-12-06
Configure rules further for better linting

axbarsan:
(Removed `no-invalid-this` since it gets triggered by class methods defined as expressions)

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -55,7 +55,10 @@
     "consistent-return": 2,
     "default-param-last": 2,
     "eqeqeq": 2,
-    "dot-notation": 2,
+    "dot-notation": [
+      2,
+      { "allowPattern": "^(([a-z]+_[a-z]+)|([A-Z][^A-Z]+))+$" }
+    ],
     "guard-for-in": 2,
     "max-classes-per-file": 2,
     "no-alert": 2,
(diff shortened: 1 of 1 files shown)


axbarsan:
- `dot-notation` is about using object properties like `Object.property` or `Object['property']`
By default, it will give us a warning for using something like `Object['property_name']` (which is not possible to write as `Object.property_name`) or `Object['PropertyName']` (which could lead to issues if used like `Object.PropertyName`, due to the confusion that it might produce)
- As for false `no-invalid-this`, take this example:
```js
class User {
  #name = '';

  constructor(name) {
    this.#name = name;
  }

  async register() {
    return UserModel.register(this.name);
  }

  login = async (password) {
    return UserModel.login(this.name, password)
  }
}
```

In this case, there is no warning for the `register` method, that uses the `this` keyword, but there is a warning for the `login` method, saying that `this` is used in an invalid context, which is not true. That's the reason for disabling it

ubergesundheit:
Thanks for the clarification :)

I don't know enough about your exact case why you're deactivating this but in the example you've posted shouldn't the rule prevent having mixed function declaration?

so

```js
  login = async (password) {
    return UserModel.login(this.#name, password);
  }
```

would become

```js
  async login (password) {
    return UserModel.login(this.#name, password);
  }
```

axbarsan:
This rule's point is to prevent the use of `this` where it shouldn't be used (aka in functions declarations where it exists, but it points to the function execution context).

The reason we're using sometimes using expressions as methods is because event handlers (we don't use `.bind()` in the constructor) run in a different execution context than where they are declared, so `this`, if used in the method's body, will be the `window` object, not the actual parent object

[You can find out more about this here, under the `this` subsection](https://www.w3schools.com/js/js_arrow_function.asp)

ubergesundheit:
Thanks for claifying the event handler usage. Didn't had this one in mind..

PR #895, opened 2019-12-09
RefreshableLabel doesn't flash on mounting

ograu:
Fixes https://github.com/giantswarm/giantswarm/issues/7983

This doesn't work for the text next to the instance type in cluster details view, because this is something that it is computed after the first render:

image

Any ideas of what we want to do here?


oponder:
Towards: https://github.com/giantswarm/giantswarm/issues/7983

PR #898, opened 2019-12-10
Remove beta tags

ograu:
Fixes https://github.com/giantswarm/giantswarm/issues/8031


ograu:
cc/ @SimonKingZA merging into master, will be in all testing installations in a while.

PR #899, opened 2019-12-10
Add Multi-AZ support for Azure clusters

axbarsan:
References https://github.com/giantswarm/giantswarm/issues/7738 , https://github.com/giantswarm/giantswarm/issues/8601

Makes Azure Multi-AZ available for release version 11.1.0

* Add support for Azure V4 Multi-AZ
* Refactor the Multi-AZ selector and move to another file
* Refactor the AZ Label to allow having zones as both, letters and numbers
* Format `index.html` with `prettier`

Functionality wise, it is not any different compared to AWS Multi-AZ.
The only difference is that the labels are numbers, as opposed to AWS, which are letters.

**Preview**

Screenshot 2020-01-14 at 16 32 12


marians:
@axbarsan Can you give a list of places in happa where this PR should have effects? e. g. cluster creation form, cluster details page, ...

axbarsan:
> @axbarsan Can you give a list of places in happa where this PR should have effects? e. g. cluster creation form, cluster details page, ...

This affects V4 cluster creation and viewing the details of existing V4 clusters

src/components/Cluster/NewCluster/AvailabilityZonesSelector.js
```diff
@@ -0,0 +1,83 @@
...
+const AvailabilityZonesSelector = ({
+  minValue,
+  maxValue,
+  value,
+  readOnly,
+  onChange,
+  requiredReleaseVersion,
+  currentReleaseVersion,
+}) => {
+  const isAvailable = cmp(currentReleaseVersion, requiredReleaseVersion) >= 0;
```
oponder: Is there a way we can change this to being computed on the cluster object (computeCapabilities) ?

Something like  `cluster.hasAvailabilityZones` ?

The goal is to move this kind of release and provider checking logic to one place.

It might mean that for this component we cant show the exact release version that would be required to support this feature, but I think that's ok.

So code like this: https://github.com/giantswarm/happa/pull/899/files#diff-d6aab637bd5ecfe4e323d3e499b95dddR77

Would join the computeCapabilities block.

axbarsan: I'll add that now 👍

axbarsan: That would work if we were fetching any clusters from the server. This check is made during cluster creation, where no clusters are fetched, so the only way we could use `computeCapabilities` is if we would use it separately, as an utility function.

oponder: Oh yeah! True. Ok lets leave it as is for now. Maybe we refactor computeCapabilities to take a release version and a provider, instead of a cluster and a provider. Because I think those are the things we really check on. That said, lets see how computeCapabilities changes after @ograu is done refactoring and then revisit this idea.

There are a few other places I think where we still do version checks in 'view' components, and I'd like to move that out. Maybe we can tackle it in a more coordinated effort.

But I'm fine for leaving it in this PR for now.

ograu: > That would work if we were fetching any clusters from the server. This check is made during cluster creation, where no clusters are fetched, so the only way we could use `computeCapabilities` is if we would use it separately, as an utility function.

We are actually fetching cluster details right after creation... check `batchedClusterCreate()`

src/components/Cluster/NewCluster/AvailabilityZonesSelector.js
```diff
@@ -0,0 +1,83 @@
...
+  minValue: PropTypes.number,
+  maxValue: PropTypes.number,
+  onChange: PropTypes.func,
+  readOnly: PropTypes.bool,
+  value: PropTypes.number,
+  requiredReleaseVersion: PropTypes.string,
+  currentReleaseVersion: PropTypes.string,
+};
+
+export default AvailabilityZonesSelector;
```
ograu: This component kind of overlaps with `AvailabilityZonesParser`, ¿Why you decided to wrote this one instead of changing that component to support this use case?

src/components/Cluster/NewCluster/AvailabilityZonesSelector.js
oponder: Yeah thats for getting to the cluster detail screen.

But this component is also used on the create cluster screen, where we don't have a cluster object yet. 


So computeCapabiltiies isn't the right function for this. I mean you could try to pass in a faked cluster object with just a release version set on it....

But what I think we should really do, is make a pure function, that doesn't modify the 'cluster' object that we pass in, but just returns a capabilities object. And then the implementing code can decide whether or not it wants to assign that to anything.

If what I'm saying is not making any sense, don't worry. I think we need to let the dust settle again a bit.

Lets talk in sig front end sync tomorrow? Or if you feel it is conflcting with the direction you're going in now, lets get on a call today.

ograu: Nono, I think this is ok, I am more concerned about the flow, I mean __where__ this code is executed, and I think that this definitely should be a pure function probably outside actions files.

src/components/Cluster/NewCluster/AvailabilityZonesSelector.js
axbarsan: I did not write this, this was taken out of the `CreateRegularCluster` component

ograu: Anyway, a lot of content is duplicated. And probably we have too many `AvailabilityZonesXxxx` components.

I'm ok with leaving it like this if you think it's not worth to modify `AvailabilityZonesParser`, but at least we could change names to something like `AZSelectorV5/V4` to make things easier.

axbarsan: I'll add an adapter to use `AvailabilityZonesParser` behind the scenes, and rename it to something like `V4AvailabilityZonesSelector`, to not over-complicate `AvailabilityZonesParser`

src/index.html
```diff
@@ -28,7 +28,7 @@
 
     <script>
       var config = {
-        apiEndpoint: 'http://localhost:8000',
+        apiEndpoint: 'https://api.g8s.godsmack.westeurope.azure.gigantic.io',
```
marians: @axbarsan That wasn't intended, was it?

axbarsan: oh 💩

marians: I think I just commented an outdated commit. Looks fine in the final version. Never mind.

axbarsan: Just checked as well, now

Scared me to death haha, great way to kickstart the day 👍

PR #901, opened 2019-12-11
Remove ESLint no-use-before-define

ograu:
Are you ok deleting this? I think it doesn't make much sense ...

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -135,7 +135,6 @@
     "no-shadow-restricted-names": 2,
     "no-undef": 2,
     "no-undef-init": 2,
-    "no-use-before-define": 2,
     "constructor-super": 2,
     "no-class-assign": 2,
     "no-confusing-arrow": 2,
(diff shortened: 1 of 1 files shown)


oponder:
This would stop us from using variables before we define them.. I think that makes a lot of sense right?

Do you have some code snippets where this rule triggers in a way you don't like?

ograu:
I don't like that it is giving us a lot of false positives in functions.

I've changed it, ¿Does it makes sense now?

ubergesundheit:
I guess I'm on the same page as @oponder that this is an useful rule to keep. Even though all the code we write in happa gets transpiled and minified down, still I think having this rule as an error for all usage befor define make sense

ograu:
> I guess I'm on the same page as @oponder that this is an useful rule to keep. Even though all the code we write in happa gets transpiled and minified down, still I think having this rule as an error for all usage befor define make sense

Not sure if I understand it, then ¿Are you ok with what I am proposing?

ubergesundheit:
> Not sure if I understand it, then ¿Are you ok with what I am proposing?

should be fine :) :+1:

PR #909, opened 2019-12-18
Move app related actions into appActions.js

oponder:
I had originally placed app related actions into the clusterActions.js file because of the pathnaming. We're installing an app on a cluster, so I felt it could be in there.

But really, an app is a concept on it's own and deserves to be in the appActions file.


oponder:
Related: https://github.com/giantswarm/giantswarm/issues/8170

PR #910, opened 2019-12-19
Use V5 app endpoints for V5 clusters.

oponder:
Builds on: https://github.com/giantswarm/happa/pull/909
Towards: https://github.com/giantswarm/giantswarm/issues/8170

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "copy-to-clipboard": "3.2.0",
     "deep-diff": "1.0.2",
     "emotion-theming": "^10.0.10",
-    "giantswarm": "git+https://github.com/giantswarm/giantswarm-js-client.git",
+    "giantswarm": "git+https://github.com/giantswarm/giantswarm-js-client.git#add-v5-app-endpoints",
     "history": "^4.7.2",
     "immer": "^3.2.0",
     "immutable": "3.8.2",
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "copy-to-clipboard": "3.2.0",
     "deep-diff": "1.0.2",
     "emotion-theming": "^10.0.10",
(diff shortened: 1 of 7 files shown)


src/index.html
```diff
@@ -1,67 +1,90 @@
-<!doctype html>
+<!DOCTYPE html>
```
oponder: this isn't actually in the commit.. I think github is caching the diff in a weird way.

src/actions/appActions.js
```diff
@@ -11,29 +13,26 @@ export function selectCluster(clusterID) {
...
+  console.log('loading1');
   return function(dispatch, getState) {
-    // This method is going to work for NP clusters, now in local dev it is not
-    // working, so early return if the cluster is a NP one.
+    const appsApi = new GiantSwarm.AppsApi();
+
     const nodePoolsClusters = getState().entities.clusters.nodePoolsClusters;
     const isNodePoolsCluster = nodePoolsClusters.includes(clusterId);
+
+    let getClusterAppsFunc;
```
oponder: Here we decide which api method to call, the v5 or the v4 version, based on whether or not this is a nodepool cluster.

A similar pattern is applied to all the other actions.

src/index.html
oponder: fixed

package.json
```diff
@@ -47,7 +47,7 @@
     "copy-to-clipboard": "3.2.0",
     "deep-diff": "1.0.2",
     "emotion-theming": "^10.0.10",
-    "giantswarm": "git+https://github.com/giantswarm/giantswarm-js-client.git",
+    "giantswarm": "git+https://github.com/giantswarm/giantswarm-js-client.git#add-v5-app-endpoints",
```
ograu: I guess this is needed by now

PR #911, opened 2019-12-20
Make app catalog chart

oponder:
towards: https://github.com/giantswarm/giantswarm/issues/8162


marians:
The happa image is public (https://quay.io/repository/giantswarm/happa?tab=settings) so we can remove the docker pull secret completely.

PR #913, opened 2019-12-20
Fix ESLint problems

axbarsan:
Fixes https://github.com/giantswarm/giantswarm/issues/8192

- Fix or ignore the problems (depends by situation)
- Add linting as a pre-commit hook
- Add linting as a task in CI
- Add ESLint rule for requiring a newline before returning


src/actions/appActions.js
```diff
@@ -19,19 +19,18 @@ export function loadApps(clusterId) {
...
+      getClusterApps = appsApi.getClusterAppsV5;
     }
 
     dispatch({
       type: types.CLUSTER_LOAD_APPS,
       clusterId,
     });
 
-    return getClusterAppsFunc(clusterId)
+    return getClusterApps(appsApi, clusterId)
```
oponder: Does this refactoring work? I'll test it in a bit. But based on my understanding, I think it shouldn't work.

The first argument to `getClusterAppsV4` or `V5` is the `clusterId`.

Before I was getting the function and using `bind` to keep the executing scope (`this`) of the function the same. Now it's being called with the wrong first argument, and it's scope will be `undefined` I think.

src/actions/clusterActions.js
```diff
@@ -632,24 +658,27 @@ export function clusterCreateKeyPair(clusterId, keypair) {
       keypair,
     });
 
-    var keypairsApi = new GiantSwarm.KeyPairsApi();
+    const keypairsApi = new GiantSwarm.KeyPairsApi();
+
     return keypairsApi
       .addKeyPair(clusterId, keypair)
-      .then(keypair => {
+      .then(pair => {
```
oponder: Just curious, why rename this?

src/actions/nodePoolActions.js
```diff
@@ -154,10 +196,10 @@ export function nodePoolsCreate(clusterId, nodePools) {
       nodePools.map(nodePool => {
         return nodePoolsApi
           .addNodePool(clusterId, nodePool)
-          .then(nodePool => {
+          .then(pool => {
```
oponder: Wondering here as well about the variable renaming. We don't really ever talk about `pools` or `pairs`

src/lib/pageVisibilityTracker.js
```diff
@@ -23,8 +23,9 @@ class PageVisibilityTracker {
   addEventListener(func) {
     if (
       typeof document.addEventListener === 'undefined' ||
-      this.hidden === undefined
+      typeof this.hidden === 'undefined'
```
oponder: saw this refactoring in other places as well. Is it really better than just checking for `=== undefined` ?

src/utils/clusterUtils.js
```diff
@@ -71,7 +77,7 @@ export function getNumberOfNodePoolsNodes(nodePools = []) {
 
 export function getMemoryTotalNodePools(nodePools = []) {
   if (!window.config.awsCapabilitiesJSON || nodePools.length === 0) {
-    return;
+    return 0;
```
oponder: Gotta make sure this doesn't have any unindented consequences.

src/lib/pageVisibilityTracker.js
ubergesundheit: `x === undefined` != `typeof x === 'undefined'`

The first one will throw an error be cause x is undefined but `typeof x === 'undefined'` will correctly return a boolean:
image

src/actions/appActions.js
axbarsan: Looks like I was too excited to go on holiday, don't know what I was thinking about 😆

src/actions/clusterActions.js
axbarsan: Because it has the same name as one of the function arguments, and although it's not a problem due to scoping, it can cause confusion, and that's the reason for having a linter rule for this

src/actions/nodePoolActions.js
axbarsan: Same here

PR #919, opened 2019-12-30
Update handlebars version to fix CVE-2019-19919

omegas27:
Towards https://github.com/giantswarm/giantswarm/issues/8212.

I only updated the minimal dependency to the one currently installed in yarn.lock which is the latest version of handlebar

I did not upgrade all depencies as I am not sure how I can test if upgrading them do not break the UI


tfussell:
Did you try `yarn test`? Not sure what it tests exactly, but it could be something. I think I got Happa running locally once, but I don't remember how complicated it was.

omegas27:
It seems I need to clone api and run it locally to play all my tests. Will do

PR #926, opened 2020-01-03
Refactor organizations list test

ograu:
I broke this test while working on another PR and decided to refactor it:

- Removed navigation part because I found it too cumbersome. Also not sure that this belongs to this test actually (imagine that we change our navigation in the future, or we want to render this component in another view with another route). Let's discuss it if you don't agree with me.
- I have used `data-testid`. If there is no other clear/nice/easy way I think that it is the way to go instead of testing implementation details.
- Now this test is checking that we are displaying the correct data


oponder:
I agree with the navigation part maybe not needing to be part of this test. But it was nice to have that covered in some way.

Can we add it back as a separate case instead of losing it entirely?

ograu:
Ok, new test with the navigation part added!

PR #928, opened 2020-01-03
Refactor cluster data flow

ograu:
Towards https://github.com/giantswarm/giantswarm/issues/7521

Sorry for the size of the PR, I have had to do a lot of changes...

In this PR:

- Added reducer for loading flags: the idea is to start removing loading flags from all other reducers and have this 'no-maintenance' reducer that will take care of all the flags as long as we follow the convention of adding `request/success/error/finished` suffixes to actions. We can use `finished` when we don't want to use `success/error`. There are some situations when errors are handled by other functions and we just want to know when the call is finished.

- Added request/success/error/finished action types

- Added options object to some action creators for setting/not setting loading flags

- Changed tests _without nodes/nps ready_ to actually check for 0 nodes and leave as todos with `.skip`s the tests with nodes/node pools

- Changed `isNodePoolsCluster` for `isV5Cluster`

- Removed the use of `.then()` in components to avoid memory leaks. And also as a good practice, because in most of the cases we want to use redux store instead of local state.


src/components/Home/Home.js
```diff
@@ -57,9 +59,13 @@ class Home extends React.Component {
    */
   title() {
     if (this.props.selectedOrganization) {
-      return `Cluster Overview | ${this.props.selectedOrganization}`;
+      return (
+        'Cluster Overview | ' +
+        this.props.selectedOrganization +
+        ' | Giant Swarm'
```
axbarsan: The `| Giant Swarm` suffix gets automatically added by the `useDocumentTitle` hook inside the `DocumentTitle` component

axbarsan:
Really like the universal loading reducer idea. 👍 
However, I think it should be heavily documented, as there's no obvious clue to its usage

ograu:
Thanks @axbarsan !

Regarding documentation, totally agree with you. Still thinking on how and where to write documentation for this. Any suggestion will be appreciated.

__tests__/V4AWSClusterManagement.js
```diff
@@ -86,10 +86,10 @@ it('renders all the v4 AWS cluster data correctly without nodes ready', async ()
   expect(getAllByText(V4_CLUSTER.id)).toHaveLength(2);
   expect(
     getByTestId('desired-nodes').querySelector('div:nth-child(2)').textContent
-  ).toBe('3');
+  ).toBe('0');
   expect(
     getByTestId('running-nodes').querySelector('div:nth-child(2)').textContent
-  ).toBe('3');
+  ).toBe('0');
```
oponder: why did this become 0?

__tests__/V4KVMClusterManagement.js
```diff
@@ -91,12 +91,20 @@ it('renders all the v4 KVM cluster data correctly without nodes ready', async ()
...
-    status: v4KVMClusterStatusResponse,
-  });
-  expect(nodes).toHaveTextContent(nodesRunning);
+  const nodes = getByText('Nodes').nextSibling.textContent;
+  expect(nodes).toBe('0');
+});
+
+/******************** PENDING TESTS ********************/
+
+it.skip('renders all the v4 KVM nodes', async () => {
```
oponder: I guess there are some changes in the tests still to do?

src/components/Layout.js
```diff
@@ -54,7 +54,7 @@ class Layout extends React.Component {
   render() {
     return (
       <DocumentTitle>
-        <LoadingOverlay loading={!this.props.firstLoadComplete}>
+        <LoadingOverlay loading={this.props.loadingClustersList !== false}>
```
oponder: Can this be `this.props.loadingClustersList` instead of `this.props.loadingClustersList !== false` ?

oponder:
```suggestion
        <LoadingOverlay loading={this.props.loadingClustersList}>
```

src/components/Organizations/Organizations.js
```diff
@@ -11,11 +12,13 @@ const Organizations = props => {
 
   return (
     <Breadcrumb data={{ title: 'ORGANIZATIONS', pathname: url }}>
-      <Switch>
-        <Route component={List} exact path={`${path}`} />
-        <Route component={Detail} path={`${path}/:orgId`} />
-        <Redirect path={`${path}*`} to={`${url}`} />
-      </Switch>
+      <LoadingOverlay loading={props.loadingClustersDetails !== false}>
```
oponder: Same suggestion. Unless I am missing something or having a boolean logic brainfart.

```suggestion
      <LoadingOverlay loading={props.loadingClustersDetails}>
```

oponder:
Thanks for the effort and hard work here so far!

__tests__/V4AWSClusterManagement.js
ograu: Same question I am asking to myself... Fixed for all v4 tests. Thanks!

__tests__/V4KVMClusterManagement.js
ograu: Test removed

src/components/Layout.js
ograu: Changed: the only downside is that now flags have to be initialized to avoid undefined flags.

oponder: Oh I'm ok if this is done with a reason. It just read a bit weird to me.

PR #930, opened 2020-01-06
Fix cluster creation payload structure for master node availability zone

marians:
Fix https://github.com/giantswarm/giantswarm/issues/8205

The attribute name submitted in the POST request for cluster creation is fixed to `availability_zone`. Before, the name `availabilityZone` was used, which caused the API to ignore the attribute as unknown.

### Business logic change

In addition, the `master` attribute is only set when the user has manually selected an availability zone to use. Otherwise it is not contained in the payload, to leave defaulting to the API.

### DIFF
--- a/src/components/Cluster/NewCluster/CreateNodePoolsCluster.js
+++ b/src/components/Cluster/NewCluster/CreateNodePoolsCluster.js
@@ -280,20 +280,22 @@ class CreateNodePoolsCluster extends Component {
       np => np.data
     );
 
+    let createPayload = {
+      owner: this.props.selectedOrganization,
+      name: this.state.name.value,
+      release_version: this.props.selectedRelease,
+      master: {},
+    };
+
+    if (this.state.hasAZLabels) {
+      // TODO: don't use array here as long as there can be only one master node.
+      createPayload.master = this.state.availabilityZonesLabels.zonesArray[0];
+    }
+
     try {
       const newCluster = await this.props.dispatch(
-        // TODO: Remove random AZ generation here, as this is done on the API side.
         clusterCreate(
-          {
-            owner: this.props.selectedOrganization,
-            name: this.state.name.value,
-            release_version: this.props.selectedRelease,
-            master: {
-              availability_zone: this.state.hasAZLabels
-                ? this.state.availabilityZonesLabels.zonesArray
-                : this.state.availabilityZonesRandom.value,
-            },
-          },
+          createPayload,
           true // is v5
         )
       );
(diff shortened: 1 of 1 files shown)


src/components/Cluster/NewCluster/CreateNodePoolsCluster.js
```diff
@@ -280,19 +280,23 @@ class CreateNodePoolsCluster extends Component {
...
     );
 
+    let createPayload = {
+      owner: this.props.selectedOrganization,
+      name: this.state.name.value,
+      release_version: this.props.selectedRelease,
+    };
+
+    if (this.state.hasAZLabels) {
+      // TODO: don't use array here as long as there can be only one master node.
```
marians: I leave this TODO comment here instead of solving it, as I don't want to make the fix bigger than it is.

src/components/Cluster/NewCluster/CreateNodePoolsCluster.js
```diff
@@ -280,19 +280,23 @@ class CreateNodePoolsCluster extends Component {
       np => np.data
     );
 
+    let createPayload = {
```
ubergesundheit: The variable declaration above is using `const`, so why not use `const` here? I've seen line 291 mutates the object but I think `const` would be the right variable declaration statement.

marians: If that's what you guys do in JS, I'll do it too ;-)

marians: After reading a but, I see that "`const` is a signal that the identifier won’t be reassigned.".

In our case, the identifier gets assigned to an object, and that object later gets modified. Is const the right thing then?

ograu: You can modify properties in a `const` object, but can't reassign the whole object.  

`const` is our default when creating variables.

marians: ACK

src/components/Cluster/NewCluster/CreateNodePoolsCluster.js
```diff
@@ -280,19 +280,23 @@ class CreateNodePoolsCluster extends Component {
...
       const newCluster = await this.props.dispatch(
         clusterCreate(
-          {
-            owner: this.props.selectedOrganization,
-            name: this.state.name.value,
-            release_version: this.props.selectedRelease,
-            master: {
-              availabilityZone: this.state.hasAZLabels
-                ? this.state.availabilityZonesLabels.zonesArray
-                : this.state.availabilityZonesRandom.value,
```
ograu: We should probably remove `this.state.availabilityZonesRandom` from everywhere as well if we are not using it.

marians: I would like to leave this for a future PR

PR #931, opened 2020-01-06
push-and-use-china-registry

calvix:
towards https://github.com/giantswarm/giantswarm/issues/7435

This should add a step that will push image to china registry on merge to master or on tag creation

I might need little bit help here. I am right that the make script is only building `dist` folder during the build process?


calvix:
@marians tested on axolotl and its working fine

PR #932, opened 2020-01-06
Fix broken images.

oponder:


oponder:
Upgraded url-loader to latest version, but it had a breaking change which affected us:

```
switch to ES modules by default (the option esModule is true by default)
```

PR #941, opened 2020-01-07
my yarn.lock after running yarn

ubergesundheit:
Seeing if this breaks ci


axbarsan:
This works for me, so I'm guessing the problem was from my side

PR #954, opened 2020-01-08
Fix app creation, wrong method after refactor

oponder:


oponder:
App catalog tests coming in soon so we can catch something like this in the future.

oponder:
fyi @axbarsan, happened here: https://github.com/giantswarm/happa/pull/913/files#diff-88613a75860525c161896b51b7dc845bR99

I didn't notice it when approving :P , but upcoming tests will catch it.

oponder:
Gonna go ahead and merge with a ping @giantswarm/sig-front-end

oponder:
Oh no I cant :P

axbarsan:
I apologize for not catching this earlier 😢

oponder:
> I apologize for not catching this earlier 😢

Nah I apologise for not having tests in that area!

PR #957, opened 2020-01-08
Improve import sorting

axbarsan:
Fixes https://github.com/giantswarm/giantswarm/issues/8293

- Improves import sorting, adds support for grouping imports
- Adds support for auto-fixing wrong import order
- Fixes existing import order problems


oponder:
Order was slightly different before and written down here: https://github.com/giantswarm/fmt/blob/master/javascript/imports.md

I personally do not mind what the order is. If it is handling it automatically like this, then I am totally ok with that.

Does it still allow us to have it 'wrong' temporarily while developing? So like, does the project still compile in dev when there is an error? It just fixes it on commit? That would be the ideal for me :D

axbarsan:
> Order was slightly different before and written down here: https://github.com/giantswarm/fmt/blob/master/javascript/imports.md
> 
> I personally do not mind what the order is. If it is handling it automatically like this, then I am totally ok with that.
> 
> Does it still allow us to have it 'wrong' temporarily while developing? So like, does the project still compile in dev when there is an error? It just fixes it on commit? That would be the ideal for me :D

FMT import sorting is no longer relevant, since it has been already changed while changing the ESLint configuration.

With this PR change, imports should be sorted similar to the initial FMT implementation, but with automatic grouping.

Linting is no longer a step of compilation, it is automatically executed pre-commit (with the auto-fixer enabled), and also executed by CI. So, imports should be auto-fixed before commiting 👍 

Also, since this is automatically checked and fixed by the linter, I propose removing it from FMT. What do you think about it?

oponder:
Yup, lets remove it from fmt. 👍

PR #963, opened 2020-01-09
More organizations tests

ubergesundheit:
This contains some changes:

- I changed the `API_ENDPOINT` to differ from the one used in local development, to prevent accidental usage of a locally running API
- Make the user response never expire in the user response mock
- Added Organization tests

### DIFF
--- a/__tests__/Organization.js
+++ b/__tests__/Organization.js
@@ -27,38 +27,21 @@ const BASE_ROUTE = '/organizations';
 
 // Responses to requests
 beforeEach(() => {
-  getMockCall('/v4/organizations/', orgsResponse);
   getMockCall('/v4/user/', userResponse);
   getMockCall('/v4/info/', AWSInfoResponse);
-  getPersistedMockCall(
-    `/v4/organizations/${ORGANIZATION}/`,
-    orgResponse
-  );
+  getPersistedMockCall(`/v4/organizations/${ORGANIZATION}/`, orgResponse);
   getMockCall('/v4/clusters/', v4ClustersResponse);
-  getMockCall(
-    `/v4/clusters/${V4_CLUSTER.id}/`,
-    v4AWSClusterResponse
-  );
+  getMockCall(`/v4/clusters/${V4_CLUSTER.id}/`, v4AWSClusterResponse);
   getMockCall(
     `/v4/clusters/${V4_CLUSTER.id}/status/`,
     v4AWSClusterStatusResponse
   );
-  getMockCall(
-    `/v4/clusters/${V4_CLUSTER.id}/apps/`,
-    appsResponse
-  );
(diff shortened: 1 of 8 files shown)


__tests__/Organization.js
```diff
@@ -1,150 +1,309 @@
...
-  requests.cluster = getPersistedMockCall(
-    `/v4/clusters/${V4_CLUSTER.id}/`,
-    v4AWSClusterResponse
-  );
-  requests.status = getPersistedMockCall(
+beforeEach(() => {
+  getMockCall('/v4/user/', userResponse);
+  getMockCall('/v4/info/', AWSInfoResponse);
+  getPersistedMockCall(`/v4/organizations/${ORGANIZATION}/`, orgResponse);
+  getMockCall('/v4/clusters/', v4ClustersResponse);
```
ograu: Is there any reason why you only persist some calls?

ubergesundheit: yes! I think this gives the tests more confidence that the requests are actually used only once.

ograu: Don't follow. Then, why persist some of them? Why don't you just use always `getMockCall`?

jest.config.js
```diff
@@ -22,7 +22,7 @@ module.exports = {
   globals: {
     // window.config object will now be available in all tests
     config: {
-      apiEndpoint: 'http://localhost:8000',
+      apiEndpoint: 'http://1.2.3.4',
```
oponder: 👍

__tests__/Organization.js
ubergesundheit: Yes good question! The ones where `getPersistedMockCall` is used, are called multiple times inside the tests, but since you've raised this, I'll change the tests to only use `getMockCall` :+1:

ograu: Thanks. This is what I would expect since you are using `beforeEach`.

oponder: There's `.times()` and some other sugary syntax for nock to specify how many times you expect the request to be made: https://github.com/nock/nock#repeat-response-n-times

ubergesundheit: yea. I was being lazy.. :) I've tried the `.times()` function with our `get...MockCall` helpers but it didn't seem to be working.

oponder: Yeah probably not possible since the .reply() is the last in the chain, and also we dont' return the nock thing anyways so chaining more onto it isn't possible even.

oponder: Oh wait we do return, but yeah we'd still have to somehow add something into the middle of that chain...

```
export const getMockCall = (endpoint, response = []) =>
  nock(API_ENDPOINT)
    .get(endpoint)
    .reply(200, response);
```

ubergesundheit:
I removed the use of `getPersistedMockCall` in the Organizations tests! If no one has any objectsions, I'll merge this later (or tomorrow)

ograu:
No objections!

Issue #759, opened 2019-09-19
Tests to add to Happa


oponder:
### Authentication
- [x] Login with password
- [x] Login with auth0
- [x] Logout
- [x] Password reset
- [x] Signup
- [x] Changing password
- [x] Changing email

### Cluster Management
- [x] Creating a cluster 
- [x] Deleting a cluster
- [x] Changing scaling settings of a cluster
- [x] Creating a keypair
- [x] Creates a v5 cluster and redirect to details view
- [x] Switch to v4 form, creates a v4 cluster and redirect to details view
- [x] The user is redirected to list when cluster doesn't exist
- [x] Cluster list shows all clusters, each one with its details and resources, for the selected organization. 

### V4 Cluster Management (AWS, KVM, Azure)
- [x] it deletes a cluster
- [x] it patches name
- [x] AWS, it renders cluster data
- [x] AWS, it scales with the scaling modal correctly
- [x] Azure render cluster data 
- [x] Azure scaling
- [x] KVM render cluster data
- [x] KVM scaling
- [x] Key pairs tab: renders data correctly
- [x] Apps tab: renders data correctly

###  V5 Cluster management
- [x] It renders cluster data without nodes ready
- [x] with nodes ready
- [x] It renders all node pools in store
- [x] It shows the menu when button is clicked
- [x] It scales a node pool correctly
- [x] It patches a np name
- [x] It sorts node pools
- [x] It deletes a V5 cluster
- [x] It deletes a node pool
- [x] It adds node pool
- [x] Key pairs tab: renders data correctly
- [x] Apps tab: renders data correctly

### Org management
- [x] Creating an org
- [x] Deleting an org
- [x] Adding a member to an org
- [x] Removing a member from an org
- [x] Viewing list of all orgs

### User management
- [x] Renders users correctly
- [x] Invites new users
- [x] Deletes existing users
- [x] Removes the expiration date from user invitations

### App catalog
- [x] App catalog viewing
- [x] App installation
- [x] Editing App

### UI Components
- [x] V4 Scale cluster modal

### Getting Started Guide
- [x] Navigating through pages
- [x] Get started button behaviour on dashboard and cluster detail page

ograu:
On SIG Front End meeting (5 Dec) we agreed on have all boxes checked before 3 months from now. This is 5 Mar 2020

oponder:
Closing this issue! Great work everyone! 81% test coverage right now, that's pretty great.

PR #964, opened 2020-01-09
Create tests for Account Settings

axbarsan:
Part of https://github.com/giantswarm/happa/issues/759

Means tests for:
- [x] Change Password Form
- [x] Change Email Form
- [x] Account Settings


__tests__/AccountSettings.js
```diff
@@ -0,0 +1,77 @@
...
+  it('renders the account settings component on the "/account-settings" route', async () => {
+    const { findByText } = renderRouteWithStore('/account-settings', {});
+
+    await findByText('Your Account Settings');
+  });
+
+  it('renders the email change form', async () => {
+    const { findByTestId } = renderRouteWithStore('/account-settings', {});
+
+    const emailChangeForm = await findByTestId(elementIDs.ChangeEmailForm);
```
ograu: I think `data-test-id` should be our last option. In case of form fields it makes more sense to use `findByDisplayValue` IMO

__tests__/AccountSettings.js
```diff
@@ -0,0 +1,77 @@
...
+
+    expect(emailChangeForm).not.toBeNull();
```
ograu: And then there is no need to make those extra `expect`s.

__tests__/AccountSettings.js
```diff
@@ -0,0 +1,77 @@
...
+  });
+
+  it('renders the password change form', async () => {
+    const { findByTestId } = renderRouteWithStore('/account-settings', {});
+
+    const passwordChangeForm = await findByTestId(
+      elementIDs.ChangePasswordForm
+    );
+
+    expect(passwordChangeForm).not.toBeNull();
```
ograu: Shouldn't we be looking for the right value to be rendered instead?

__tests__/AccountSettings.js
axbarsan: There is no need for the `expect`s anyway, because `findByTestId` throws an error if the element is not found. I added them because it's easier for us to go through and understand the tests

__tests__/AccountSettings.js
axbarsan: I'm not sure what you mean by that
Actual tests for the `PasswordChangeForm` are in a separate file, this just checks that it's there

__tests__/AccountSettings.js
axbarsan: This actually checks if the wrapper exists, actual form tests are separate for both `ChangeEmailForm` and `ChangePasswordForm`

__tests__/AccountSettings.js
ograu: I find it confusing and not adding a lot of value to a test, especially the `not.toBeNull()` part.

To be clear, my two points are: I think there are too many`not.toBeNull`s and `findById`s in these tests.

__tests__/AccountSettings.js
ograu: Yes, I picked a bad example here, I am ok with using this for wrappers.

__tests__/AccountSettings.js
axbarsan: I agree that `not.toBeNull()` is not adding much value, other than readibility. But the `findById` calls are checking if the elements are rendered, so they definately have a value.

__tests__/AccountSettings.js
```diff
@@ -0,0 +1,77 @@
...
+    expect(emailChangeForm).not.toBeNull();
+  });
+
+  it('renders the email change form label and description', async () => {
+    const { findByTestId } = renderRouteWithStore('/account-settings', {});
+
+    const emailLabel = await findByTestId(elementIDs.ChangeEmailForm);
+    expect(emailLabel).not.toBeNull();
+
+    const emailDescription = await findByTestId(elementIDs.ChangeEmailForm);
```
axbarsan: Just realized that these are the wrong IDs

axbarsan:
> Would be good to use getbylabel where possible indeed.
> 
> Also I'm not to sure about these type of things at the start of tests:
> 
> ```
> const elementIDs = {
>   input: 'change-email/input',
>   button: 'change-email/button',
> };
> ```
> 
> ```
> const statusMessages = {
>   Success: 'Saved Succesfully',
>   EmailAlreadyInUse:
>     'This e-mail is in already in use by a different user. Please choose a different e-mail address',
>   ServerError:
>     'Something went wrong while trying to update your e-mail address.',
> };
> ```
> 
> I appreciate the desire to factor code nicely and have strings in a starting block, but does this really make the 'story' of the code more readable? To me it's just a layer of indirection. Is there some benefit I am missing?

Using `getByLabelText` didn't make sense to me before, but now, after reading this https://testing-library.com/docs/guide-which-query, it finally does. So we're kind of testing around what the user should see on the screen, kind of like in a more UI testing way, rather than an unit test. I will update my stuff to use it.

I think we should include this in our FMT.

As with having IDs and messages separate, I believe they really contribute to reducing the clutter in tests, making them easier to read, and also, for easy editing in the future, since you know you only have to change one string, and the tests will check for the correct thing.

ograu:
> So we're kind of testing around what the user should see on the screen, kind of like in a more UI testing way, rather than an unit test. I will update my stuff to use it.
> 
> I think we should include this in our FMT.

I think we already have it in `giantswarm/fmt`: 

- https://github.com/giantswarm/fmt/blob/master/javascript/testing.md#statements
- https://github.com/giantswarm/fmt/blob/master/javascript/testing.md#avoid-implementation-details

Please feel free to add anything if you think this is not clear enough.

__tests__/AccountSettings.js
```diff
@@ -0,0 +1,67 @@
...
+    requests.orgs = getPersistedMockCall('/v4/organizations/');
+    requests.clusters = getPersistedMockCall('/v4/clusters/');
+    requests.catalogs = getPersistedMockCall('/v4/appcatalogs/');
+  });
+
+  afterAll(() => {
+    for (const req of Object.values(requests)) {
+      req.persist(false);
+    }
+  });
```
ograu: We should probably get rid of this and use the pattern @ubergesundheit  used for organizations from now on.

PR #966, opened 2020-01-10
Update react-datepicker and other deps.

oponder:


oponder:
Found a break due to react-datepicker, so instead of just fixing it, I also added a test that failed first.

I think that's a good habit to start adding tests that guard for a regression or bug before fixing the bug.

PR #967, opened 2020-01-10
Always use latest v12 nodejs in circle ci

ubergesundheit:
It should be safe to use automatically use the latest node 12 LTS in CI runs.


ubergesundheit:
I like your suggestion of failing CI on outdated node versions! I'll see how I could do this

PR #982, opened 2020-01-13
Fetch cluster details only for selected organization clusters

ograu:
Currently, in clusters list view, all clusters details owned by all organizations are fetched. This means that  we fetch data that the user doesn't need in this view.

This PR implements cluster filtering by selected organization, so Happa will fetch details only for clusters that are rendered. Cluster details will be fetched when needed: in cluster details view and in organizations list.


ograu:
Can't reproduce this issue...

ograu:
I think this is solved:

- I've added cluster details and node pools load on cluster details view. I though that we could avoid this calls here because we are doing them in parent `Layout` but it turned out that it isn't possible
- I've added a `dispatch` in organization detail view that sets cluster details loading flag to true, otherwise rendering of the details start with the loading flag for cluster details to `false`

PR #985, opened 2020-01-13
Fixed keypairs loading and tests

ograu:
Fix key pair loading in v5 clusters and fix `nock` warning in tests for `apps` endpoint not being called.


ograu:
great, thanks! :-)

PR #993, opened 2020-01-14
Kubesystem for ingress app namespace

oponder:
This ensures that the  `nginx-ingress-controller-app` can only be installed to the `kube-system` namespace.

### DIFF
--- a/src/components/AppCatalog/AppDetail/__tests__/InstallAppForm.js
+++ b/src/components/AppCatalog/AppDetail/__tests__/InstallAppForm.js
@@ -1,37 +1,24 @@
 import '@testing-library/jest-dom/extend-expect';
 
-import { render } from '@testing-library/react';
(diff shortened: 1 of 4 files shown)


oponder:
namespace

src/components/AppCatalog/AppDetail/__tests__/InstallAppForm.js
```diff
@@ -0,0 +1,38 @@
...
+
+import { render } from '@testing-library/react';
+import { ThemeProvider } from 'emotion-theming';
+import React from 'react';
+import theme from 'styles/theme';
+
+import InstallAppForm from '../InstallAppForm';
+
+it('renders without crashing', () => {
+  render(
```
axbarsan: You can use the `renderWithTheme` utility in `renderUtils` for this 👍

PR #994, opened 2020-01-14
add organization member removal test

ubergesundheit:
This ticks the last box (Removing a member from an org) in #759 under Organizations

### DIFF
--- a/__tests__/Organization.js
+++ b/__tests__/Organization.js
@@ -168,10 +168,11 @@ describe('Organizations basic', () => {
   });
 
   it('shows organization details correctly', async () => {
+    getMockCallTimes(`/v4/organizations/${ORGANIZATION}/`, orgResponse, 1);
     const {
       findByText,
       getByText,
-      getByTestId,
+      findByTestId,
       queryByTestId,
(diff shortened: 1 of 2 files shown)


__tests__/Organization.js
```diff
@@ -249,6 +249,65 @@ describe('Organizations basic', () => {
...
+    );
+
+    // users
+    const usersTable = await findByTestId('org-detail-users-wrapper');
+    expect(usersTable.querySelector('tbody > tr > td').textContent).toBe(
+      orgResponse.members[0].email
+    );
+
+    const removeUserTableButton = usersTable.querySelector(
+      'tbody > tr > td:nth-of-type(3) > button'
```
axbarsan: Why not use a test id, as the structure might change in the future?

ograu: +1

ubergesundheit: The tables on the organization detail component uses an external component. I'll look into adding test-ids

ubergesundheit: removed querySelector from this and the organization detail view test

PR #998, opened 2020-01-14
Add cluster scaling tests

axbarsan:
Part of https://github.com/giantswarm/happa/issues/759

- [x] Azure cluster scaling
- [x] KVM cluster scaling
- [x] V4 cluster scaling modal


src/components/Cluster/ClusterDetail/__tests__/ScaleClusterModal.js
```diff
@@ -0,0 +1,620 @@
...
+import { Providers } from 'shared';
+import { StatusCodes } from 'shared/constants';
+import {
+  API_ENDPOINT,
+  AWSInfoResponse,
+  v4AWSClusterResponse,
+  v4AzureClusterResponse,
+  v4KVMClusterResponse,
+} from 'testUtils/mockHttpCalls';
+import { getComponentWithStore } from 'testUtils/renderUtils';
```
oponder: I think this is something I wanted to avoid for 'component' tests. I'd like to eventually have a clear separation between components that need the store and wrappers that wire the store up to the components. For now, to get this tested , I suppose its ok. It even serves as a signal. But it's something to think about.

axbarsan: I agree with that 100%. It would make testing so easy 👍

__tests__/V4KVMClusterManagement.js
```diff
@@ -1,74 +1,66 @@
...
-    `/v4/organizations/${ORGANIZATION}/`,
-    orgResponse
-  );
-  requests.clusters = getPersistedMockCall('/v4/clusters/', v4ClustersResponse);
-  requests.cluster = getPersistedMockCall(
-    `/v4/clusters/${V4_CLUSTER.id}/`,
-    v4KVMClusterResponse
-  );
-  requests.status = getPersistedMockCall(
+  nock.disableNetConnect();
```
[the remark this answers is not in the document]
axbarsan: Praise @ubergesundheit for that 😁

axbarsan:
I agree that testing connected components is really painful. But I believe we need to get the current functionalities tested as they are, as it will make refactoring much easier later

ograu:
> I agree that testing connected components is really painful. But I believe we need to get the current functionalities tested as they are, as it will make refactoring much easier later

I agree and also I think that the more we focus on unit testing, the more painful refactors/additions of new features are going to be.

PR #1006, opened 2020-01-15
Validate organization creation input

ubergesundheit:
Towards giantswarm/giantswarm#8403

Validates the input field for creating organizations using `minlength`, `maxlength` and `pattern` fields and only allows to click the button if validation succeeds

### DIFF
--- a/src/components/Modals/Modals.js
+++ b/src/components/Modals/Modals.js
@@ -20,6 +20,7 @@ import ClusterIDLabel from 'UI/ClusterIDLabel';
 class Modals extends React.Component {
   state = {
     emailValid: false,
+    organizationNameValid: false,
   };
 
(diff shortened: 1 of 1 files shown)


axbarsan:
This is cool and does what it should do, however we don't currently rely on HTML5 form validation. The main problem is that every browser has a different implementation, and the warning messages have a different appearance on each one of them.

Currently we use the `InputField` component (you can find it in `src/components/shared/InputField.js`), that accepts custom validation errors, and which will keep a consistent UI experience across our forms. This implies that you implement the validation manually (length and validation pattern), but that functionality is so common that I believe it should be included in the actual `InputField` component at some point

ubergesundheit:
I'll look into using `InputField`

PR #1007, opened 2020-01-15
Fix pre-commit formatting not running on staged files

axbarsan:
- Remove the `pretty-quick` dependency
- Use `lint-staged` to only run tasks on staged files before commiting

Now both linting and formatting are running just on staged files, which increases development speed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,9 +16,9 @@
     "test:clean-cache": "yarn test --clearCache",
     "test:coverage": "yarn test --collect-coverage",
     "format": "./node_modules/.bin/prettier \"src/**/*.js\"",
-    "format:staged": "pretty-quick --staged --pattern ./src/**/*.js",
     "prettier": "yarn format --write",
-    "validate-prettier": "yarn format --version; yarn format --no-color --check"
+    "validate-prettier": "yarn format --version; yarn format --no-color --check",
+    "precommit": "./node_modules/.bin/lint-staged"
   },
   "prettier": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -141,7 +141,14 @@
...
     "hooks": {
-      "pre-commit": "yarn lint:fix && yarn format:staged"
+      "pre-commit": "yarn precommit"
     }
+  },
+  "lint-staged": {
+    "*.js": [
+      "yarn lint:fix",
+      "yarn prettier",
+      "git add"
```
ubergesundheit: Does this automatically run `git add`?

ubergesundheit:
Could you also upgrade husky in this PR to [4.0.10](https://www.npmjs.com/package/husky/v/4.0.10)?

package.json
axbarsan: Yes, on the changed files (which are the staged files, anyway), so nothing breaks

ubergesundheit: `changed` or `staged` files? I very much only use `git add -p` for adding only the changes I really want into the staging area. As I understand it, unstaged changes would be added to the staging area with this change?

axbarsan: I apologize for the confusion. I actually meant that it runs on staged files, and after making fixes, it adds the files again, to be able to commit the new changes.

If ran without `git add` as a final task, the changes from the tasks would not be included in the commit.

ubergesundheit: yeah nice :+1:

PR #1009, opened 2020-01-16
Minor refactor cluster status call

ograu:
Several things going on here as a previous step for moving resource computations to `clusterActions`:

- Removed old `clusterLoadStatus()` because it was useless as we just call status endpoint for v4 clusters
- Added `NOT_FOUND` suffix to `loadingReducer.js` that will set loading flag to false (I owe you documentation explaining how this works, coming this week)
- And inlining action objects: I find that this is best for readibility, specially in big files. WDYT? Are you ok with this?


ubergesundheit:
I am pro inlining action objects

PR #1019, opened 2020-01-17
Change cancel button in cluster form to use danger style

ograu:
Before:

before

After:

after

Weird enough I saw this in `gauss` but this cancel button is not present in 0.10.8 release.


marians:
Sorry, didn't see this while it was in review.

Why the "Danger" style? We use that color do mark a desctructive process.

In my opionion, losing a bit of form input that can be reproduced in abuot a minute is not a destructive process.

ograu:
Just saw it was red and changed to the red color we were using in other buttons.

marians:
That's the wrong fix in my opinion. If `bsStyle='default'` turns the button red, we need to fix that instead.

PR #1021, opened 2020-01-17
Refactor computeCapabilities and add hasOptionalIngress

oponder:
Based on what we discussed yesterday, also added some tests. Perhaps over tested.

### DIFF
--- a/__tests__/ClusterManagement.js
+++ b/__tests__/ClusterManagement.js
@@ -1,28 +1,28 @@
 import '@testing-library/jest-dom/extend-expect';
+
+import { fireEvent, wait } from '@testing-library/react';
+import nock from 'nock';
 import {
   API_ENDPOINT,
   appCatalogsResponse,
-  getPersistedMockCall,
+  appsResponse,
   AWSInfoResponse,
+  getPersistedMockCall,
+  nodePoolsResponse,
   ORGANIZATION,
   orgResponse,
   orgsResponse,
   releasesResponse,
   userResponse,
-  v5ClustersResponse,
-  v4AWSClusterStatusResponse,
-  appsResponse,
   V4_CLUSTER,
-  V5_CLUSTER,
-  v4ClustersResponse,
   v4AWSClusterResponse,
+  v4AWSClusterStatusResponse,
+  v4ClustersResponse,
+  V5_CLUSTER,
   v5ClusterResponse,
-  nodePoolsResponse,
+  v5ClustersResponse,
 } from 'testUtils/mockHttpCalls';
-import { fireEvent, wait } from '@testing-library/react';
 import { renderRouteWithStore } from 'testUtils/renderUtils';
-import nock from 'nock';
-import { within } from '@testing-library/dom';
 
 // Tests setup
 const requests = {};
@@ -155,7 +155,7 @@ details view`, async () => {
     v4AWSClusterResponse
   );
 
-  const { getByText, getByTestId, getAllByText } = renderRouteWithStore(
+  const { findByText, findByTestId, getAllByText } = renderRouteWithStore(
     '/organizations/acme/clusters/new/'
   );
 
@@ -173,10 +173,10 @@ details view`, async () => {
     `/v4/clusters/${V4_CLUSTER.id}/key-pairs/`
   );
 
-  await wait(() => getByText('Details and Alternatives'));
-  fireEvent.click(getByText('Details and Alternatives'));
+  fireEvent.click(await findByText('Details and Alternatives'));
 
-  await wait(() => document.querySelector('.modal-content'));
+  // Wait for the modal to pop up.
+  await findByText('Release Details');
 
   // Find the second button which is the 8.5.0
   // TODO Improve this, check with cmp that this is not a node pools release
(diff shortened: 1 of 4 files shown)


__tests__/ClusterManagement.js
```diff
@@ -187,13 +187,13 @@ details view`, async () => {
...
 
-  // Are we in the v4 cluster creation form
-  await wait(() => expect(getByTestId('cluster-creation-view')));
-  fireEvent.click(getByText('Create Cluster'));
-  await wait(() => getByTestId('cluster-details-view'));
+  // Click the create cluster button.
+  fireEvent.click(await findByText('Create Cluster'));
+
+  // Wait till we're on the cluster detail page.
+  await findByTestId('cluster-details-view');
```
ograu: nice simplification, more readable ❤️

src/actions/clusterActions.js
```diff
@@ -417,7 +397,10 @@ export function clusterLoadDetails(clusterId, { withLoadingFlags }) {
       if (isV5Cluster) cluster.nodePools = [];
 
       const provider = getState().app.info.general.provider;
-      cluster.capabilities = computeCapabilities(cluster, provider);
+      cluster.capabilities = computeCapabilities(
+        cluster.release_version,
+        provider
+      );
```
ograu: Why are we computing this in two places? Wouldn't be enough to do it just in one place?

ograu: We are comptaiong capabilities in `clustersList`... ah, ok, this for direct access to the cluster

ograu: Maybe we can do it in _only_ in `clusterLoadDetails`?

oponder: I think in clusterList then only should be enough?

But maybe after upgrading a cluster, we dont call cluster list, but we do call cluster detail, so that keeps it fresh? 

In any case, I'm fairly certain we need it in the cluster listing.

src/utils/clusterUtils.js
```diff
@@ -133,3 +136,31 @@ export const clusterNodePools = (nodePools, cluster) => {
...
+export function computeCapabilities(releaseVersion, provider) {
+  const capabilities = {
+    canInstallApps: false,
+    hasOptionalIngress: false,
+  };
+
+  // Installing Apps
+  // Must be AWS or KVM and larger than 8.1.0
+  // or any provider and larger than 8.2.0
+  if (
```
ubergesundheit: I was also checking this (and other places where release versions were compared) out and thought why not drop this stuff if no one uses old versions any more (See https://github.com/giantswarm/giantswarm/blob/master/ops-recipes/list-tenant-clusters.md). The oldest tenant cluster version I found was `8.2.0`

src/actions/clusterActions.js
oponder: The app catalog for example doesn't need cluster details, but it does need a full list of clusters to see which can install apps on it.

So right now it does do all the cluster details calls eventually (and possibly filtered by organisation), but it doesn't wait for all the details to load in order to show the app detail page, which I think is right.

If we want to have it in clusterLoadDetails only, then we'd have to load all clusters and their details, instead of just the ones that belong to the selected organisation, when showing the app install modal.

ograu: didn't know that... why do we need to call it on cluster listing?

oponder: So I think if we're going to have to have it in only one place, the best place is `clustersList`, because that always happens before everything. 

The only tricky part is when creating a new cluster. Then I think we do not do a clustersList call to refresh the state, instead we do only the clusterLoadDetails for the new cluster and add the new cluster to the state manually right?

src/utils/clusterUtils.js
oponder: Yeah if the version doesn't exist anymore, we can totally drop the logic that supports it.

src/actions/clusterActions.js
ograu: Yes, but there is also the refreshing thing. If this is something that can be updated/modified, then I think that it makes more sense to have it in `clusterLoadDetails`.

oponder: > why do we need to call it on cluster listing?

The install app modal shows a list of all clusters in all organisations, but it filters that list on `capabilities.canInstallApp`, because we don't want to show clusters in this list that can't install apps.

Screenshot 2020-01-17 at 2 37 58 PM

If we do it only on clusterLoadDetail, that only happens for clusters that are in the currently selected organisation.

Otherwise we'd have to load the detail for every cluster always again.

What I'm saying is, I think we can remove it from clusterLoadDetail if we are doing it now in cluster listing. But we need to check closely the case where we are creating a new cluster.

src/utils/clusterUtils.js
ubergesundheit: alright. I am investigating cluster-service not returning enough releases im my development setup which does not let me create clusters through happa, which also uses one of these comparisons for a very old version (6). Ill see later if you have already removed these here..

src/actions/clusterActions.js
oponder: > Yes, but there is also the refreshing thing. If this is something that can be updated/modified, then I think that it makes more sense to have it in clusterLoadDetails.

Then I think we _might_ need it in both places. 

Unless we think of a different place entirely to put this computeCapabilities. Like maybe a clusterComputeCapabilities action, which computes these values for all clusters in the current entities list, that we can then add to right place in the batched action calls.

ograu: I'm ok with leaving this in both places then. DO you think it's useful to add a comment on why we are doing it in `clustersList`? Cause in my mind this belongs to cluster details action. WDYT?

oponder: I'll add a comment 👍

PR #1028, opened 2020-01-20
Remove deprecated version checks

ubergesundheit:
This pull request removes tennant cluster release versions from some places in happa.

I checked tennant cluster versions with lowest to be `8.0.0` (I've done this with [these](https://github.com/giantswarm/giantswarm/blob/master/ops-recipes/list-tenant-clusters.md) instructions)

I am sure you'll find this change to be very controversial in regard to removing these kind of guards. In my opinion removing these checks is a good thing because it removes code that isn't executed anyway when no one is using these old cluster versions.

### DIFF
--- a/src/components/Cluster/NewCluster/CreateRegularCluster.js
+++ b/src/components/Cluster/NewCluster/CreateRegularCluster.js
@@ -68,7 +68,7 @@ const FlexWrapperAZDiv = styled.div`
 
 class CreateRegularCluster extends React.Component {
   static isScalingAutomatic(provider) {
-    return provider !== Providers.AWS;
+    return provider === Providers.AWS;
   }
 
   state = {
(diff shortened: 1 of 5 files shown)


src/components/Cluster/NewCluster/CreateRegularCluster.js
```diff
@@ -68,14 +67,8 @@ const FlexWrapperAZDiv = styled.div`
...
-  static isScalingAutomatic(provider, releaseVer) {
-    if (provider !== Providers.AWS) {
-      return false;
-    }
-
-    // In order to have support for automatic scaling and therefore for scaling
-    // limits, provider must be AWS and cluster release >= 6.3.0.
-    return cmp(releaseVer, '6.2.99') === 1;
+  static isScalingAutomatic(provider) {
+    return provider !== Providers.AWS;
```
axbarsan: This should be the other way around. Autoscaling is only available on AWS at the moment, and this check is disabling it for AWS clusters

axbarsan:
This is awesome 👍

src/components/Cluster/NewCluster/CreateRegularCluster.js
ubergesundheit: good catch! :+1: thanks

oponder:
By the way I wasn't able to check versions on 3 installations specifically, but I trust you!

PR #1031, opened 2020-01-21
Move all URLs to an enum-like data structure and provide helpers for easily manipulating them

axbarsan:
Fixes https://github.com/giantswarm/giantswarm/issues/7803

- Move all internal routes into a single place, for easier editing
- Create an utility class for easily managing paths and their parameters
- Clean up stuff in a couple of tests to make the linter happy

Tests pass fine, also tested manually and everything's looking good


src/lib/routePath.js
```diff
@@ -0,0 +1,257 @@
...
+   */
+  static create(template, parameters = {}) {
+    const newRoute = new RoutePath(template);
+    newRoute.params = parameters;
+
+    return newRoute;
+  }
+
+  /**
+   * Generate an usable string path with parameters
```
oponder:
```suggestion
   * Generate a usable string path with parameters
```

PR #1033, opened 2020-01-22
Computing resources with reselect

ograu:
In this PR I am moving the code that computes resources to memoized selectors so it is only recomputed when there are changes in the state that this derived data is computed from.

A side (and intended) benefit is that components are cleaner and this logic is in one place instead of being spread and repeated across the codebase.

I am using [Reselect](https://github.com/reduxjs/reselect) for this.

There is a flash in dashboard items computed info that I have already solved in an upcoming PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,6 +80,7 @@
     "react-window": "1.8.5",
     "redux": "4.0.5",
     "redux-thunk": "2.3.0",
+    "reselect": "^4.0.0",
     "semver-compare": "1.0.0",
     "serialize-javascript": "2.1.2",
     "superagent": "5.2.1",
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
     "react-window": "1.8.5",
     "redux": "4.0.5",
     "redux-thunk": "2.3.0",
-    "reselect": "^4.0.0",
+    "reselect": "4.0.0",
     "semver-compare": "1.0.0",
     "serialize-javascript": "2.1.2",
     "superagent": "5.2.1",
--- a/src/actions/clusterActions.js
+++ b/src/actions/clusterActions.js
@@ -15,6 +15,9 @@ const clustersApi = new GiantSwarm.ClustersApi();
 function clustersLoadArrayToObject(clusters, provider) {
   return clusters
     .map(cluster => {
+      // Remove cluster's create_date because we are loading it in cluster details call
+      delete cluster.create_date;
+
       return {
         ...cluster,
         lastUpdated: Date.now(),
--- a/src/components/Cluster/ClusterDetail/ClusterDetailView.js
+++ b/src/components/Cluster/ClusterDetail/ClusterDetailView.js
@@ -274,7 +274,6 @@ class ClusterDetailView extends React.Component {
                           showScalingModal={this.showScalingModal}
                           showUpgradeModal={this.showUpgradeModal}
                           workerNodesDesired={this.getDesiredNumberOfNodes()}
-                          workerNodesRunning={getNumberOfNodes(cluster)}
                         />
                       )}
 
--- a/src/components/Cluster/ClusterDetail/NodesRunning.js
+++ b/src/components/Cluster/ClusterDetail/NodesRunning.js
@@ -9,24 +9,18 @@ const NodesRunning = ({ workerNodesRunning, RAM, CPUs, nodePools }) => {
 
   return (
     <div data-testid='nodes-running'>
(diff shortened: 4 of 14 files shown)


ograu:
I think that this approach is better than adding resources to the store as I mentioned some days ago. Basically what we get this way is that we keep store state minimal and we can memoize calculations.

src/actions/clusterActions.js
```diff
@@ -15,6 +15,9 @@ const clustersApi = new GiantSwarm.ClustersApi();
 function clustersLoadArrayToObject(clusters, provider) {
   return clusters
     .map(cluster => {
+      // Remove cluster's create_date because we are loading it in cluster details call
+      delete cluster.create_date;
```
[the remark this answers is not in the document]
ograu: This is because the `create_date` field in the clusters list and cluster details APIs response have different time format, one has seconds rounded and the other has nanosedconds I believe. 

As they are different strings, we update the value each time `clustersList()` and `clusterLoadDetails()` run. Then the selector can't memoize because it is receving different inputs every time this calls are performed, so it recomputes after every single call.

We could format the two strings to be exactly the same value, or modify responses in the API repo to use the same format, but as we don't need this to be present during initialization I just removed the key and let the cluster detail action to populate it in store.

ograu:
> **1. I was wondering why they are called selectors, but I guess that's just what it is.**
> 
> This helped me understand a bit.
> https://medium.com/@matthew.holman/what-is-a-redux-selector-a517acee1fe8

As you might know they are called selectors as they select bits of data. They could have been called getters, but I suppose getters is used in so many contexts and somebody thought that it would be cool to use another name.

ograu:
> **3. A note on function naming:**
> 
> Should we name selectors after the field they populate?
> 
> For example, since the selector is populating the `resources` field, should we call the function `selectResources` instead of `selectComputedResources` ?

Totally ok with that. 

> And also is there a way we can avoid the V4 / V5 prefix and leave that as a concern for the selector?
> If that gets too weird in the selector, then I am fine with `selectResourcesV4` and `selectResourcesV5`

I thought that this was cleaner, but I agree with you in that this would make it everything easier in the component side, which is something that I like. We totally can make it this way if you like it more. 

> **4. Performance nitpick**
> 
> > Selectors created with createSelector have a cache size of 1. This means they always recalculate when the value of an input-selector changes, as a selector only stores the preceding value of each input-selector.
> 
> I'm not sure we're getting the performance benefits of memoization because we're still having to recompute each time because it's always based on the clusterID at the start right.
> 
> So if we use computed resources multiple times on the same page, yes we avoid doing the calculation again.
> 
> But if we're switching from one cluster to another cluster, it'll recompute again.
> 
> Still an improvement! Just making sure my mental model is right.

Yes, that's true, this is something I've done having clusters list and `ClusterDashboardItem` in mind, every time you go to a cluster details view, this is recomputed. 

Keep in mind that in clusters list is where memoization is more interesting because there can be n clusters and hence n recomputing of resources on every refresh. But yeah, we can iterate on this and memoize values for all ids. I don't know if this is worth it but we can do it if we want.

ograu:
Sorry, but I'm not totally sure about this memoization thing we are discussing, I will experiment this afternoon with this.

package.json
```diff
@@ -80,6 +80,7 @@
     "react-window": "1.8.5",
     "redux": "4.0.5",
     "redux-thunk": "2.3.0",
+    "reselect": "^4.0.0",
```
axbarsan: Don't forget to pin this version, as we do with the others 👍

axbarsan:
How do you feel about grouping selectors in separate files, as we do with reducers? So we'd have files like `clusterSelectors.js` inside of `/src/selectors/`?

package.json
ograu: Upps, thanks!

ograu:
> How do you feel about grouping selectors in separate files, as we do with reducers? So we'd have files like `clusterSelectors.js` inside of `/src/selectors/`?

Actually I like this more than use a `index.js` 👍

ograu:
- Changed the names of functions and renamed the selector file as well.
- Didn't put everything inside a `selectResources` function because the two functions are different and it doesn't make sense, it makes code more difficult IMO. I think this way it is simpler. Anyway, I'm happy if someone wants to refactor it.
- Changed output selectors using the factory function approach suggested in the docs, so now they produce a brand new function for every component using them. BUT didn't manage to make memoization work yet, will iterate on this after this PR is merged.

Issue #1034, opened 2020-01-23
Data flow: Merge onto existing state or split up store (cluster details and cluster list?)


oponder:
Starting a async discussion here instead of in slack :)

I've noticed that the `CLUSTER_LIST_SUCCESS` reducer replaces the `entities.cluster.items` state with what it got from the API. 

That means that if we had any cluster details already loaded, they will be removed.

<img width="381" alt="Screenshot 2020-01-23 at 8 22 56 AM" src="https://user-images.githubusercontent.com/455309/72965700-4d1bc400-3dbd-11ea-89bf-2675101d8756.png">

As a result, it is basically not safe to do a `clustersList` _without_ following it with `clustersDetails` to repopulate the removed fields (like `api_endpoint`)

Currently when we delete a cluster, we do a `clustersList` to just refresh the list of clusters, but the side effect is that we remove all details from other clusters too. 

This results in the `TypeError: Failed to construct 'URL': Invalid URL` on the KeyPairs component.

Currently in the reducer we have:

```
    case types.CLUSTERS_LIST_SUCCESS:
      draft.items = action.clusters;
      draft.v5Clusters = action.v5ClusterIds;

      return;
```

I would change it for something like:

```
    case types.CLUSTERS_LIST_SUCCESS:
      let clusters = {};
      Object.keys(action.clusters).forEach(clusterID => {
        clusters[clusterID] = Object.assign({}, draft.items[clusterID], action.clusters[clusterID])
      })

      draft.items = clusters;
      draft.v5Clusters = action.v5ClusterIds;
```

But does that bring back a pattern we were trying to get rid of?

An alternative to 'augmenting' existing state objects, would be to store the cluster list, and the individual cluster details as separate things. I think that wouldn't be an easy change to make, but would it make things easier to reason about?

ping @ograu @giantswarm/sig-front-end

ograu:
IMO we should probably get rid of that call to `clustersList()` and let react and redux do their job, probably we will have to make some adjustments somewhere, but now that I am reading it, if we just delete that line, the cluster will be removed from the store and as we are working with immutable updates, this should be reflected to all the views:

```javascript
export const batchedClusterDeleteConfirmed = cluster => async dispatch => {
  try {
    dispatch(push(`/organizations/${cluster.owner}`));
    await dispatch(clusterActions.clusterDeleteConfirmed(cluster));
    dispatch(modalActions.modalHide());
    // ensure refreshing of the clusters list
[...]

PR #1035, opened 2020-01-23
Adjust deletion behaviour.

oponder:
Instead of deleting the cluster object, I am setting the `deleted_date`, which is more like how the API works too.

That way we don't remove it from the list, and then have it suddenly pop back in again in case it is taking a while to be deleted.

Related to: https://github.com/giantswarm/happa/issues/1034

### DIFF
--- a/src/reducers/clusterReducer.js
+++ b/src/reducers/clusterReducer.js
@@ -129,7 +129,7 @@ const clusterReducer = produce((draft, action) => {
       return;
 
     case types.CLUSTER_DELETE_SUCCESS:
-      delete draft.items[action.clusterId];
+      draft.items[action.clusterId].delete_date = Date.now();
       draft.lastUpdated = Date.now();
 
       return;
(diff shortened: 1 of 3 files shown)


src/reducers/clusterReducer.js
```diff
@@ -129,7 +129,7 @@ const clusterReducer = produce((draft, action) => {
       return;
 
     case types.CLUSTER_DELETE_SUCCESS:
-      delete draft.items[action.clusterId];
+      draft.items[action.clusterId].delete_date = Date.now();
```
ograu: `Date.now()` makes this reducer impure (heresy!)  😛 

As we are not writing unit tests for reducers I am ok with this, but it is a no-no in reduxland: https://github.com/reduxjs/redux/issues/1639#issuecomment-212107524

oponder: I could pass it in from the action. Let's try to do the right thing :) 

There's still those last updated things with Date.now, but I don't wanna make the problem worse.

oponder:
@ograu Please take a second look

ograu:
LGTM!

PR #1036, opened 2020-01-23
Clusters list with spinners

ograu:
Happa's first view - clusters list - has to wait for the details of all the clusters owned by the selected organization to be fetched and stored before being able to render. Thanks to changes introduced in this PR the view will be rendered in two steps:

1. After `clustersList()` is done, `DashboardItems`are rendered, but without details (spinners)
2. When `clusterLoadDetails()` is finished and its respective flag set to `false`, details are rendered

Benefits are that it takes less time to render and the user kind of have information of what’s going on. 

TODOs: we don’t have dedicated flags for each cluster, that means that all cluster details have to be fetched before all spinners disappear (v5 clusters will also have to wait for node pools to be fetched).

Details:

- Added cluster status flag
- Add/modify flags to achieve desired behavior
- Removed `clustersList()` in `batchedRefreshClusters()`
- Added `initilializeNodePools` flag to avoid reseting nodePools array in cluster object (an approach that we could use for https://github.com/giantswarm/happa/issues/1034)
- Added `LoadingOverlayWithoutStyles` component for a less opinionated loading spinner


src/components/UI/LoadingOverlayWithoutStyles.js
```diff
@@ -0,0 +1,27 @@
...
+//
+// Usage:
+//
+// <LoadingOverlay loading={something.hasLoaded}>
+//    <h1>{something.that.only.is.here.when.fully.loaded}</h1>
+// </LoadingOverlay>
+
+function LoadingOverlayWithoutStyles(props) {
+  if (props.loading) {
+    return <img className='loader' src='/images/loader_oval_light.svg' />;
```
axbarsan: Can you please import the image as a module, [as specified in fmt](https://github.com/giantswarm/fmt/blob/master/javascript/general_style_guide.md) ?

src/index.html
```diff
@@ -16,22 +16,18 @@
...
-    happaVersion: 'development',
-  };
+var config = {
+  // apiEndpoint: "https://api.g8s.seal.us-east-1.aws.gigantic.io",
+  // apiEndpoint: "https://api.g8s.gauss.eu-central-1.aws.gigantic.io",
+  apiEndpoint: "https://gs-api.g8s.eu-central-1.aws.cps.vodafone.com",
```
axbarsan: Don't forget these in 😬

ograu: already changed 🤦‍♂  forgot to push, doing it now

ubergesundheit:
I think its really nice to see that things are loading, but I think the round spinners inside the clusters list are not that pleasing to (at least my) eye.. I think animated text (opacity maybe) like `loading` would be suited better here

ograu:
We can use skeletons/placeholders or whatever we want.

marians:
> we don’t have dedicated flags for each cluster

@ograu What does that mean?

ograu:
> > we don’t have dedicated flags for each cluster
> 
> @ograu What does that mean?

We have general flags for each action type. For instance: we have a `CLUSTER_DETAILS_LOAD` action type that will generate a loading flag `loadingFlags.CLUSTER_DETAILS_LOAD` but all clusters share this same flag, we don't have something as `loadingFlags.CLUSTER_DETAILS_LOAD_n78kp`.

Does that answer your question?

marians:
Regarding the behaviour. With high bandwidth and the API reacting fast, it's hardly noticeable where the spinner appears and disappears. There is a lot going on quickly.

I had to throttle the bandwidth to 3G to actually be able to see where it appears.

1. Is there a straightforward way to only show the spinner when loading takes more than ~100 ms? That would be to avoid too much movement.

2. I see the spinner appear where we show the creation time (e. g. `Created 4 months ago`). Why do we have to wait for that info? It is part of the [cluster list response](https://docs.giantswarm.io/api/#operation/getClusters) so when the name of the cluster is known, the creation date should be known, too.

3. The spinner appears a bit too big for me in this context. It seems to take the maximum line height. I think it would be more appropriate if it had roughly the height and position of an uppercase letter in the text line where it appears.

> Does that answer your question?

Yes, thanks.

ograu:
> 1. Is there a straightforward way to only show the spinner when loading takes more than ~100 ms? That would be to avoid too much movement.

The usual way in most apps is show a placeholder, usually a box with rounded corners and a different color. That would be my preferred solution for this.

> 2. I see the spinner appear where we show the creation time (e. g. `Created 4 months ago`). Why do we have to wait for that info? It is part of the [cluster list response](https://docs.giantswarm.io/api/#operation/getClusters) so when the name of the cluster is known, the creation date should be known, too.

True, fixing it.

> 3. The spinner appears a bit too big for me in this context. It seems to take the maximum line height. I think it would be more appropriate if it had roughly the height and position of an uppercase letter in the text line where it appears.

Let's remove the spinner and use placeholders?

marians:
Yeah, let's try the placeholders. That should be much simpler.

ograu:
ok, I will wait for @oponder's take on this before removing spinners.

oponder:
I'm on board with the other opinions about how the spinning loader look, the code changes are great and going in a good direction.

If you wanna hang out about the look and feel, I'd be happy to.

ograu:
Are you ok if I merge this before more conflicts arise? And keep working towards the placeholders idea in an upcoming PR?

PR #1037, opened 2020-01-23
Show a 'n/a' label when there are no nodes loaded yet

axbarsan:
Fixes https://github.com/giantswarm/giantswarm/issues/8567

- Create a `ScalingNodeCounter` component, that it's reused for all providers, for consistent styling
- Show a `n/a` label instead of showing `0` nodes if the cluster is not loaded yet
- *BONUS:* Lower the precision of the total memory displayed in the cluster detail, as it would sometimes display 64-bit floats

**Preview**

Screenshot 2020-01-23 at 15 50 10

### DIFF
--- a/src/components/Cluster/ClusterDetail/WorkerNodesAzure.js
+++ b/src/components/Cluster/ClusterDetail/WorkerNodesAzure.js
@@ -21,12 +21,18 @@ export const LineDiv = styled.div`
   }
 `;
 
+export const ScalingNodeCounter = styled(RefreshableLabel)`
+  margin-right: 25px;
+`;
+
 function WorkerNodesAzure({ instanceType, nodes, showScalingModal }) {
   const instanceTypeText = instanceType
     ? // prettier-ignore
       // eslint-disable-next-line no-magic-numbers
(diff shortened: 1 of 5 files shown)


src/components/Cluster/ClusterDetail/WorkerNodesAzure.js
```diff
@@ -21,13 +21,19 @@ export const LineDiv = styled.div`
   }
 `;
 
+export const ScalingNodeCounter = styled(RefreshableLabel)`
+  margin-right: 25px;
+`;
+
```
ograu: ❤️ now I get it!

src/components/UI/RefreshableLabel.js
```diff
@@ -41,12 +41,15 @@ const Wrapper = styled.div`
  */
 
 // As this hook has styles, we are going to pass a styles prop to overwrite/add any styles
-function RefreshableLabel({ children, value, style }) {
+function RefreshableLabel({ children, value, style, className }) {
```
ograu: We can remove `style` if we are going to use this approach

PR #1052, opened 2020-01-28
Activate docker layer caching for happa

marians:


oponder:
Just curious, does it have some noticeable / measurable improvement?

marians:
I didn't see an improvement. But then it's also a total mystery whether this PR actually activated layer caching, or whether it was already active.

PR #1055, opened 2020-01-28
Fix memory leaks and errors in tests console

ograu:
My intention was to fix those potential memory leaks that tests were revealing, and end up removing what I believe is useless code. 

We don't need to check for cluster details being loaded in this nested component, we do it in one of the parents.

Please correct me if I am wrong. Thanks!


axbarsan:
This is really cool 🎈

PR #1056, opened 2020-01-29
Let produceNodePools reset local nodePools in NodePools component

ograu:
When a user deletes the last node pool in a cluster, Happa shows a spinner instead showing the no node pools message.

This PR fixes this behaviour.

Also it changes the initialization of local nodePools so you can see at a glance its type.

### DIFF
--- a/src/components/Cluster/ClusterDetail/V5ClusterDetailTable.js
+++ b/src/components/Cluster/ClusterDetail/V5ClusterDetailTable.js
@@ -339,9 +339,6 @@ class V5ClusterDetailTable extends React.Component {
       .join(' ');
     this.setState({
       availableZonesGridTemplateAreas: `"${availableZonesGridTemplateAreas}"`,
-    });
-
-    this.setState({
       nodePools,
       loading: false,
     });
--- a/src/components/Cluster/ClusterDetail/V5ClusterDetailTable.js
+++ b/src/components/Cluster/ClusterDetail/V5ClusterDetailTable.js
@@ -306,9 +306,9 @@ class V5ClusterDetailTable extends React.Component {
     }
   }
 
-  // TODO Move this to the action creator so it will be triggered on every NPs load.
   produceNodePools = () => {
-    // If there are no node pools, set local state to an empty array
+    // If there are no node pools or after all node pools are removed, set local
+    // state to an empty array
     if (Object.keys(this.props.nodePools).length === 0) {
       this.setState({
         nodePools: [],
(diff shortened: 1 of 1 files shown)


src/components/Cluster/ClusterDetail/V5ClusterDetailTable.js
```diff
@@ -306,37 +306,45 @@ class V5ClusterDetailTable extends React.Component {
...
+      .sort()
+      .join(' ');
+    this.setState({
+      availableZonesGridTemplateAreas: `"${availableZonesGridTemplateAreas}"`,
+    });
+
+    this.setState({
+      nodePools,
+      loading: false,
+    });
```
axbarsan: Can you please merge together these 2 calls?

src/components/Cluster/ClusterDetail/V5ClusterDetailTable.js
```diff
@@ -306,37 +306,45 @@ class V5ClusterDetailTable extends React.Component {
...
+    // If there are no node pools or after all node pools are removed, set local
+    // state to an empty array
+    if (Object.keys(this.props.nodePools).length === 0) {
       this.setState({
-        availableZonesGridTemplateAreas: `"${availableZonesGridTemplateAreas}"`,
-      });
-
-      this.setState({
-        nodePools,
+        nodePools: [],
```
oponder: Hey Oriol! This is something I just noticed, I know it's a comment on something that existed already before this PR, so whatever comes out of it might not be in scope of these changes... but:

Why is nodepools in the state? Calling this produceNodePools in componentDidMount feels a bit weird to me.

I think this is a different pattern than what we normally do right?

Is there a way to move this code to mapStateToProps?

Or is the actual work being done here so 'heavy' that we want to do it after the component mounts so we can show a loading spinner for just that?

ograu: Yes, you are totally right, this shouldn't be here. And no, this is not an expensive computation. 

This `produceNodePools` function does two things:

1. It selects cluster node pools 
2. It generates a AZs array for each NP

What I want to do is put 1 in `mapStateToProps`, using a selector (that we already have in `clustersSelectors`) and substitute `state.nodePools` using instead of it  something like `state.nodePoolsAZs`

I just haven't had the time to do it.

oponder: Ok cool! No worries! Thanks for explaining :)

ograu: Thanks for noticing it! I love to discuss this kind of things :-)

PR #1061, opened 2020-01-31
[CHANGE] Produce node pools to selector

ograu:
- Node pools injected through `mapStateToProps` using a selector instead being filtered in `V5ClusterDetailsTable`
- Removed useless code
- Removed `selectors/index.js` (duplicate of `clusterSelectors`)
- Using deep equality checks in memoized selectors


axbarsan:
What do you think about computing `availableZonesGridTemplateAreas` in the render method, instead of using state?

What I'm thinking about is removing the calls to `this.produceAZGrid()` from `componentDidMount()` and `componentDidUpdate()`, and refactoring `this.produceAZGrid()` to have memoized returned values, and also, returning the `availableZonesGridTemplateAreas` directly.

I know we talked about getting rid of the "setting the state from props" pattern at some point

ograu:
> I know we talked about getting rid of the "setting the state from props" pattern at some point

Yes, good idea. IMO this is the way to go. I'll give it a try.

The only thing I am skipping for now is memoization, I am having some troubles with creating memoized selectors with Immer.

axbarsan:
> The only thing I am skipping for now is memoization, I am having some troubles with creating memoized selectors with Immer.

I was thinking about memoizing the `this.produceAZGrid()` method, not the selector, which can be done in 2 ways:
- If you want to keep using a class component, using the `memoize` function in underscore.js
- If you want to refactor to a functional component, using the `useMemo` hook in React

ograu:
Thanks, but my idea was to derive data in selectors, this is the pattern I am working with.

I think that it is interesting to establish patterns and stick to them, at least they make everything easier for me, so I'd rather find a way (I was actually thinking in underscore's `memoize`) to make it work with selectors.

ograu:
We can talk about it if you think your approach is better.

axbarsan:
> Thanks, but my idea was to derive data in selectors, this is the pattern I am working with.
> 
> I think that it is interesting to establish patterns and stick to them, at least they make everything easier for me, so I'd rather find a way (I was actually thinking in underscore's `memoize`) to make it work with selectors.

So you want to move the template grid generation to the selector? That's even better 👍

src/index.html
```diff
@@ -17,21 +17,17 @@
...
-    azureCapabilitiesJSON: '{"Standard_A2_v2":{"additionalProperties":{},"description":"This is some description","maxDataDiskCount":4,"memoryInMb":4294.967296,"name":"Standard_A2_v2","numberOfCores":2,"osDiskSizeInMb":1047552,"resourceDiskSizeInMb":21474.83648},"Standard_A4_v2":{"additionalProperties":{},"description":"Here is a longer description that might be too long for the field","maxDataDiskCount":8,"memoryInMb":8589.934592,"name":"Standard_A4_v2","numberOfCores":4,"osDiskSizeInMb":1047552,"resourceDiskSizeInMb":42949.67296},"Standard_A8_v2":{"additionalProperties":{},"description":"Another VM size description text","maxDataDiskCount":16,"memoryInMb":17179.869184,"name":"Standard_A8_v2","numberOfCores":8,"osDiskSizeInMb":1047552,"resourceDiskSizeInMb":85899.34592}, "Standard_D2s_v3":{"additionalProperties":{},"description":"Dsv3-series, general purpose, 160-190 ACU, premium storage supported","maxDataDiskCount":4,"memoryInMb":8589.935,"name":"Standard_D2s_v3","numberOfCores":2,"osDiskSizeInMb":1047552,"resourceDiskSizeInMb":17179.87}}',
-    happaVersion: 'development',
-  };
+  apiEndpoint: "https://api.g8s.gauss.eu-central-1.aws.gigantic.io",
```
axbarsan: Don't forget this here 👍

src/components/Cluster/ClusterDetail/ClusterDetailView.js
```diff
@@ -67,7 +67,7 @@ class ClusterDetailView extends React.Component {
     this.loadDataInterval = this.props.setInterval(
       this.refreshClusterData,
       // eslint-disable-next-line no-magic-numbers
-      30 * 1000 // 30 seconds
+      3 * 1000 // 30 seconds
```
axbarsan: Don't forget this here 👍

axbarsan:
Looking really good 👍

ograu:
Spent all the afternoon trying to avoid the deep equal solution. In the end this is what I am using in this PR for memoized selectors.

ograu:
> Looking really good 👍
> 
> Do you mind adding a few lines to FMT about data processing? (that it has to be done in the Redux layer)

Sure! Will do it tomorrow :-)

PR #1062, opened 2020-01-31
Fix flashes in cluster details

ograu:
Fix `batchedRefreshClusters` so it doesn't reset node pools array when refreshing cluster details view


src/components/Cluster/ClusterDetail/ClusterDetailView.js
```diff
@@ -67,7 +67,7 @@ class ClusterDetailView extends React.Component {
     this.loadDataInterval = this.props.setInterval(
       this.refreshClusterData,
       // eslint-disable-next-line no-magic-numbers
-      30 * 1000 // 30 seconds
+      3 * 1000 // 30 seconds
```
axbarsan:
```suggestion
      3 * 1000 // 3 seconds
```

ograu: 🤦‍♂ thanks Adrian

PR #1079, opened 2020-02-05
[FEATURE] Explain why the ingress column is empty on som AWS clusters

oponder:
This adds an explanation for the empty ingress column on `> 10.x.x AWS`  clusters

Screenshot 2020-02-05 at 3 47 31 PM


src/components/Cluster/ClusterDetail/ClusterApps.js
```diff
@@ -309,22 +312,63 @@ class ClusterApps extends React.Component {
...
+                  {this.preinstalledApps().management.map(app => {
+                    return (
+                      <div className='cluster-apps--app' key={app.name}>
+                        <img alt={`${app.title} icon`} src={app.logoUrl} />
+                        {app.name}
+                        <small>{app.version}&nbsp;</small>
+                      </div>
+                    );
+                  })}
+                </div>
```
axbarsan: These 2 are really similar. Do you mind creating a separate component for them?

src/components/Cluster/ClusterDetail/ClusterApps.js
```diff
@@ -309,22 +312,63 @@ class ClusterApps extends React.Component {
...
+                        <br />
+                        <br />
+                        Read more in our{' '}
+                        <a href='https://docs.giantswarm.io/guides/installing-optional-ingress-controller/'>
+                          installing an ingress controller guide.
+                        </a>
+                      </p>
+                    </OptionalIngressNotice>
+                  ) : (
+                    undefined
```
axbarsan: No need to return `undefined` here, you can just write
```js
{this.props.hasOptionalIngress && (
  <Component>
  </Component
)}
```

src/components/Cluster/ClusterDetail/ClusterApps.js
```diff
@@ -309,22 +312,63 @@ class ClusterApps extends React.Component {
...
+                  {this.props.hasOptionalIngress ? (
+                    <OptionalIngressNotice>
+                      <p>
+                        The ingress controller is optional on this cluster.
+                        <br />
+                        You can install one using our app catalog.
+                        <br />
+                        <br />
+                        Read more in our{' '}
+                        <a href='https://docs.giantswarm.io/guides/installing-optional-ingress-controller/'>
```
axbarsan:
```suggestion
                        <a target='_blank' href='https://docs.giantswarm.io/guides/installing-optional-ingress-controller/'>
```

src/components/Cluster/ClusterDetail/ClusterApps.js
axbarsan: The ingress column is similar, as well

oponder: ok! will fix that

PR #1083, opened 2020-02-06
Prevent v4 to v5 upgrade.

oponder:
This prevents users from seeing the "upgrade available" message when looking at a `< 10.0.0` cluster that is targeting a `>= 10.0.0` release.

Towards: https://github.com/giantswarm/giantswarm/issues/8696


src/components/Cluster/ClusterDetail/ClusterDetailView.js
```diff
@@ -142,19 +142,6 @@ class ClusterDetailView extends React.Component {
     return 'Not found';
   }
 
-  // Determine whether the current cluster can be upgraded
-  canClusterUpgrade() {
```
oponder: Extracted this to clusterUtils so that I can unit test it.

src/utils/clusterUtils.js
```diff
@@ -1,9 +1,28 @@
...
+import { Constants, Providers } from 'shared/constants';
 
 // Here we can store functions that don't return markup/UI and are used in more
 // than one component.
 
+// Determine whether a cluster at a certain version can be upgraded to a target version.
+export function canClusterUpgrade(currentVersion, targetVersion, provider) {
+  // Cluster must have a release_version.
+  if (currentVersion === '') return false;
+  if (Boolean(currentVersion) !== true) return false;
```
axbarsan: Why do we need both of these checks here?

Shouldn't `if (!currentVersion) return false` be enough?

oponder: I think theres some weird API corner case we once had to guard against, where we'd get a empty string back. It might not be a issue anymore though, but I am reluctant to change that right now 😅

axbarsan: `if (!currentVersion) return false;` guards against an empty string, `null`, `undefined`, `0`, `NaN` and `false` so we should be safe 👍

src/components/Cluster/ClusterDetail/ClusterDetailView.js
ograu: 👍I am ok with this, but I think from now on it would be interesting to compute derived data on selectors, though.

src/utils/clusterUtils.js
oponder: also I do prefer `(!currentVersion)` over `Boolean(currentVersion) !== true)`

But you started it :P ;)

https://github.com/giantswarm/happa/blame/228a39be86d3e2b3e5470570999e1079c9041884/src/components/Cluster/ClusterDetail/ClusterDetailView.js#L151

src/components/Cluster/ClusterDetail/ClusterDetailView.js
oponder: Ah yeah, I can make a selector! Only thing is I would probably still like an extracted function that the selector then uses, because the inputs of that function are a bit clearer (currentVersion, targetVersion, provider) instead of (state, props)

src/utils/clusterUtils.js
axbarsan: Don't throw me under the bus! I just converted `!!this.props.targetRelease` to `Boolean(this.props.targetRelease)` because it was upsetting the linter 😬

src/components/Cluster/ClusterDetail/ClusterDetailView.js
oponder: Actually are there any rules about the inputs to selectors? We dont have to pass the top level state or props always right?

ograu: We don't have rules for it. Some people say that it's best to pass top level state because if you normalize the store, then you just have to change arguments in one place. And I think that's what I am doing. 

In case of props I think it's best to pass the value you need instead of top level props, because props can  vary from one component to another.

oponder: Ok I start to understand things more :)

src/components/Cluster/ClusterDetail/ClusterDetailView.js
```diff
@@ -343,7 +331,7 @@ class ClusterDetailView extends React.Component {
                 />
               )}
 
-              {targetRelease && (
+              {canClusterUpgrade && (
```
oponder: somewhat unrelated. Just felt that we shouldn't render the modal at all if we cant upgrade the cluster anyways.

It used to be that we always could upgrade when we had a next release (which is what targetRelease represents)

But that's no longer the case.

src/components/Cluster/ClusterDetail/ClusterDetailView.js
oponder: If you can take a look and let me know if the selector I made kinda makes sense, that would be great!

ograu: I think it is ok! :-)

PR #1086, opened 2020-02-06
Delete cluster on 404

ograu:
When a cluster is finally removed (excluded from response in `/v4/clusters` after been marked as `deleted`) Happa has to remove the cluster from the store, otherwise it will be trying to fetch data for a cluster that no longer exists and triggering error messages.

### DIFF
--- a/src/actions/clusterActions.js
+++ b/src/actions/clusterActions.js
@@ -452,15 +452,17 @@ export function clusterLoadDetails(
     } catch (error) {
       if (error.status === StatusCodes.NotFound) {
         new FlashMessage(
-          'This cluster no longer exists.',
+          `Cluster <code>${clusterId}</code> no longer exists.`,
           messageType.INFO,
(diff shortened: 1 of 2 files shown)


src/actions/clusterActions.js
```diff
@@ -452,15 +452,17 @@ export function clusterLoadDetails(
     } catch (error) {
       if (error.status === StatusCodes.NotFound) {
         new FlashMessage(
-          'This cluster no longer exists.',
+          `Cluster <code>${clusterId}</code> no longer exists.`,
           messageType.INFO,
-          messageTTL.MEDIUM,
-          'Redirecting you to your organization clusters list'
+          messageTTL.MEDIUM
         );
```
ograu: Removed this because this message can appear in cluster list, which doesn't make sense at all.

axbarsan:
Saw this today, awesome!

PR #1100, opened 2020-02-10
Add error reducer

ograu:
Towards https://github.com/giantswarm/giantswarm/issues/6716

The idea is to centralise our error flags/messages into an `error` object in the store, and use them in error boundaries.

### DIFF
--- a/src/reducers/errorReducer.js
+++ b/src/reducers/errorReducer.js
@@ -1,0 +1,29 @@
+import produce from 'immer';
+
+const initialState = {};
+
+const errorReducer = produce((draft, action) => {
+  const { type, payload } = action;
+  const matches = /(.*)_(SUCCESS|ERROR)/.exec(type);
+
+  // not a *_SUCCESS / *_ERROR actions, so we ignore them
+  if (!matches) return;
+
+  const [, requestName, requestState] = matches;
+
+  switch (requestState) {
+    case 'ERROR':
+      draft[requestName] = payload?.message ? payload.message : '';
+
+      return;
+
+    // If the actions is a SUCCESS one, lets remove the key and keep the error object as
+    // minimal as possible.
+    case 'SUCCESS':
+      if (draft[requestName]) {
+        delete draft[requestName];
+      }
+  }
+}, initialState);
+
+export default errorReducer;
--- a/src/reducers/errorReducer.js
+++ b/src/reducers/errorReducer.js
@@ -13,7 +13,7 @@ const errorReducer = produce((draft, action) => {
 
   switch (requestState) {
     case 'ERROR':
-      draft[requestName] = payload?.message ? payload.message : '';
+      draft[requestName] = payload?.message ?? '';
 
       return;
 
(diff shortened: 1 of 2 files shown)


src/reducers/errorReducer.js
```diff
@@ -0,0 +1,29 @@
...
+  const matches = /(.*)_(SUCCESS|ERROR)/.exec(type);
+
+  // not a *_SUCCESS / *_ERROR actions, so we ignore them
+  if (!matches) return;
+
+  const [, requestName, requestState] = matches;
+
+  switch (requestState) {
+    case 'ERROR':
+      draft[requestName] = payload?.message ? payload.message : '';
```
axbarsan: [There's actually an easier way of doing this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#Combining_with_the_nullish_coalescing_operator)

So it would be something like 
```js
draft[requestName] = payload?.message ?? '';
```

ograu: Wow, cool, thanks Adrian! 👍 didn't know that one

src/reducers/errorReducer.js
```diff
@@ -0,0 +1,29 @@
...
+  const matches = /(.*)_(SUCCESS|ERROR)/.exec(type);
+
+  // not a *_SUCCESS / *_ERROR actions, so we ignore them
+  if (!matches) return;
+
+  const [, requestName, requestState] = matches;
+
+  switch (requestState) {
+    case 'ERROR':
+      draft[requestName] = payload?.message ?? '';
```
oponder: This means we store nothing in case of an error without a message?

Or will we have to make sure that we always dispatch a `*_ERROR` action with a message? 

I'm trying to understand the flow of error data here... when an error occurs, it will hit the nearest error boundary right, and then we dispatch a general error from that boundary? And then the component that the error boundary is rendering looks into the state to show some error information?

ograu: > This means we store nothing in case of an error without a message?

> Or will we have to make sure that we always dispatch a `*_ERROR` action with a message?
> 

We will always want store something. This is just a guard against us forgetting to add the error to the action.

> I'm trying to understand the flow of error data here... when an error occurs, it will hit the nearest error boundary right, and then we dispatch a general error from that boundary? And then the component that the error boundary is rendering looks into the state to show some error information?

The error boundary will dispatch nothing, but in some cases we might want to dispatch something. Error boundaries is more about showing a fallback UI, and logging to slack. 

We have to think about error boundaries as consecuences of dispatching an error, not as error dispatchers.

oponder: Ok, it's just in some cases there will be errors that happen with no `*_ERROR` action. Just like uncaught errors and mistakes in components themselves. 

And I think that's the more common case that an error boundary will become visible. 

So I'm trying to understand in what situations and what components will try to render data from this error reducer. 

But it's ok, I don't see the big picture yet.

PR #1101, opened 2020-02-10
Add entity error reducer and ErrorFallback component

ograu:
Introducing `ErrorFallback` and `entityErrorReducer`.

I want to display meaningful info in the view when something did go wrong. For example, when for any reason we can't fetch node pools, I don't want `0 nodes`(which is not true)  in the place when resources should be detailed in clusters list. I'd prefer to have a message telling something to the user:

Captura de pantalla 2020-02-10 a las 17 25 56

- `ErrorFallback` is like `LoadingOverlay` but for errors.

- `entityErrorReducer` is used to keep track of errors that happen to every cluster/node pools/entity. Errors are stored in an `errorsByEntity` object so we can fetch and use them quickly in components.

ps: forget about ErrorBoundaries, and sorry for mixing concerns. Error Boundaries are for catching errors in the view. I am more concerned by how we handle errors from API calls.

### DIFF
--- a/src/components/Home/ClusterDashboardItem.js
+++ b/src/components/Home/ClusterDashboardItem.js
@@ -9,11 +9,15 @@ import React from 'react';
 import ButtonGroup from 'react-bootstrap/lib/ButtonGroup';
 import { connect } from 'react-redux';
 import { Link } from 'react-router-dom';
-import { selectClusterNodePools } from 'selectors/clusterSelectors';
+import {
+  selectClusterNodePools,
+  selectClusterNodePoolsErrorsById,
+} from 'selectors/clusterSelectors';
 import { OrganizationsRoutes } from 'shared/constants/routes';
 import { Dot } from 'styles';
 import Button from 'UI/Button';
 import ClusterIDLabel from 'UI/ClusterIDLabel';
+import ErrorFallback from 'UI/ErrorFallback';
 import RefreshableLabel from 'UI/RefreshableLabel';
 
 import ClusterDashboardResources from './ClusterDashboardResources';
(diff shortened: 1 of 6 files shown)


ograu:
Other considerations:

- `ErrorFallback` maybe has to be style agnostic, and can have other style. This is just a POC regarding what this component renders should look like.

- Maybe I went to far with this. Let me know if you think so please.

axbarsan:
I really like the way you approached this topic. Keep up the awesome work! 👍

oponder:
actually I am thinking a bit about this...

I'm ok to merge this for now.. but I'd like to ponder a bit together.

Because we are deciding kind of now that a component is not responsible for being reasonably renderable in the case of an error.

This error fallback component can be useful, but I think it maybe should be something put _inside_ of `ClusterDashboardResources` instead of _around_ it.

ograu:
> I'm ok to merge this for now.. but I'd like to ponder a bit together.
> 
> Because we are deciding kind of now that a component is not responsible for being reasonably renderable in the case of an error.

This component is unopinionated about where can be used. We can use it wherever we want, like `LoadingOverlay`. But you are right in the sense that we are setting a precedent here.

> This error fallback component can be useful, but I think it maybe should be something put _inside_ of `ClusterDashboardResources` instead of _around_ it.

I think it makes sense to not render a component that is not going to be rendered, performance-wise. This is like an early return for cases you don't want to cover in a function.  Why do you think we should put the error fallback inside the component?

ograu:
Merging now so I can keep working on this, but happy to keep the discussion going @oponder

PR #1103, opened 2020-02-11
Fallback message when nodes are not ready yet

ograu:

We want to give relevant information in the resources components when the cluster is being created. Now we are showing this:

image

In this PR I am proposing this:

image

And this:

image

ps: not sure if the message is really appropiate

### DIFF
--- a/__tests__/V5ClusterManagement.js
+++ b/__tests__/V5ClusterManagement.js
@@ -76,7 +76,7 @@ afterAll(() => {
 
 /************ TESTS ************/
 
-it('renders all the v5 cluster data correctly with 0 nodes ready', async () => {
+it('renders all the v5 cluster data correctly', async () => {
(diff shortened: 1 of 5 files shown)


__tests__/V5ClusterManagement.js
```diff
@@ -76,7 +76,7 @@ afterAll(() => {
 
 /************ TESTS ************/
 
-it('renders all the v5 cluster data correctly with 0 nodes ready', async () => {
+it('renders all the v5 cluster data correctly', async () => {
```
ograu: I was going to fix the test, but there was nothing to be fixed: the name was wrong

ograu:
> Code is looking good.
> 
> About the appearance, I think it would make sense if the fallback message had a lower opacity than the real value, just to show that it is temporary

I like the idea 👍 
Done, and screenshots updated in the PR description.

PR #1104, opened 2020-02-11
Change ErrorFallback so that it accepts an array

ograu:
And so we can pass different errors at once.

### DIFF
--- a/src/components/UI/ErrorFallback.js
+++ b/src/components/UI/ErrorFallback.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import React from 'react';
 
 // ErrorFallback takes an error string or an array of error strings and will render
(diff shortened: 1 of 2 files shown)


src/components/UI/ErrorFallback.js
```diff
@@ -2,15 +2,32 @@ import styled from '@emotion/styled';
 import PropTypes from 'prop-types';
 import React from 'react';
 
-// ErrorFallback takes one string prop 'errorMessage' and will render the
-// message if it evaluates to true and the children otherwise.
+// ErrorFallback takes an error string or an array of error strings and will render
+//  the message/s. If receives a falsy value or an array of falsy values it will
```
axbarsan:
```suggestion
//  the message/s. If it receives a falsy value or an array of falsy values it will
```

PR #1109, opened 2020-02-12
Adapt Happa to use new error/loading reducers in clusterReducer and cluster views

ograu:

- Removed action handling in `clusterReducer`when we were just storing flags because errors reducers and loading reducer are taking care of this
- Adjust tests because `ClusterApps` is now a connected component
- Fix flag not working in 'ClusterApps``
- Added an _universal_ selector for all kinds of entity errors


src/components/Cluster/ClusterDetail/ClusterDetailView.js
```diff
@@ -45,10 +45,6 @@ const WrapperDiv = styled.div`
   }
 `;
 class ClusterDetailView extends React.Component {
-  state = {
-    errorLoadingApps: false,
-  };
-
```
ograu: This was always false, removed cause now we are using errors in store

PR #1116, opened 2020-02-12
Temporary fix for last updated label

ograu:
I noticed that the last updated label has stop working: this is because I removed `draft.lastUpdate = Date.now()` in the reducer because it was updating the store and triggering rerenders each and every time an action with 'lastUpdate`was dispatched, no matter if new state was added to the store or not.

This is a temporary solution that updates the value every time that a refresh happens. This is not adding alot of value to the view, it just saying that a refresh happened (neither was doing so before IMO).

We should probably think about this label and how to provide more useful information to the user. Maybe use error object, or maybe add info about when was the last time the cluster was modified.


src/index.html
```diff
@@ -17,21 +17,18 @@
...
-    azureCapabilitiesJSON: '{"Standard_A2_v2":{"additionalProperties":{},"description":"This is some description","maxDataDiskCount":4,"memoryInMb":4294.967296,"name":"Standard_A2_v2","numberOfCores":2,"osDiskSizeInMb":1047552,"resourceDiskSizeInMb":21474.83648},"Standard_A4_v2":{"additionalProperties":{},"description":"Here is a longer description that might be too long for the field","maxDataDiskCount":8,"memoryInMb":8589.934592,"name":"Standard_A4_v2","numberOfCores":4,"osDiskSizeInMb":1047552,"resourceDiskSizeInMb":42949.67296},"Standard_A8_v2":{"additionalProperties":{},"description":"Another VM size description text","maxDataDiskCount":16,"memoryInMb":17179.869184,"name":"Standard_A8_v2","numberOfCores":8,"osDiskSizeInMb":1047552,"resourceDiskSizeInMb":85899.34592}, "Standard_D2s_v3":{"additionalProperties":{},"description":"Dsv3-series, general purpose, 160-190 ACU, premium storage supported","maxDataDiskCount":4,"memoryInMb":8589.935,"name":"Standard_D2s_v3","numberOfCores":2,"osDiskSizeInMb":1047552,"resourceDiskSizeInMb":17179.87}}',
-    happaVersion: 'development',
-  };
+  apiEndpoint: "https://api.g8s.gauss.eu-central-1.aws.gigantic.io",
```
axbarsan: 🦉

ograu: 💩 🙈

PR #1120, opened 2020-02-13
Add tests for app installation and editing

axbarsan:

This also introduces the `forceRemoveAll()` helper function for flash messages.
It clears all flash messages queues and removes existing flash messages, because they remain open between tests and can cause undesired behavior


__tests__/AppCatalog.js
```diff
@@ -22,162 +26,633 @@ import {
...
+    getMockCall(
+      `/v4/clusters/${V4_CLUSTER.id}/status/`,
+      v4AWSClusterStatusResponse
+    );
+    getMockCall(`/v4/organizations/${ORGANIZATION}/credentials/`);
+    getMockCall('/v4/appcatalogs/', appCatalogsResponse);
+  });
+
+  afterEach(() => {
+    expect(nock.isDone());
```
oponder: I noticed we have this in a few places, we're not actually checking anything here 😅 

```suggestion
    expect(nock.isDone()).toBeTruthy()`;
```

or `toBe(true)` if you prefer.

getting strict on the requests doesn't have to be in the scope of this PR though if changing this causes you a lot more work right now.

__tests__/AppCatalog.js
```diff
@@ -22,162 +26,633 @@ import {
...
+    await findAllByText(V4_CLUSTER.name);
+    await findByText(/kubernetes endpoint uri/i);
+
+    /**
+     * Manually removing the toast notification to prevent
+     * it from showing up in other tests
+     */
+    document.querySelector('#noty_layout__topRight').remove();
+
+    installAppRequest.done();
```
oponder: `.done()` on a nock scope doesn't actually do anything. It returns true or false based on whether the interceptor has intercepted a call or not. It can be used in an assertion, but on it's own, no need for it.

__tests__/AppCatalog.js
```diff
@@ -22,162 +26,633 @@ import {
...
+        name: 'secret.json',
+      }
+    );
+    fireEvent.change(fileInput, {
+      target: {
+        files: [file],
+      },
+    });
+
+    await waitForDomChange();
```
oponder: what are we actually asserting with `waitForDomChange`? Is this enough to say that Happa did what we expected it to do?

axbarsan:
Thank you for the comments, @oponder, guess I'll start reworking the tests
image

As for asserting if the functionality actually works, I could watch for the flash messages.

And about the requests, I will handle the ones in the app catalog tests in this PR and handle the rest in  a separate one. That should resolve all memory leaks in our test, so really good catch!

src/lib/flashMessage.js
```diff
@@ -62,6 +62,18 @@ export function clearQueue(queueName) {
   Noty.closeAll(queueName);
 }
 
+/**
+ * Remove all messages from all queues without waiting for the animation
+ * to finish
+ */
+export function forceRemoveAll() {
```
oponder: Since this is a new concept, it would be nice to mention it in the PR, what made you introduce it, and possibly for bonus points, to introduce it in a PR of it's own.

src/components/Cluster/ClusterDetail/AppDetailsModal.js
```diff
@@ -42,41 +42,41 @@ const AppDetailsModal = props => {
...
-  function dispatchDeleteAppConfig(app, clusterId, dispatch) {
-    dispatch(deleteAppConfig(app.metadata.name, clusterId))
+  function dispatchDeleteAppConfig(app, clusterId, dispatch, closeModal) {
+    return dispatch(deleteAppConfig(app.metadata.name, clusterId))
       .then(() => {
         return dispatch(loadApps(clusterId));
       })
       .then(() => {
-        onClose();
+        closeModal();
```
oponder: I'm not sure about this change here, how did this ever work before :P I know I wrote it... but I see there is no onClose() being passed in to this function... weird.

axbarsan: It works before `onClose()` is declared in this file, so it's in scope.
I just wanted it to be the same as the other `dispatch*` methods in the file, because I had some issues and I thought it would solve them, which they didn't. Left them like that for consistency, though

src/lib/flashMessage.js
axbarsan: You're right, I added a description for that

src/components/Cluster/ClusterDetail/AppDetailsModal.js
ograu: Late to the party, sorry.

It worked because it is called on confirmation in `batchedClusterDeleteConfirmed()`. 

It would be nice if we clean these `then().catch()` blocks in components to avoid memory leaks, race conditions, etc. Generally we want them in actions

axbarsan: A refactor is on the way, but that will come in a different PR 👍

ograu: 🎉

PR #1125, opened 2020-02-13
Entity loading reducer

ograu:
Fixes https://github.com/giantswarm/giantswarm/issues/8815

Added `entityLoadingReducer` for more granular control on loading statuses, and added check for cluster specific loading flag in `ClusterDetailView`.

Also refactored `entityErrorReducer` cause it was overwriting errors instead of adding them.


axbarsan:
I think the `kind/bug` label is enough. The issue might be added twice to the changelog if there are multiple labels present

ograu:
and thanks @axbarsan for your quick reviews!

PR #1128, opened 2020-02-14
Revert "Change nodes not ready condition"

ograu:
Reverts giantswarm/happa#1108 as cluster data doesn't include timezone, so date comparisons are not possible.


oponder:
Hey, just writing quickly from the airport. 

We should be able to count on dates in our system being in UTC always. 

So you can force the dates to be UTC and then compare against local time with some confidence

PR #1129, opened 2020-02-14
Revert "Fallback message when nodes are not ready yet"

ograu:
Reverts giantswarm/happa#1108 as cluster data doesn't include timezone, so date comparisons are not possible.

### DIFF
--- a/__tests__/V5ClusterManagement.js
+++ b/__tests__/V5ClusterManagement.js
@@ -76,7 +76,7 @@ afterAll(() => {
 
 /************ TESTS ************/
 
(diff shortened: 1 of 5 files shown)


marians:
Not sure what happened here. However dates coming from the backend, if missing time zone information, can always be assumed to be UTC.

axbarsan:
> Not sure what happened here. However dates coming from the backend, if missing time zone information, can always be assumed to be UTC.

This was re-done with UTC in mind 😬 
https://github.com/giantswarm/happa/pull/1137

PR #1130, opened 2020-02-16
Update production build

ubergesundheit:
This pull request contains the following chages:

- Upgrade nginx image from 1.14 to 1.16
- gzip compress all static files in container build
- Adds some security headers to the ingress
- Some tweaks in the nginx configuration file
- Move all assets to the `assets` subdirectory in order to utilize caching directives present in nginx config

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,18 +1,13 @@
 FROM nginx:1.16-alpine
 
-RUN apk --no-cache add jq findutils gzip
+RUN apk --no-cache add jq findutils gzip && \
+  rm -r /etc/nginx/conf.d && \
+  mkdir -p /etc/nginx/config
 
-COPY dist /www
-
-RUN rm -r /etc/nginx/conf.d
-RUN mkdir -p /etc/nginx/config
 COPY nginx.conf /etc/nginx/config
-
-COPY VERSION /
 COPY scripts/start.sh /
-
-# Test certifiates will be overwritten in production by configmap
-COPY test/certs /etc/nginx/certs
+COPY dist /www
+COPY VERSION /
 
 RUN find /www \
   -type f -regextype posix-extended \
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,4 +22,6 @@ RUN find /www \
 
 RUN chown -R nginx:nginx /www
 
+USER nginx
+
 CMD ["/start.sh"]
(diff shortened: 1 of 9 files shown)


Dockerfile
```diff
@@ -1,20 +1,27 @@
...
+COPY nginx.conf /etc/nginx/config
 
-ADD VERSION /
-ADD scripts/start.sh /
+COPY VERSION /
+COPY scripts/start.sh /
 
 # Test certifiates will be overwritten in production by configmap
-ADD test/certs /etc/nginx/certs
+COPY test/certs /etc/nginx/certs
```
ubergesundheit: What about this? Was this intended to be used with https? Could we remove this also?

oponder: This is a self signed cert for when you run this container on your own machine, that way you can at least get a response from it. Otherwise nginx doesn't start up since it complains about the missing files.

Also I think we need it for the 'smoke test' in the CI to work. (If that is still there). I think there is a step where we just check if the container starts up and is able to serve a request.

ubergesundheit: In the current configuration, the nginx container serves plain http. The CI configuration does not contain https checking and the deployment in the helm chart does not inject anything into the pod. I've checked the pods of happa on gauss, the certificate & key there is the same that is copied here into the container.

For me both is now fine: Removing this or fixing https serving from the happa container. What do you suggest?

ubergesundheit: For the record: [This commit] removes the ssl stuff from nginx config

[This commit]: https://github.com/giantswarm/happa/commit/890aeb3953657c822f53bdb6be680d7073daf7d2#diff-daaf2fceca49ed034eace91c4e875653https://github.com/giantswarm/happa/commit/890aeb3953657c822f53bdb6be680d7073daf7d2#diff-daaf2fceca49ed034eace91c4e875653

oponder: Oh! 😅 yeah it looks like these files are totally not needed then!

ubergesundheit: :) alright I'll remove this then

ubergesundheit:
I've removed the certificates & moved the VERSION file into dist to reduce layers in the container image.

I'll now make some tests with a deployment on ginger and post results here.

ubergesundheit:
Before:
- ~2.5s loading for /

After:
- Browser caches static assets like js & css files
- ~2s loading for /
- Security headers present

PR #1137, opened 2020-02-17
Fallback message when nodes are not ready yet with UTC

ograu:
This is basically the same that I merged and then reverted last week -  https://github.com/giantswarm/happa/pull/1103 - except that in `clusterUtils.js` now we are using utc time.

Sorry for the back and forth.

### DIFF
--- a/src/utils/clusterUtils.js
+++ b/src/utils/clusterUtils.js
@@ -1,3 +1,4 @@
+import moment from 'moment';
 import cmp from 'semver-compare';
 import { Constants, Providers } from 'shared/constants';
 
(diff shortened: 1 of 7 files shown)


src/utils/clusterUtils.js
```diff
@@ -148,3 +149,12 @@ export function computeCapabilities(releaseVersion, provider) {
       provider === Providers.AWS && cmp(releaseVersion, '10.0.99') === 1,
   };
 }
+
+export const isClusterYoungerThanOneHour = createDate => {
+  // eslint-disable-next-line no-magic-numbers
```
axbarsan:
```suggestion
```

This shouldn't be needed, as `1` should be an exception in the linter

PR #1145, opened 2020-02-17
Create model layer for the control plane API

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/8622

This is a replacement for the GiantSwarm API Client we currently use. This is an open discussion and I believe we should work together to get this in the right direction.

The code shown in the PR is not production ready. It lacks functionality, testing and documentation.

# Proposed architecture

### Overview (see explanation below)

Happa architecture

### Purpose
* Make it easier for us to use different clients (if we ever want to switch to GraphQL or we want to use WebSockets for certain endpoints)
* Reducing complexity in our Redux actions
* Faster development time (knowing the good spot to place new code in case of implementing a new feature, so less moving stuff around)
* 100% control (customizable, extendable, debuggable)

### Implementation

The proposed architecture is pure MVVM. I tried to keep the Model implementation as light as possible, to be easily debuggable. The model layer is the only thing that needs to be developed in a different way, compared to the way we do now, and we'll only do that for new features, so it should be pretty straightforward.

### Terms

* Model layer = Data fetching/storage/processing
  Parts of the model layer:
  * Clients = API request handlers (different server protocols, ex: _HTTP, WebSockets,_; they can be extended to create special clients for each API - to have specific settings by default)
  * Services = The place to use clients to create API requests and process data
  * Integration Controller = The place where services are used, where storage (persistent or not) is controlled (ex: _redux_)
  * Volatile Storage = Data that gets stored in memory (ex: _Redux store_)
  * Persistent Storage = Data that gets stored on the client side (ex: _Cookies, Local Storage_)
* ViewModel = Smart components linked to the model (they get data from the model and make requests for new data)
* View = Dumb UI components that just return markup from injected data
* Binding between ViewModel and Model = The glue between ViewModel and Model (what passes the data and data request methods from the Model to the ViewModel layer)

### Data flow

* Requesting data
Route gets mounted
=> a smart component from the ViewModel layer gets rendered
=> the component calls a method bound from the model
=> an action is triggered in the integration controller
=> a method from a service is called with the client injected from the action creator
=> the service asks for the data from the API, using the injected client

* Receiving data
The API gives the response to the client
=> the client receives the data
=> the service processes the data
=> the data is passed back to the integration controller
=> the data is stored by the integration controller (either in a volatile or peristent container)
=> the data is passed through the binding to the ViewModel component that wants it
=> the data is passed to multiple dumb UI components in the View layer
=> the markup is rendered and shown on screen

### Examples (keep in mind that these are just to illustrate the proof of concept, real implementations might differ)

**`Giant Swarm Service - info requests`**
```javascript
export function getInstallationInfo(httpClient) {
  httpClient.setRequestMethod(HttpRequestMethods.GET);
  httpClient.setURL(paths.Info);

  return httpClient.execute();
}
```

**`getInfo()` action creator**
```javascript
function getInfo() {
  return async (dispatch, getState) => {
    dispatch({ type: types.INFO_LOAD_REQUEST });

    try {
      const authToken = await selectAuthToken(dispatch, getState());
      const httpClient = new GiantSwarmClient(authToken);
      const info = await getInstallationInfo(httpClient);

      dispatch({
        type: types.INFO_LOAD_SUCCESS,
        info: info,
      });
    } catch (error) {
      // eslint-disable-next-line no-console
      console.error('Error loading installation info:', error);

      dispatch({
        type: types.INFO_LOAD_ERROR,
        error: error,
      });

      throw error;
    }
  };
}
```

**`selectAuthToken()` token selector - get the current auth token, and renew it if it's expired**
```javascript
export async function selectAuthToken(dispatch, state) {
  let currentToken = state.app.loggedInUser.auth.token;

  try {
    if (isJwtExpired(currentToken)) {
      const newAuthData = await Auth.getInstance().renewToken();
      currentToken = newAuthData.accessToken;

      await dispatch(userActions.auth0Login(newAuthData));
    }

    return currentToken;
  } catch (err) {
    const newErr = Object.assign({}, err, { status: StatusCodes.Unauthorized });

    throw newErr;
  }
}
```


ograu:
Wow, you did a lot of thinking!  💪 👏 

We can talk about this on SIG Frontend.

axbarsan:
We discussed it in the chapter-front-end sync and decided to go forward with the proposal, and iterate on it until we get the implementation that we want to work with.

Next steps for me would be to implement some real requests (such as the info request) to see how everything would look in the wild

axbarsan:
* Implemented `info` endpoint request using the new model
* Replaced `info` endpoint mock in tests with new model mock
* Added automatic token renewal state update

Tested manually and it's working pretty good.

The PR is getting pretty big, so I propose keeping this implementation in our codebase as a proof of concept and merging this now.

I will follow up with PRs for testing each component of the model layer (especially the `HttpClient`)

__tests__/AccountSettings.js
```diff
@@ -29,8 +30,8 @@ describe('AccountSettings', () => {
   });
 
   beforeEach(() => {
+    getInfo.mockResolvedValueOnce(AWSInfoResponse);
```
ograu: ❤️

__tests__/AdminLogin.js
```diff
@@ -137,7 +138,8 @@ describe('AdminLogin', () => {
 
   it('redirects to homepage if the user has been previously logged in', async () => {
     getMockCall('/v4/user/', userResponse);
-    getMockCallTimes('/v4/info/', AWSInfoResponse, 2);
+    getInfo.mockResolvedValueOnce(AWSInfoResponse);
+    getInfo.mockResolvedValueOnce(AWSInfoResponse);
```
ograu: My thinking is that tests is not the place to figure out how many calls we are doing, we should keep track of this while we are developing, or on performance audits.

I am ok if you guys want to do it here, it's just that I would love to forget about this in tests and make them lighter.

axbarsan: I am with you on that, it makes testing more annoying than it should be.

However, here I just try to make a drop-in replacement. We should definitely discuss removing this pattern, but I believe the change should happen in another PR

ograu: Agree :-)

oponder: I think there are some benefits to having a strict watch on the request behaviour... We can catch regressions or issues that we wouldn't otherwise if we would just set up a set of persistent requests or mocks of these new functions without checking how often they are called.

That said I'm happy to discuss it with everyone. 

Also I think the integration tests are getting close to complete? We shouldn't be adding much more integration tests after this set. 

There are some refactorings to do to fix a bit of flappyness, but I think the focus should start shifting to testing the components individually.

I did the Version Picker component in a kind of test driven way, and it was pretty nice to do it like that. And in those more focused contexts, there is no need to worry about mocking or nocking any requests at all.

axbarsan: Let's discuss it on Friday 🚀

oponder:
Hey Adrian, thanks for the effort here! 

I'm gonna write out the thoughts that popped into my head as I'm kind of having them. Most likely there is a lot more to this setup that you envision, and I would need a bit more time to understand the benefits.

But for me, I think the boxes in the diagrams make things feel a bit more independent than they really are.

I would call the stack of a [client + gateway + service] as a client, and I think it would be good to see that reflected in the code and folder structure. These three components are not really interchangeable or composable, and we should keep related things together.

Right now it's kind of spread out and for me a bit harder to grasp things in one 'brain stack'.

Like I'm used to seeing something like an InfoClient and a UsersClient that then make use of the same underlying BaseHttpClient, or something like that.

Also it feels a bit wrong to me that things like the `ConfigService` can reach into the `ControlPlaneGateway` to set things. 

Lastly, I'd consider renaming the ControlPlaneGateway to GiantSwarmAPIGateway, and reserve ControlPlaneGateway for when we start talking to the K8S api.

The kind of functional `getInfo()` function in `src/model/gateways/ControlPlaneGateway/info.js` is also a bit strange to me. I'm not sure what the intention is here, but I think we should go either with classes and just make things methods of that class, or try a purely functional approach (though Im not totally sure how to approach that in this case), but probably the ControlPlaneGateway instance should be one of the arguments that is passed into the `getInfo()` function, instead of making it something that the function fetches itself. 

Alternatively you could add the getInfo() function to the prototype of the ControlPlaneGateway, and rely on `this` more. There was a fun library I used for a while called Stampit that you might be interested in : https://github.com/stampit-org/stampit

Anyways I think we're mixing some paradigms here in a not so useful way just yet. 

And it's also quite possible that I am not seeing the benefit yet of this level of separation of concerns... Wanna pair a bit so you can show me around some more?

Let me know what you think!

Thanks!

axbarsan:
After discussing with @oponder, we decided to simplify the logic:
- Get rid of the Gateway layer
- Use services for handling clients and requests
- Convert services to functions
- Have a single service per API
- Inject clients from the action creators
- Mock service functions in tests

Additionally, to make them all work together:
- Extend clients to have clients specific for each API (to have bound URLs and settings)
- Create a "selector" for authentication tokens, which will get the token from the store and, if it's expired, renew it and return the new one

Updated the code, will update the diagram and the document.

Thank you for working together on this 👍

src/model/clients/HttpClient.js
```diff
@@ -0,0 +1,133 @@
...
+  DELETE: 'DELETE',
+};
+
+export class HttpClient {
+  static get(url, config) {
+    const boundConfig = Object.assign({}, config, {
+      url,
+      method: HttpRequestMethods.GET,
+    });
+    const newClient = new HttpClient(boundConfig);
```
ubergesundheit: Is it really required to instantiate a new HttpClient for each request?

ubergesundheit:
All in all I like this direction! I think it will make it easy to switch out implementations later when happa speaks to the control plane.

One comment about your choice of http client. In my previous projects I was really happy with using [`ky`](https://github.com/sindresorhus/ky). Its "better" than axios in being able to retry requests and being smaller meaning less data to transfer for users. (Also see [this tweet](https://twitter.com/sindresorhus/status/1037763588826398720) from the creator of `ky` where he explains the difference). I'm fine with axios though..

src/model/clients/HttpClient.js
ubergesundheit: these are only for one-shot requests right? Maybe these shoudn't be supported like this. My main concern would be to have too many floating around clients but I'm unsure if this will be a problem for browsers :)

axbarsan: The clients are declared inside function bodies, so they'll get marked for garbage collection once the function gets done executing (in this case when the promise resolves or rejects), so memory usage is not a real issue.
Both `ky` and `axios` do the same thing behind de scenes, it's not a big deal.

With the current implementation you could re-use one if you really wanted, but since the configuration of the request is stored as an instance property of the class, you'd have to reset it for every use.

Reusing would look like this:
```javascript
const client = new HttpClient({
  method: HttpRequestMethods.GET,
  url: requestURL,
});
client.setAuthorizationToken(authToken);

const firstResponse = await client.execute();

client.setRequestConfig({
  method: HttpRequestMethods.POST,
  url: requestURL,
  body: data,
});
client.setAuthorizationToken(authToken);

const secondResponse = await client.execute();

```

Please keep in mind that the static methods are factories for quickly executing a request, they are not the only way to use them.

We could have the client as a function, and then pass everything to it. And for the `GiantSwarmClient`, we would just create a new function and return the client function with a few bound values.
For me, there are 2 problems with that approach:
* The developer experience would not be as good as using instance methods for configuration
* The interface would not be the same for other clients (such as a WebSockets client - where we would need to keep a reference of the connection, so we could stop it on demand)

I'm not disregarding the idea of switching to a function, but please take the things above into consideration

ubergesundheit: Cool thanks for the detailed explanation :+1:

axbarsan:
Hi @ograu , thank you for taking a look at this.

The purpose of this refactor did not change since we last talked. This is only a replacement for the Giant Swarm client.
In the PR description there's a diagram showing the full architecture of happa, with the proposed changes. That is to show where the changes made in this PR will fit into our current system.

As of this PR, we're only adding Services and Clients, the rest remains the same.

Most of the proposed structure that we saw in the last standup has been converted to functional, now we're only using OOP in clients. I actually explained to @ubergesundheit why I believe OOP suits clients better than procedural/functional.

The services have been converted to different functions that will make the API calls. These functions get the desired client (such as the giant swarm client) as an argument so they'll be easier to test, because we could inject a mock client into them and test services separately from the client.

Please let me know if there's still something that doesn't seem right. We can sort it out here or at the standup, tomorrow 👍

ograu:
We are on the same page, then. 
I'm ok with everything, good job @axbarsan !

src/reducers/appReducer.js
```diff
@@ -41,25 +42,25 @@ const makeAppReducer = () => {
 
         return;
 
-      case types.LOGIN_SUCCESS:
-        // TODO This is a Side effect.
-        // Is there a better place for setUserToStorage()?
+      case types.LOGIN_SUCCESS: {
         setUserToStorage(action.userData);
```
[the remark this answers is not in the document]
axbarsan: I actually thought about this a couple of times, here's what my logic was:

Redux uses the event-driven pattern
To better understand this, think of events and events listeners in a web page.

When you click on something, a click event is dispatched in the browser.
If you want to listen to those events, you'll set up an event listener and you'll bind a callback to it.
So every time that event is dispatched, the callback you set up will be executed.
Now imagine you'd want to set up the same event callback for multiple event listeners.

It will probably look like this:
```javascript
function handleEvent(event) {
  switch (event.type) {
    case "mousemove":
      doSomething();
      break;

    case "click":
      doSomethingElse();
      break;
  }
}
```

Now what does that look like? (hint: redux reducer)

That being said, to do something when an event is dispatched, we'd do it in the callback, not in the dispatcher, right?

So as a parallel to that, redux does the same thing, we dispatch events, and the reducers act as event listeners. So it will only make sense to implement 3rd party functionality in the reducer, right?

Also, this is only valid because `setUserToStorage();` performs 3rd party functionality on the local storage, not on the redux store. If it was performing it on the redux store, then it would've been bad practice.

Hope this makes as much sense to you as it did to me 😬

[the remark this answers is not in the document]
ograu: > So as a parallel to that, redux does the same thing, we dispatch events, and the reducers act as event listeners. So it will only make sense to implement 3rd party functionality in the reducer, right?

I don't agree with this. 
As I said we can do it as we are doing here in this specific case. Redux is unopinionated about this. It's just that we decided to keep reducers simple are pure. It's considered a best practice. There's a lot of literature about this.

We use thunks and action creators for this kind of side effects.

> Also, this is only valid because setUserToStorage(); performs 3rd party functionality on the local storage, not on the redux store. If it was performing it on the redux store, then it would've been bad practice.
 
🤔 Still a side effect. We just want to pass data to reducers and write a minimal amount of logic if it's needed. But we want them to be pure.

Happy to discuss it tomorrow if it doesn't make sense to you. This is just one approach, maybe the more "functional" one, but as I said, Redux gives plenty of freedom when it comes to decide these kind of things.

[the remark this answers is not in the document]
axbarsan: By looking at how the pattern is supposed to be implemented, to me still makes more sense to do it in reducers.
But that's just my opionion, I'm willing to do whatever we decide on. The "recommended" redux way seems to be by using middleware, which kind of makes sense

PR #1146, opened 2020-02-19
Change organization tests to not use persisted nock mocks.

oponder:
Work in progress.

### DIFF
--- a/__tests__/Organization.js
+++ b/__tests__/Organization.js
@@ -11,6 +11,7 @@ import {
   AWSInfoResponse,
   generateRandomString,
   getMockCall,
+  getMockCallTimes,
   ORGANIZATION,
   orgResponse,
   orgsResponse,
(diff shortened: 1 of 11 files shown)


__tests__/Organization.js
```diff
@@ -33,302 +31,315 @@ afterAll(() => {
...
+  if (!nock.isDone()) {
+    console.error('Nock has pending mocks:', nock.pendingMocks());
+  }
+
   nock.cleanAll();
 });
 
-describe('Navigation', () => {
-  it('navigation has selected the right page when in organization list route', async () => {
+describe('', () => {
```
axbarsan: Why does this have to be in a separate `describe` block?

oponder: The beforeEach with all the `getMockCalls` applies to most but not all tests, so I wrapped them in a describe

oponder: Specifically, the organization deletion tests add an organization to the organization list response manually, in order to test deleting it.

Thinking about it a bit, I could change the test so that we remove `acme` instead of the randomly generated organization we insert into the response, but I didn't want to make too much of a change to the test logic itself.

axbarsan: Nevermind, I didn't see the indentation and I thought this describe block was separated from the others, and couldn't understand how it would work 😅

PR #1151, opened 2020-02-21
toggleChecked only returns function if it exists

ograu:
Fixes https://github.com/giantswarm/giantswarm/issues/9033

Just fixing it for now,  probably we should find an easier/clearer/simpler solution for this component now that it seems clear which use cases it has to cover, but not doing it because in the new HA Masters mockups we are not using label selectors in forms, so we can simplify this `AvailabilityZonesLabel` component when changing the views.


src/components/UI/AvailabilityZonesLabel.js
```diff
@@ -104,9 +104,17 @@ function AvailabilityZonesLabel({
...
+    // to change form state, and in cluster details view where we do not need to
+    // perform any action onClick.
+    // We only want to return onToggleChecked if it is passed.
+    if ((!isMaxReached || isChecked || isRadioButtons) && onToggleChecked) {
+      return onToggleChecked(!isChecked, { title, value, label });
     }
+
+    // Returning a function for consistency
+    // eslint-disable-next-line no-empty-function
+    return () => {};
```
axbarsan: Unless ‘onToggleChecked’ returns a function, this is does not really bring consistency.
Please keep in mind that in the case below this one, the function is returning the result of executing ‘onToggleChecked’, not the actual function

src/components/UI/AvailabilityZonesLabel.js
```diff
@@ -104,9 +104,17 @@ function AvailabilityZonesLabel({
...
 
   const toggleChecked = () => {
-    if (!isMaxReached || isChecked || isRadioButtons) {
-      onToggleChecked(!isChecked, { title, value, label });
+    // This component is used in forms, where we do pass an onToggleChecked function
+    // to change form state, and in cluster details view where we do not need to
+    // perform any action onClick.
+    // We only want to return onToggleChecked if it is passed.
+    if ((!isMaxReached || isChecked || isRadioButtons) && onToggleChecked) {
+      return onToggleChecked(!isChecked, { title, value, label });
```
axbarsan: You could get rid of that `&& onToggleChecked` and use `onToggleChecked?.(...)` 👍

src/components/UI/AvailabilityZonesLabel.js
ograu: Yeah, it works, but you are right, let me rewrite it. 👍

src/components/UI/AvailabilityZonesLabel.js
ograu: Didn't know about this! I will leave it as I have it now, ESLint is yelling at me

axbarsan: Unfortunatelly ESLint is still stupid about optional chaining, but you can disable the rule for the line if you want to use it 👍

ograu: Niice, thanks for keeping me updated about this new stuff! :-)

PR #1152, opened 2020-02-21
Fix 'cannot read property 'displayname' of null' error in dev.

oponder:
This fixes a `cannot read property 'displayname' of null` error for me in development.

Did anyone else get this all the time too?


axbarsan:
I suppose this started happening after the changes to fit the new ESLint config. Probably a class component that was converted to a functional one

oponder:
4W6a7sg

PR #1154, opened 2020-02-21
Add reset password tests

axbarsan:

`PasswordForgot` file has been renamed to `ForgotPassword`, but it shows up as a new file

* Add some fixing for the `PasswordField` component, which was not behaving correctly. (also fixed the tests that use it)
* Mock the delete auth token request in `Logout` test because it somehow started breaking??


src/components/ForgotPassword/SetPassword.js
```diff
@@ -308,7 +316,7 @@ class SetPassword extends React.Component {
           again for verification purposes.
         </p>
         <div className='textfield'>
-          <label>Email</label>
+          <label htmlFor='email'>Email</label>
```
oponder: 👍

__tests__/ForgotPassword.js
```diff
@@ -0,0 +1,319 @@
...
+    nock.enableNetConnect();
+  });
+
+  afterEach(() => {
+    if (!nock.isDone()) {
+      // eslint-disable-next-line no-console
+      console.error('Nock has pending mocks:', nock.pendingMocks());
+    }
+
+    expect(nock.isDone()).toBeTruthy();
```
ograu: I'm trying these approaches in the tests I am writing/modifying and I think that `afterEach` callback should be an `async` function, and this expectation should be _awaited_:

```javascript
await wait(() => expect(nock.isDone()).toBe(true));
```

Otherwise you can experience different results or race conditions depending on the tests that runs before.

axbarsan: Makes sense, I'll make the change 👍

__tests__/ForgotPassword.js
```diff
@@ -0,0 +1,320 @@
...
+    nock.enableNetConnect();
+  });
+
+  afterEach(async () => {
+    if (!nock.isDone()) {
+      // eslint-disable-next-line no-console
+      console.error('Nock has pending mocks:', nock.pendingMocks());
+    }
+
+    await wait(() => expect(nock.isDone()).toBeTruthy());
```
ograu: Sorry, I'm being late again.

One thing I am seeing now in the tests I am refactoring is that this pattern (console.error and then await fro the expect) will trigger a lot of false positives, it will log a lot of `console.error`s that shouldn't be logged, nor consiered as errors. 

If we expect here to all nocks to be done, a lot of async actions will appear in the _Nock has pendings mock: _ message, and then you might think (I did) that you are intercepting too many calls, but what happens is that these calls are not yet matched, which is something expected actually. 

The `console.error` makes sense when debugging, if you want to find out which calls you are intercepting but happa is not using. But this two snippets together doesn't make sense and can lead to confusion IMO.

axbarsan: Yeah, you're right. Actual checking for `isDone()` for printing pending mocks should be in the `wait` callback. Guess it's too late now 💣

ograu: It's ok like this, it works, it's just that we might get too many logs and that can be pretty confusing, especially when some tests are failing and you are debugging. Just something to be aware of.

PR #1164, opened 2020-02-24
Add a version picker component.

oponder:
To be implemented in the app creation and app editing forms.

But for now, this is just the component and its tests.


src/components/UI/StyleGuide.js
```diff
@@ -38,12 +39,56 @@ Header.propTypes = {
 
 // eslint-disable-next-line react/no-multi-comp
 const StyleGuide = () => {
+  const [selectedVersion, setVersion] = useState('1.0.5');
+
+  const handleSetVersion = version => {
```
ubergesundheit: Why not letting the VersionPicker component call `setVersion` directly?

oponder: This is like the callback that the VersionPicker exposes, that way whatever is implementing it can decide what to do when someone picks a version.

ubergesundheit: But is there a difference in
```
<VersionPicker
  onChange={setVersion}
  ... />
```
and
```<VersionPicker
  onChange={handleSetVersion}
  ... />
```

Right now it just adds another function which just calls the `setVersion` function withougt changing parameters or anything

oponder: Oh :) heh yeah. I guess I was just trying to show a pattern for the recommended implementation.

oponder:
versionpicker

src/components/UI/VersionPicker/VersionPicker.js
```diff
@@ -0,0 +1,220 @@
+import styled from '@emotion/styled';
+import PropTypes from 'prop-types';
+import React, { useState } from 'react';
+import DropdownMenu from 'UI/DropdownMenu';
+
+// AppVersionPicker takes a list of app versions and a currently selected version
```
axbarsan: I believe it would make more sense to have this comment right on top of the `AppVersionPicker` component.

Also, a topic for discussion would be switching to only using JSDoc style comments, instead of using a mix between JSDoc and Golang comments. But that's not in the scope of this PR

src/components/UI/VersionPicker/__tests__/VersionPicker.js
```diff
@@ -0,0 +1,124 @@
...
+
+import { fireEvent, render, within } from '@testing-library/react';
+import { ThemeProvider } from 'emotion-theming';
+import React from 'react';
+import theme from 'styles/theme';
+import AppVersionPicker from 'UI/VersionPicker/VersionPicker';
+
+it('renders without crashing', () => {
+  render(
+    <ThemeProvider theme={theme}>
```
axbarsan: You can use the `renderWithTheme()` helper in `testUtils/renderUtils` if you don't want to add `ThemeProvider` in every test 👍

src/components/UI/VersionPicker/VersionPicker.js
oponder: I'll bring this up in the next chapter sync. I'd think JSDoc style is the way to go. And i'll make a change here in this PR in the meanwhile.

PR #1165, opened 2020-02-24
Redirect when cluster id doesn't exist

ograu:
Realized while writing tests that Happa just redirects if the cluster is in store, otherwise it just breaks.


src/components/Cluster/ClusterDetail/ClusterDetail.js
```diff
@@ -1,38 +1,57 @@
...
+    new FlashMessage(
+      `Cluster <code>${props.clusterId}</code> doesn't exist.`,
+      messageType.INFO,
+      messageTTL.MEDIUM
+    );
+
+    dispatch(push(AppRoutes.Home));
+
+    return null;
+  }
```
ubergesundheit: I think this should be inside an `useEffect` hook like this:

```js
useEffect(() => {
  if (!props.cluster) {
    new FlashMessage(
      `Cluster <code>${props.clusterId}</code> doesn't exist.`,
      messageType.INFO,
      messageTTL.MEDIUM
    );

    dispatch(push(AppRoutes.Home));
},[]) // empty dependencies to make it behave like componentDidMount
```

src/selectors/clusterSelectors.js
```diff
@@ -123,6 +123,9 @@ export const selectAndProduceAZGridTemplateAreas = () =>
 export const selectCanClusterUpgrade = (state, clusterID, targetVersion) => {
   const cluster = state.entities.clusters.items[clusterID];
 
+  if (!cluster) return;
```
axbarsan: I believe this should at least return `null` if nothing is found

From ECMAScript2015 spec:

> **4.3.10 undefined value**
> primitive value used when a variable has not been assigned a value
>
> **4.3.12 null value**
> primitive value that represents the intentional absence of any object value

That was my thinking when I introduced that linter rule

src/components/Cluster/ClusterDetail/ClusterDetail.js
ograu: `useEffect` runs after every render, and I want an early return ¿why wait to render when we can redirect before?

src/selectors/clusterSelectors.js
ograu: I am thinking that if we have to return somethiung here, maybe we should return `false` as `canClusterUpgrade` returns a boolean

src/components/Cluster/ClusterDetail/ClusterDetail.js
ubergesundheit: Good point, with `useEffect` the conditional would be executed only once which saves the component from evaluating the if clause on every render.

Maybe checking this here is not the right way to go anyway?

ograu: Maybe :-) but, ¿Why? And where would you place the check?

ubergesundheit: After some thinking I couldn't find a better place ;D

My comment about `useEffect` still stands. When used with an empty array as dependencies, the function inside gets executed only once.

See https://reactjs.org/docs/hooks-reference.html#conditionally-firing-an-effect

ograu: If I were to add a `useEffect` I would have to check before (after in the code, before in time) in render that the prop exist, otherwise it throws an error because the render goes first, so the condition will be evaluated in every render as well.

So my conclusion is that there is nothing we get from adding the `useEffect` ¿Or maybe am I missing something?

ubergesundheit: oh yeah you're right. I didn't read the rest of the component 😮

PR #1167, opened 2020-02-25
Added a missing word

glitchcrab:
I spotted a missing word so I added it - now the sentence makes sense (unlike this one).


glitchcrab:
Not too sure about label selection - enhancement makes the most sense I think.

PR #1168, opened 2020-02-25
Add integration tests for the Users page and modals

axbarsan:

### DIFF
--- a/__tests__/Users.js
+++ b/__tests__/Users.js
@@ -182,14 +182,16 @@ describe('Users', () => {
 
     const { findByText, getByText } = renderRouteWithStore(UsersRoutes.Home);
 
-    const selectedEmailCell = await findByText(/developer4@giantswarm.io/i);
+    const selectedEmailCell = await findByText(new RegExp(desiredEmail, 'i'));
     let deleteButton = within(selectedEmailCell.parentNode).getByText(
       /delete/i
     );
     fireEvent.click(deleteButton);
 
     expect(
-      getByText(/Are you sure you want to delete developer4@giantswarm.io?/i)
+      getByText(
+        new RegExp(`Are you sure you want to delete ${desiredEmail}?`, 'i')
+      )
     ).toBeInTheDocument();
 
     deleteButton = getByText(/delete user/i);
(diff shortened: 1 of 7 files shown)


__tests__/Users.js
```diff
@@ -0,0 +1,237 @@
...
+
+describe('Users', () => {
+  beforeAll(() => {
+    nock.disableNetConnect();
+  });
+
+  afterAll(() => {
+    nock.enableNetConnect();
+  });
+
```
ograu: I like this approach of placing the mock call inside a `describe` so we can have different calls defined for each test suite, but ¿Is there a reason for placing `nock.disableNetConnect()` and `nock.enaableNetConnect()` here? Don't you think we want to have these at top level or even in a global a config file?

__tests__/Users.js
```diff
@@ -0,0 +1,237 @@
...
+        // eslint-disable-next-line no-console
+        console.error('Nock has pending mocks:', nock.pendingMocks());
+      }
+      expect(isDone).toBeTruthy();
+    });
+
+    nock.cleanAll();
+
+    forceRemoveAll();
+  });
```
ograu: This will trigger unnecessary `console.error`s ?

__tests__/Users.js
```diff
@@ -0,0 +1,237 @@
...
+    getMockCall(`/v4/organizations/${ORGANIZATION}/credentials/`);
+    getMockCall(`/v4/organizations/giantswarm/credentials/`);
+    getMockCall('/v4/appcatalogs/', appCatalogsResponse);
+
+    nock(global.config.passageEndpoint)
+      .get('/invites/')
+      .reply(StatusCodes.Ok, invitesResponse);
+  });
+
+  // Stop persisting responses
```
ograu: We can remove this comment here

__tests__/Users.js
ograu: I am using this:

```javascript
afterEach(async () => {
  await wait(() => expect(nock.isDone()).toBe(true));
  nock.cleanAll();
  forceRemoveAll();
});
```

ograu: I think this will result in unnecessary `console.error`s.
I am using this:

```javascript
afterEach(async () => {
  await wait(() => expect(nock.isDone()).toBe(true));
  nock.cleanAll();
  forceRemoveAll();
});
```

__tests__/Users.js
axbarsan: The `describe` block doesn't do anything functionality wise. Jest isolates every file from the others automatically. I only use `describe` blocks just because they give a better output while running tests.

I added `nock.disableNetConnect();` and `nock.enableNetConnect();` because they give us consistent results on both our local machines and in CI. Having them in a global config is a good idea, but I believe we should move them after we finish refactoring the old tests, since they will probably break some of the tests.

We already had them in a couple of tests, so I didn't introduce them now:
image

ograu: Yeah, we can do it after we refactor all tests 👍 

> The `describe` block doesn't do anything functionality wise. Jest isolates every file from the others automatically. I only use `describe` blocks just because they give a better output while running tests.

I thought that it made a difference, check it out: https://jest-bot.github.io/jest/docs/api.html#aftereachfn

axbarsan: Sorry, I meant that in this case the describe block doesn't do anything, because it wraps all the tests in the file. So having the `beforeAll` call outside the `describe` block would be the same thing as having it inside it

PR #1169, opened 2020-02-25
Error handling in cluster creation views

ograu:
Fixes https://github.com/giantswarm/giantswarm/issues/9088

This fixes some quirks in v4/v5 creation form: https://gigantic.slack.com/archives/CQ465CFG8/p1582631553009400 . Basically, no error message and an endless spinner.

Now it will show a flash message and the actual error message instead the buttons:

Captura de pantalla 2020-02-25 a las 17 17 06

This is a quick fix. IMO it's not ideal and we should think how do we want to translate these kind of errors into meaningful pieces of UI. Maybe adding a button for retry would be a better option.

### DIFF
--- a/src/actions/clusterActions.js
+++ b/src/actions/clusterActions.js
@@ -273,11 +273,9 @@ export function clusterCreate(cluster, isV5Cluster) {
     try {
       dispatch({ type: types.CLUSTER_CREATE_REQUEST });
 
-      const method = isV5Cluster
-        ? 'addClusterV5WithHttpInfo'
-        : 'addClusterWithHttpInfo';
-
-      const data = await clustersApi[method](cluster);
+      const data = isV5Cluster
+        ? await clustersApi.addClusterV5WithHttpInfo(cluster)
+        : await clustersApi.addClusterWithHttpInfo(cluster);
 
       const location = data.response.headers.location;
       if (typeof location === 'undefined') {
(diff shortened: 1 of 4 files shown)


src/actions/clusterActions.js
```diff
@@ -269,57 +269,61 @@ function clusterLoadStatus(clusterId, { withLoadingFlags }) {
...
+      dispatch({ type: types.CLUSTER_CREATE_REQUEST });
 
-    return clustersApi[method](cluster)
-      .then(data => {
-        const location = data.response.headers.location;
-        if (typeof location === 'undefined') {
-          throw new Error('Did not get a location header back.');
-        }
+      const method = isV5Cluster
+        ? 'addClusterV5WithHttpInfo'
```
axbarsan: I know you didn't introduce this in the PR, but can you please change these to something like:
```javascript
const method = isV5Cluster
  ? clustersApi.addClusterV5WithHttpInfo
  : clustersApi.addClusterWithHttpInfo
```

Having the function names as strings makes it impossible for our code editors to check if we used the wrong names.
One wrong letter and it will break cluster creation for everyone...

ograu: > I know you didn't introduce this in the PR, but can you please change these to something like:

Actually I think I did! 😂 
Yeah, and this is more straigthforward, changing it 👍

PR #1170, opened 2020-02-26
Ellipsis in members list and as a composable

ograu:
Fixes https://github.com/giantswarm/giantswarm/issues/9083 in organization view.

Before:

image

After:

image

And adds `Ellipsis` as a composable styles that you can throw anywhere you want this behaviour as long as its parent is a `block` element and has `white-space: no-wrap`


marians:
Thanks @ograu !!!

PR #1172, opened 2020-02-26
Cluster name input autofocus and autoselect

ograu:
Fixes https://github.com/giantswarm/giantswarm/issues/7801#issuecomment-590897544

### DIFF
--- a/src/components/Cluster/NewCluster/CreateNodePoolsCluster.js
+++ b/src/components/Cluster/NewCluster/CreateNodePoolsCluster.js
@@ -223,7 +223,10 @@ class CreateNodePoolsCluster extends Component {
     },
   };
 
+  clusterNameInput = React.createRef();
+
   componentDidMount() {
+    this.clusterNameInput.current.select();
     this.isValid();
   }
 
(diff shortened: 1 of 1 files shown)


src/components/Cluster/NewCluster/CreateNodePoolsCluster.js
```diff
@@ -223,7 +223,10 @@ class CreateNodePoolsCluster extends Component {
     },
   };
 
+  clusterNameInput = React.createRef();
+
   componentDidMount() {
+    this.clusterNameInput.current.select();
```
axbarsan: `current` can be null, let’s keep it safe 😬

```suggestion
    this.clusterNameInput.current?.select();
```

ograu: haha it's so cheap to use it!  But is there a situation when `render()` will not render the input? I mean, if there's an error `ComponentDidMount()` doesn't run...

axbarsan: Well refs get created before `componentDidMount()` executes, so we're safe on that. My only concern is forgetting about the ref while refactoring this, or while replacing the `input` element with a custom component and not passing the ref, and  then spamming the `exceptions` channel again 😆 

It is not necessary, but it wouldn't hurt to have a safety net. I got used to checking if the ref is there from using strict TypeScript, here's the definition for the ref object:
```typescript
interface RefObject<T> {
  readonly current: T | null;
}
```

ograu: Interesting, thanks, I think I will start using this conditional chaining in lots of places 🙂

PR #1173, opened 2020-02-26
Try to catch the api endpoint that is triggering the error

ograu:


src/components/Cluster/ClusterDetail/KeyPairs.js
```diff
@@ -79,9 +79,13 @@ class KeyPairs extends React.Component {
...
   componentDidMount() {
-    this.apiEndpointHostname = new URL(
-      this.props.cluster.api_endpoint
-    ).hostname;
+    try {
+      this.apiEndpointHostname = new URL(
+        this.props.cluster.api_endpoint
+      ).hostname;
+    } catch (error) {
+      throw Error(`Api endpoint: ${this.props.cluster.api_endpoint}`);
```
ograu: Will this be logged in exceptions channel? ping @oponder

oponder: I think it will!

oponder: But you can also manually send messages even when there is no exception condition.

I can make an example if you want, but it be possible if you can get a handle on the exception notification instance.

ograu: Ok, thanks, for now we just want to log it when it throws an error, but sometimes it can be very useful to log stuff 👍

PR #1174, opened 2020-02-27
Increase timeout to 20 seconds.

oponder:
Not sure if this is the right approach to fix our problems. But making the PR to have a place to discuss and decide.

- Did we want to increase the timeout for all installations?
- Do we want timeouts to be configurable per installations?

or

- Do we want to possibly even reduce the timeout and implement retries for certain requests?

### DIFF
--- a/src/components/Layout.js
+++ b/src/components/Layout.js
@@ -33,7 +33,7 @@ import Users from './Users/Users';
 
 const defaultClient = GiantSwarm.ApiClient.instance;
 defaultClient.basePath = window.config.apiEndpoint;
-defaultClient.timeout = 10000;
+defaultClient.timeout = 20000;
 const defaultClientAuth =
   defaultClient.authentications['AuthorizationHeaderToken'];
 


ubergesundheit:
A configurable timeout per installation would be really awesome in my opinion. Generally we should reduce the time users have to wait for anything (be it an error or the success of loading something)

axbarsan:
I'd go with different timeouts for each installations, as well. Retries would have to be configured selectively for each request (as in some would use it, and some not - to not create the same object a couple of times, for example), which would be pretty cumbersome to implement

ograu:
+1 for the configurable timeout for installation.

Regarding retry buttons, I think we should think in fallbacks for recurring situations and I think a retry button could be a nice solution in some scenarios, ie when cluster creation fails.

marians:
Retrying is a IMO a huge topic I'd like to avoid now. A default timeout plus the option to override it per installation config would sound much better as a first step.

oponder:
closing in favour of an implementation that takes values from the installations repo.

PR #1177, opened 2020-02-27
Make the whole catalog repo clickable

axbarsan:
Fixes giantswarm/giantswarm#8817

- Removed the `Browse Apps` button
- Wrapped the external links in the description with an `object` tag (quick hack to make nested links work)
- Exported the component to another file, refactored and cleaned it up

**Preview**

* Idle state
image

* Hovered state
image


oponder:
I think I like your hover state more than my fuzzy shadowy one and we should probably apply it also to the app hover state on the next page for some consistency.

axbarsan:
> I think I like your hover state more than my fuzzy shadowy one and we should probably apply it also to the app hover state on the next page for some consistency.

+1 for that

Also added to apps

**Preview**

* Idle state:
image

* Hovered state:
image

PR #1179, opened 2020-02-28
Bake in a static releases list for dev only.

oponder:
Hey guys,

What do you think of this? It bakes the release endpoint response as a config parameter in `index.html`, which `releaseAction.js` then checks for. `start.sh` removes it in production so that everything works as normal there.


src/actions/releaseActions.js
```diff
@@ -8,8 +8,20 @@ export function loadReleases() {
...
-      .getReleases()
+    let getReleaseMethod = releasesApi.getReleases.bind(releasesApi);
+
+    // Do not make an API call to the releases endpoint if we have
+    // config.staticReleases set. This is to ease development. "start.sh" removes
+    // the staticReleases in production.
+    if (window.config.staticReleases) {
+      getReleaseMethod = () => {
+        return Promise.new(resolve => {
+          resolve(window.config.staticReleases);
```
axbarsan: You can shorten this to
```javascript
return Promise.resolve(window.config.staticReleases);
```

src/index.html
```diff
@@ -30,6 +30,7 @@
...
... 400MB."},{"component":"coredns","description":"Fix HPA manifests for Kubernetes 1.16."},{"component":"external-dns","description":"Updated external-dns app version to v0.5.18."},{"component":"kiam","description":"Updated kiam app version to v3.5."}],"components":[{"name":"app-operator","version":"1.0.0"},{"name":"aws-operator","version":"8.1.1"},{"name":"calico","version":"3.10.1"},{"name":"cert-operator","version":"0.1.0"},{"name":"cluster-operator","version":"2.1.1"},{"name":"containerlinux","version":"2247.6.0"},{"name":"docker","version":"18.06.1"},{"name":"etcd","version":"3.3.17"},{"name":"kubernetes","version":"1.16.3"},{"name":"vault","version":"0.10.3"},{"name":"cert-exporter","version":"1.2.1"},{"name":"cert-manager","version":"0.9.0"},{"name":"chart-operator","version":"0.11.4"},{"name":"cluster-autoscaler","version":"1.16.2"},{"name":"coredns","version":"1.6.5"},{"name":"external-dns","version":"0.5.18"},{"name":"kiam","version":"3.5.0"},{"name":"kube-state-metrics","version":"1.9.0"},{"name":"metrics-server","version":"0.3.3"},{"name":"net-exporter","version":"1.5.1"},{"name":"node-exporter","version":"0.18.1"}],"timestamp":"2020-02-05T12:00:00Z","version":"11.0.1"}],
```
marians: Can we make that easier to read and maintain? At some point we will have to remove a release and add a new one, I'm sure.

oponder: hmm.. I was thinking what we should do generally is just copy the full release endpoint output from a valid installation, and not manage it too much manually.

oponder:
Small problem, still can't create clusters though, because the `api` has its own idea of what releases are there and uses that to validate the cluster creation. So we end up with 

`{/go/src/github.com/giantswarm/api/server/endpoint/v5/cluster/creator/endpoint.go:121: release version '11.0.1' does not exist: invalid request error}`

I guess if we have to bake in a dev response for the releases endpoint somewhere it should be in the `api` 🤔

oponder:
I'll close this for now.. it doesn't really solve the problem unfortunately. At lets us get to the cluster create screen, but not actually create any clusters.

PR #1180, opened 2020-02-28
Clean up automated tests

axbarsan:
Hopefully this will fix all the weird test behaviour

* Moved all the common test behaviour that we would normally just copy and paste around to a global config
* You can now enable seeing which pending requests hang in which tests by uncommenting some lines in the global test config

### DIFF
--- a/src/components/SignUp/SignUp.js
+++ b/src/components/SignUp/SignUp.js
@@ -17,6 +17,8 @@ import TermsOfService from './TermsOfService';
 const passage = new Passage({ endpoint: window.config.passageEndpoint });
 
 export class SignUp extends React.Component {
+  isComponentMounted = false;
+
   state = {
     statusMessage: 'verify_started',
     // eslint-disable-next-line react/no-unused-state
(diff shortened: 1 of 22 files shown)


src/components/SignUp/SignUp.js
```diff
@@ -68,11 +71,13 @@ export class SignUp extends React.Component {
         });
 
         setTimeout(() => {
-          this.setState({
-            statusMessage: 'enter_password',
-          });
+          if (this.isComponentMounted) {
```
axbarsan: Introduced this because the timeout callback could get executed after the component was unmounted, which was throwing an error in one of the tests

testUtils/setupTests.js
```diff
@@ -0,0 +1,27 @@
...
+
+afterAll(() => {
+  nock.enableNetConnect();
+});
+
+afterEach(async () => {
+  await wait(() => {
+    const done = nock.isDone();
+
+    // Uncomment the next lines to debug hanging requests
```
oponder: Cool!

PR #1181, opened 2020-02-28
Get started button tests

oponder:
Towards: https://github.com/giantswarm/happa/issues/759

We're already indirectly testing that the get started button shows up in the Getting Started tests. I added a test now to check that it goes away when the cluster is older than 30 days.

Also refactored to non-persisted nock calls.

### DIFF
--- a/__tests__/GettingStarted.js
+++ b/__tests__/GettingStarted.js
@@ -26,8 +26,10 @@ beforeEach(() => {
   getMockCall(`/v4/organizations/${ORGANIZATION}/`, orgResponse);
   getMockCall(`/v4/organizations/${ORGANIZATION}/credentials/`);
   getMockCall('/v4/appcatalogs/', appCatalogsResponse);
-  // prettier-ignore
-  getMockCall(`/v4/clusters/${V4_CLUSTER.id}/status/`,v4AWSClusterStatusResponse);
+  getMockCall(
+    `/v4/clusters/${V4_CLUSTER.id}/status/`,
+    v4AWSClusterStatusResponse
+  );
 });
 
 afterEach(async () => {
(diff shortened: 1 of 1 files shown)


__tests__/GettingStarted.js
```diff
@@ -19,56 +18,36 @@ import {
...
+
+  await wait(() => expect(nock.isDone()).toBeTruthy());
+
+  nock.cleanAll();
 });
 
 /************ TESTS ************/
 
 it('lets me get there from the dashboard and go through the pages', async () => {
+  // prettier-ignore
```
axbarsan: Why is this needed?

__tests__/GettingStarted.js
```diff
@@ -93,3 +72,22 @@ it('lets me get there from the dashboard and go through the pages', async () =>
...
+it('the get started button does not show up if the cluster is older than 30 days', async () => {
+  const futureDate = new Date();
+  futureDate.setMonth(futureDate.getMonth() + 2);
+
+  const clusterCreateDate2MonthsLater = futureDate.toISOString();
+
+  // prettier-ignore
+  getMockCall('/v4/clusters/', [Object.assign({}, v4ClustersResponse[0], {create_date: clusterCreateDate2MonthsLater})]);
+  // prettier-ignore
+  getMockCall(`/v4/clusters/${V4_CLUSTER.id}/`,Object.assign({}, v4AWSClusterResponse, {create_date: clusterCreateDate2MonthsLater,}));
```
axbarsan: Can you please extract these to separate variables?

__tests__/GettingStarted.js
oponder: I like the kind of single scan line that forms from those getMockCall's being in a nice line... when prettier starts wrapping it all I get a bit sad.

But I can remove it and let prettier just do it's thing.

PR #1195, opened 2020-03-02
Add support for 'giantswarm' authorization schemes

axbarsan:


src/selectors/authSelectors.js
```diff
@@ -1,22 +1,26 @@
...
 
-    return currentToken;
+    return [currentToken, scheme];
   } catch (err) {
-    const newErr = Object.assign({}, err, { status: StatusCodes.Unauthorized });
+    const newErr = Object.assign({}, err, {
+      status: StatusCodes.Unauthorized,
+    });
 
     throw newErr;
```
ograu: Do we need to throw an error here?

axbarsan: Yes, otherwise we wouldn't know that authorization failed. We always wrap these promises into a try/catch block inside actions, so it's not a problem 👍

ograu: I think we should log these error in their respective reducers and not let them propagate throught the app. We get nothing from this.

ograu: But anyway, probably this is another discussion, approving this

axbarsan: image

This is used inside of action creators, by not rejecting the promise, we don't know that the authorization has failed, therefore it will not propagate to the error reducer. Not rejecting the promise will cause the promise to hang, and we'll be stuck with an infinite loading screen.

ograu: I have another mental model, here is mine: when control jumps to the `catch` block, the promise has already been rejected, so no need to throw.

Also we decide how and when loading screens are rendered, and we are using the dispatch for this, so no need to throw, right?

axbarsan: Please check this out:
https://repl.it/@axbarsan/Promise-reject

The promise is resolved with undefined when we catch the promise early and don't reject again

ograu: I think my mental model is not wrong.

This happens because you have a `try/catch`block in the `HttpClient` `execute` methid and then you are again doing `try/catch` in thunks which is a bit crazy IMO and not needed at all. If we are catching API errors at thunks level ¿Why are we catching them in `HttpClient` also?  

I think we should only do this in thunks, and then we don't need to throw error once they are catched, cause they will end up catched somewhere else and break something.  Instead what we want to do is catch them -> log them in respective reducer ->  tell the user what happened.

axbarsan: The reason for catching here https://github.com/giantswarm/happa/blob/174934bb6a357aa8fa8cf67778bbb766b99cdc69/src/model/clients/HttpClient.js#L129
is to reject with the actual response from the server, not the axios object, because we want to have an universal interface for our clients, to be able to easily swap HTTP libraries.

The reason for catching in the files pointed by you, is that auth0 rejects with a response with code `200`,  if the authorization fails. So I'm catching it, changing the response code to `401`  and rejecting again, so we could be able to display the error properly.

The only reason for catching in these situations, is to reject with a different response. As far as I know, there's no other way of doing this, but maybe we're talking about different things here.

ograu: I will dig a bit into it when I have time and share my findings wiht you. I think that we are missing something here.

Thanks!

PR #1196, opened 2020-03-02
Error handling only in thunks

ograu:
Related to https://github.com/giantswarm/happa/pull/1195


ograu:
What about this @axbarsan ?

src/model/clients/HttpClient.js
```diff
@@ -112,21 +112,15 @@ export class HttpClient {
...
-        data,
-      });
-
-      return response.data;
-    } catch (err) {
-      return Promise.reject(err.response.data);
-    }
+    await this.onBeforeRequest(this.requestConfig);
+
+    return axios({
```
axbarsan: By removing the try/catch block, it means that this promise will resolve/reject with objects created by `axios`, which means that every file which uses `HttpClient` or any class that extends it will depend on `axios`.

when `axios()` resolves, it resolves with an object of type `AxiosResponse` [(see type)](https://github.com/axios/axios/blob/521444513969a08ec5ef943c41ba0812845ed4f9/index.d.ts#L73), and it rejects with an object of type `AxiosError` [(see type)](https://github.com/axios/axios/blob/521444513969a08ec5ef943c41ba0812845ed4f9/index.d.ts#L82).

The original's try/catch block's purpose is to resolve/reject the promise with the response that we get from the server, so no files will actually depend on `axios`, except for this one.

This is a breaking change, the try/catch block is needed in the current implementation.

ograu: Ok, I see... then we could do the same you were doing but without returning another resolved promise

axbarsan: What do you mean? Can you please provide an example?

ograu: Pushed already

src/model/clients/HttpClient.js
```diff
@@ -114,13 +114,21 @@ export class HttpClient {
...
+        timeout,
+        headers,
+        url,
+        method,
+        data,
+      });
+
+      return response.data;
+    } catch (err) {
+      return err.response.data;
```
axbarsan: Please beware that this will return a resolved promise, not a rejected one.

ograu: And this is what I don't get... AFAIK when a promise return an error it is automatically rejected, right?

Another consideration would be that here we will probably need to return the whole response if we want to chech statuses in thunks, so we will have either to depend a little bit on axios or keep providing abstractions for whatever use cases we want.

axbarsan: Check this out:
https://repl.it/@axbarsan/failed-promises
Because we are catching the failed promise, it is no longer rejected.  If we return in the `catch` block, it is the same as returning in the `try` block. The error is already caught, so the promise will resolve.

ograu: True.

I prefer my first proposal, but if you don't want to use `.data` in actions, which I find reasonable, then we can throw here and catch later:

```javascript
try {
// ...
} catch (err) {
      throw Error(err.response.data);
    }
```

But I think we should be throwing `err.response` so we pass the full error to action creators to do whatever check we want to do there.

axbarsan: To be able to use response codes and what else in actions, and not depend on `axios`, can you please create a custom response class, that we can pass around and use?

oponder:
Just wanna say I'm gonna review this and form an opinion by tomorrow. Thanks!

oponder:
I'm a bit out of the loop on what the problem is... maybe this is one for a session where we identify the things we want to achieve and then discuss approaches for achieving it together. 

We could call it "Exceptions and Error handling"

Sounds like an idea?

ograu:
Yeah, sure. I just wanted to avoid rethrowing in catch blocks. And yeah, things are not as easy as I assumed them to be.

My points are:

1. I would like to find a better solution if we need to reject a promise than throwing an error that we have already catched and stored in state. Any reactions to this error in app should be done using state. It is true that react now is quite resilient to errors catched in `render()` but I already have removed some `throw` statements in `catch` blocks that were literally breaking the app, meaning a blue page with nothing rendered in it.

2. We need to find a way with the new http client to make errors fully available in thunks, otherwise no way to save relevant error data to state, or for instance, to check status in thunks.

And also, we should decide which data we want to have available in store regarding errors. There’s job to be done in that regards, cause now we are logging the whole error object in some situations, which is something we should probably avoid.

axbarsan:
I opened a PR based on this, to better support server responses:
https://github.com/giantswarm/happa/pull/1226

ograu:
> I opened a PR based on this, to offer better support for server responses:
> #1226

Then, I'm closing this one.

PR #1200, opened 2020-03-03
Chart version table

oponder:
This adds a list of chart versions and what app version they will install to the app detail page.

Related design discussion: https://github.com/giantswarm/giantswarm/issues/9185


oponder:
It looks like this: 

Screenshot 2020-03-03 at 11 50 10 PM

src/components/UI/AppDetails/AppDetails.js
```diff
@@ -88,8 +95,62 @@ const Install = styled.div`
...
+
+  td {
+    vertical-align: top;
+    code {
+      ${Ellipsis}
+      max-width: 150px;
+      display: inline-block;
+    }
+
+    .copyable {
```
axbarsan: This references a className that is assigned in a different component. If the `Copyable` component changes its structure, this will not work in the same way.

Would you consider extending `Copyable`?
Ex:
```javascript
const StyledCopyable = styled(Copyable)`
  ...styles
`
```

And then using it as normal: `<StyledCopyable />`.

(Please note that in order to do this, the `className` prop must be able to be passed to the root element inside the `Copyable` component, so passing the `className` prop to the root `div` inside `Copyable` should do the job)

oponder: The thing is I want to target a copyable inside a certain other element, and not the other copyables.

I could make two different StyleCopyables, but is there a better way?

I was looking at this but I can't seem to make it work: https://emotion.sh/docs/styled#targeting-another-emotion-component

axbarsan: Yes, that is actually a good idea, and it should work, since we added the babel transformer for emotion.

Don't forget to pass the `className` prop, so it could work
Just need to do this:
image

oponder: Somehow can't get it to work, will go for two styled `Copyable`s instead.

The error I got was `invalid hook call`, so it's like it's trying to render the component somehow.

axbarsan: Ahh yes, forgot about this. To make it work:
- Pass the `className` property to the root component of `Copyable`
- First, define a styled copy of `Copyable`
```javascript
const StyledCopyable = styled(Copyable)``
```

And then target that one inside other styled components
```javascript
${Copyable} {
  ...styles
}
```

oponder: It works! w00t

PR #1203, opened 2020-03-04
Write tests and comments for the new model layer

axbarsan:


src/model/clients/__tests__/HttpClient.js
```diff
@@ -0,0 +1,336 @@
...
+        TestHeader2: 'test value',
+        TestHeader3: 'test value',
+      },
+    };
+
+    try {
+      await HttpClient.get({ url: '/api/test' }, config);
+    } catch (err) {
+      expect(err).toStrictEqual(
+        `This is embarrassing, we couldn't execute this request. Please try again in a few moments.`
```
oponder: what's the error code in this case? is it a 404 or something like that? Or is it that there is no server at all to talk to?

oponder:
Oh this was a draft.. sorry for the early approval :P

axbarsan:
> Oh this was a draft.. sorry for the early approval :P

No worries man, thank you! I'll ask for your review again once I finish it

src/model/clients/__tests__/HttpClient.js
axbarsan: There's no code here, this is just if you set up the client the wrong way

```javascript
// Wrong
await HttpClient.get({ url: '/api/test' }, config);

// Correct
await HttpClient.get('/api/test', config);
```

axbarsan:
The beauty of JSDoc comments
image

src/selectors/__tests__/authSelectors.js
```diff
@@ -0,0 +1,114 @@
...
+    });
+
+    it(`doesn't renew the token if the authentication scheme is 'giantswarm'`, async () => {
+      helpers.isJwtExpired.mockReturnValue(true);
+
+      const state = {
+        app: {
+          loggedInUser: {
+            email: 'developer@giantswarm.io',
+            auth: { scheme: 'giantswarm', token: 'an-expired-token' },
```
oponder: technically giantswarm tokens never expire. I'm not sure about the implementation in the selectAuthToken in this case.. if the scheme is giantswarm, we don't even need to ever call isJwtExpired. The `giantswarm` auth token is not even a jwt token :) 

But that might not be in scope of this PR.

axbarsan: `isJwtExpired` is not actually called. I made the test that way to be sure that even if the `isJwtExpired` is setup to return `true`, it would still work as expected for `giantswarm` auth tokens.

I will change the token, it is indeed a bit confusing

PR #1204, opened 2020-03-04
Upgrade notice in clusters list

ograu:
Towards https://github.com/giantswarm/giantswarm/issues/9095

Upgrade notice is now shown in home cluster list.

image

Still to be added in list in organiation detail view.

### DIFF
--- a/src/components/Cluster/ClusterDetail/ClusterDetail.js
+++ b/src/components/Cluster/ClusterDetail/ClusterDetail.js
@@ -5,8 +5,11 @@ import React from 'react';
 import { Breadcrumb } from 'react-breadcrumbs';
 import { connect, useDispatch } from 'react-redux';
 import { Redirect, Route, Switch } from 'react-router-dom';
-import { selectCanClusterUpgrade } from 'selectors/clusterSelectors';
-import cmp from 'semver-compare';
+import {
+  selectCanClusterUpgrade,
+  selectTargetRelease,
+} from 'selectors/clusterSelectors';
+// import cmp from 'semver-compare';
 import { AppRoutes } from 'shared/constants/routes';
 import _ from 'underscore';
 
@@ -77,44 +80,11 @@ ClusterDetail.propTypes = {
 function mapStateToProps(state, ownProps) {
   const clusterID = ownProps.match.params.clusterId;
   const cluster = state.entities.clusters.items[clusterID];
-  // eslint-disable-next-line init-declarations
-  let release;
-  // eslint-disable-next-line init-declarations
-  let targetReleaseVersion;
-  let isV5Cluster = false;
-
-  if (cluster) {
-    if (cluster.release_version && cluster.release_version !== '') {
-      release = state.entities.releases.items[cluster.release_version];
-    }
-
-    const activeReleases = _.filter(state.entities.releases.items, x => {
-      return x.active;
-    });
-
-    const availableVersions = activeReleases.map(x => x.version).sort(cmp);
-
-    // Guard against the release version of this cluster not being in the /v4/releases/
-    // response.
-    // This will ensure that Happa can calculate the target version for upgrade
-    // correctly.
-    if (availableVersions.indexOf(cluster.release_version) === -1) {
-      availableVersions.push(cluster.release_version);
-      availableVersions.sort(cmp);
-    }
-
-    if (
-      availableVersions.length >
-      availableVersions.indexOf(cluster.release_version)
-    ) {
-      targetReleaseVersion =
-        availableVersions[
-          availableVersions.indexOf(cluster.release_version) + 1
-        ];
-    }
-
-    isV5Cluster = state.entities.clusters.v5Clusters.includes(cluster.id);
-  }
+  const release = cluster
+    ? state.entities.releases.items[cluster.release_version]
+    : null;
+  const isV5Cluster = state.entities.clusters.v5Clusters.includes(clusterID);
+  const targetReleaseVersion = selectTargetRelease(state, cluster);
 
   return {
     credentials: state.entities.credentials,
@@ -125,11 +95,7 @@ function mapStateToProps(state, ownProps) {
     nodePools: state.entities.nodePools.items,
     provider: state.app.info.general.provider,
     release: release,
-    canClusterUpgrade: selectCanClusterUpgrade(
-      state,
-      clusterID,
-      targetReleaseVersion
-    ),
+    canClusterUpgrade: selectCanClusterUpgrade(state, clusterID),
     targetRelease: state.entities.releases.items[targetReleaseVersion],
     user: state.app.loggedInUser,
     region: state.app.info.general.datacenter,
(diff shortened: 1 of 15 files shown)


src/components/Cluster/ClusterDetail/ClusterDetail.js
```diff
@@ -5,8 +5,11 @@ import React from 'react';
 import { Breadcrumb } from 'react-breadcrumbs';
 import { connect, useDispatch } from 'react-redux';
 import { Redirect, Route, Switch } from 'react-router-dom';
-import { selectCanClusterUpgrade } from 'selectors/clusterSelectors';
-import cmp from 'semver-compare';
+import {
+  selectCanClusterUpgrade,
+  selectTargetRelease,
+} from 'selectors/clusterSelectors';
+// import cmp from 'semver-compare';
```
axbarsan: I think we can remove this

src/styles/index.js
```diff
@@ -109,3 +109,27 @@ export const Dot = styled.span`
 export const FallbackSpan = styled.span`
   opacity: 0.5;
 `;
+
+export const UpgradeNoticeStyles = ({
```
axbarsan: Is there any reason why this is a mixin and not a component?

ograu: Wanted to decouple styles and markup/functionality to avoid what has happenend with AZLabels.

axbarsan: But it's being used like a component. The separate `text-decoration` values can be implemented by extending the component

ograu: Yes it can, but then we need to pass a function and we need to add this function to the component and check that is not null, etc.

I prefer this approach because is the simpler by now. And I'm not done BTW.

axbarsan: I'm not sure I follow... extending is pretty easy with styled components

@oponder  had a similar situation in this file, where he extended the component (Check the `Copyable` component):
https://github.com/giantswarm/happa/blob/master/src/components/UI/AppDetails/ChartVersionsTable.js

ograu: I wasn't following you either  😆 🤦‍♂ So I was doing a mixin instead of and extend because my initial idea was to have a bunch of styles to throw anywhere we need them, then I noticed that I had to customize one property and did it this way, just because sass. Any special reason.

Do we have guidelines about this?

ograu: or opinions?

axbarsan: We haven't documented this yet.

I believe we should only use mixins when they can make reusability easier across different contexts, and when they can be composed with other styles to create components. Ex: a `CenteredFlexContainer` mixin would return the styles:
```css
display: flex;
justify-content: center;
align-items: center;
```

ograu: Can we have this discussion in an a dedicated PR instead? Or in `fmt` repo? So we can mege this

src/components/Home/UpgradeNotice.js
```diff
@@ -0,0 +1,45 @@
...
+import { UpgradeNoticeStyles } from 'styles';
+import LoadingOverlay from 'UI/LoadingOverlay';
+
+const UpgradeWrapperDiv = styled.div`
+  display: inline-block;
+  ${UpgradeNoticeStyles({ textDecoration: 'none' })}
+`;
+
+// This component receiveis a cluster id, finds if this cluster is 'upgradable' and
+// in case it is, outputs an uograde notice,
```
oponder:
```suggestion
// This component receives a cluster id, finds if this cluster is 'upgradable' and
// in case it is, outputs an upgrade notice.
```

src/components/Cluster/ClusterDetail/RegionAndVersions.js
```diff
@@ -85,12 +87,9 @@ function RegionAndVersions({
       </div>
       {canUpgrade && (
         <a className='upgrade-available' onClick={showUpgradeModal}>
-          <Upgrade>
-            <span>
-              <i className='fa fa-warning' />
-              Upgrade available
-            </span>
-          </Upgrade>
+          <UpgradeWrapperDiv>
```
oponder: Any reason we can't use the new component here? Instead making a new styled component and injecting the styles?

src/styles/index.js
oponder: Yeah I think there is an easier to follow way that I also recently learned on how to deal with these cases.

If you wanna pair on a refactor Im happy to, then we can also make a fmt PR.

I think we can take the time to 'get it right' in this PR though. Or is this change is urgent?

src/styles/index.js
```diff
@@ -109,3 +109,27 @@ export const Dot = styled.span`
...
+    display: unset;
+    font-size: 16px;
+    font-weight: 300;
+    &:hover {
+      text-decoration: ${textDecoration};
+    }
+  }
+  &:before {
+    display: inline-block;
+    content: '\f019';
```
oponder: Is this the warning icon?

This is kinda cool, but at the same time also more cryptic to know what it is.

The span with `fa fa-warning` was pretty clear what it does... maybe we should extract that into a 'warning icon' component, or maybe add some comments here about what `\f019` means.

axbarsan: I believe that's the unicode character code of the icon in the font-awesome font. As you'd have `\u26be` for a ⚾️emoji, for example.

There's a cheatsheet here
https://fontawesome.com/cheatsheet

ograu:
Changed the approach. Now using a component that outputs the notice with or without the onClick function depending of if it is passed. It's not something that I love, but it's the best approach I could think of for now.

So no css content and no mixins, just a regular react component.

Waiting for your comments.

src/components/Home/UpgradeNotice.js
```diff
@@ -7,22 +7,45 @@ import {
...
   selectLoadingFlagByAction,
 } from 'selectors/clusterSelectors';
-import { UpgradeNoticeStyles } from 'styles';
 import LoadingOverlay from 'UI/LoadingOverlay';
 
 const UpgradeWrapperDiv = styled.div`
   display: inline-block;
-  ${UpgradeNoticeStyles({ textDecoration: 'none' })}
+  color: ${props => props.theme.colors.gold};
+  cursor: ${({ hasFunction }) => (hasFunction ? 'pointer' : 'inherit')};
```
axbarsan: Why not use the `onClick` prop for this check?

ograu: good catch!

src/components/Home/UpgradeNotice.js
```diff
@@ -0,0 +1,69 @@
...
+      </UpgradeWrapperDiv>
+    </LoadingOverlay>
+  );
+}
+
+UpgradeNotice.propTypes = {
+  clusterId: PropTypes.string,
+  canClusterUpgrade: PropTypes.bool,
+  loadingReleases: PropTypes.bool,
+  showUpgradeModal: PropTypes.func,
```
oponder: I think this prop could just be called "onClick". This component doesn't / shouldn't know that its job is to show a modal. It just has to execute a function when it is clicked.

PR #1205, opened 2020-03-04
error.response.status instead error.status

ograu:
It looks like we are looking into the wrong property?

Just want to check with you before change this everywhere in case I am missing something.


oponder:
Is it not working right now with `error.status` ? That's weird... I'd have to check the actual error object to tell you whether this is right or wrong, but I assume you noticed it wasn't working as intended?

ograu:
I noticed that the `error.status` when a cluster is not found is actually `401` whilst `error.response.status` is `404`. The browser identifies it as a `404` so I just assumed that we have to check this one instead.

oponder:
I think it's worth it to try to figure out why that 401 is in error.status. Possibly it has to do with the auth token refreshing?

ograu:
but then the network tab will show a 401, while when I saw it yesterday it was loggued as a 404

oponder:
I did a bit of digging because I happened to be here in a different PR. 

You can recreate this codepath by removing the `/v5/clusters/${V5_CLUSTER.id}/` nock from the `keyPairs` test for example.

When I do that I get this:

```
  console.error node_modules/@testing-library/react/dist/act-compat.js:52
    Error in batchedLayout TypeError: Cannot read property 'status' of undefined
        at /Users/oponder/code/giantswarm/happa/src/actions/clusterActions.js:191:26
        at process._tickCallback (internal/process/next_tick.js:68:7)
```

When I investigated the error object I did see that `error.status` was `404` though, but `error.response` was undefined.

Maybe it means there is a difference in the errors when nock is doing the request handling, but I think that would be unlikely, since it's still the same client. It would be worth it to test that, but would need a bit of time to create a situation in the live API where the cluster is in the list, but the cluster detail request returns a 404.

But right now I'm a but unsure about approving the PR.

ograu:
Closing because can't reproduce this error anymore.

ograu:
Keep working on it because I think that this is the source of half of the exceptions we are having. 

Do you think this check I've added makes sense? Do you think it is somehow related to this? : https://github.com/giantswarm/api/blob/c692177cc698a99d65f9e7dcc7fb8cd7f178c89a/server/endpoint/v5/cluster/searcher/endpoint.go#L102

src/actions/clusterActions.js
```diff
@@ -186,7 +186,10 @@ export function clusterLoadDetails(
 
       return cluster;
     } catch (error) {
-      if (error.status === StatusCodes.NotFound) {
+      if (
+        error.status === StatusCodes.NotFound ||
+        error.status === StatusCodes.Unathorized
```
axbarsan: There's no need for the double check, because in the snippet you sent, the API returns a 404 response even if the user is unauthorized.
It is explained why in the comment there
```
// Returning 404 NOT FOUND on unauthorized request is common
// practice and aligns with v4 implementation as well.
```

ograu: Yes, but check this...

> I noticed that the `error.status` when a cluster is not found is actually `401` whilst `error.response.status` is `404`. The browser identifies it as a `404` so I just assumed that we have to check this one instead.

This happens when a cluster is deleted and Happa tries to refresh the clusters list, when I found this in the API I couldn't help to think that this could be somehow related. Maye it is related with the auth token refresh instead, as Oliver is suggesting.

Also, removing a cluster when the user is not authorized and the cluster is in the store, makes sense. We should obviously change the message, but what happens after this check is valid for a `401 `, so what I am proposing is: let's merge this as an temporary fix (or temporary test) and see if we get less _x is undefined_ exceptions in the cluster list.

ograu: BTW if someone wants to work on this, you can reproduce the error simply by creating a cluster, then deleting it and just leave the cluster list opened for a while.

axbarsan: The server returns a `404` error, so the problem is on the client-side
image

And IMO, the status code change happens in 2 places, and because of the structure of the response object, this  change can only happen in `/src/lib/giantswarmClientPatcher.js`, inside the catch call, where we change the status code to `401`

ograu: Makes sense! 

Should we merge this so we can get rid of these errors? or you prefer to leave it like this for now?

ograu:
Not merging, closing it because I think my initial approach was the correct one, opened this new PR: https://github.com/giantswarm/happa/pull/1249

PR #1206, opened 2020-03-05
Fix triggering of update cluster modal

ograu:
Removed the conditional because we are not doing the check here. 

Not sure what is cheaper: checking if the cluster can update or rendering the modal.


ograu:
Yeah, let's keep it simple for now? Or should I add the conditional again?

axbarsan:
> Yeah, let's keep it simple for now? Or should I add the conditional again?

Let's keep it simple 👍

PR #1207, opened 2020-03-05
Clusters are created in the currently selected organization, instead of a picked organization

axbarsan:
Right now, we have the following scenario:
* Go to the organizations page
* Click on an organization that is not your currently selected one
* Press on the `Create Cluster` button
* Create a cluster

The new cluster will not be created in the organization that you clicked on, but on the currently selected organization (in the top dropdown)

This PR makes clusters be created in the organization that you clicked on 👍


axbarsan:
> I thought that this was a desired behaviour though.

It's pretty weird though.
You'll visit `https://happa.g8s.ghost.westeurope.azure.gigantic.io/organizations/Adrian/clusters/new` thinking you'd add a cluster into that organization. But you'll actually not...

PR #1208, opened 2020-03-05
Upgrade available icon in org details

ograu:
Towards giantswarm/giantswarm#9095

Now upgrade notice in clusters list. 
I don't know which is the best way to do it. As the table is quite cluttered, I decided to just add the icon, the text is shown when the icon is hovered. It is not responsive, yet:

UpgradeAvailable


axbarsan:
Looking good 👍 What do you think about using a tooltip for showing the descriptive text, instead of expanding the label?

ograu:
I thought about it. 

I followed this approach for convenience, but if you prefer a tooltip, I'm ok with this as well  👍

marians:
How about showing the icon next to the release version number?

For me, the position outside the table is a bit strange and looks more like a general warning regarding this cluster. The proximity to the version number could help to frame it better. Using a tooltip would help in that position, too.

src/components/Organizations/Detail/View.js
```diff
@@ -24,6 +27,14 @@ const MembersTable = styled.div`
...
 `;
 
+const UpgradeNoticeWrapperDiv = styled.div`
+  cursor: pointer;
+  width: 22px;
+  span {
+    display: none !important;
+  }
+`;
+
```
ograu: If I wrap the component into an styled one `styled(UpgradeNotice)` for some reason the tooltip doesn't shoew up...

src/components/Organizations/Detail/Detail.js
```diff
@@ -69,9 +68,9 @@ function mapStateToProps(state, ownProps) {
   const allClusters = state.entities.clusters.items;
   let clusters = [];
 
-  clusters = _.filter(allClusters, cluster => {
-    return cluster.owner === ownProps.match.params.orgId;
-  });
+  clusters = Object.keys(allClusters)
+    .filter(id => allClusters[id].owner === ownProps.match.params.orgId)
+    .map(id => allClusters[id]);
```
axbarsan:
```suggestion
  clusters = Object.values(allClusters)
    .filter(cluster => cluster.owner === ownProps.match.params.orgId);
```

src/components/Organizations/Detail/View.js
```diff
@@ -86,6 +97,11 @@ class OrganizationDetail extends React.Component {
           return cmp(a, b);
         },
       },
+      {
+        dataField: 'path',
+        text: '',
+        formatter: upgradeNoticeIcon.bind(this),
```
axbarsan: Can you please use an arrow function here, so we could get rid of using `this` inside a function?

ograu: I think it won't work as the function is declared outside the class

ograu:
This is the hovered state now:

Captura de pantalla 2020-03-09 a las 9 47 33

I have alligned releases to right so the are closer to upgrade available icons.

src/components/Organizations/Detail/View.js
axbarsan: What I'm suggesting is to reach an implementation similarto:
```javascript
  formatter: (_, cluster) =>  upgradeNoticeIcon(cluster, ...someOtherProps)
```
+ modifying `upgradeNoticeIcon` accordingly

So we wouldn't have to use `this` inside the `upgradeNoticeIcon`.

marians:
> I have alligned releases to right so the are closer to upgrade available icons.

Good thinking. For my eyes, it could even be closer, so that the distance between version number and icon is much smaller than the distance between icon and the "CREATED" content.

src/components/Organizations/Detail/View.js
```diff
@@ -264,12 +265,12 @@ function clusterIDCellFormatter(cell) {
 }
 
 // eslint-disable-next-line react/no-multi-comp
-function upgradeNoticeIcon(_, cluster) {
+function upgradeNoticeIcon(_, cluster, orgId) {
   const clusterDetailPath = RoutePath.createUsablePath(
     OrganizationsRoutes.Clusters.Detail,
     {
       // eslint-disable-next-line react/no-this-in-sfc
```
axbarsan:
```suggestion
```

src/index.html
```diff
@@ -28,20 +28,20 @@
...
-        azureCapabilitiesJSON: '{"Standard_A2_v2":{"additionalProperties":{},"description":"This is some description","maxDataDiskCount":4,"memoryInMb":4294.967296,"name":"Standard_A2_v2","numberOfCores":2,"osDiskSizeInMb":1047552,"resourceDiskSizeInMb":21474.83648},"Standard_A4_v2":{"additionalProperties":{},"description":"Here is a longer description that might be too long for the field","maxDataDiskCount":8,"memoryInMb":8589.934592,"name":"Standard_A4_v2","numberOfCores":4,"osDiskSizeInMb":1047552,"resourceDiskSizeInMb":42949.67296},"Standard_A8_v2":{"additionalProperties":{},"description":"Another VM size description text","maxDataDiskCount":16,"memoryInMb":17179.869184,"name":"Standard_A8_v2","numberOfCores":8,"osDiskSizeInMb":1047552,"resourceDiskSizeInMb":85899.34592}, "Standard_D2s_v3":{"additionalProperties":{},"description":"Dsv3-series, general purpose, 160-190 ACU, premium storage supported","maxDataDiskCount":4,"memoryInMb":8589.935,"name":"Standard_D2s_v3","numberOfCores":2,"osDiskSizeInMb":1047552,"resourceDiskSizeInMb":17179.87}}',
-        happaVersion: 'development',
-      };
+  // apiEndpoint: "https://api.g8s.gauss.eu-central-1.aws.gigantic.io",
```
axbarsan: 🙈

ograu:
image

marians:
Late here, but LGTM too! Thanks @ograu !

PR #1210, opened 2020-03-05
Refactor InstallAppForm to remove memory leak

axbarsan:
Refactor the form abilities logic inside of `InstallAppForm` to get rid of memory leak. `updateNamespace()` was ran continuously for the `nginx-ingress-controller-app` app.

This was also the problem that started giving us errors in the tests.


axbarsan:
> My bad! Thank you for finding the fix.

No worries, I should've found it in the PR...

PR #1212, opened 2020-03-06
Refactor the app detail modal in preparation of editing app version.

oponder:
I've done a lot of refactors on the app detail modal, would like to get those reviewed individually instead of also adding the editing app version implementation into the mix.

There is a placeholder edit app version pane, but no way to get there yet (the button has been commented out). In another PR I continue with this.


src/components/Cluster/ClusterDetail/AppDetailsModal/EditChartVersionPane.js
```diff
@@ -0,0 +1,7 @@
+const EditChartVersionPane = () => {
+  return 'EditChartVersionPane';
```
oponder: This is the placeholder pane.

src/components/Cluster/ClusterDetail/AppDetailsModal/InitialPane.js
```diff
@@ -0,0 +1,147 @@
...
+        </div>
+
+        <div className='labelvaluepair'>
+          <div className='labelvaluepair--label'>CHART VERSION</div>
+          <div className='labelvaluepair--value code'>
+            {/* eslint-disable-next-line no-magic-numbers */}
+            <span>{truncate(props.app.spec.version, 20)}</span>
+          </div>
+          {/* <br />
+          <Button onClick={props.showEditChartVersionPane}>
```
oponder: Commented out the way to get to the new placeholder pane for now.

src/components/Cluster/ClusterDetail/AppDetailsModal/AppDetailsModal.js
```diff
@@ -0,0 +1,254 @@
...
+  deleteAppSecret: 'deleteAppSecret',
+  deleteApp: 'deleteApp',
+  editChartVersion: 'editChartVersion',
+  initial: 'initial',
+};
+
+const AppDetailsModal = props => {
+  const [pane, setPane] = useState(modalPanes.initial);
+
+  if (props.app === null || typeof props.app === 'undefined') {
```
axbarsan:
```suggestion
  if (!props.app) {
```

src/components/Cluster/ClusterDetail/AppDetailsModal/AppDetailsModal.js
```diff
@@ -0,0 +1,254 @@
...
+    showPane(modalPanes.initial)();
+    props.onClose();
+  }
+
+  async function loadAppsAndClose() {
+    await props.dispatch(loadApps(clusterId));
+    onClose();
+  }
+
+  async function _deleteAppConfig() {
```
axbarsan: Why the `_` at the beginning of these functions' names? 😬

src/components/Cluster/ClusterDetail/AppDetailsModal/AppDetailsModal.js
```diff
@@ -0,0 +1,254 @@
...
+    done();
+  }
+
+  async function _updateAppSecret(values, done) {
+    await props.dispatch(updateAppSecret(appName, clusterId, values));
+    await loadAppsAndClose();
+    done();
+  }
+
+  function deleteConfirmFooter(cta, onConfirm) {
```
axbarsan: Can you please extract this to a different file?

src/components/Cluster/ClusterDetail/AppDetailsModal/AppDetailsModal.js
```diff
@@ -0,0 +1,254 @@
...
+        </>
+      );
+
+      modalBody = (
+        <>
+          Are you sure you want to delete {props.app.metadata.name}&nbsp; on{' '}
+          <ClusterIDLabel clusterID={props.clusterId} />?
+          <br />
+          <br />
+          There is no undo.
```
axbarsan: Scary stuff 🎃

src/components/Cluster/ClusterDetail/ClusterApps.js
```diff
@@ -10,7 +10,7 @@ import { AppCatalogRoutes } from 'shared/constants/routes';
 import Button from 'UI/Button';
 import ClusterDetailPreinstalledApp from 'UI/ClusterDetailPreinstalledApp';
 
-import AppDetailsModal from './AppDetailsModal';
+import AppDetailsModal from './AppDetailsModal/AppDetailsModal';
```
axbarsan: I think it would be sensible to rename `AppDetailsModal.js` to `index.js`, to be able to import it like before:
`import AppDetailsModal from './AppDetailsModal'`

src/components/Cluster/ClusterDetail/AppDetailsModal/AppDetailsModal.js
oponder: They clash otherwise with the imported action names above.

src/components/Cluster/ClusterDetail/ClusterApps.js
oponder: I'd be ok with that, but we agreed at some point to stop doing it: https://github.com/giantswarm/fmt/blob/master/javascript/general_style_guide.md#naming

axbarsan: I was unaware of that. A workaround would be creating an `index.js` file that just exports `AppDetailsModal` 😅

oponder: Hmm, not a big concern for me right now. I'd rather have less files and noise to be honest.

src/components/Cluster/ClusterDetail/AppDetailsModal/AppDetailsModal.js
oponder: Right now it's a little jsx fragment that is repeated otherwise in a few places, I don't feel like it warrants being in it's own file / component. Do you feel strongly about it?

oponder: Also resisting because then I end up with more lines due to how prettier does the new lines.

So it goes from this one liner:

```
modalFooter = deleteConfirmFooter('Delete App', _deleteApp);
```

to

```
modalFooter =  <DeleteConfirmFooter
        cta={Delete App}
        onConfirm={_deleteApp}
        onCancel={showPane(modalPanes.initial)}
/>
```

oponder: I could keep the function and extract the jsx fragment if you think it makes things better.

```
  function deleteConfirmFooter(cta, onConfirm) {
    return (
      <DeleteConfirmFooter
        cta={cta}
        onConfirm={onConfirm}
        onCancel={showPane(modalPanes.initial)}
      />
    );
  }
```

src/components/Cluster/ClusterDetail/AppDetailsModal/AppDetailsModal.js
axbarsan: Can we rename them, maybe? Or if not, creating aliases for the actions by adding the `Action` suffix to them?
Ex.:

```javascript
import { deleteAppConfig as deleteAppConfigAction } from "..."
```

src/components/Cluster/ClusterDetail/AppDetailsModal/AppDetailsModal.js
axbarsan: I like to have components in separate files, but not forcing it if you want to do otherwise. 
But I believe we should at least extract `deleteConfirmFooter()` from the main function body, because we're basically creating a new component with every update. If it was separate, react will be able to reuse it

axbarsan: I saw now that you extracted it. The comment below is no longer applicable 👍

PR #1223, opened 2020-03-09
Refactor keypair test and fix overloaded action type.

oponder:
This refactors the Keypair tests to not use persisted mocks and uncovered a little bug when a cluster returns 404. 

We fire off a `CLUSTER_DELETE_REQUEST` action to try to remove the cluster form the store, but that also triggers the cluster delete modal to show up.  The modal reducer was already using `CLUSTER_DELETE_REQUEST` as its trigger to show the delete modal.

This was the cause of the occasional error below:

```
TypeError: Cannot read property 'id' of undefined
 at id (webpack:///./src/components/Modals/Modals.js:325:66)
```


__tests__/KeyPairs.js
```diff
@@ -25,47 +26,27 @@ import { renderRouteWithStore } from 'testUtils/renderUtils';
...
+  getMockCall('/v4/appcatalogs/', appCatalogsResponse);
+  getMockCall(`/v4/clusters/${V5_CLUSTER.id}/key-pairs/`);
 });
 
 // Stop persisting responses
-afterAll(() => {
+afterEach(() => {
   Object.keys(requests).forEach(req => {
     requests[req].persist(false);
   });
```
axbarsan: We don't need this anymore 😬 
```suggestion
```

__tests__/KeyPairs.js
```diff
@@ -109,3 +90,6 @@ it('lets me open and close the keypair create modal', async () => {
   // Then the modal should be gone.
   expect(modal).not.toBeInTheDocument();
 });
+
+// it.only('lets me create a keypair', async () => {
+// });
```
axbarsan: I think you forgot this here

oponder: Ah yeah, this PR started as something else :)  thx

src/actions/actionTypes.js
```diff
@@ -75,6 +75,7 @@ export const CLUSTER_DELETE_REQUEST = 'CLUSTER_DELETE_REQUEST';
 export const CLUSTER_DELETE_CONFIRMED = 'CLUSTER_DELETE_CONFIRMED';
 export const CLUSTER_DELETE_SUCCESS = 'CLUSTER_DELETE_SUCCESS';
 export const CLUSTER_DELETE_ERROR = 'CLUSTER_DELETE_ERROR';
+export const CLUSTER_REMOVE_FROM_STORE = 'CLUSTER_REMOVE_FROM_STORE';
```
oponder: fyi @ograu I introduced `CLUSTER_REMOVE_FROM_STORE ` because `CLUSTER_DELETE_REQUEST` is actually the trigger to show the cluster delete modal.

We might have to rename `CLUSTER_DELETE_REQUEST` to align it with what the `_REQUEST` actions actually mean now, but for now this was my short path to fixing this bug.

ograu: 👍

PR #1226, opened 2020-03-09
Create custom response object

axbarsan:
Related to https://github.com/giantswarm/happa/pull/1196

This introduces custom response objects.

This allows us to check status codes and other properties that were in the actual response, in the redux actions, while also making switching `axios` to another client really simple.

### DIFF
--- a/src/model/clients/__tests__/HttpClient.js
+++ b/src/model/clients/__tests__/HttpClient.js
@@ -113,7 +113,7 @@ describe('HttpClient', () => {
     const client = new HttpClient(config);
     const result = await client.execute();
 
-    expect(result).toStrictEqual({
+    expect(result.data).toStrictEqual({
       testResponse: 'test value',
     });
   });
@@ -147,7 +147,7 @@ describe('HttpClient', () => {
 
     expect(beforeReqHook).toHaveBeenCalledWith(config);
 
-    expect(result).toStrictEqual({
+    expect(result.data).toStrictEqual({
       testResponse: 'test value',
     });
   });
@@ -171,7 +171,7 @@ describe('HttpClient', () => {
     };
     const result = await HttpClient.get('/api/test', config);
 
-    expect(result).toStrictEqual({
+    expect(result.data).toStrictEqual({
       testResponse: 'test value',
     });
   });
@@ -200,7 +200,7 @@ describe('HttpClient', () => {
     };
     const result = await HttpClient.post('/api/test', config);
 
-    expect(result).toStrictEqual({
+    expect(result.data).toStrictEqual({
       testResponse: 'test value',
     });
   });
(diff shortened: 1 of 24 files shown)


src/model/clients/__tests__/HttpClient.js
```diff
@@ -282,12 +282,12 @@ describe('HttpClient', () => {
...
     const result = await HttpClient.delete('/api/test', config);
 
-    expect(result).toStrictEqual({
+    expect(result.data).toStrictEqual({
       testResponse: 'test value',
     });
   });
 
-  it('provides the response body as an error if the request fails', async () => {
+  it('provides the an error if the request fails', async () => {
```
oponder:
```suggestion
  it('provides an error if the request fails', async () => {
```

axbarsan: Thank you 😬

ograu:
Great job! 

Just wanted to say that I find difficult to follow things like this:

```javascript
getInstallationInfo.mockResolvedValueOnce(mockAPIResponse(AWSInfoResponse));
```

But I think the methods naming is great, so if you are ok, I am ok.

axbarsan:
> Great job!
> 
> Just wanted to say that I find difficult to follow things like this:
> 
> ```js
> getInstallationInfo.mockResolvedValueOnce(mockAPIResponse(AWSInfoResponse));
> ```
> 
> But I think the methods naming is great, so if you are ok, I am ok.

The problem is that we expect a response of type `GenericResponse` out of `getInstallationInfo()`. So instead of creating a custom implementation in every test, I decided to create the helper.

However, we could wrap `mockResolvedValueOnce` and use the `mockAPIResponse(getInstallationInfo, AWSInfoResponse)`. What do you think about that?

oponder:
How about wrapping `AWSInfoResponse` with `mockAPIResponse` at the source? So before it even gets imported in `testUtils/mockHttpCalls` ?

axbarsan:
> How about wrapping `AWSInfoResponse` with `mockAPIResponse` at the source? So before it even gets imported in `testUtils/mockHttpCalls` ?

Do you mean having mocks in the actual model? Or having separate methods for mocking each response? (Like having `mockAWSInfoResponse()`)

oponder:
I mean like here: 

https://github.com/giantswarm/happa/blob/e43cba3ad7c658815b5b41a2226bfae45dc191bb/testUtils/mockHttpCalls/info.js#L3

We could instead do: 

```
export const AWSInfoResponse = mockAPIResponse({
  general: {
   ...
  },
});
```

Is that too naive?

axbarsan:
> I mean like here:
> 
> https://github.com/giantswarm/happa/blob/e43cba3ad7c658815b5b41a2226bfae45dc191bb/testUtils/mockHttpCalls/info.js#L3
> 
> We could instead do:
> 
> ```
> export const AWSInfoResponse = mockAPIResponse({
>   general: {
>    ...
>   },
> });
> ```
> 
> Is that too naive?

That's actually a very good idea, I'll do it 👍Thank you!

Teamwork 💪

PR #1227, opened 2020-03-10
Add a way to change the app version of an installed app.

oponder:
This turns the 'Chart Version' on the App Detail modal into a drop down from which a user can pick a different version, leading to a confirmation dialogue with some important warning info.

There was some good UX notes given, I've applied some of the easier / quicker to apply ones in this PR already. For the other changes we discussed, I'd like to go back to the speccing phase and make another PR to sort that out properly if possible.

versionpicker


src/actions/appActions.js
```diff
@@ -228,3 +228,50 @@ export function deleteApp(appName, clusterID) {
       });
   };
 }
+
+/**
+ * Takes an app and a cluster id and some values and tries to update the app to those values. Dispatches CLUSTER_UPDATE_APP_SUCCESS
+ * on success or CLUSTER_UPDATE_APP_ERROR on error.
```
axbarsan:
```suggestion
 * Takes an app and a cluster id and some values and tries to update the app to those values. Dispatches `CLUSTER_UPDATE_APP_SUCCESS`
 * on success or `CLUSTER_UPDATE_APP_ERROR` on error.
```

src/actions/appActions.js
```diff
@@ -228,3 +228,50 @@ export function deleteApp(appName, clusterID) {
...
+ *
+ * @param {Object} appName The name of the app
+ * @param {Object} clusterID Where to delete the app.
```
axbarsan:
```suggestion
 * @param {string} appName The name of the app
 * @param {string} clusterID Where to delete the app.
```

src/actions/appActions.js
```diff
@@ -228,3 +228,50 @@ export function deleteApp(appName, clusterID) {
...
+    const isV5Cluster = v5Clusters.includes(clusterID);
+
+    let modifyApp = appsApi.modifyClusterAppV4.bind(appsApi);
+
+    if (isV5Cluster) {
+      modifyApp = appsApi.modifyClusterAppV5.bind(appsApi);
+    }
+
+    return modifyApp(clusterID, appName, { body: values })
+      .then(() => {
```
axbarsan: Not sure if we actually agreed to only using `async/await` at some point.

src/components/Cluster/ClusterDetail/ClusterApps.js
```diff
@@ -417,20 +411,22 @@ class ClusterApps extends React.Component {
...
-          onClose={this.hideAppModal}
-          visible={this.state.appDetailsModal.visible}
-        />
+        {this.state.appDetailsModal.appName && (
+          <AppDetailsModal
+            // Instead of just assigning the selected app to the state of this component,
+            // this ensures any updates to the apps continue to flow down into the modal.
+            app={
+              this.props.installedApps &&
+              this.props.installedApps.find(
```
axbarsan:
```suggestion
              this.props.installedApps?.find(
```

axbarsan:
Nothing crazy, just a few pointers 👍

src/actions/appActions.js
oponder: Yeah we agreed not to! But I'm kinda copy pasting from other actions in this file. I'll make a todo to refactor this.

oponder:
The failing test reveals to me that there are some error cases that I haven't covered very well.

I'll try to address that in this PR still.

oponder:
Actually, I'll do my fixes in separate PR's so that we can discuss them individually.

PR #1229, opened 2020-03-11
Error message above button in cluster creation when no releases

ograu:
Towards https://github.com/giantswarm/giantswarm/issues/6862

What about replacing the _Create Cluster_ button with a message? 

cc/ @marians 

image

BTW this is an example of how easy and flexible it can be to react to errors in UIs if we have errors in store IMHO.


ograu:
Another approach (probably better approach) would be to completely hide the form.

axbarsan:
I'd go for displaying the error, but also displaying the button as disabled, rather than hiding it. I believe it's a more common behaviour.

As for reacting to errors, pretty cool 👍

ograu:
Will fix tests once we agreed on what to show.

src/components/Cluster/NewCluster/CreateRegularCluster.js
```diff
@@ -681,10 +691,8 @@ function mapStateToProps(state) {
 
   return {
     ...propsToPush,
-    clusterCreateError: selectErrorByAction(
-      state,
-      actionTypes.CLUSTER_CREATE_REQUEST
-    ),
+    clusterCreateError: selectErrorByAction(state, CLUSTER_CREATE_REQUEST),
+    releasesLoadError: selectErrorByAction(state, RELEASES_LOAD_REQUEST),
```
oponder: 🤯
👍 that's very cool

oponder:
I think keeping the disabled button would be nice.

The message could explain why the button is disabled: "It is not possible to create a cluster at this time. There are no active releases available for this platform."

Instead of saying "The cluster can't be created ... ", which is more something you'd say after someone already tried to create a cluster.

ograu:
This is how it looks now:

image

Also, I have modified `ErrorFallback` to accept a `className` prop so we can customize styles using `styled`

PR #1234, opened 2020-03-12
Make user-installed app list auto refresh

axbarsan:
Fixes https://github.com/giantswarm/giantswarm/issues/8819

* Make apps animate in/out when added or removed
* Export user installed apps to a different component
* Refactor installed apps as styled components

### Previews (Low quality)

**Adding an app**

captured

**Removing an app**

captured (1)


oponder:
Based on the bug fixed in https://github.com/giantswarm/happa/pull/1235/, what happens if the AppDetailModal is open while an app is removed due to an auto refresh?

In my PR, I didn't make the AppDetailModal resilient against not having an actual app to show.

I think we'll run into a similar bug.

axbarsan:
> Based on the bug fixed in #1235, what happens if the AppDetailModal is open while an app is removed due to an auto refresh?
> 
> In my PR, I didn't make the AppDetailModal resilient against not having an actual app to show.
> 
> I think we'll run into a similar bug.

Good catch! It blows up 😬 
Will push a fix in a few minutes

axbarsan:
So now, if the app gets deleted while the detail modal is open, the modal will close automatically and a flash message saying "The app you were looking at was deleted from the cluster by someone else." will be displayed 👍

PR #1237, opened 2020-03-13
Split cluster resources component

ograu:
In this PR I am doing the following:

- Splitting resources component in the clusters list into two components: one for v4 clusters and the other for v5 clusters
- Extract number of nodes into its own component.

More verbose, but I think I have improved the readibility and maintainibility here, basically because now we have less conditionals in the markup, so it's easier to follow IMHO.


src/components/Home/ClusterDashboardResourcesV4.js
```diff
@@ -0,0 +1,78 @@
...
+        <ClusterDashboardNodes
+          numberOfNodes={numberOfNodes}
+          createDate={cluster.create_date}
+        />
+      </RefreshableLabel>
+      {cluster.kvm && (
+        <span>
+          <Dot style={{ paddingLeft: 0 }} />
+          <RefreshableLabel value={storage}>
+            <span>{storage ? storage : '0'} GB storage</span>
```
axbarsan:
```suggestion
            <span>{storage ?? '0'} GB storage</span>
```

src/components/Home/ClusterDashboardResourcesV5.js
```diff
@@ -0,0 +1,94 @@
...
+          <span>{cluster.nodePools.length} node pools, </span>
+        </RefreshableLabel>
+      )}
+      <RefreshableLabel value={numberOfNodes}>
+        <ClusterDashboardNodes
+          numberOfNodes={numberOfNodes}
+          createDate={cluster.create_date}
+        />
+      </RefreshableLabel>
+      {numberOfNodes !== 0 && hasNodePools && (
```
axbarsan: Since we're doing these 2 checks together in both cases, I think it would be sensible to add the `numberOfNodes !==` check to `hasNodePools`. What do you think?

axbarsan:
Looks good. Nothing too special from my part, just a few nitpicks

src/components/Home/ClusterDashboardResourcesV5.js
```diff
@@ -0,0 +1,94 @@
...
+        <>
+          <Dot style={{ paddingLeft: 0 }} />
+          <RefreshableLabel value={cores}>
+            <span>{cores ? cores : '0'} CPU cores</span>
```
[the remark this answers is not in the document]
ograu: Thanks for the suggestions @axbarsan , in this case I will keep this one cause I want `"0"` in case of any falsy value is passed

src/components/Home/ClusterDashboardResourcesV5.js
ograu: 👍

PR #1246, opened 2020-03-17
Fix admin login if switching endpoints

axbarsan:
Fixes an issue where if you switched API endpoints and the token expired in the meantime, you wouldn't be able to login anymore, without manually clearing localstorage

Now deletion is done automatically, and you get redirected to the admin login, so it speeds up the authentication process.


axbarsan:
I have upset the test gods, I must fix them now 😬

PR #1247, opened 2020-03-17
Fix for undefined API endpoint in cluster details view

ograu:
Fix for `undefined` `api_endpoint` in cluster details view: https://gigantic.slack.com/archives/C3MCF5EJK/p1584444878000300

Adding the specific cluster flag for the cluster because when accessing it directly it doesn't exists in the store, so this:

```javascript
<LoadingOverlay loading={this.props.loadingCluster}>
  <KeyPairs cluster={cluster} />
</LoadingOverlay>
```

can't do its job.

### DIFF
--- a/src/components/Cluster/ClusterDetail/ClusterDetailView.js
+++ b/src/components/Cluster/ClusterDetail/ClusterDetailView.js
@@ -224,13 +224,16 @@ class ClusterDetailView extends React.Component {
       region,
       genericLoadingCluster,
       loadingNodePools,
+      loadingCluster,
     } = this.props;
+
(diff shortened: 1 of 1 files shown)


axbarsan:
Awesome that you found the fix for this 🚀 

Should we also remove this?
https://github.com/giantswarm/happa/blob/c840815f3f851b239f88bf9482ef3987890863ba/src/components/Cluster/ClusterDetail/KeyPairs.js#L87

ograu:
> Awesome that you found the fix for this 🚀

Hopefully... 😅 

> Should we also remove this?

Would like to keep it until we are sure this is fixed. Adding a note in my board to not forget about this.
Thanks for noticing it 🙂

PR #1248, opened 2020-03-17
Optimistic update on cluster name change

axbarsan:
Fixes https://github.com/giantswarm/giantswarm/issues/9428

### Changes
* Wrote unit tests for initial component, to not refactor blindly
* Refactored the component, disconnected it from the store
* Made the `OK` button have the success look (to be different from the cancel button)
* Removed the success message after editing cluster/node pool names, we're going full optimistic here (if the request fails, an error flash message would be shown, and the value will be reverted to the one before the change)
* Updated the tests that asserted those messages
* Instead of displaying a flash message with an error if the input validation fails, we're displaying a tooltip with the error message, giving the input a red border, and disabling the submit button

### How it looks
* **Error state**
image

* **All bueno**
image

### DIFF
--- a/src/components/Cluster/ClusterDetail/AppDetailsModal/AppDetailsModal.js
+++ b/src/components/Cluster/ClusterDetail/AppDetailsModal/AppDetailsModal.js
@@ -119,6 +119,7 @@ const AppDetailsModal = props => {
     done();
   }
 
+  // eslint-disable-next-line react/no-multi-comp
   function deleteConfirmFooter(cta, onConfirm) {
     return (
       <DeleteConfirmFooter
--- a/src/components/Cluster/ClusterDetail/NodePool.js
+++ b/src/components/Cluster/ClusterDetail/NodePool.js
@@ -14,6 +14,21 @@ import AvailabilityZonesWrapper from './AvailabilityZonesWrapper';
 import NodePoolDropdownMenu from './NodePoolDropdownMenu';
 import ScaleNodePoolModal from './ScaleNodePoolModal';
 
+const NPViewAndEditName = styled(ViewAndEditName)`
+  input[type='text'] {
+    font-size: 15px;
+    line-height: 1.8em;
+    margin-bottom: 0;
+  }
+  .btn-group {
+    top: 0;
+  }
+  button {
+    font-size: 13px;
+    padding: 4px 10px;
+  }
+`;
+
 const NodesWrapper = styled.div`
   width: 36px;
   height: 30px;
@@ -108,13 +123,12 @@ class NodePool extends Component {
         <NameWrapperDiv
           style={{ gridColumn: isNameBeingEdited ? '2 / 9' : null }}
         >
-          <ViewAndEditName
-            cssClass='np'
-            entity={nodePool}
-            entityType='node pool'
+          <NPViewAndEditName
+            name={nodePool.name}
+            type='node pool'
             onSubmit={this.editNodePoolName}
             ref={viewEditName => (this.viewEditNameRef = viewEditName)}
-            toggleEditingState={this.toggleEditingState}
+            onToggleEditingState={this.toggleEditingState}
           />
         </NameWrapperDiv>
         {/* Hide the rest of fields when editing name */}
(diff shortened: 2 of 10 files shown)


src/components/Cluster/ClusterDetail/AppDetailsModal/AppDetailsModal.js
```diff
@@ -119,6 +119,7 @@ const AppDetailsModal = props => {
     done();
   }
 
+  // eslint-disable-next-line react/no-multi-comp
```
axbarsan: This was messing up my linter 😬

testUtils/setupTests.js
```diff
@@ -1,3 +1,5 @@
+import '@testing-library/jest-dom/extend-expect';
```
axbarsan: Added this globally, since we're pretty much adding it manually to every test

src/components/Cluster/ClusterDetail/NodePool.js
```diff
@@ -14,6 +13,21 @@ import AvailabilityZonesWrapper from './AvailabilityZonesWrapper';
...
+  }
+  .btn-group {
+    top: 0;
+  }
+  button {
+    font-size: 13px;
+    padding: 4px 10px;
+  }
+`;
+
```
axbarsan: Extracted this from the `ViewAndEditName` component, since that should be more generic

src/components/Cluster/ClusterDetail/NodePool.js
```diff
@@ -53,28 +67,13 @@ class NodePool extends Component {
...
-
-          return reject(error);
-        });
-    });
+    try {
+      this.props.dispatch(
+        nodePoolActions.nodePoolPatch(cluster.id, nodePool, { name })
+      );
+    } catch {
+      // Do nothing on purpose, flash message is already displayed inside the action
```
oponder: I think if are handling errors at the action level, then we should just leave out the `try/catch` block here and remove the `throw` from the action. 

And then possibly we can do a manual post to slack using the error reporting library if we still want to be aware of cases where nodepool names fail to update.

So it becomes always the action layer's responsibility to catch any errors and not throw them any further up into the UI. 

If somehow an error does happen in the UI (like imagine maybe we this.props.dispatch is undefined for whatever reason) then I think the UI should fail, possibly with an error boundary, and we should get a notification about it.

Like this we could make a mistake (possible the compiler, linter, or tests, will catch it, but who knows...) where something is undefined or otherwise wrong on one of these lines, and we wouldn't know about it.

axbarsan: Yes, that is a very valid opinion. The problem is is with how the patch functions are currently created, and how they throw errors back into the UI. I started refactoring the patch requests when I starting working on this issue, but it just seemed to be too big of a change (and a risk) for this PR (for example, the cluster patch functions are also used for scaling and upgrading clusters).

Just removing the `throw` from the actions makes some tests fail, so I'd prefer to do that refactor after I get done with this PR.

However, knowing if the feature fails is very important, so I'll add a manual error report (for now, until I refactor it) 👍

axbarsan:
Turns out our app doesn't like exporting things from the main component file, so I'll have to find a way around that

```javascript
/home/circleci/project/src/styles/app.sass:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,global,jest){@import vendor/clearfix
                                                                                             ^

    SyntaxError: Invalid or unexpected token

       7 | import 'noty/lib/noty.css';
       8 | import 'react-datepicker/dist/react-datepicker.css';
    >  9 | import 'styles/app.sass';
         | ^
      10 | 
      11 | import { Notifier } from '@airbrake/browser';
      12 | import monkeyPatchGiantSwarmClient from 'lib/giantswarmClientPatcher';

      at Runtime.createScriptFromCode (node_modules/jest-runtime/build/index.js:1059:14)
      at Object.<anonymous> (src/components/index.js:9:1)

```

axbarsan:
Wrote an error reporter singleton to not have to export from the main component file anymore, since that breaks everything

PR #1249, opened 2020-03-18
Using error.response instead superagent custom message

ograu:
I think that the root of these errors is that superagent is creating this custom error and passing just the message as the `error` sometimes: https://github.com/visionmedia/superagent/blob/295dfcdacedd45e43a22d250bcaac6bf3d0a9229/lib/client.js#L425

Also, today in the support channel: https://gigantic.slack.com/archives/C02EVLE9W/p1584536120338300

I've seen this message in a lot of places in Happa. This is because we are getting this custom message instead looking at error.response, which is what we should be doing IMO.

As for the `401` I saw in the network tab, I think this is only adding confusion, my fault for mixing concerns. This is a different topic, and yes, it has to be something related to authentication. Probably a temporal 401 just before the token has been refreshed. This is out of the scope of this PR.


src/actions/clusterActions.js
```diff
@@ -186,7 +186,7 @@ export function clusterLoadDetails(
 
       return cluster;
     } catch (error) {
-      if (error.status === StatusCodes.NotFound) {
+      if (error.response.status === StatusCodes.NotFound) {
```
axbarsan: I don’t feel safe replacing this altogether. What do you think about having both conditions, but with a logical OR?

src/components/Cluster/ClusterDetail/ClusterDetailView.js
```diff
@@ -114,7 +113,6 @@ class ClusterDetailView extends React.Component {
     this.props.dispatch(
       batchedRefreshClusterDetailView(clusterID, this.props.isV5Cluster)
     );
-    this.props.dispatch(loadApps(clusterID));
```
ograu: Removed this and placed this disapatch in its corresponding batched function. 

The whole point of batched functions is to centralize series of API calls that a view needs. We decided to do this as a way of self-documenting the code, and also for debugability. 

If place dispatches consecutively in the view, as we had before, we are cheating ourselves.

Also, this was triggering an error for deleted clusters, because the thunk was dispatched no matter if the cluster existed or not.

src/reducers/clusterReducer.js
```diff
@@ -65,10 +65,10 @@ const clusterReducer = produce((draft, action) => {
 
       return;
 
-    // This is the action that we dipatch in order to actually remove a cluster from the store.
+    // This is the action that we dispatch in order to actually remove a cluster from the store.
     case types.CLUSTER_REMOVE_FROM_STORE:
       delete draft.items[action.clusterId];
-      if (action.isV5) {
+      if (action.isV5Cluster) {
```
ograu: This never worked actually!

src/actions/clusterActions.js
axbarsan: Something like 
```suggestion
      if (error.status === StatusCodes.NotFound || error.response.status === StatusCodes.NotFound) {
```

ograu: Please check this out: https://github.com/visionmedia/superagent/issues/1139#issuecomment-267698065

ograu: It's pretty old, but IIRC, with `axios` is the same thing, you check `error.response`

axbarsan: My only fear is that this will no longer work once we switch to the Cluster API, and we'll use our custom client. 
Our custom response objects don't have the `response` property

ograu: Legit. 

But then you will have all this noisy checks in these ORs.  I'm ok with everything, you decide :-)
For now, this does the job, if you want we can add these extra checks for the future.

src/components/Cluster/ClusterDetail/ClusterDetailView.js
```diff
@@ -79,7 +78,7 @@ class ClusterDetailView extends React.Component {
   loadDetails = () => {
     const { cluster, clusterId, organizationId, dispatch } = this.props;
 
-    if (typeof cluster === 'undefined') {
+    if (typeof cluster === 'undefined' || cluster.delete_date) {
```
ograu: We don't want the user to access details of a deleted clusters

PR #1250, opened 2020-03-18
Create custom error object with detailed stack traces

axbarsan:
We decided to add this in our [error handling discussion](https://docs.google.com/document/d/1d5_91P3r399qUhxkfUgFYCfKgyhMq01rrhFpZJcJ2mA/edit)

* Moved `ErrorReporter` helper in the `errors` package, as well
* Created a new error which supports:
  * Custom types
  * Custom stack traces (we can also edit them if we really want to)
  * Support for detailed stack traces (the helper fetches the sourcemaps and resolves the code based on them, it also tries to guess anonymous function names) - This will not happen if the user is offline
  * Easy object/string/json serialization/deserialization
  * Built-in error reporting with our 

How I believe this should work:
* We only store `CustomError` objects in the store
* We get the UI to react to that (I also added selectors for getting only error messages, as the old ones will get the whole error object)

As for how to actually use the new error object, you can check the tests 👍


oponder:
I like that it tries to make a good looking stack trace by even getting the sourcemap, this would let us remove the need for `api` to try and do that.

However I am somewhat concerned about the size of sourcemaps, and the behaviour of `stacktrace-gps`.  Have we tested it a bit? 

Some questions I can try to answer myself, but perhaps you know already:

- Will it try to fetch the sourcemap multiple times, or is it smart and does it get cached? 
- Does it only try to fetch the sourcemap if there is actually an error to report? 
- What happens if it can't find the source map?

axbarsan:
> I like that it tries to make a good looking stack trace by even getting the sourcemap, this would let us remove the need for `api` to try and do that.
> 
> However I am somewhat concerned about the size of sourcemaps, and the behaviour of `stacktrace-gps`. Have we tested it a bit?

`stacktrace-gps` is part of `stacktrace.js, which is thoroughly tested. I initially wanted to write something similar, but why reinvent the wheel if this is the go-to library for this (AFAIK there aren't any other libraries that do this). I've also tested it myself with the error reporter, and it works as expected.

> * Will it try to fetch the sourcemap multiple times, or is it smart and does it get cached?

I haven't found any info about it using any type of cache.
We could write our own solution for that. However, my only concern is that stack traces are pretty large, and storing them into memory could impact our performance.
We could however fetch them when the user loads the app, store them in session storage, and use them when we need it.

> * Does it only try to fetch the sourcemap if there is actually an error to report?

It only fetches the sourcemap when we call the `generateDetailedStack()` method.

> * What happens if it can't find the source map?

[There's actually a test for that](https://github.com/giantswarm/happa/blob/f77f5cece9d48a69372b02abe2cf0e2a7dc43d03/src/lib/errors/__tests__/CustomError.js#L137)

It will return the original stack trace, but with the message `Couldn't get the detailed stack: {errorName}` prepended to it

axbarsan:
> Looks good to me! Thanks for answering the questions.

What's your opinion on the cache, though? Is that something we should focus on? Or should we get some real-world results first?

oponder:
I think we should get some real world results first, let's start using this in certain places, perhaps discuss making some specific Custom Errors that we can reuse in certain places. And then see what the impact is.

PR #1253, opened 2020-03-20
Address objects inside of the error fallback helper

axbarsan:
Fixes an exception raised in our channel
https://gigantic.slack.com/archives/C3MCF5EJK/p1584687346000100

Addressess the case that the error is not an array (it usually is an error object or a server response)

### DIFF
--- a/src/components/Home/Home.js
+++ b/src/components/Home/Home.js
@@ -1,3 +1,4 @@
+import * as actionTypes from 'actions/actionTypes';
 import { batchedRefreshClusters } from 'actions/batchedActions';
 import * as clusterActions from 'actions/clusterActions';
 import * as nodePoolActions from 'actions/nodePoolActions';
@@ -12,6 +13,7 @@ import { Link } from 'react-router-dom';
 import ReactTimeout from 'react-timeout';
 import { CSSTransition, TransitionGroup } from 'react-transition-group';
 import { bindActionCreators } from 'redux';
+import { selectErrorByAction } from 'selectors/clusterSelectors';
 import { OrganizationsRoutes } from 'shared/constants/routes';
 import Button from 'UI/Button';
 import ClusterEmptyState from 'UI/ClusterEmptyState';
@@ -162,7 +164,10 @@ function mapStateToProps(state) {
   const selectedOrganization = state.main.selectedOrganization;
   const organizations = state.entities.organizations.items;
   const allClusters = state.entities.clusters.items;
-  const errorLoadingClusters = state.errors.CLUSTERS_LIST;
+  const errorLoadingClusters = selectErrorByAction(
+    state,
+    actionTypes.CLUSTERS_LIST_REQUEST
+  );
   const v5Clusters = state.entities.clusters.v5Clusters;
   const nodePools = state.entities.nodePools.items;
 
@@ -176,7 +181,7 @@ function mapStateToProps(state) {
   return {
     clusters,
     organizations,
-    errorLoadingClusters,
+    errorLoadingClusters: Boolean(errorLoadingClusters),
     selectedOrganization,
     v5Clusters,
     nodePools,
--- a/src/components/UI/ErrorFallback.js
+++ b/src/components/UI/ErrorFallback.js
@@ -2,9 +2,6 @@ import styled from '@emotion/styled';
 import PropTypes from 'prop-types';
 import React from 'react';
 
-const DEFAULT_MESSAGE =
-  'There was a problem. Please try again later and, if the problem persists, contact Giant Swarm support.';
-
 // ErrorFallback takes an error string or an array of error strings and will render
 //  the message/s. If it receives a falsy value or an array of falsy values it will
 // render its children.
@@ -14,54 +11,17 @@ const ErrorWrapperSpan = styled.span`
   font-weight: 300;
 `;
 
-/**
- * Helper that finds the best way to return a pretty-printed
- * version of the error object passed to it
- * @param {?(string[]|string)} errors - Error message(s)
- * @param {string} [separator=". "] - Separator between messages (if passing an error list)
- * @return {?string}
- */
-function parseErrors(errors, separator = '. ') {
-  if (typeof errors === 'string' || !errors) return errors;
-
-  let message = DEFAULT_MESSAGE;
-
-  if (Array.isArray(errors)) {
-    // Remove nullable values and join values in a string
-    // TODO maybe we want to show errors in a different manner.
-    message = errors.reduce((acc, currErr) => {
-      if (!currErr) return '';
-
-      let newAcc = acc;
-      if (newAcc !== '') {
-        newAcc += separator;
-      }
-
-      newAcc += currErr;
-
-      return newAcc;
-    }, '');
-
-    if (message === '') return null;
+function ErrorFallback({ errors, children }) {
+  if (errors) {
+    return <ErrorWrapperSpan>{errors}</ErrorWrapperSpan>;
   }
 
-  return message;
-}
-
-function ErrorFallback({ errors, children, separator }) {
-  const errorMessage = parseErrors(errors, separator);
-
-  return errorMessage ? (
-    <ErrorWrapperSpan>{errorMessage}</ErrorWrapperSpan>
-  ) : (
-    children
-  );
+  return children;
 }
 
 ErrorFallback.propTypes = {
-  errors: PropTypes.oneOfType([PropTypes.string, PropTypes.array]),
+  errors: PropTypes.string,
   children: PropTypes.node,
-  separator: PropTypes.string,
 };
 
 export default ErrorFallback;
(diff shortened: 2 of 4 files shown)


src/components/UI/ErrorFallback.js
```diff
@@ -11,18 +14,29 @@ const ErrorWrapperSpan = styled.span`
...
+    // Remove nullable values and join values in a string
+    // TODO maybe we want to show errors in a different manner.
+    const notNullErrors = errors.filter(error => error);
+    if (notNullErrors.length === 0) return null;
+
+    message = notNullErrors.join('. ');
+  } else {
+    // Maybe we're lucky and this is an error object/response object
+    message = errors.message || DEFAULT_MESSAGE;
+  }
```
ograu: propTypes is string or array... don't understand this: my idea was to pass messages (strings) to this component, as this is just responsible to render messages.

src/components/UI/ErrorFallback.js
```diff
@@ -50,7 +50,11 @@ function ErrorFallback({ errors, children }) {
 }
 
 ErrorFallback.propTypes = {
-  errors: PropTypes.oneOfType([PropTypes.string, PropTypes.array]),
+  errors: PropTypes.oneOfType([
+    PropTypes.object,
+    PropTypes.string,
+    PropTypes.array,
+  ]),
   children: PropTypes.node,
```
ograu: Sorry, I wasn't clear enough: Why do we want to pass the full object here? What do we get from this?

src/components/UI/ErrorFallback.js
axbarsan: Good catch about the proptypes.

This addresses objects as well, because there are cases that use objects for errors.

For example:
https://github.com/giantswarm/happa/blob/077c47676ff688532cdcb3de8759165a4443707a/src/actions/nodePoolActions.js#L57

dispatches an error object, and then

https://github.com/giantswarm/happa/blob/077c47676ff688532cdcb3de8759165a4443707a/src/components/Home/ClusterDashboardItem.js#L177
uses the error object, which triggers an error.

ograu: Yes, my point is that I think we shouldn't be doing this, and if we do we should fix instead letting this component accept objects as props IMHO.

axbarsan: Alright, it wasn't clear at first about how it should work, so I thought that was on purpose. I'll fix it

ograu: Thanks 🙂

src/components/Home/Home.js
```diff
@@ -162,7 +164,10 @@ function mapStateToProps(state) {
   const selectedOrganization = state.main.selectedOrganization;
   const organizations = state.entities.organizations.items;
   const allClusters = state.entities.clusters.items;
-  const errorLoadingClusters = state.errors.CLUSTERS_LIST;
+  const errorLoadingClusters = selectErrorByAction(
```
axbarsan: Added the selector because it was getting the error state manually

src/components/UI/ErrorFallback.js
```diff
@@ -11,22 +14,43 @@ const ErrorWrapperSpan = styled.span`
...
+function parseErrors(errors, separator = '. ') {
   if (typeof errors === 'string' || !errors) return errors;
 
-  // Remove null values and joins values in a string
-  // TODO maybe we want to show errors in a different manner.
-  const notNullErrors = errors.filter(error => error);
-  if (notNullErrors.length === 0) return null;
+  let message = DEFAULT_MESSAGE;
+
+  if (Array.isArray(errors)) {
```
axbarsan: Kept the check here because we'd get problems again if we want to use the `ErrorFallback` component with actions that already dispatch error objects

src/components/UI/ErrorFallback.js
```diff
@@ -38,6 +62,7 @@ function ErrorFallback({ errors, children }) {
 ErrorFallback.propTypes = {
   errors: PropTypes.oneOfType([PropTypes.string, PropTypes.array]),
   children: PropTypes.node,
+  separator: PropTypes.string,
```
axbarsan: Added this for easy customization of the output message

oponder:
Is this error component a bit too fancy, with its separator handling and stuff like that? 

I think if as a follow up, the intention is to simplify and get to a place where we are just sending a single string, that would probably be the best!

But otherwise the PR looks good to me.

Will let @ograu give the final opinion here since he has a bit more into the error handling and stuff like that right now.

axbarsan:
> Is this error component a bit too fancy, with its separator handling and stuff like that?
> 
> I think if as a follow up, the intention is to simplify and get to a place where we are just sending a single string, that would probably be the best!
> 
> But otherwise the PR looks good to me.
> 
> Will let @ograu give the final opinion here since he has a bit more into the error handling and stuff like that right now.

I'm all for simplifying it, but I guess what Oriol had in mind with the list of errors was a scenario like this:

* The action throws a generic error, for actions that get used in multiple places (such as patch actions): `The cluster could not be updated`
* At the component level we would pass another message that is more specific to that scenario, such as `The cluster name was not changed.`
* The final message would be something like `The cluster could not be updated. The cluster name was not changed.`

We could do the message manipulation at component level, but in the end that code would just get duplicated every time we reach a situation like this.

oponder:
Hmm.. I see. 

In order for that to work we do need some mechanism to inform the thing that initiated the action that an error occurred though right? So then we'd be back doing a throw and try/catching in the UI?

axbarsan:
> Hmm.. I see.
> 
> In order for that to work we do need some mechanism to inform the thing that initiated the action that an error occurred though right? So then we'd be back doing a throw and try/catching in the UI?

That already works with how we're doing stuff now. With the error selectors, the injected error is `null` if there's no error, and changed to a string value if an error occurs.

ograu:
> That already works with how we're doing stuff now. With the error selectors, the injected error is `null` if there's no error, and changed to a string value if an error occurs.

Exactly we already have this. The flow is:

- The action fails and dispatch an error
- The error reducer logs the error in the store
- Components are subscribed to relevant errors and show messages from the store if they happen to exist

ograu:
> Is this error component a bit too fancy, with its separator handling and stuff like that?

Maybe there's a better approach, but my initial idea was to have a component with the only responsability to print error messages. I wrote one that accepted a string, the error message. 

Then I found situations in wich more than one api call where involved, so I decided that this component had to accept arrays.

Now I think I was wrong: a simpler approach would be for this component to only accept strings, one error message at a time, and when more than one error, two for instance have to be _listened_., then just use the component twice.

ograu:
From a maintanibility and readibility perspective, I think this is the best we can do.

axbarsan:
> > Is this error component a bit too fancy, with its separator handling and stuff like that?
> 
> Maybe there's a better approach, but my initial idea was to have a component with the only responsability to print error messages. I wrote one that accepted a string, the error message.
> 
> Then I found situations in wich more than one api call where involved, so I decided that this component had to accept arrays.
> 
> Now I think I was wrong: a simpler approach would be for this component to only accept strings, one error message at a time, and when more than one error, two for instance, have to be _listened_, then just use the component twice.

But what about the case in the comment above? When the error message from the action is too generic and we want to go in more deeply?

ograu:
My thinking is that in those situations we can write custom messages at component level and pass them to this error fallback component. So we can keep `ErrorFallback` simple. WDYT?

axbarsan:
I don't think we get any benefit from removing a couple of lines of code from the `ErrorFallback` component, and adding more lines to other components.

We're essentially going to have something like this in multiple components:
```javascript
let errorMessage = null;

if (props.someError) {
  errorMessage = `${props.someError}. Some other message.`;
}
```
instead of a couple of extra lines instead of `ErrorFallback`.

ograu:
For me this is easier to grasp, in fact this was what I was doing here: https://github.com/giantswarm/happa/pull/1229/files#diff-d6aab637bd5ecfe4e323d3e499b95dddR417

But I am ok with whatever you decide :-)

oponder:
I'm ok with whatever you guys agree with here, I think if there was a good plan, don't let me interrupt that! 

Though if you ask me, I kinda like the explicitness of the example you gave, even if it is more lines. 

I don't mind the duplication in this case, I consider it maybe as coincidental duplication. (https://williamboles.me/coincidental-duplication-costs-can-really-add-up/)

There might be times when we don't want this behaviour, and then it's useful that the error fallback is just a dumb component, instead having to keep in mind that it takes strings or arrays and does other (minimally) fancy things.

ograu:
> There might be times when we don't want this behaviour, and then it's useful that the error fallback is just a dumb component, instead having to keep in mind that it takes strings or arrays and does other (minimally) fancy things.

Agree 100% 👍

axbarsan:
Alright guys, thank you for the insights. I updated the component and made it as dumb as possible 😬

PR #1254, opened 2020-03-20
Responsive clusters list

ograu:
in


src/components/Home/ClusterDashboardItem.js
```diff
@@ -45,29 +79,27 @@ const WrapperDeleted = styled.div`
...
   flex: 0 0 210px;
+  display: flex;
+  justify-content: flex-end;
+  ${mq('medium')} {
+    flex: unset;
+    position: absolute;
+    top: 21px;
+    right: 8px;
+  }
+  ${mq('725')} {
```
axbarsan:
```suggestion
  ${mq(725)} {
```

src/components/Home/ClusterDashboardItem.js
```diff
@@ -45,29 +79,27 @@ const WrapperDeleted = styled.div`
...
+    position: relative;
+    width: 100%;
+    top: 9px;
+    margin-bottom: 9px;
+    justify-content: flex-start;
+    padding-left: 98px;
+  }
+  ${mq('small')} {
```
axbarsan: What do you think about extracting these device types to an enum?

src/components/Home/ClusterDashboardItem.js
```diff
@@ -25,6 +25,21 @@ import ClusterDashboardResourcesV4 from './ClusterDashboardResourcesV4';
 import ClusterDashboardResourcesV5 from './ClusterDashboardResourcesV5';
 import UpgradeNotice from './UpgradeNotice';
 
+const LabelWrapper = styled.div`
+  width: 90px;
```
axbarsan: I believe this would be more future-proof. What do you think?
```suggestion
  min-width: 90px;
  max-width: 100%;
```

axbarsan:
Wooow, looking good 💪

src/components/Home/ClusterDashboardItem.js
ograu: Here I was thinking in types already: `breakpoint` argument is a string, that's why I passed a string.

src/components/Home/ClusterDashboardItem.js
ograu: Well, I don't know... In which scenarios do you think it would be useful?

src/components/Home/ClusterDashboardItem.js
axbarsan: I think it would raise confusion, as someone might think "should this be `'725'` or `'725px'`?".
As for types, the argument type can be an union (`string | number`), and accept both strings and numbers.

src/components/Home/ClusterDashboardItem.js
axbarsan: In case the ID get longer (which will happen, since we'll let users set their own IDs in the near future"

src/shared/constants/cssBreakpoints.js
```diff
@@ -0,0 +1,5 @@
+export const CSSBreakpoints = {
+  small: 'small',
+  medium: 'medium',
+  large: 'large',
+};
```
ograu: It's not clear to me whether we are importing everything to our `index.js` in `constants` folder or we are ok with having this and importing the module where we need it

axbarsan: Well for the rest of the constants we export from the `index.js` file in the `constants` folder, and then we import something like:
```javascript
import { CSSBreakpoints } from 'shared/constants';
```

axbarsan: Not a blocker for me, it doesn't really make any difference

ograu: ok, changing it 👍

ograu:
I decided to use directly `theme.breakpoints` instead, it doesn't make sense to use an enum if we already have this. WDYT?

axbarsan:
> I decided to use directly `theme.breakpoints` instead, it doesn't make sense to use an enum if we already have this. WDYT?

Not sure about this. In theory, we could have multiple themes, and each theme might choose to implement a breakpoint or not. If we ever want to programmatically change the theme (ex: light/dark themes), we would have the wrong theme imported.

Why not have the breakpoints enum, and then have breakpoint names from enums inside the theme? Ex:
```javascript
  breakpoints: {
    [CSSBreakpoints.small]: 540,
    [CSSBreakpoints.medium]: 825,
    [CSSBreakpoints.large]: 1200,
  },
```

ograu:
> Not sure about this. In theory, we could have multiple themes, and each theme might choose to implement a breakpoint or not. If we ever want to programmatically change the theme (ex: light/dark themes), we would have the wrong theme imported.
> 

We can easily implement  something like this whenever we have different themes with different brekpoints. I never been in such a situation, though.

> Why not have the breakpoints enum, and then have breakpoint names from enums inside the theme? Ex:
> 
> ```js
>   breakpoints: {
>     [CSSBreakpoints.small]: 540,
>     [CSSBreakpoints.medium]: 825,
>     [CSSBreakpoints.large]: 1200,
>   },
> ```

Because it is redundant and makes the code harder to follow IMO.

ograu:
> I never been in such a situation, though.

Talking about breakpoints

axbarsan:
What I'm trying to say is that we shouldn't rely on something that might not always be there, or which might have a different structure than we expect

ograu:
Sorry, not following @axbarsan , same applies with `CSSBreakpoints.small`, you are relying on something that might not be there...

axbarsan:
Sorry for not being clear enough. What I mean is that we don't have any guarantee that the `small` (for example) property is in the `breakpoints` property of the current theme, that's why it's not a good reason to assume that all themes have the same structure, because we cannot enforce it

ograu:
Ok, let's do it. 

I don't really see a need for this now, but if you think you probably will need more/different `breakpoints` objects in future, it's definitely interesting to have this now 👍

axbarsan:
Thank you! 🚀

PR #1266, opened 2020-03-24
Version picker error handling

oponder:
This is a draft because I am not too sure about the new pattern of error handling now.

The thing with communicating errors to the UI using only redux and having no local state is that we'd then also need some way to clear the error in some UI components.

In this PR I am again resorting to using local state to set the error that this component should show. But instead of using a throw in the action and try/catch in the component, I am returning a kind of result object, which could contain an error. The UI component then decides what to do based on that.

I tried at first to use `clusterUpdateAppError: selectErrorByAction(state, CLUSTER_UPDATE_APP_ERROR)` in `mapStateToProps`

but the problem is that once an error occurs, this component will then always show that error.

There's no way to reset it when I close the modal and open it again. 

I thought about adding some other action that clears errors or something like that, but it started to feel like introducing a lot of back and forth communication between the component and the redux state which to me did not seem to be the intent of this pattern.

Bringing this up for @ograu and @axbarsan , maybe what I am doing is ok, and I am just misinterpreting the result of our error handling meeting. Let me know if this pattern is fine or if it clashes with what we decided at the end of this meeting: https://docs.google.com/document/d/1d5_91P3r399qUhxkfUgFYCfKgyhMq01rrhFpZJcJ2mA/edit

>Decisions
	We handle errors in the data layer use redux to communicate those errors to the UI.
	We want UI to not have to do try / catch
	Fmt additions to explain


axbarsan:
This is a very valid point, as there's no easy way to clear an error right now.

I don't think it's a good idea to switch to local state, as we know how complicated can that get, especially with nested components.

I think we should create a custom React hook for errors, which we can use like
`const { message, clear } = useError('ERROR_TYPE');`
(we could also have more useful stuff there, that's why we're destructuring from an object, not a tuple)

The hook would use Redux's `useSelector` under the hood, and would also create the error clear dispatcher in there, so we wouldn't need to bind props and other crazy stuff.

What do you think about that?

ograu:
+1 with what Adrian is proposing, I think it is the way to go if we want to have errors in store. Think about how clean, nice and (relatively) easy it will be to use this. 

> I thought about adding some other action that clears errors or something like that, but it started to feel like introducing a lot of back and forth communication between the component and the redux state which to me did not seem to be the intent of this pattern.

I understand that you may find a little bit more convoluted conceptually than having local state manage everything but:

> I don't think it's a good idea to switch to local state, as we know how complicated can that get, especially with nested components.

I totally agree with this, there's a lot of different errors than can happen and you will need to have those errors availablle in various UI layers in Happa: so a lot of errors are global state actually and that's why I think this kind of logic doesn't belong to the UIs.

oponder:
Ok! I will try to make that hook and make this work without the local state. Thanks for the input!

src/actions/appActions.js
```diff
@@ -263,15 +263,30 @@ export function updateApp(appName, clusterID, values) {
...
+          appName,
+        });
+
+        return {
+          error: '',
+        };
       })
       .catch((error) => {
-        new FlashMessage(
-          `Something went wrong while trying to update your app. Please try again later or contact support: support@giantswarm.io`,
```
oponder: Not dispatching flashMessages from actions anymore.

src/actions/actionTypes.js
```diff
@@ -294,3 +294,6 @@ export const RELEASES_LOAD_ERROR = 'RELEASES_LOAD_ERROR';
 export const RELEASE_SELECTED = 'RELEASE_SELECTED';
 
 export const ROUTER_LOCATION_CHANGE = '@@router/LOCATION_CHANGE';
+
+// Error store management
+export const SINGLE_ERROR_CLEAR = 'SINGLE_ERROR_CLEAR';
```
oponder: I called it "SINGLE_ERROR_CLEAR" because otherwise calling it the more straightforward "CLEAR_ERROR" matches with the request / error reducers that are looking for things that end with "_ERROR"

src/hooks/errors.js
```diff
@@ -0,0 +1,20 @@
...
+export const useError = (errorType) => {
+  const dispatch = useDispatch();
+
+  return useSelector((state) => {
+    const errorMessage = state.errors[typeWithoutSuffix(errorType)] ?? null;
+
+    const clear = function () {
+      dispatch({
+        type: SINGLE_ERROR_CLEAR,
+        errorType: typeWithoutSuffix(errorType),
```
oponder: This is what we dispatch to clear an error.

```
{
  type: "SINGLE_ERROR_CLEAR"
  errorType: "CLUSTER_UPDATE_APP"
}

File: src/reducers/entityLoadingReducer.js [hunk 4]
```diff
@@ -10,7 +10,7 @@ const initialState = {
 // item that we are having problems with getting its info, ie getCluster or getNodePool
 const entityLoadingReducer = produce((draft, action) => {
   const { type, id } = action;
-  const matches = /(.*)_(REQUEST|SUCCESS|ERROR|FINISHED|NOT_FOUND)/.exec(type);
+  const matches = /(.*)_(REQUEST|SUCCESS|ERROR|FINISHED|NOT_FOUND)$/.exec(type);
```
oponder: To make sure we don't thing `SINGLE_ERROR_CLEAR` is a request, I added this `$` so that only actions ending in one of these words are considered for the loading reducer.

oponder:
Alright! I think I did it. Nice idea guys! I think I can see how this will keep things more stable :)

src/hooks/errors.js
```diff
@@ -0,0 +1,20 @@
+import { SINGLE_ERROR_CLEAR } from 'actions/actionTypes';
+import { useDispatch, useSelector } from 'react-redux';
+import { typeWithoutSuffix } from 'selectors/selectorUtils';
+
+export const useError = (errorType) => {
+  const dispatch = useDispatch();
+
+  return useSelector((state) => {
+    const errorMessage = state.errors[typeWithoutSuffix(errorType)] ?? null;
```
axbarsan: We can reuse our existing selector here, to not duplicate logic. It would look something like:
```javascript
const errorMessage = useSelector((state) => selectErrorByAction(state, errorType));
```

oponder: ok! done!

src/hooks/errors.js
```diff
@@ -1,12 +1,13 @@
 import { SINGLE_ERROR_CLEAR } from 'actions/actionTypes';
 import { useDispatch, useSelector } from 'react-redux';
+import { selectErrorByAction } from 'selectors/clusterSelectors';
 import { typeWithoutSuffix } from 'selectors/selectorUtils';
 
 export const useError = (errorType) => {
   const dispatch = useDispatch();
 
   return useSelector((state) => {
```
axbarsan: Just wondering, why do you return the output of `useSelector` like that?

Why not 
```javascript
  const dispatch = useDispatch();
  const errorMessage = useSelector(state =>
    selectErrorByAction(state, errorType)
  );

  const clear = () => {
    dispatch({
      type: SINGLE_ERROR_CLEAR,
      errorType: typeWithoutSuffix(errorType)
    });
  };

  return { errorMessage, clear };
};
```

Does it provide any additional benefit?

oponder: Even I'm not sure what I was smoking

oponder: Cleaned it up :)

PR #1268, opened 2020-03-24
Hackathon Q1: Introducing TypeScript

axbarsan:
# 👋 Hello! Let's build better code.

### Why TypeScript?
* Static typing -> harder for us developers to make mistakes, since the compiler (or even the code editor) will give us error before we get to do anything that might break in some edge case
* Code completion -> we know which properties to provide, which types, etc.
* Features are introduced quicker -> get every new feature ~6 months before they get introduced into JavaScript (and they all have polyfills, as well)
* Better OOP support (interfaces, explicit member accessibility -> `private`/`protected` methods, custom types, generics)
* Code editor on steroids -> the editor understands your code better, and it can provide actual improvements and automatic refactorings (example below)

This
image
becomes this
image

Now that we got that out of the way, let's write some TypeScript.

### This PR includes:
* Add typescript dependencies
* Add babel config for supporting typescript files
* Add webpack config for supporting typescript files
* Add useful linter rules (had to disable some built-in, because are handled by typescript plugin for both JS and TS) + fix the issues caused by the new rules
* Convert index.js and App.js to typescript
* Write types for theme, convert style utils to typescript, add global theme config with the theme interface (now we get auto-complete with theme fields everywhere)
* Write examples for TypeScript React components

### Learning materials
Good starting points:
* https://www.typescriptlang.org/docs/handbook/typescript-in-5-minutes.html
* https://www.typescriptlang.org/docs/handbook/basic-types.html
* https://rmolinamir.github.io/typescript-cheatsheet/

If you want to start from scratch or understand everything better, here's a good crash course:
* https://www.youtube.com/watch?v=BwuLxPH8IDs

### The direction
As we are starting out migrating the codebase to TypeScript, if we aim for 100% strictness and correctness it will actually slow us down, instead of helping us.

That being said, I propose going this route:
* Every new feature we develop from now on should be written in TypeScript
* We use strict types where possible, but we don't waste time being stuck trying to find the perfect type for something that is still in the old JS code: throw an `any` in there and we'll address it later
* Slowly start migrating our existing files to TypeScript (great learning exercise) -> same as the point above, do it quick and dirty, we'll iterate and improve
* Migrate all the tests
* Add more strictness (tighter linter rules -> start fixing files and removing `// @ts-ignore` comments)
* Polishing phase -> if you see something that can be improved, do it

### Examples in the PR
We got 2 basic examples:

* The happy case, the UI component:  https://github.com/giantswarm/happa/blob/b0ddbf14359ef9b9fe04de5d83a59cc91990c099/src/components/UI/VersionPicker/VersionPicker.tsx
😃Everything's nice and easy, we can be strict and clean

* The less happy case, the component connected to the redux store, with all the bells and whistles:
https://github.com/giantswarm/happa/blob/b0ddbf14359ef9b9fe04de5d83a59cc91990c099/src/components/Auth/Login.tsx
🤐 Erm...

As you can tell, we have a lot of `// @ts-ignore` comments and some linter rules disabled. Since we have a lot of moving parts here (redux state properties, actions, components etc.), we'll change a lot of files just to get this component "production-ready". This will take ages, and it will be really frustrating.
What I believe will work for us is going for something in between -> some `any` types here and there, some linter/compiler ignoring, and then slowly refining the codebase over time.

### Where to start?
- Easy: constants, lib, selectors, utils, model, reducers, etc.
- Medium: actions, UI components
- Hard: Store-connected components

For constants, there is an example on how they should look as TypeScript enums:
https://github.com/giantswarm/happa/blob/b0ddbf14359ef9b9fe04de5d83a59cc91990c099/src/shared/constants/cssBreakpoints.ts
(You can check the diff for its usage)

### Style
We can talk about this for days, but for starters:
* Don't type styled components, unless we provide custom props for them -> waste of time
* Interfaces over types for Map-like data structures (basically plain JS objects) -> interfaces can be implemented, extended and merged


src/components/Auth/Login.tsx
```diff
@@ -7,27 +7,67 @@ import {
...
+interface ILoginProps extends IStateProps, IDispatchProps {}
+
+interface ILoginState {
+  email: string;
+  password: string;
+  authenticating: boolean;
+}
+
+class Login extends React.Component<ILoginProps, ILoginState> {
+  public static propTypes: IStateProps & IDispatchProps = {
```
ograu: `public` is not the default label? Should we label everything as `public` then?

axbarsan: `public` is indeed the default label. I prefer to specifically write it for consistency, since we're writing `private` and `protected`, as well

src/components/UI/VersionPicker/VersionPicker.tsx
```diff
@@ -101,25 +101,47 @@ const Body = styled.div`
...
+  // An array of versions to pick from.
+  versions: IVersion[];
+
+  // The currently selected version, will be highlighted in bold in the list.
+  selectedVersion: string;
+
+  // A callback function that gets called when a version is selected by the user.
+  onChange?: (newVersion?: string) => void;
+}
+
```
ograu: Do you follow an y rules when deciding to use a `type` or an `interface`? Interfaces are more powerful, but we could use the convention to use `type`s whenever something is not meant to be reused/extended. WDYT?

src/components/UI/VersionPicker/VersionPicker.tsx
```diff
@@ -216,9 +238,9 @@ const VersionPicker = ({ onChange, selectedVersion, versions, ...props }) => {
 };
 
 VersionPicker.propTypes = {
+  selectedVersion: PropTypes.string.isRequired,
+  versions: PropTypes.array.isRequired,
   onChange: PropTypes.func,
```
ograu: We can get rid of that one, right?  Or maybe we want to check this at runtime as well?

src/components/UI/VersionPicker/VersionPicker.tsx
axbarsan: > Interfaces over types for Map-like data structures (basically plain JS objects) -> interfaces can be implemented, extended and merged

or taken from the TypeScript documentation:
> Because an ideal property of software is being open to extension, you should always use an interface over a type alias if possible.

I'd say we should be consistent and always use interfaces for objects with a map-like structure, example:
```javascript
const someObj = {
  someKey: 'someValue',
}
```

This way there would not be any arguing in PRs whether to use types or interfaces. 
Also, theoretically speaking, objects in JS are all the same type, so it makes sense to define them with interfaces, since they just respect a certain blueprint.

There are special cases where only `type` works (such as mapped types):
```typescript
// Built-in Record type
type Record<K extends keyof any, T> = {
    [P in K]: T;
}
```

There's a small doc about this in the TypeScript documentation:
https://www.typescriptlang.org/docs/handbook/advanced-types.html#interfaces-vs-type-aliases

src/components/UI/VersionPicker/VersionPicker.tsx
axbarsan: You mean getting rid of proptypes completely? I wanted to discuss that 😬

ograu: 👍  I  meant getting rid only of `onChange` prop since we are checking it in `IVersionPickerProps` interface

axbarsan: We'll get linter errors if we don't include all the props in the `propTypes` 😅

src/components/Auth/Login.tsx
```diff
@@ -7,27 +7,67 @@ import {
...
+     */
+    // @ts-ignore
+    dispatch: PropTypes.func,
+    // @ts-ignore
+    flashMessages: PropTypes.object,
+    // @ts-ignore
+    actions: PropTypes.object,
+  };
+
+  public state: ILoginState = {
```
axbarsan:
```suggestion
  public readonly state: ILoginState = {
```

axbarsan:
> LGTM!
> 
> Should we test a deploy before merging? Just to make sure the production build still works ok? I noticed some dockerfile changes. When the CI passes, it usually is all ok, but wouldn't hurt to check.

I did test builds locally before submitting this for the hackathon, but I'll test deploy to an installation now, to be sure

PR #1124, opened 2020-02-13
Various testing changes

ubergesundheit:
I'm trying out code splitting with Suspense

PR #1269, opened 2020-03-25
Fix test router redirecting

axbarsan:
This should fix the problem with the tests in https://github.com/giantswarm/happa/pull/1124/
This was branched off of that brach, and is meant to be merged back into it

The problem was when rendering a route during tests:
* The `/` path would be the initial path
* We would push to the desired path provided to the `renderRouteWithStore` function

The problem in the `SignUp` tests was that we would first push to `/` with an user in the storage, so the app would deserialize it and act like it was logged in, which would mount the `Layout` component and trigger the API calls. Then we would redirect to `SignUp`, but we would cut the test before the api calls were actually done.

The reason why this only surfaced now is that lazy rendering is asynchronous, which gave the API requests time to initialize.

### DIFF
--- a/__tests__/AdminLogin.js
+++ b/__tests__/AdminLogin.js
@@ -63,16 +63,18 @@ helpers.validateOrRaise = jest.fn();
 const renderRouteWithStore = (
   initialRoute = AppRoutes.Home,
   state = {},
-  storage = initialStorage,
-  history = createMemoryHistory()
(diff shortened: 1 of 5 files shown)


axbarsan:
Whoops, gotta fix something in AdminLogin tests

ubergesundheit:
<3

PR #1272, opened 2020-03-27
Playing with TypeScript

ograu:
The purpose of this PR and [this other PR](https://github.com/giantswarm/happa/pull/1268) is to kickstart a discussion to decide whether we finally want to add TypeScript, as [this is somthing still to be decided](https://gigantic.slack.com/archives/CQ465CFG8/p1585140108003100) to the mix.

In this PR:

- I converted some small UI components to `.ts`, and then
- Converted `NodePool` component, which is a connected one
- Created a `types.ts` file for reuse interfaces
- Also there is a minor fixing for `CSSBreakpoints` importing

After two days hacking with TypeScript, these are my thoughts:

- It is clear that TS offer some great advantatges, being the most remarkable self-documentation of the code and getting rid of type errors, IMO.
- I'm not used to typed languages, so using TS feels superweird to me. I think it adds a lot of noise and it makes more difficult to focus on getting things done. __I think this is just a phase that I would overcome after some weeks working with TS__.

And my opinion about if we should migrate to TS or not:

- From a purely selfish poin of view, I would say yes. I would like to experiment with this, especially taking into account that usage of TS in frontend frameworks in becaming mainstream.
- Is this good for Happa? Yes, sure.
- Is this needed for Happa? No, we can live without this.
- Would the benefits of using TS outweigh the drawbacks? In the short/medium term, I don't think so. In the long term, I would say yes.

Finally, food for touhgt, you will find controversial and subjective (but totally valid IMO) opinions in this relatively short vid (20 min)  from MPJ, they talk also about linters: https://www.youtube.com/watch?v=_ECuXk4MX6Y

This vid pretty much sums up how I feel right now! 😬


src/shared/types.ts
```diff
@@ -0,0 +1,26 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  availability_zones: any;
+  id: string;
+  name: string;
+  node_spec: object;
+  scaling: IScaling;
+  status: INodePoolStatus;
+}
+
+export interface IDispatchProps extends DispatchProp {
```
axbarsan: It's common practice to name props coming from `mapDispatchToProps` `IDispatchProps`, and this name kind of collides with that. Maybe rename to something more action-related?

src/components/UI/ErrorFallback.tsx
```diff
@@ -1,26 +1,29 @@
...
+
 /**
  * This component will return the error passed to it,
  * if it exists, or the children, if it does not
  */
-function ErrorFallback({ errors, children, ...rest }) {
-  if (errors) {
-    return <ErrorWrapperSpan {...rest}>{errors}</ErrorWrapperSpan>;
-  }
+function ErrorFallback({ error, children }: IErrorFallbackProps) {
```
axbarsan: This is usually typed with the React Function Component generic:
```typescript
const ErrorFallback: React.FC<IErrorFallbackProps> = ({ error, children }) => { ... }
```

src/components/Home/UpgradeNotice.tsx
```diff
@@ -30,21 +30,29 @@ const UpgradeWrapperDiv = styled.div`
   }
 `;
 
+interface IUpgradeNoticeProps {
+  canClusterUpgrade: boolean;
+  loadingReleases: boolean;
+  onClick?: () => void;
```
axbarsan: It's common practice to have optional parameters at the end of the interface declaration

src/components/Home/ClusterDashboardNodes.tsx
```diff
@@ -4,8 +4,16 @@ import { FallbackMessages } from 'shared/constants';
...
 import { isClusterYoungerThanOneHour } from 'utils/clusterUtils';
 
+interface IClusterDashboardNodes {
+  numberOfNodes: number;
+  createDate: string;
+}
+
 // This component outputs 'Nodes not ready' or the actual number of nodes in the cluster.
-function ClusterDashboardNodes({ numberOfNodes, createDate }) {
+function ClusterDashboardNodes({
```
axbarsan: Same here as below, with the `React.FC` generic

src/components/Cluster/ClusterDetail/NodePool.tsx
```diff
@@ -52,41 +53,108 @@ const NameWrapperDiv = styled.div`
...
 `;
 
-class NodePool extends Component {
-  state = {
+interface INPViewAndEditName {
+  activateEditMode: () => boolean;
+  name: string;
+}
+
+type TNPViewAndEditName = React.Component<INPViewAndEditName>;
```
axbarsan: Types don't have to be prefixed with the `T` letter, only interfaces need to have the `I` prefix

src/components/Cluster/ClusterDetail/NodePool.tsx
```diff
@@ -52,41 +53,108 @@ const NameWrapperDiv = styled.div`
...
+
+interface IStateProps {
+  availableZonesGridTemplateAreas: string | undefined;
```
axbarsan:
```suggestion
  availableZonesGridTemplateAreas?: string;
```

src/components/Cluster/ClusterDetail/NodePool.tsx
```diff
@@ -52,41 +53,108 @@ const NameWrapperDiv = styled.div`
...
+  cluster: object;
```
axbarsan: Object is the base type of everything in JavaScript, a `Record` would be more suitable

src/components/Cluster/ClusterDetail/NodePool.tsx
```diff
@@ -52,41 +53,108 @@ const NameWrapperDiv = styled.div`
...
+interface INodePoolProps extends IStateProps, IDispatchProps {}
+
+interface IScaleNodePoolModal {
+  reset: () => void;
+  show: () => void;
+  setNodePool: (nodePool: INodePool) => void;
+}
+
+class NodePool extends Component<INodePoolProps, INodePoolsState> {
+  static propTypes: IStateProps & IDispatchProps = {
```
axbarsan:
```suggestion
  static propTypes: INodePoolProps = {
```

src/components/Cluster/ClusterDetail/NodePool.tsx
```diff
@@ -52,41 +53,108 @@ const NameWrapperDiv = styled.div`
...
+    provider: PropTypes.string,
+  };
+
+  public state: INodePoolsState = {
     isNameBeingEdited: false,
   };
 
-  toggleEditingState = (isNameBeingEdited) => {
+  private viewEditNameRef: React.RefObject<TNPViewAndEditName> | null = null;
+  private scaleNodePoolModal: React.RefObject<
```
axbarsan: These aren't actually of type `RefObject`. `RefObject` is the type of ref we use with `ref.current`.
These are component refs, which make the class instance of the component be accessible from the parent component. That's why we can use `this.viewEditNameRef.activateEditMode()`.

Having these as `TNPViewAndEditName | null` would be more suitable

src/components/Cluster/ClusterDetail/NodePool.tsx
```diff
@@ -134,13 +203,15 @@ class NodePool extends Component {
...
+                  current < desired ? theme.colors.goldBackground : undefined,
               }}
             >
               {current}
             </NodesWrapper>
             {/* Applying style here because is super specific for this element and can't use nth-child with emotion */}
             <NodePoolDropdownMenu
+              // This is a div, but still TS doesn't like 'style' attr...
+              // @ts-ignore
               style={{ justifySelf: 'right' }}
```
axbarsan: The error is there because the component doesn't actually use that prop
https://github.com/giantswarm/happa/blob/318e6957209e571a06abb68776cbb9f6e7aad2c9/src/components/Cluster/ClusterDetail/NodePoolDropdownMenu.js

src/components/Cluster/ClusterDetail/NodePool.tsx
```diff
@@ -52,41 +53,108 @@ const NameWrapperDiv = styled.div`
...
+    }),
+    // @ts-ignore
+    nodePoolActions: PropTypes.object,
+    // @ts-ignore
+    dispatch: PropTypes.func,
+    // @ts-ignore
+    provider: PropTypes.string,
+  };
+
+  public state: INodePoolsState = {
```
axbarsan: I forgot to add this to the example

```suggestion
  public readonly state: INodePoolsState = {
```

src/shared/types.ts
ograu: 👍

src/components/UI/ErrorFallback.tsx
ograu: Thanks!

ograu: I haven't found any exemples with function declarations instead of function expressions. Is there a way to do it with function declarations?

axbarsan: That's because it isn't the "recommended" way of using it

I suppose something like `function ErrorFallback({ error, children }: IErrorFallbackProps): React.ReactElement` would work, but I haven't checked

src/components/Cluster/ClusterDetail/NodePool.tsx
ograu: We already have a `NPViewAndEditName` for the styled component... Any other onvention we can use here?

src/components/Cluster/ClusterDetail/NodePool.tsx
ograu: Cool!

src/components/Cluster/ClusterDetail/NodePool.tsx
axbarsan: Maybe `NPViewAndEditNameComponent`?

ograu: I meant for types, sorry. Any arguments against using T prefix for types? I assume it's not common practice, but ¿Why we shouldn't use T?

axbarsan: Well since any other class or function represent a type, with that logic, we'd need to have the `T` prefix for classes and functions, as well... which is kind of annoying.

With other words: we'd use the `User` name when creating a `User` class (which also has the `User` type), it would be kind of weird and unnecessary to use `TUser`. Same with types

ograu: Makes sense. TY!

oponder:
> I'm not used to typed languages, so using TS feels superweird to me. I think it adds a lot of noise and it makes more difficult to focus on getting things done. I think this is just a phase that I would overcome after some weeks working with TS.

I was in the same boat while learning Go, coming from Ruby and PHP. It felt really weird and kind of annoying that I had to specify everything up front.

But after doing that for a while and then going back to working on Happa in Javascript,  I started to miss some of the guarantees and editor features that a typed language gave me.

What I think is also really neat about Typescript is that we can apply the typing where we need it, remaining pragmatic, and getting kind of 'best of both worlds' in some way.

src/components/Cluster/ClusterDetail/NodePool.tsx
ograu: If I don't want to specify the shape because I haven't an interface for it yet, isn't `object` best than `any`?

axbarsan: Oops, I was mistaking the `object` type with the `Object` type. The `Object` type is the one that defines all objects, and the `object` type defines all non-primitive types. So both `Record` and `object` alright here, for now

ograu: hehe, at least I got something right! 😅

ograu:
> I was in the same boat while learning Go, coming from Ruby and PHP. It felt really weird and kind of annoying that I had to specify everything up front.

My emotional side is telling me all the time _Typescript sucks!!!_  😡,  while my head is like _Take it easy, it will take some time_ 🎶 ☯️ 😌 

> What I think is also really neat about Typescript is that we can apply the typing where we need it, remaining pragmatic, and getting kind of 'best of both worlds' in some way.

Yes, that's definitely supercool, we can define how much TS we want to use. Otherwise migrations would become impossible.

ograu:
> Looking really good for now! Just left a couple of pointers

Thanks @axbarsan!
Finally it was more than "a couple"! 😬 🙏

src/components/UI/ErrorFallback.tsx
```diff
@@ -16,7 +16,10 @@ interface IErrorFallbackProps {
  * This component will return the error passed to it,
  * if it exists, or the children, if it does not
  */
-function ErrorFallback({ error, children }: IErrorFallbackProps) {
+function ErrorFallback({
```
axbarsan: Not a blocker from me, but why don't you like using the function expression? 😬

src/components/Cluster/ClusterDetail/NodePool.tsx
```diff
@@ -176,7 +178,7 @@ class NodePool extends Component<INodePoolProps, INodePoolsState> {
         <NameWrapperDiv
           style={{ gridColumn: isNameBeingEdited ? '2 / 9' : undefined }}
         >
-          <NPViewAndEditName
+          <NPViewAndEditNameStyled
             // @ts-ignore
```
axbarsan: Is this needed?

src/components/Cluster/ClusterDetail/NodePool.tsx
```diff
@@ -52,41 +53,110 @@ const NameWrapperDiv = styled.div`
...
-  toggleEditingState = (isNameBeingEdited) => {
+  private viewEditNameRef: NPViewAndEditName | null = null;
+  private scaleNodePoolModal: IScaleNodePoolModal | null = null;
+
+  toggleEditingState = (isNameBeingEdited: boolean): void => {
     this.setState({ isNameBeingEdited });
   };
 
   triggerEditName = () => {
+    // @ts-ignore
```
axbarsan: I believe this isn't needed anymore

src/components/Cluster/ClusterDetail/NodePool.tsx
```diff
@@ -52,41 +53,110 @@ const NameWrapperDiv = styled.div`
...
     this.viewEditNameRef.activateEditMode();
   };
 
-  editNodePoolName = (name) => {
+  editNodePoolName = (name: string): void => {
     const { cluster, nodePool } = this.props;
 
     try {
       this.props.dispatch(
+        // @ts-ignore
```
axbarsan: Same here

src/components/UI/ErrorFallback.tsx
ograu: I'm ok with both, I tend to use function declarations in general because of the hoisting and because I found them more readable at the top level. That's it :-)

src/components/Cluster/ClusterDetail/NodePool.tsx
```diff
@@ -165,39 +234,20 @@ class NodePool extends Component {
...
-    }),
-  }),
-  nodePoolActions: PropTypes.object,
-  dispatch: PropTypes.func,
-  provider: PropTypes.string,
-};
-
-function mapStateToProps(state, ownProps) {
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function mapStateToProps(state: Record<string, any>, ownProps: INodePoolProps) {
```
axbarsan:
```suggestion
function mapStateToProps(state: Record<string, any>, ownProps: INodePoolProps): IStateProps {
```

src/components/UI/ErrorFallback.tsx
axbarsan: I don't think we'll have any problems with using it this way, but syntactically, it's no longer a react component, it's just a function that returns a react element 🤔... Guess we'll find out in the future if something breaks, it's not a hard fix

ograu: A function that returns a react element it is a component, a functional component, isn't it? That'a the whole move of hooks AFAIU.

And it is basically the exact same than a function expression that returns a react element, with some little differences.

src/components/Cluster/ClusterDetail/NodePool.tsx
ograu: `IStateProps` have more properties defined...

Does it makes sense to have a type for this if we are defining this prop in the props interface?

src/components/Cluster/ClusterDetail/NodePool.tsx
ograu: It looks like...

src/components/UI/ErrorFallback.tsx
axbarsan: Functionality-wise, expressions vs function components are the same.

What I meant that for TypeScript, `React.FC`, which is the function component type, is not the same as the type of the function declared here. So any function/generic type/etc that takes a react component (being it `FC` or `ClassComponent`), won't work with the function above.

That's why I said that we should see in the future if it will throw any errors

src/components/Cluster/ClusterDetail/NodePool.tsx
axbarsan: Probably because the `ViewAndEditName` UI component, isn't typed 😢

src/components/Cluster/ClusterDetail/NodePool.tsx
axbarsan: But `IStateProps` is supposed to contain only the props that are retrieved from the redux state.
It is useful to have them defined separately for each component (`INodePoolProps`, `IStateProps`, `IDispatchProps`), because it's easier to know while developing where the props are coming from (if it's an own prop, or an action, or a prop coming from the state)

src/components/UI/ErrorFallback.tsx
ograu: It looks like TypeScript doesn't support this: https://github.com/microsoft/TypeScript/issues/22063

So we have to use function expressions.

src/components/Cluster/ClusterDetail/NodePool.tsx
ograu: 👍 ok, I get it, thanks

src/components/Cluster/ClusterDetail/NodePool.tsx
ograu: Yep

PR #1283, opened 2020-03-30
(version picker) Show the button loading when the request is in flight.

oponder:
This puts the button in the loading state after pressing submit.


oponder:
> Awesome job @ograu !

Indeed! Thanks @ograu :D

ograu:
🤗

PR #1284, opened 2020-03-30
Fix kubernetes release label

ubergesundheit:
This PR fixes the stuck "kubernetes version" label in the cluster creation components. This happened because the state containing the version was only updated on component creation.

This PR also contains some style refactorings. It removes some unused html tags and inline style.


src/components/Cluster/NewCluster/ReleaseSelector.js
```diff
@@ -23,22 +23,37 @@ const FlexRowDiv = styled.div`
...
+      kubernetesVersion: this.extractKubernetesVersion(),
+      loading: false,
+    });
+  }
 
-    this.setState({ kubernetesVersion, loading: false });
+  componentDidUpdate({ selectedRelease: prevRelease }) {
+    if (prevRelease !== this.props.selectedRelease) {
+      this.setState({ kubernetesVersion: this.extractKubernetesVersion() });
+    }
```
axbarsan: Not a blocker from me but this is the perfect use case for React's built-in `gerDerivedStateFromProps`
https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops

Also, the `loading` state can be `false` by default

src/components/Cluster/NewCluster/ReleaseSelector.js
```diff
@@ -23,22 +23,37 @@ const FlexRowDiv = styled.div`
...
   }
 
+  extractKubernetesVersion = () => {
+    const { selectedRelease, releases } = this.props;
+
+    return releases[selectedRelease]?.components.find(
+      (component) => component.name === 'kubernetes'
+    ).version;
```
axbarsan: Can the component named `kubernetes` not be there? Asking because `find()` will return `undefined` in that case, and trying to get the`version` property will throw an error

src/components/Cluster/NewCluster/ReleaseSelector.js
axbarsan: Now that I think about it more, I think we can actually remove the `kubernetesVersion` property from the local state completely.
The release version could be extracted in the `render()` method, and the `extractKubernetesVersion` method could be memoized, to prevent unnecessary updates.

What do you think?

src/components/Cluster/NewCluster/ReleaseSelector.js
ubergesundheit: I think it should always be there, but I'll add `?.` for good measure

ubergesundheit:
Suggestion: I'll try to refactor this into an functional component with `useMemo` for the kubernetes version label and `useRef` for the `ReleaseDetailsModal` ref

oponder:
> Suggestion: I'll try to refactor this into an functional component with useMemo for the kubernetes version label and useRef for the ReleaseDetailsModal ref

Go for it if you want to do that! 

Though for me, just the little change of getting rid of local state for that kubernetesVersion is already good enough!

ubergesundheit:
Yeah, the complete refactor can be done sometime else. For now I removed the local state including `loading` since it seemed to serve no purpose here.

PR #1285, opened 2020-03-30
Refactor appActions to be more in line with current code style.

oponder:
Refactors `appActions.js` to not use promises or throw any errors. Using a work around of returning an error message object for now.


oponder:
this has drifted a bit. will do a new PR

PR #1287, opened 2020-04-01
Responsive nav with hamburger

ograu:
mobile-nav


src/components/UI/Navigation/MainMenu.js
```diff
@@ -0,0 +1,187 @@
...
+    &:last-child {
+      margin-right: 0;
+    }
+
+    &:hover {
+      color: ${(props) => props.theme.colors.white1};
+    }
+  }
+
+  @media only screen and (max-width: ${(props) =>
```
axbarsan, line 36: Why is this not the same as the component below?
```javascript
${mq(CSSBreakpoints.Large)} {
  display: flex;
}
```

ograu: because 🤦‍♂!

src/components/UI/Navigation/Hamburger.js
```diff
@@ -0,0 +1,79 @@
...
+    margin: 2px 1px;
+  }
+
+  &.open {
+    .first,
+    .third {
+      width: 17.5px;
+    }
+    .first {
+      transform: rotate(35deg) translate(2.3px, 4.9px);
```
axbarsan, line 39: image

Currently when the hamburger becomes an `X`, the `X` is not centered anymore into the button.

This worked for me:
```suggestion
      transform: translate3d(0, 6px, 0) rotate(35deg);
      transform-origin: center;
```
(using `translate3d` because it is GPU accelerated)

and then for the `.third`:
```css
transform: translate3d(0, -6px, 0) rotate(-35deg);
transform-origin: center;
```

axbarsan:
It looks like it's not the same padding in both sides of the navigation bar (iPhone X screen size)
image

Also, the menu entries are a bit too small to push with the finger, in my opinion. Maybe I just have fat fingers 🤷‍♂

src/components/UI/Navigation/MainMenu.js
```diff
@@ -0,0 +1,187 @@
...
+import Hamburger from './Hamburger';
+
+const NavDiv = styled.div`
+  float: left;
+  padding-left: 20px;
+  height: 40px;
+
+  & > a {
+    text-decoration: none;
+    color: #eee;
```
axbarsan, line 24: You can use the theme to store this

ograu:
> It looks like it's not the same padding in both sides of the navigation bar (iPhone X screen size)
> image
> 
> Also, the menu entries are a bit too small to push with the finger, in my opinion. Maybe I just have fat fingers 🤷‍♂

Yeah, I noticed it...  Will try to add different styles for < medium

ograu:
What about this @axbarsan ?

image

src/components/UI/DropdownMenu.js
```diff
@@ -61,12 +61,12 @@ const MenuWrapper = styled.div`
...
 function DropdownMenu(props) {
-  const [isOpen, setIsOpen] = useState(true);
+  const [isOpen, setIsOpen] = useState(false);
   let timeOutId = null;
 
   const onBlurHandler = () => {
     timeOutId = setTimeout(() => {
-      setIsOpen(true);
+      setIsOpen(false);
     });
```
axbarsan: What should this timeout do? It has no timeout period

ograu: IIRC this is to put this call at the end of the queue, this way we can first check if any child has focus before closing the menu. This is an old JS hack.

ograu: If you don't specify time, it is the same as to say "on the next tick", check it out: https://reactjs.org/docs/accessibility.html

axbarsan: I know what setTimeout(..., 0) does, but I didn't know that `0` was the default value of `setTimeout` 👍

marians:
Great to have the responsive menu!

There are two design decisions in this PR which I am not so fond of:

- The hamburger icon where the middle bar is wider than the top and bottom. To my eyes this looks weird and broken.
- Once the menu is open, the close button shape is wider than it's high. Also the icon inside has a rectangular shape that is wieder than high. I would appreciate a square shape with the icon being as high as it's wide.

oponder:
Marian made the comments i wanted to make :) other than that it looks good to me.

ograu:
@marians please, take a look at this whenever you can. Thanks!

squared-hamburger (1)

oponder:
Oh I did just notice that the version picker styles are a bit broken now, at least at the version that was deployed to ginger.

https://happa.g8s.ginger.eu-west-1.aws.gigantic.io/styleguide
Screenshot 2020-04-02 at 2 57 51 PM

For reference, here's what they looked like before:

Screenshot 2020-04-02 at 2 58 30 PM

ograu:
> Oh I did just notice that the version picker styles are a bit broken now, at least at the version that was deployed to ginger.
> 
> https://happa.g8s.ginger.eu-west-1.aws.gigantic.io/styleguide
> Screenshot 2020-04-02 at 2 57 51 PM
> 
> For reference, here's what they looked like before:
> 
> Screenshot 2020-04-02 at 2 58 30 PM

Thanks, fixing it.

PR #1290, opened 2020-04-02
Run typechecking before committing and as a CI step

axbarsan:
This is an extra step, to make sure that our TypeScript code has no issues.


package.json
```diff
@@ -183,7 +184,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "yarn precommit"
+      "pre-commit": "yarn typecheck && yarn precommit"
```
axbarsan, line 187: Doing it like this because `tsc` does not take files as input, so it cannot be used by `lint-staged`

PR #1291, opened 2020-04-02
Happa home mobile changes

ograu:
Just a couple of minor changes in mobile view to make better use of available view height:

- I've moved the _get started_ button to the top and to the right, and have made it smaller
- I've reduced the space above the _launch new cluster_ button

Before:

image

After:

image

### DIFF
--- a/src/components/Home/ClusterDashboardItem.js
+++ b/src/components/Home/ClusterDashboardItem.js
@@ -121,6 +121,10 @@ const ClusterDetailsDiv = styled.div`
   ${mq(CSSBreakpoints.Medium)} {
     height: unset;
   }
+  ${mq(CSSBreakpoints.Small)} {
+    display: inline-block;
+    transform: translateY(-1px);
+  }
 `;
 
 function ClusterDashboardItem({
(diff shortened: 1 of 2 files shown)


ograu:
Just pushed a change in the media query to reduce top padding in app when the menu is changed to the hamburger menu:

image

oponder:
I have a small design comment, other than that looks good to me.

I feel like the padding of the box should be maintained, right now the Get Started button feels a bit close to the edges for me.

Screenshot 2020-04-02 at 10 06 19 PM

oponder:
And... maybe we should just get rid of the get started button for narrow screens. Since you probably won't be able to 'Get Started' from such a small screen anyways.

axbarsan:
What do you think about making `Launch New Cluster` have a sticky position? (remain on top while scrolling)

axbarsan:
As for the `Get Started` button, I would remove the text from it, and just leave the button with the `play` icon

oponder:
> I would remove the text from it, and just leave the button with the play icon

That would make me think a bit too long about what this button does. I'd prefer to make it smaller and have less padding than removing the button text totally

ograu:
I think a play icon is not super self-explanatory, it is not obvious what it does.

As for the sticky _launch new cluster_ I think it is an interesting idea, the question for me is ¿Is this button so important that we want it to be _always_ in screen?

ograu:
What about this?

image

PR #1292, opened 2020-04-03
Add tooltips to explain DESIRED node count

marians:
Related to https://github.com/giantswarm/giantswarm/issues/8932

### Previews

Node pools table

image

V4 cluster details

image

### DIFF
--- a/src/shared/constants/index.js
+++ b/src/shared/constants/index.js
@@ -16,4 +16,8 @@ export const Constants = {
   DEFAULT_NODEPOOL_NAME: 'Unnamed node pool',
   AWS_V5_VERSION: '10.0.0',
   AZURE_MULTI_AZ_VERSION: '11.1.0',
+
+  // UI labels
+  DESIRED_NODES_EXPLANATION:
+    'Autoscaler&apos;s idea how many nodes would be required for the workloads',
 };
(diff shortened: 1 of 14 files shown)


src/shared/constants/index.js
```diff
@@ -16,4 +16,8 @@ export const Constants = {
   DEFAULT_NODEPOOL_NAME: 'Unnamed node pool',
   AWS_V5_VERSION: '10.0.0',
   AZURE_MULTI_AZ_VERSION: '11.1.0',
+
+  // UI labels
+  DESIRED_NODES_EXPLANATION:
+    'Autoscaler&apos;s idea how many nodes would be required for the workloads',
```
ograu, line 22: Backticks would do the trick:

```suggestion
`Autoscaler's idea how many nodes would be required for the workloads`
```

marians: Maybe, but the character you are suggesting there is not the apostrophe I want. It's a single quotation mark, which doesn't make for typographically appealing apostrophe.

marians: I could just use the unicode character, but think this here is more robust, no?

marians: Damn, now I see that it actually fails to render properly.... didn't even look at my own screenshot.

oponder: The grammar here was a bit weird for me, but I was too late to comment.

I think it should've been "Autoscaler's idea _of_ how many many..."

PR #1293, opened 2020-04-03
Clarify app configuration fields.

oponder:
I've been working on some documentation around app configuration and some domain language is emerging that I'd like to carry through in the UI as well.

The way I labeled the configuration fields was confusing at best and misleading at worst.

This might be a lot more verbose, but the intention is to explain that you are uploading a file containing values for a specific configuration level. I will eventually add something like "click here for an example file" or "click here for docs" to explain that further.

The toasts are also adjusted to be very explicit about what is happening.


oponder:
Towards: https://github.com/giantswarm/giantswarm/issues/9035

PR #1310, opened 2020-04-06
Clusters list performance improvements

ograu:
PoC for this: https://github.com/giantswarm/giantswarm/issues/9909

Basically here I am demonstrating points 1 and 2

~~FYI: I've removed the code that sorted  clusters based on names. We can add this later if we really want to merge this.~~ Already added sorting.

Before (each number represents a re-render):

image

After:

image

For context, Happa has two clusters in this list:

image


oponder:
Nice improvement!!

The thing I am really curious about (and not sure how to answer) is do these re-renders represent an actual "Draw", because it's my understanding that that is the costly part. 

Or do these re-renders just represent React generating a new shadow dom diff, to see that nothing changed, and so no costly draws are executed? 

Still great to do less code execution in general.

ograu:
Good question! AFAIK the reconcilation process have to steps: render an commit. The commit  is the one were DOM mutations are made, so not sure that every re-render actually means a re-painting. Will dig further tomorrow, now I'm AFK.

ograu:
Yes, [`render()`does not actually re-render the elements](https://reactjs.org/docs/reconciliation.html#motivation), it recreates the (virtual) dom tree so React can do the diffing thing and decide if a re-paint has to happen or not. So probably, if there isn't anything noticeable from the user's perspective, we should not pay too much attention to the number of times `render()` is fired.

ograu:
Besides that, another thing to keep in mind is that React batches DOM changes when state or props are modified in a short period of time.

oponder:
> we should not pay too much attention to the number of times render() is fired

It's still worth it to reduce the number of unnecessary code executions. It might be that you fixed something that could really spiral out of control when there are very many clusters or something in the future. So I think it's worth it! Thanks!

src/components/Home/Home.js
```diff
@@ -77,61 +80,61 @@ class Home extends React.Component {
   }
 
   render() {
+    console.log('hey');
```
axbarsan, line 83:
```suggestion
```

src/selectors/clusterSelectors.js
```diff
@@ -13,8 +13,20 @@ import {
...
+  return state.entities.clusters.items[id];
+};
+
+const selectOrganizationClusterNames = (state) => {
+  const clusters = state.entities.clusters.items;
+  const clusterIds = Object.keys(clusters);
+
+  return clusterIds
+    .filter((id) => clusters[id].owner === state.main.selectedOrganization)
+    .reduce(
```
axbarsan, line 26: We can do the filtering in `reduce`'s lambda body, so we won't do the extra loop

ograu:
> LGTM
> 
> Also, there's a more elegant way of checking, instead of console logs:
> https://github.com/welldone-software/why-did-you-render

I think that this is for PureComponents, IIRC.

src/components/Home/Home.js
ograu: 🤦‍♂ not the best way definetily, but the quicker :-P
Thanks for noticing!

src/selectors/clusterSelectors.js
ograu: I've been about to not use `reduce`, just for readibility. If we add the check inside I think this will be less readable...

axbarsan:
> I think that this is for PureComponents, IIRC.

It works for regular ones, as well

https://github.com/welldone-software/why-did-you-render/blob/master/README.md#tracking-components

ograu:
Yeah, it requires some setup, though. I've actually had enough info using the profiler extension (and the `console`, I admit it 😬 )

src/selectors/clusterSelectors.js
axbarsan: To me this is equally as readable, and we're not executing any extra loop

```javascript
const selectOrganizationClusterNames = (state) => {
  const { selectedOrganization } = state.main;
  const clusters = state.entities.clusters.items;
  const clusterIDs = Object.keys(clusters);

  let cluster = {};

  return clusterIDs.reduce(
    (accumulator, id) => {
      cluster = clusters[id];

      if (cluster.owner !== selectedOrganization) {
        return accumulator;
      }

      return [...accumulator, { id, name: cluster.name }];
    },
    [],
  );
};
```

ograu: I thinks that this is less readable, less semantic and less elegant. And from a performance point of view, we are not looping through a huge array here, so I really think it makes no difference 🤷‍♂

PR #1312, opened 2020-04-08
Add more tooltips to node pools list

marians:
This adds more column header tooltips to the node pools table.


marians:
Failing with 

```
please upgrade the NODE_VERSION variable to v12.16.2
```

I will try to do that in CircleCI.

PR #1313, opened 2020-04-08
Change the wording from "nodePool" to "node pool" in the node pool deletion confirmation dialog

marians:


ograu:
I think we have already changed this in the other modal PR

PR #1314, opened 2020-04-08
Node pools messages in delete modal

ograu:
This is something requested by a user: [check the slack thread](https://gigantic.slack.com/archives/C03LYSJL0/p1586276256002400) where this was discussed.

Before (no matter how many node pools the cluster has):

image

After with two or more node pools:

image

With just one node pool:

image

### DIFF
--- a/__tests__/V5ClusterManagement.js
+++ b/__tests__/V5ClusterManagement.js
@@ -91,7 +91,7 @@ it('renders all node pools in store', async () => {
 
   await waitFor(() => findAllByTestId('node-pool-id'));
 
-  nodePoolsResponse.forEach(nodePool => {
+  nodePoolsResponse.forEach((nodePool) => {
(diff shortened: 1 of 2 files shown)


marians:
The sentence "This node pool is not sharing any node labels with other node pools." should be removed. It was specked with a feature (custom node labels per node pool) in mind that has not made it into the scope.

marians:
I like how you are moving the node pool ID and name out of the dialog title into the body! This helps a lot.

Actually I think we should also show that information when deleting the only existing node pool. It can still be valuable when users try to reason if they should really do this.

ograu:
Modified message and added label to the multiple node pools modal. The images in the description are updated, ping @marians

oponder:
> I like how you are moving the node pool ID and name out of the dialog title into the body! This helps a lot.

That's indeed a really nice change!

PR #1316, opened 2020-04-08
Add new action for refreshing clusters list

ograu:
This PR address this: https://github.com/giantswarm/giantswarm/issues/9874

Currently, clusters list refreshes when clusters are deleted, but doesn't when clusters are added by another user/interface. This is because we decided to remove clusters list fetching from the refresh batched function because we were overwriting the store each time we fetched.

This PR attemps to change this - i.e., make clusters list responsive to any addition - by adding a dedicated action that will add new clusters to the store _without_ touching the clusters that are already in store.


ograu:
> Though I think at some point we could maybe try to solve the problem while having only one 'clustersList' action. That would be ideal for me, and my gut tells me it should be possible.

I have gave it a try.

It is possible, but after been playing with the idea of merging whatever we get in the new `getClusters` with what we already have in the store, I think that it's not something advisable. Yes, we could have one `clustersList` function that would handle everything, but this would result in a long function with quite a lot of conditional logic in it. 

This function would have to:

- dispatch different actions depending of we are on init or refreshing
- keep track of properties inside every cluster that we do not want to overwrite, basically what's inside of `clusterLoadDetails`: `status`, `workers`, 'capabilities`, `provider` and `scaling`

OR

We could copy what I've done in `refreshClustersList` and paste it inside the other function, resulting in a function that would be bigger than the sum of this two.

PR #1317, opened 2020-04-09
App readme "backend" changes.

oponder:
This adds the actions and logic for fetching readmes when visiting the App Detail page.

It doesn't implement showing the README in the component yet, wanted to split the PR's to separate the concerns a bit.

It also changes the AppDetail page's URL to include the version of the App, and makes the component actually use that appVersion. But it does not include a way for users to change the version from the interface yet. Though that'll be relatively easy to do now that the component cares about the version in the URL. (for another PR)

towards: https://github.com/giantswarm/giantswarm/issues/8828


src/reducers/catalogReducer.js
```diff
@@ -36,6 +36,16 @@ const catalogReducer = produce((draft, action) => {
...
+
+      return;
+
+    case types.CLUSTER_LOAD_APP_README_SUCCESS:
+      if (!draft.items[action.catalogName]) return;
+      if (!draft.items[action.catalogName].apps[action.appVersion.name]) return;
+
+      draft.items[action.catalogName].apps[action.appVersion.name].find(
+        (av) => av.version === action.appVersion.version
+      ).readme = action.readmeText;
```
axbarsan, line 48:
```suggestion
      draft.items[action.catalogName]?.apps[action.appVersion.name]?.find(
        (av) => av.version === action.appVersion.version
      )?.readme = action.readmeText;
```

src/components/AppCatalog/AppDetail/AppDetail.js
```diff
@@ -29,18 +40,42 @@ class AppDetail extends React.Component {
...
+  componentDidMount() {
+    this.loadReadme();
+  }
+
+  componentDidUpdate() {
+    this.loadReadme();
+  }
+
+  // Dispatch an action to load readme associated with this app if it has one and
+  // it hasn't been loaded yet.
```
axbarsan, line 52:
```suggestion
  /**
   * Dispatch an action to load readme associated with this app if it has one and
   * it hasn't been loaded yet.
   */
```

src/actions/__tests__/appActions.js
```diff
@@ -0,0 +1,143 @@
+import '@testing-library/jest-dom/extend-expect';
```
axbarsan, line 1: I believe these tests only assert that `redux` works, which we don't really need to do, as `redux` is an external library and has its own tests in its codebase.

Why not actually test the actions together with the components that use them?

oponder: Thanks for reviewing!

These tests actually run the thunk and check that it dispatches actions with the expected contents depending on the arguments.

Redux and the reducer isn't tested here.

So this is a nice little isolated test of the logic in the action, and exercises mostly our own business logic, like the bits of code that check for a README url, and change the README url if it is a URL of a test app.

oponder:
I'll fix the tests tmw :P

src/actions/__tests__/appActions.js
axbarsan: Thunks and action dispatching are still part of redux. So testing if specific actions are dispatched with the right payload still means we're testing redux.

The only part of business logic that is being tested here is that an action does in fact trigger an API request. I don't believe we actually need this, since we'll test the same functionality in component tests.

However, if for some reason we actually need to test the actions, we can treat them as regular functions, there's no need for mock stores.
Example:
https://github.com/giantswarm/happa/blob/d7837f028dc1820e95c74e1a92e650d0f1316151/src/actions/appActions.js#L303
Here we return a higher order function, which we use to inject a mock `dispatch` function, which we can use for testing.


Maybe there's something that I'm missing

oponder: > Thunks and action dispatching are still part of redux. So testing if specific actions are dispatched with the right payload still means we're testing redux.

I don't think we're testing redux. But maybe we say the same thing in different ways. In my point of view, "testing redux" would mean having a test that doesn't really touch any of the app logic and just exercises some trivial part of redux's logix.

But in this case, the type of actions dispatched from this Thunk, the order in which they are dispatched, and the actual values they contain are all being tested here, which is really all business logic specific to this app, and has nothing to do with the redux library (which is not under test anyways, what we are using here is a mock of the redux store, so that we can check our expectations on it at the end of the test).

I think we could do that by mocking dispatch as well if you prefer, but this was the documented way of testing actions, and it works fine from a mental model point of view, to test things this way.

Are you opposed to the extra dependency on the `redux-mock-store`?

The only thing we're "using" from a redux implementation point of view is that actions dispatched against a store get processed in order. The mock library lets us fetch them all using `getActions` and compare them against an expected array of actions. 

>  I don't believe we actually need this, since we'll test the same functionality in component tests.

I think these focused 'action tests' we have a chance to get into the real nitty gritty of what the action should do,  checking all sorts of edge cases, without firing up the heavier component tests.

In the component (but actually integration) test we tend to focus on the happy path, and some general error cases, to save some cpu cycles.

axbarsan: I was sure I actually replied to this, but I guess I didn't press the big green button.

I was actually wrong in my previous comment, as I actually just ran into a similar problem now, with the happa update warning, where it would be more reliable and efficient to test the actions directly.

I'm all for this now 👍 

Also, about `redux-mock-store`, it is a bit unnatural for me to test this way, but I guess if we still need to take middleware into consideration, we have to do it this way.

But some food for thought for the future:
For our future implementations (with the proprietary model layer), I believe it would be safer to have business logic to the service part of the model layer, as having it all in actions is a bit too fragile (imagine the pain of refactoring all tests once we replace redux)

oponder: Yeah I realised as I was writing the code here that this wasn't really in line with the new architecture, I think I'll try to make a service for this still in this PR so that I get some practice.

Then indeed the specific readme URL processing tests could move to the service, and then these action tests would become pointless.

src/reducers/catalogReducer.js
```diff
@@ -40,12 +40,9 @@ const catalogReducer = produce((draft, action) => {
       return;
 
     case types.CLUSTER_LOAD_APP_README_SUCCESS:
-      if (!draft.items[action.catalogName]) return;
```
oponder: I'll put this back.

It's a bit clearer what the intent is in such a long line. And also easier syntax to get right.

oponder:
@axbarsan or @giantswarm/chapter-front-end can I get an official approval?

PR #1318, opened 2020-04-09
Restructure CI to run checks in parallel

ubergesundheit:
Refactoring of the ci process with the goal to make it faster.

Changes (so far):

- don't use make
- separate linting/typechecking/stylechecking/testing/building into separate jobs
- change workflow to execute jobs in parallel
- execute jobs before `container-build` in docker executor on circle ci
- execute `push-happa-to-aliyun-master` only for tags and master builds
- update architect-orb to 0.8.9
- remove duplicated docker run of `latest` container (I think this is a leftover from lighthouse testing)

This removes the version check we had before this.

Once merged, the branch protection should require `build-container` instead of `build`.

It could be possible that increasing the `resource_class` for all jobs could speed this up.


ubergesundheit:
CI now builds in parallel:

image

oponder:
A place to keep an eye on build times over time: https://circleci.com/build-insights/gh/giantswarm/happa/master

oponder:
should we specify parallelism now? https://circleci.com/docs/2.0/parallelism-faster-jobs/

I think it's still running possibly as one at a time?

ubergesundheit:
image

The jobs in the middle column run in parallel. The `parallelism` mentioned in the blog post relates to parallelism in steps inside jobs but I think the steps are already dependant on each other.

We should definetely try to test this though

PR #1320, opened 2020-04-10
Add ingress instructions tab to cluster detail view

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/9872

This adds the new Ingress tab in cluster details, as structured in the ticket above

## Preview

**AWS/AZURE**

localhost_7000_organizations_giantswarm_clusters_n88rh(iPad Pro)

**KVM**

localhost_7000_organizations_giantswarm_clusters_n88rh(iPad Pro) (1)


oponder:
Nice work implementing, 

Seeing it like this though the page feels really quite dense and difficult for my eye to grab onto anything that gives me context about what I am looking at.

Some suggestions:

1. can we limit the width of the view to at most 900 px?
2. I think the vertical spacing could use some scrutiny, but that would be a bit hard to write out.
3. the "A" and "B" circles position feel slightly off balance

I know this is a bit vague feedback.. if you see what I see, then feel free to tackle it. 

Otherwise I can do a bit of css changes myself, or possible draw some lines to see if I can make my alignment issues clearer.

oponder:
One other thing popped up in my mind: Not all clusters have an ingress installed by default.

We would need to check based on the release version if there is an ingress by default or not,
and then look into the apps installed on the cluster to see if the nginx-ingress-controller is installed and then adjust the Ingress tab to show something like "No ingress installed yet" with al link to here: https://docs.giantswarm.io/guides/installing-optional-ingress-controller/

axbarsan:
The screenshots I initially provided were kind of off (I took them on a 27-inch screen). I updated the description with the new screenshots.
I also added a bit of vertical breathing room and a maximum width (so it would look relatively OK on bigger screens).

As for selectively showing the `Ingress` tab, I will create a separate issue for that, to not over-complicate what we got here.

axbarsan:
https://github.com/giantswarm/giantswarm/issues/10076

PR #1333, opened 2020-04-13
Redirect from details after deletion

ograu:
Fixes https://github.com/giantswarm/giantswarm/issues/9879

I added support for the case when the cluster is deleted: it is not `undefined`, but we don't want the user to access the details view. 

Now, on delete, the user will be redirected and a toast will notify him that the cluster has been deleted or it is not in her organizations.

Also removed redirection from `clusterActions` because I think this is not the place for this, as this action can be dispatched in different contexts.

### DIFF
--- a/src/components/Cluster/ClusterDetail/ClusterDetailView.js
+++ b/src/components/Cluster/ClusterDetail/ClusterDetailView.js
@@ -78,27 +78,37 @@ class ClusterDetailView extends React.Component {
     );
   };
 
+  itDoesntExist = (isDeleted) => {
+    const { clusterId, organizationId, dispatch } = this.props;
+
+    const organizationDetailPath = RoutePath.createUsablePath(
+      OrganizationsRoutes.Detail,
+      {
+        orgId: organizationId,
+      }
+    );
+
+    const text = isDeleted
+      ? 'This cluster has been deleted'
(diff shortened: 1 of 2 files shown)


src/components/Cluster/ClusterDetail/ClusterDetailView.js
```diff
@@ -78,27 +78,38 @@ class ClusterDetailView extends React.Component {
...
-        {
-          orgId: organizationId,
-        }
-      );
+    const organizationDetailPath = RoutePath.createUsablePath(
+      OrganizationsRoutes.Detail,
+      {
+        orgId: organizationId,
+      }
+    );
```
ograu, line 90: Do we really want to redirect the user to the organization view?  Why not redirect him to the home view?

oponder:
I think this could be a big unexpected context switch for the user. 

As a step to fix the bug, I think it's ok.

But as a user experience thing, I think the better solution would be to 'gray out' the page and make everything inactive/unusable while showing the notification, and not automatically redirect them.

They might be in the middle of reading a sentence or looking at something, and I think I would feel confused and interrupted if it to suddenly changed my context, even though the information i was looking at became useless.

src/components/Cluster/ClusterDetail/ClusterDetailView.js
oponder: I think the home view would probably be nicer yeah.

PR #1334, opened 2020-04-13
Add warning if there is a newer version of the app available

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/9708

Don't be discouraged by the number of files changed, that's only because I had to add mocks to all tests.

In this PR:
* Moved the footer out of the static index.html to a separate component
* Created a new http client for self requests (getting files from happa)
* Created a new service for getting happa metadata (such as version)
* In the `start.sh` script, we're now writing the version in the `metadata.json` file
* Added the functionality to fetch the app version from metadata after the application bootstraps
* Added the functionality to periodically (10 min intervals now) check for updates to the `metadata.json` file
* Added the functionality to update the app from a button in the app footer (which will also display a short loading animation before refreshing the page)

### Preview

**Regular hover state**

image

**Update available hover state**

image

**Updating hover state**

image

(when not hovering, the tooltips are not shown)

**Update available flash message**

image


marians:
Looking nifty! design-wise my only wish would be to give `Update now` an actual button look and feel. Ideally an outline style, and keep it really small. I think that would increase the likelihood that it's perceived as clickable. The dot between button and warning symbol should be removed then.

marians:
This thought just popped up: Users might not exactly get what is being updated. Especially when looking at a tenant cluster details page, it could be confused with the cluster being upgraded.

Can we make the wording more clear so that it indicates what is being updated? E. g. name the button `Upgrade Happa now`

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,97 @@
...
+  const releaseURL: string = getReleaseURL(currentVersion);
+  const isUpdateReady: boolean = hasUpdateReady(currentVersion, newVersion);
+
+  const handleUpdate = useCallback(() => {
+    dispatch(executeUpdate());
+  }, [dispatch]);
+
+  useEffect(() => {
+    dispatch(metadataActions.setInitialVersion());
+    metadataActions.registerUpdateChecker(dispatch);
```
oponder, line 55: Is it possible that we are registering multiple 'setTimeouts' like this?

I think we should guard against that and take the timeout ID and store it somewhere. Then check if it already exists, and either clear the existing one and set a new one, or just not set any new timeout.

axbarsan:
I made the update button look like a button and updated the screenshots

src/components/Footer/Footer.tsx
axbarsan: About the request mocks not being manually added to all tests that need it, I remember us talking about having a more granular control over which and how many requests are executed, and extracting some requests to a separate helper function kind of goes against that. We could re-address that, though.

As for the timeouts, I would usually store the timeout, but in this case, does it really make a difference? The `Footer` component is rendered at the root of the app, and the timeout is created when the `Footer` component is mounted. The only situation where the timeout would be created again would be if the whole app would be unmounted and mounted again. Is that possible though?

axbarsan: However, if our logic ever changes and we want to change where the `Footer` component is rendered, we will 99% sure forget about the timeout thing, so it would probably be safer to store it, to not have to deal with it later

oponder: True, the footer only mounts once and that effect would only fire again if 'dispatch' changes in the props right.

I guess the only remaining argument if is in some future we decide to change something in the footer or that effect which could potentially cause re-rendering, then we might forget that the effect creates a timeout and introduce an unintended side effect.

marians:
Tiny thing: I'd like to remove the exclamation mark. Not needed in interaction element labels, like buttons.

Should that button label be defined in a single place to be re-usable?

axbarsan:
> Tiny thing: I'd like to remove the exclamation mark. Not needed in interaction element labels, like buttons.
> 
> Should that button label be defined in a single place to be re-usable?

That's actually a good idea, I've updated the code and the screenshots, as well 👍

axbarsan:
Btw, I've changed to using this technique for testing timers:
https://jestjs.io/docs/en/timer-mocks

This way we don't actually have to wait a lot of time before timers finish, and we can also clear them after every test.

PR #1336, opened 2020-04-16
Fix app creation

oponder:
I broke it in a previous PR, causing the action not to have the 'catalog name' anymore.

I was actually changing `repo` to `catalogName` in some places, but reverted that change because it muddied the PR, but apparently didn't revert it good enough.


__tests__/AppCatalog.js
```diff
@@ -231,7 +231,14 @@ describe('Apps and App Catalog', () => {
         message: `We're installing your app called 'test-app' on ${V4_CLUSTER.id}`,
       };
       nock(API_ENDPOINT)
-        .intercept(`/v4/clusters/${V4_CLUSTER.id}/apps/test-app/`, 'PUT')
+        .intercept(`/v4/clusters/${V4_CLUSTER.id}/apps/test-app/`, 'PUT', {
```
oponder, line 234: This is a regression test for what happened. It makes sure that the request that Happa makes includes the expected JSON body

PR #1350, opened 2020-04-19
Use templated endpoints in index.html

ubergesundheit:
I noticed sometimes commits which changed `apiEndpoint` and `passageEndpoit` sneaked past human review. 

This PR should make it easier to change the andpoints in development by setting `HAPPA_API_ENDPOINT` and `HAPPA_PASSAGE_ENDPOINTS`. Webpack will now take care of setting the desired values. If the environment variables are missing, default values will be used.

### DIFF
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -6,12 +6,27 @@ const webpack = require('webpack');
 const HtmlWebpackPlugin = require('html-webpack-plugin');
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const { CleanWebpackPlugin } = require('clean-webpack-plugin');
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const process = require('process');
+
+const makeEndpoints = () => {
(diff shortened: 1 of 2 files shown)


webpack.common.js
```diff
@@ -6,12 +6,27 @@ const webpack = require('webpack');
...
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const { CleanWebpackPlugin } = require('clean-webpack-plugin');
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const process = require('process');
+
+const makeEndpoints = () => {
+  const { HAPPA_API_ENDPOINT, HAPPA_PASSAGE_ENDPOINT } = process.env;
+
+  return {
+    apiEndpoint: HAPPA_API_ENDPOINT
```
oponder, line 16: I think this will break when it hits this startup script right? https://github.com/giantswarm/happa/blob/master/scripts/start.sh

ubergesundheit: Nope, the start script operates on `index.html`. The regex does not care about whats coming after the `: `.

axbarsan:
This is cool 👍

PR #1351, opened 2020-04-20
On the App Detail Page, hide most chart versions, but make them expandable.

oponder:
Clicking on the "+ 7 more" label expands them all. With no way to 'recollapse' them. I think this is something that very few people will interact with and once it's open they probably don't want to close it again.

Screenshot 2020-04-20 at 9 57 25 AM

### DIFF
--- a/src/components/UI/AppDetails/VersionsRow.js
+++ b/src/components/UI/AppDetails/VersionsRow.js
@@ -1,0 +1,87 @@
+import styled from '@emotion/styled';
+import PropTypes from 'prop-types';
+import React, { useState } from 'react';
+import Copyable from 'shared/Copyable';
+import Truncated from 'UI/Truncated';
+
+const INITIAL_MAX_CHART_VERSIONS = 2;
+
+const ChartVersion = styled(Copyable)`
+  margin-bottom: 10px;
+  float: left;
+`;
+
+const AppVersion = styled(Copyable)`
+  display: inline-block;
+  float: none;
+  position: relative;
+  left: 8px;
+  code {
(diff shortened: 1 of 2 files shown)


src/components/UI/AppDetails/VersionsRow.js
```diff
@@ -0,0 +1,87 @@
...
+                copyText={appVersionObject.version}
+              >
+                <Truncated numStart={12} as='code'>
+                  {appVersionObject.version}
+                </Truncated>
+              </ChartVersion>
+            );
+          })}
+        </div>
+        <br style={{ clear: 'both' }} />
```
ubergesundheit, line 62: Could you move this `br` into the conditional or even better remove it entirely? Or do you intend to have the `+ N versions` button on a new line? I think it should be possible then to give it 100% width to achieve the same effect.

src/components/UI/AppDetails/VersionsRow.js
```diff
@@ -0,0 +1,87 @@
...
+        {entries.length > expandVersions ? (
+          <ExpandVersions onClick={expand}>
+            {`+ ${(entries.length - expandVersions).toString()} more ...`}
```
ubergesundheit, line 65:
```suggestion
            + {(entries.length - expandVersions).toString()} more &hellip;
```

src/components/UI/AppDetails/VersionsRow.js
oponder: I made some changes in general to improve the alignment and allow it fit on the same line

oponder: Screenshot 2020-04-20 at 3 14 17 PM

PR #1357, opened 2020-04-20
Show the readme if it is there and make it look ok.

oponder:
The chart versions table length is being fixed in: https://github.com/giantswarm/happa/pull/1351

localhost_7000_app-catalogs_giantswarm-test_nginx-ingress-controller-app_1 6 8-a70edadceed0c36c9ea933df39ba8e796fc6b057__q=

### DIFF
--- a/src/components/UI/AppDetails/AppDetails.js
+++ b/src/components/UI/AppDetails/AppDetails.js
@@ -126,9 +126,9 @@ function flatten(text, child) {
 }
 
 function HeadingRenderer(headingProps) {
-  var children = React.Children.toArray(headingProps.children);
-  var text = children.reduce(flatten, '');
-  var slug = text.toLowerCase().replace(/\W/g, '-');
+  const children = React.Children.toArray(headingProps.children);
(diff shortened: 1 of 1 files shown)


src/components/UI/AppDetails/AppDetails.js
```diff
@@ -95,6 +96,47 @@ const Install = styled.div`
...
+function flatten(text, child) {
+  return typeof child === 'string'
+    ? text + child
+    : React.Children.toArray(child.props.children).reduce(flatten, text);
+}
+
+function HeadingRenderer(headingProps) {
+  var children = React.Children.toArray(headingProps.children);
+  var text = children.reduce(flatten, '');
+  var slug = text.toLowerCase().replace(/\W/g, '-');
```
ubergesundheit, line 131:
```suggestion
  const children = React.Children.toArray(headingProps.children);
  const text = children.reduce(flatten, '');
  const slug = text.toLowerCase().replace(/\W/g, '-');
```

PR #1358, opened 2020-04-20
Fix a bug in the upgrade modal.

oponder:
Fix `TypeError: n.forEach is not a function`


oponder:
Now the little error case is able to trigger correctly and the modal can be rendered.

Screenshot 2020-04-20 at 4 42 25 PM

PR #1270, opened 2020-03-25
add spot instance ui stuff

ubergesundheit:
Towards giantswarm/giantswarm#8347

Related to https://github.com/giantswarm/api/pull/999

PR #1359, opened 2020-04-20
Cluster labeling

ubergesundheit:
Adds cluster label display to the Cluster list and Cluster Details. Towards  giantswarm/giantswarm#1693

Screenshots 6489ede
localhost_7000_(Laptop) (1)
localhost_7000_(Laptop)


ubergesundheit:
Since I'll be doing #1270 this week, @oponder @axbarsan could you look at this and see if the technical side is good for now.

src/components/Modals/ReleaseDetailsModal.js
```diff
@@ -74,15 +74,17 @@ class ReleaseDetailsModal extends React.Component {
...
-                          name={component.name}
-                          version={component.version}
-                        />
-                      );
-                    })}
+                    {_.sortBy(release.components, 'name').map(
+                      (component, i) => {
+                        return (
+                          <ReleaseComponentLabel
+                            key={`${component.name}-${release.version}-${i}`}
```
axbarsan, line 81: Not ideal (talking about the index), but since I believe these aren't filtered, it shouldn't be a problem for now

src/components/UI/ReleaseComponentLabel.js
```diff
@@ -91,15 +54,19 @@ const ReleaseComponentLabel = (props) => {
...
-      <VersionLabel
-        isAdded={isAdded}
-        isRemoved={isRemoved}
-        newVersion={version}
-        oldVersion={oldVersion}
-      />
-    </Wrapper>
+    <ValueLabel
+      className={className}
+      color='#395b70'
```
axbarsan, line 59: Please use the theme to fetch this color.

You can use this hook to get the theme in a non-styled component:
https://emotion.sh/docs/theming#usetheme-hook

src/components/UI/ValueLabel.tsx
```diff
@@ -0,0 +1,62 @@
...
+const CommonCSS = css`
+  padding: 5px 8px;
+  font-size: 14px;
+  font-weight: 300;
+`;
+
+const LabelWrapper = styled.span`
+  ${CommonCSS};
+  border: 1px solid ${(props) => props.color ?? '#397049'};
+  background-color: ${(props) => props.color ?? '#397049'};
```
axbarsan, line 24: Same here, these should come from the theme

src/components/Modals/ReleaseDetailsModal.js
ubergesundheit: This is only required because aws release 9.2.3 contains [`coredns`](https://github.com/giantswarm/releases/blob/master/aws.yaml#L378-L379) [twice](https://github.com/giantswarm/releases/blob/master/aws.yaml#L408-L409) and we're putting everything together (`compoents` vs `apps` in the release). This would go away over time so I'm fine with reverting this to its original state.

ubergesundheit: AWS release 9.2.4 will fix this inconsitency. I'll revert this later..

oponder:
Looks good to me, though I am not loving the green label color. So far we try to use green for confirmation buttons, could we use a different color? One of the shades of blue perhaps?

ubergesundheit:
> Looks good to me, though I am not loving the green label color. So far we try to use green for confirmation buttons, could we use a different color? One of the shades of blue perhaps?

Yeah totally :) I'll post more screenshots

ubergesundheit:
I noticed the blue foreground in the cluster overview & in cluster details were different so I took the liberty of adding another color to the theme to unify them. (See d56e8ac)

image

oponder:
Nice looking good so far! 1 more thing is the vertical spacing between the elements in that box. 

it looks like the labels have a bit more margin bottom than the text elements, so like the line heights are a bit off, ya know what I mean?

ubergesundheit:
> Nice looking good so far! 1 more thing is the vertical spacing between the elements in that box.
> 
> it looks like the labels have a bit more margin bottom than the text elements, so like the line heights are a bit off, ya know what I mean?

yeah I noticed it! I'll change them to use less margin vertically.

> Approving so you can keep moving!

Thanks, but I didn't thought far enough ahead. I need to implement label editing in here as well and also merge https://github.com/giantswarm/giantswarm-js-client/pull/109 :)

Much appreciated, though! :+1:

ubergesundheit:
I just pushed rudimentary cluster label editing in https://github.com/giantswarm/happa/pull/1359/commits/9bd7ddb0a19f95979c8242cb2307360550791a51 

This isn't styled yet and is also missing updating the store after label editing.

I'll continue to wire up the updating in the store, write some tests & then make it look beautiful.

src/components/Cluster/ClusterDetail/EditClusterLabelsModal/AddClusterLabel.tsx
```diff
@@ -0,0 +1,39 @@
+import PropTypes from 'prop-types';
+import React, { useState } from 'react';
+import Button from 'UI/Button';
+import EditableValueLabel from 'UI/EditableValueLabel';
+
+interface IAddClusterLabelProps {
+  onSave({ label, value }: { label: string; value: string | null }): void;
+}
+
+const AddClusterLabel = ({ onSave }: IAddClusterLabelProps) => {
```
axbarsan, line 10: There's a more common way to do it
```suggestion
const AddClusterLabel: React.FC<IAddClusterLabelProps> = ({ onSave }) => {
```

src/components/Cluster/ClusterDetail/EditClusterLabelsModal/EditClusterLabelsModal.tsx
```diff
@@ -0,0 +1,173 @@
...
+    updateLabel({ label: '', value: '', reset: true });
+  };
+
+  const saveLabels = useCallback(() => {
+    actions.clusterLabelsPatch(clusterId, modifiedLabels).then(() => {
+      closeModal();
+    });
+  }, [actions, clusterId, modifiedLabels]);
+
+  return open ? (
```
axbarsan, line 92: Can we use our new "return early" pattern here? (ex: if not open, return the edit button, and then the rest)

src/components/UI/LabelsList.tsx
```diff
@@ -2,22 +2,32 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import ValueLabel from 'UI/ValueLabel';
 
+interface ILabel {
```
axbarsan, line 5: I noticed you're duplicating this in multiple places. Do you mind storing it in a common place (such as model, or maybe it is already in the js client, and you can import it from there)?

axbarsan: I'd say the best place for it is in the js client, as we'll probably use all the types in there when we'll migrate our redux logic to typescript

src/components/Cluster/ClusterDetail/EditClusterLabelsModal/AddClusterLabel.tsx
axbarsan: Btw this is present in multiple places, not only in this component

ubergesundheit:
I spoke to @marians about this. We agreed to the following:

- Only display user defined labels (For now this will be implemented by hiding labels with `giantswarm.io` in keys)
- Show labels only in cluster details
- Label editing shouldn't happen in a modal, instead will happen in place
- Key & value modification will be done through two input fields

For now I'll briefly stop working on this PR and create a wireframe based on giantswarm/giantswarm#9077 

Some useful changes have been merged into master with #1413

ubergesundheit:
I'm going to close this and will do smaller PRs

PR #1366, opened 2020-04-21
Use tabs in the app detail view.

oponder:
This adds two tabs to the app detail view: "About" and "Other Versions".

The table of versions has moved fully into the "Other Versions" tab, and the "About" tab shows the Readme if it is there with the metadata in a side pane.

Screenshot 2020-04-21 at 1 54 10 PM
Screenshot 2020-04-21 at 1 54 19 PM


ubergesundheit:
looks good :+1:

PR #1381, opened 2020-04-23
Fix a race condition when opening the cluster detail view.

oponder:
This PR changes two things:

1. It makes sure that multiple entity load results are kept instead of overwritten.

For example a request to load cluster detail, followed by a request to load nodepools, the state would end up looking like

`{loadingFlagsByEntity: {abc12: {CLUSTER_NODEPOOLS_LOAD: false}}`

And we lost the fact that it used to also contain:

`{loadingFlagsByEntity: {abc12: {CLUSTER_DETAILS_LOAD: false}}`

With the change to the entityLoadingReducer, we now see:

`{loadingFlagsByEntity: {abc12: {CLUSTER_NODEPOOLS_LOAD: false, CLUSTER_DETAILS_LOAD: false}}`

2. The selectLoadingFlagByIdAndAction returns 'true' by default for any entity loading requesst that has not ever been loaded yet. 

This fixes a race condition in the cluster detail view.


oponder:
Tested, seems to work ok. 

One problem or thing to keep in mind though if we carry this 'return true for never loaded' thing is that we now can't rely on the loading flag to check if a request is already in progress. 

I don't think we do that anywhere for the cluster detail loading, but there might be some other parts of the app that check first if a request is already in progress before firing off a request action.

PR #1387, opened 2020-04-24
Adjust the guide to support optional ingress controller clusters.

oponder:
towards: https://github.com/giantswarm/giantswarm/issues/7783

A step gets inserted into the getting started guide for clusters that have an optional ingress controller:

Screenshot 2020-04-27 at 3 40 01 PM

Which presents a bit of an explanation and a convenient button for installing the ingress controller immediately.

Screenshot 2020-04-27 at 3 50 05 PM

Screenshot 2020-04-27 at 3 50 26 PM

### DIFF
--- a/src/components/GettingStarted/GettingStarted.js
+++ b/src/components/GettingStarted/GettingStarted.js
@@ -3,62 +3,143 @@ import RoutePath from 'lib/routePath';
 import PropTypes from 'prop-types';
 import React from 'react';
 import { Breadcrumb } from 'react-breadcrumbs';
+import { connect } from 'react-redux';
 import { Redirect, Route, Switch } from 'react-router-dom';
 import { OrganizationsRoutes } from 'shared/constants/routes';
 
-import Page1_ConfigureKubeCTL from './Steps/ConfigureKubectl';
-import Page3_NextSteps from './Steps/NextSteps';
-import Page0_Overview from './Steps/Overview';
-import Page2_SimpleExample from './Steps/SimpleExample';
+import ConfigureKubeCTL from './Steps/ConfigureKubectl';
+import InstallIngress from './Steps/InstallIngress';
+import NextSteps from './Steps/NextSteps';
+import Overview from './Steps/Overview';
+import SimpleExample from './Steps/SimpleExample';
 
-const GettingStarted = (props) => (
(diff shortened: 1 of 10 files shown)


oponder:
Would like some sig-ux feedback on this if possible! ping @giantswarm/sig-ux @marians

oponder:
I changed the layout of the overview page, screenshot has been updated.

src/components/GettingStarted/GettingStarted.js
```diff
@@ -3,62 +3,143 @@ import RoutePath from 'lib/routePath';
...
+        }}
+      >
+        <div>
+          <Switch>
+            <Route
+              render={() => <Overview steps={steps} {...props} />}
+              exact
+              path={OrganizationsRoutes.Clusters.GettingStarted.Overview}
+            />
+            {steps.map(({ component, routePath, title }, i) => (
```
axbarsan, line 106:
```suggestion
            {steps.map(({ component: Component, routePath, title }, i) => (
```

If we destructure the component like this, then we can use it like a regular component (ex: `<Component {...someProps} />`), instead of relying on `React.createElement`

axbarsan: Here's a doc about it, for more context: https://reactjs.org/docs/jsx-in-depth.html#user-defined-components-must-be-capitalized

src/components/GettingStarted/GettingStarted.js
```diff
@@ -3,62 +3,143 @@ import RoutePath from 'lib/routePath';
...
+        'Enable your Kubernetes CLI to access your Kubernetes cluster at Giant Swarm',
+      url: clusterGuideConfigurationPath,
+      routePath: OrganizationsRoutes.Clusters.GettingStarted.ConfigureKubeCtl,
+      component: ConfigureKubeCTL,
+    },
+
+    {
+      title: 'Run a simple example',
+      description:
+        'To make sure everything works as expected, let&apos;s start a hello world application',
```
axbarsan, line 55: Btw, this is treated as a string when used in the component, so the `&apos;` is escaped in the output.

image

oponder: Well spotted! thanks!

oponder:
Thanks for the review everyone!

PR #1422, opened 2020-05-01
Revert moment dependency upgrade.

oponder:
Something is not working will with the latest version of moment introduced here: https://github.com/giantswarm/happa/pull/1418

Reverting it.

I'd like to analyse why the tests passed and what we can do to fix it.


oponder:
Merged without review.

PR #1428, opened 2020-05-04
Handle error if Key Pair creation fails due to PKI backend not being ready

axbarsan:
Fixes https://github.com/giantswarm/giantswarm/issues/10023

* Refactored the `KeyPairCreateModal` component and split it into multiple smaller components
* Wrote unit tests for the whole functionality
* Added a custom error state for 503 errors
* Added a shake animation for when the keypair creation fails

## Preview

**Regular error**

image

**503 Error**

image

**503 Error, hovering on the PKI acronym**

image

**503 Error, demo with failure animation**

captured (4)


src/shared/types.ts
```diff
@@ -21,4 +21,11 @@ export interface INodePool {
   status: INodePoolStatus;
 }
 
+export interface IKeyPair {
```
axbarsan, line 24: This location is temporary, until we switch to a more domain-oriented folder structure

src/components/Cluster/ClusterDetail/KeypairCreateModal/KeyPairCreateModal.js
```diff
@@ -119,19 +118,11 @@ const KeyPairCreateModal = (props) => {
...
-
-    return cnPrefix;
-  };
-
   const cnPrefixValidation = (value) => {
     let error = null;
     if (value !== '') {
       const endRegex = /[a-zA-Z0-9]$/g;
-      const regex = /^[a-zA-Z0-9][a-zA-Z0-9@\.-]*$/g;
+      const regex = /^[a-zA-Z0-9][a-zA-Z0-9@.-]*$/g;
```
oponder: Example: 

The commit is called: "Move AddKeyPair template into a separate component"

But it is also changing this validation regex. I think in this case it's ok, I guess we don't need to escape the `.` inside a match list like that (TIL), but it makes me think a bit and would be best in a separate commit or PR even.

axbarsan:
I appreciate you taking the time to review this. 

My commits were a bit sloppy this time, I usually pay more attention to what I put in each of them, but this time I did bigger chunks of work and then staged and commited parts of it.

The failure animation was something that Marian and I briefly discussed, but I guess we can discuss it some more in sig-ux. I removed it for now.

I will try to keep my PRs more focused in the future, so it will be easier for you to review. Thank you

marians:
BTW, looking good to me, too. Thanks! 🌷

PR #1462, opened 2020-05-18
Add a boilerplate reducing middleware and action creator. Implement catalog listing with it.

oponder:
WIP: Figuring out the Typescript and proper placement of some files still.


oponder:
This PR is a bit of a mess, but it fixes a few things, which were a kind of natural consequence of the refactoring. 

- No more double requests to fetch app catalogs
- Only load the index of the catalog that is relevant if you open the app with a deep link to a catalog or an app detail
- Load catalogs individually


Still a WIP. Tomorrow I will see if I can untangle or clean up the PR a bit.

oponder:
The PR is split into two commits. If you want possibly an easier time reviewing, then you can look at the commits separately.

First commit is about the new middleware and action creator, and an implementing it for listing catalogs. 

The new middleware does not let me do the trick of taking the response from the thunk into the UI layer, which is a pattern we wanted to avoid anyways, so I had to make some changes throughout the UI as well.

The second commit is about making it work in the UI.

src/actions/catalogActions.js
```diff
@@ -100,17 +100,35 @@ export function catalogsLoad() {
 }
 
 export function catalogLoadIndex(catalog) {
-  return function (dispatch) {
+  return function (dispatch, getState) {
+    if (getState().entities.catalogs.items[catalog.metadata.name].apps) {
+      // Skip if we already have apps loaded.
+      return new Promise((resolve) => {
+        resolve();
+      });
```
axbarsan, line 108:
```suggestion
      return Promise.resolve();
```

src/stores/appcatalog/actions.ts
```diff
@@ -0,0 +1,33 @@
+import GiantSwarm from 'giantswarm';
+
+import { createAsynchronousAction } from '../asynchronousAction';
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const listCatalogs = createAsynchronousAction<any, any, any>({
```
oponder, line 6: These are PayloadType, StateType, and ResponseType.

I haven't made interfaces for that yet, mostly because StateType is the whole redux state, and I'm not ready to tackle that yet, and it would be out of scope of this PR.

src/stores/callAPIMiddleware.js
```diff
@@ -0,0 +1,38 @@
...
+        });
+      })
+      .catch((error) => {
+        dispatch({
+          error,
+          type: types.error,
+        });
+      });
+
+    return next(action);
```
axbarsan, line 36: I believe this is unreachable because of returning on line `22`

oponder:
Thanks for the nice review :+1:

src/stores/callAPIMiddleware.js
axbarsan: You forgot this in, the middlewares placed after this one in the list will not be called

oponder: No it's gone in one of the later commits, possibly github was out of date. And it shouldn't be called anyways :+1:  since the action isn't a proper action and the next middlewares won't know what to do with it.

PR #1471, opened 2020-05-18
Create an general Input component from one in AppCatalog

ubergesundheit:
While working on cluster labelling I noticed I need a styled `input` component.

There are several in the codebase. The one from AppCatalog looked nice enough to be re-used. So I've converted it into a more general (& typed) one.

@giantswarm/chapter-front-end WDYT is it worth our time to look through all components and try to replace all other `input` implementations with one (or few) central one(s)?


ubergesundheit:
Yes, I wanted to make use of the composability of components. For now the `FileInput` and `Input` is merged because they were so similar.

ubergesundheit:
My Plan would be to create a meta-issue which lists all `inputs` in the application with checkboxes for everyone to pick up. ;)

oponder:
:+1: would be happy to help with that, and thankful if you arrange the delegation

ubergesundheit:
Nice, I'll create the issue soon.

PR #1474, opened 2020-05-19
Cluster label editing UI (without adding)

ubergesundheit:
This PR implements some parts of https://github.com/giantswarm/giantswarm/issues/10541

For now, this PR only implements the bare minimum to make interaction work. Next steps are: Implementing deletion interaction, Implementing label creation interaction & wiring everything with redux.

This is how it looks in commit 1f72344: It will not show up in builds because the corresponding feature flag is set to `false`.

ping @giantswarm/sig-ux for awareness

cluster-labelling-edit


src/components/UI/EditableValueLabel.tsx
```diff
@@ -0,0 +1,199 @@
...
+  value,
+}: IEditableValueLabel) => {
+  const valueElement = useRef<HTMLSpanElement>(null);
+  const inputElement = useRef<HTMLInputElement>(null);
+
+  const [editing, setEditing] = useState(isNew);
+  const [
+    { internalValue, isValid, validationError },
+    setInternalValue,
+  ] = useReducer<Reducer<IReducerState, string>>(
```
oponder, line 88: I didn't know about this hook yet :P 
took me a while to get it.

Maybe for understandability, could we extract the arguments into named variables like `reducer` and `initialValue` or something like that?

I don't think prettier makes it easier to read this stuff either.. what do you guys think? Is it something I just have to get used to?

ubergesundheit: This is indeed hard to read. I can try to move this in separate functions.

ubergesundheit:
> I'm not sure about the styling of the error tooltip, but we have time to think and talk about that before it is officially visible to customers.

This is exactly what this PR is for :) Do you have concrete ideas?

src/components/UI/EditableValueLabel.tsx
```diff
@@ -0,0 +1,199 @@
...
+
+const EditableValueLabel = ({
+  allowEdit,
+  isNew,
+  onCancel,
+  onEdit,
+  onSave,
+  validationFunc,
+  value,
+}: IEditableValueLabel) => {
```
axbarsan, line 80: You can use the same `React.FC` generic here for defining the component, as you used in other components

src/components/UI/EditableValueLabel.tsx
```diff
@@ -0,0 +1,199 @@
...
+              type='text'
+              onChange={({ target: { value: newRawValue } }) =>
+                setInternalValue(newRawValue)
+              }
+              value={internalValue}
+              onKeyUp={keyHandler}
+            />
+            <Overlay
+              // @ts-ignore
+              target={inputElement.current}
```
axbarsan, line 143: This should work
```suggestion
              target={inputElement.current as HTMLInputElement}
```

src/components/UI/EditableValueLabel.tsx
```diff
@@ -0,0 +1,199 @@
...
+};
+
+EditableValueLabel.propTypes = {
+  allowEdit: PropTypes.bool,
+  isNew: PropTypes.bool,
+  onCancel: PropTypes.func,
+  onEdit: PropTypes.func,
+  onSave: PropTypes.func,
+  validationFunc: PropTypes.func,
+  value: PropTypes.string.isRequired,
```
axbarsan, line 187: What do you think about always having required props first, and then the optional ones? (same as in the props interface)

ubergesundheit: I've sorted by first letter, but I like your idea more :+1:

src/components/UI/EditableValueLabel.tsx
ubergesundheit: Yeah I've tried, but this didn't go well with the PropTypes.

For example with the `allowEdit` prop:
```
Type 'Requireable<boolean>' is not assignable to type 'Validator<boolean>'.
  Types of property '[nominalTypeHack]' are incompatible.
    Type '{ type: boolean | null | undefined; } | undefined' is not assignable to type '{ type: boolean; } | undefined'.
      Type '{ type: boolean | null | undefined; }' is not assignable to type '{ type: boolean; }'.
        Types of property 'type' are incompatible.
          Type 'boolean | null | undefined' is not assignable to type 'boolean'.
            Type 'undefined' is not assignable to type 'boolean'.ts(2322)
```

I think this is related to this issue https://github.com/DefinitelyTyped/DefinitelyTyped/issues/28015 but couldn't figure this out.

axbarsan: That error is probably because these props are defined as optional in PropTypes, but as mandatory in the props interface.
```
  allowEdit: PropTypes.bool,
  isNew: PropTypes.bool,
  onCancel: PropTypes.func,
  onEdit: PropTypes.func,
  onSave: PropTypes.func,
  validationFunc: PropTypes.func,
```

In the interface:

```
  allowEdit: boolean;
  isNew: boolean;
  onCancel(): void;
  onEdit(): void;
  onSave(value: string): void;
  validationFunc: IValidationFunction;
```

In order for that to work, you have to either make the props required in PropTypes, or make them optional in the props interface

ubergesundheit:
Thank you very much for your comments! <3

PR #1480, opened 2020-05-20
update react-router

oponder:
Dependabot PR's failed, probably because these need to go hand in hand.

Just testing it manually still to make sure.


oponder:
going with a self approved merge. fyi @giantswarm/chapter-front-end

PR #1483, opened 2020-05-20
Add radio input for HA Masters

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/10831

Looks and behaves similar to the implementation we have for the `Master node availability zones selection` section in the cluster creation page, but this one is a stand-alone component.

I tried to keep the customization level really low, to keep the component extendable, while also keeping the native functionality (same event callback types etc), but only added a label and a custom style to the checkbox.

The PR also disables the `no-confusing-arrow` eslint rule since it contradicts with the recommended way of using styled components

## Preview

**Normal**

image

**Disabled**

image


oponder:
how would you wire in the onChange and Checked props for this? would it be like this?

```
<RadioInput
  bulletProps={onChange:..., checked:...}
/>
```

Could you try to implement something simple in the example in the styleguide that shows how the wiring should be done?

And also if it is like `bulletProps`, then I would argue maybe to bring `onChange` and `checked` up to the props of RadioInput, instead of grouping it in bulletProps, that's a bit unintuitive to me.

In any case it could also use some comments explaining usage in the component itself.

Kinda like `UI/Button.js`

axbarsan:
This is meant to be used as a native input. It supports all props that are supported by the native `input` element.
```
<RadioInput onChange={handleOnChange} />
```

Here I'm extending the native input props interface:
https://github.com/giantswarm/happa/blob/f58598e7a0249ea9e98daa6e99067de9317f55a7/src/components/UI/Inputs/RadioInput.tsx#L58

And then I'm passing them to the input element here:
https://github.com/giantswarm/happa/blob/f58598e7a0249ea9e98daa6e99067de9317f55a7/src/components/UI/Inputs/RadioInput.tsx#L77

`bulletProps` are there just for customizing the look of the bullet (a regular `span` html element), they don't do anything to the checkbox functionality.

I will add some comments for the props, maybe that will clear it up 👍

axbarsan:
I like to implement UI components this way so they won't need documentation. You also get code completion for all possible input-related props:

image

I will add comments mentioning that

oponder:
> I like to implement UI components this way so they won't need documentation. You also get code completion for all possible input-related props:

I like it a lot! this is a really good approach. 

I do still think it would be extra nice if the example in the styleguide could be wired up so that you have a proper usage example, even if it is used in basically the same way as a native input element.

axbarsan:
> > I like to implement UI components this way so they won't need documentation. You also get code completion for all possible input-related props:
> 
> I like it a lot! this is a really good approach.
> 
> I do still think it would be extra nice if the example in the styleguide could be wired up so that you have a proper usage example, even if it is used in basically the same way as a native input element.

But then as we add many more components like that, `StyleGuide` will become really clogged up with different state values and other code.

What do you think about switching to [Storybook](https://storybook.js.org/) for these kinds of things in the future?

oponder:
Storybook seems really worth investigating yeah! :+1: 

Kind of on a low priority, if someone feels called to try it out feel free to do so, kind of way.

PR #1487, opened 2020-05-22
Move Happa to app collection deployment method

wtty-fool:
Towards https://github.com/giantswarm/giantswarm/issues/10177

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,35 +56,97 @@ workflows:
             - build
           <<: *job_filters
 
-      - architect/push-to-docker-legacy:
-          name: push-happa-to-aliyun-master
-          image: 'registry-intl.cn-shanghai.aliyuncs.com/giantswarm/happa'
-          username_envar: 'ALIYUN_USERNAME'
-          password_envar: 'ALIYUN_PASSWORD'
+      - architect/push-to-docker:
+          name: push-happa-to-quay
+          context: architect
+          image: "quay.io/giantswarm/happa"
+          username_envar: "QUAY_USERNAME"
+          password_envar: "QUAY_PASSWORD"
           requires:
             - build-container
+          # Needed to trigger job also on git tag.
+          filters:
+            tags:
+              only: /^v.*/
+
+      - architect/push-to-docker:
+          name: push-happa-to-aliyun
+          context: architect
+          image: "registry-intl.cn-shanghai.aliyuncs.com/giantswarm/happa"
+          username_envar: "ALIYUN_USERNAME"
+          password_envar: "ALIYUN_PASSWORD"
+          requires:
+            - build-container
+          # Needed to trigger job only on git tag.
           filters:
             branches:
               only: master
             tags:
-              only: /^v[0-9]+\.[0-9]+\.[0-9]+$/
-      - deploy-to-all:
+              only: /^v.*/
+
+      - architect/push-to-app-catalog:
+          name: push-happa-to-control-plane-app-catalog
+          context: architect
+          app_catalog: "control-plane-catalog"
+          app_catalog_test: "control-plane-test-catalog"
+          chart: "happa"
           requires:
-            - build-container
-            - push-happa-to-aliyun-master
+            - push-happa-to-quay
+          # Needed to trigger job also on git tag.
           filters:
             tags:
-              only: /^v[0-9]+\.[0-9]+\.[0-9]+$/
-            branches:
-              ignore: /.*/
-      - deploy-to-testing:
+              only: /^v.*/
+
+      - hold-push-happa-to-aliyun-pr:
+          type: approval
           requires:
             - build-container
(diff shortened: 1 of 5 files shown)


helm/happa/values.yaml
```diff
@@ -0,0 +1,16 @@
+name: happa
+namespace: giantswarm
+image:
+  repository: giantswarm/happa
+  tag: [[ .Version ]]
+
+userID: 101
+groupID: 101
+
+Installation:
```
corest, line 10: this `Installation.x` can be removed from here?

.circleci/config.yml
```diff
@@ -56,35 +56,97 @@ workflows:
...
+
+      - hold-push-happa-to-aliyun-pr:
+          type: approval
           requires:
             - build-container
-            - push-happa-to-aliyun-master
+          # Needed to prevent job from being triggered on master branch.
           filters:
+            branches:
+              ignore: master
```
kopiczko, line 107: I don't think we need that guy anymore.

helm/happa/templates/deployment.yaml
```diff
@@ -0,0 +1,105 @@
+apiVersion: apps/v1
+kind: Deployment
```
kopiczko, line 2: Why do we have all those new tamplates?

helm/happa/values.yaml
kopiczko: Otherwise you'll need to define it for testing which is a bit silly.

helm/happa/templates/deployment.yaml
wtty-fool: https://github.com/giantswarm/happa/pull/1500

.circleci/config.yml
```diff
@@ -56,35 +56,86 @@ workflows:
...
+          name: push-happa-to-aliyun
+          context: architect
+          image: "registry-intl.cn-shanghai.aliyuncs.com/giantswarm/happa"
+          username_envar: "ALIYUN_USERNAME"
+          password_envar: "ALIYUN_PASSWORD"
           requires:
             - build-container
-            - push-happa-to-aliyun-master
+          # Needed to trigger job only on git tag.
+          filters:
```
kopiczko, line 81: Nit: The master branch filter can stay IMO. If someone needs to test it there they can remove it in a PR.

oponder:
do we retain the current release and preview functionality?

merges to master are deployed right now automatically but only to the test installations
and then once we tag a vx.x.x, it gets automatically deployed to all installations

kopiczko:
> merges to master are deployed right now automatically but only to the test installations

That wouldn't be retained.

> and then once we tag a vx.x.x, it gets automatically deployed to all installations

That would be retained.

oponder:
sad... but I guess there is no way around that yet?

wtty-fool:
I'm not aware of any ready solution for the moment being.

kopiczko:
> sad... but I guess there is no way around that yet?

Not really...

oponder:
Is there a big benefit or upcoming change that necessitates this? Can we delay it?

It's actually a 'regression' in developer experience for me and other front end devs to do this. I would like to delay it if possible, until a solution exists that lets us do what we had before.

If not, I understand. Just want to check if it's super duper needed immediately.

kopiczko:
We're getting rid of draughtsman so `architect deploy` will stop working. This is the last item on the team's story so I don't see use being postponing it for too long: https://github.com/giantswarm/giantswarm/issues/10177. W/o GH deployment events this scenario will be difficult to achieve because environments we're deploying to are not controlled from the build configuration anymore and push based deployment would be very difficult because of VPN.

I'm with you this is totally backwards for you, but keeping previous pipeline for this single reason seems to be an overkill.

But I'd try creating an issue for ability to deploy to a certain environments. We already are using tricks to not deploy route53-manager and oidc app to all installations so that would be useful. I could create the issue myself but I feel it would get more traction if someone outside the team was pushing it.

oponder:
Yeah Im just raising my stake here and I recognise it's not the right place. 

I'll try to raise the point somewhere. I think some of this tooling is technically in batman's domain as well anyways right? Or what team do you think should want to own and improve the deployment automation, and make that more delightful for all?

When you guys feel happy with this PR, I am ok with it too.

oponder:
question though, how does the automated master deploying work then?

> and then once we tag a vx.x.x, it gets automatically deployed to all installations
> That would be retained.

like.. what is doing the automatic deploying  of those tagged versions?

wtty-fool:
@oponder This is the part filtering what makes it into app collections:
https://github.com/giantswarm/happa/pull/1487/files#diff-1d37e48f9ceff6d8030570cd36286a61R109-R112
This particular one ignores all branches and pushes only tags starting with `v.`.

I'm going to go ahead an merge this, but let's keep the conversation going either here or in an issue somewhere.

PR #1499, opened 2020-05-25
Cluster labelling UI for adding

ubergesundheit:
Followup to #1474 Implements giantswarm/giantswarm#10541

This PR:

- Increases the granularity of validation messages of cluster keys & values
- Adds a `onClick` prop to the `ValueLabel` component
- Removes the validation error tooltips. Validation errors are now displayed inside the edit tooltip
- Changes editing of labels to allow changes of key and value of a label at the same time
- Move some styles of subcomponents in to own files

This is how it looks (as of 811ab69):
cluster-ui-2

PR #1501, opened 2020-05-26
Cluster labelling wiring

ubergesundheit:
Followup to #1474 and #1499  Implements giantswarm/giantswarm#10541

This also adds some tests for cluster labels.

I tried to follow the new approach introduced in #1462


src/reducers/index.js
```diff
@@ -18,13 +19,14 @@ import users from './userReducer';
...
   catalogs,
+  clusterLabels,
   clusters,
   credentials,
   invitations,
+  nodePools,
   organizations,
   releases,
   users,
-  nodePools,
```
ubergesundheit, line 27: Sorted a-z. Or this the order intentional?

src/components/Cluster/ClusterDetail/ClusterLabels/ClusterLabels.tsx
```diff
@@ -2,19 +2,25 @@ import styled from '@emotion/styled';
...
+  getClusterLabelsLoading,
+} from 'selectors/clusterLabelSelectors';
+import { updateClusterLabels } from 'stores/clusterlabels/actions';
 import LabelWrapper from 'UI/ClusterLabels/LabelWrapper';
 
 import DeleteLabelButton from './DeleteLabelButton';
 import EditLabelTooltip from './EditLabelTooltip';
 
-interface IClusterLabelsProps {
+interface IClusterLabelsProps extends HTMLAttributes<HTMLDivElement> {
```
axbarsan, line 16: `HTMLAttributes` is not exactly the most suitable type because those are HTML attributes, not React attributes (some attributes have different names)
```suggestion
interface IClusterLabelsProps extends React.ComponentPropsWithoutRef<'div'> {
```

src/reducers/index.js
axbarsan: Let's hope not 😬

src/stores/clusterlabels/reducer.ts
```diff
@@ -0,0 +1,36 @@
...
+};
+
+const clusterLabelsReducer = produce((draft: IState, action) => {
+  const { types } = updateClusterLabels();
+  switch (action.type) {
+    case types.request:
+      draft.requestInProgress = true;
+      draft.error = null;
+
+      return;
```
axbarsan, line 22:
```suggestion
      break;
```

src/stores/clusterlabels/reducer.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    case types.success:
+      draft.requestInProgress = false;
+      draft.error = null;
+
+      return;
```
axbarsan, line 28:
```suggestion
      break;
```

__tests__/V5ClusterManagement.js
```diff
@@ -447,3 +449,127 @@ it('renders an error message if there was an error loading apps', () => {
...
+    }
+  );
+  const { findByText, findByTestId } = renderRouteWithStore(clusterDetailPath);
+
+  await findByText('Labels:');
+  fireEvent.click(await findByTestId('add-label-button'));
+
+  const keyInput = await findByTestId('label-key-input');
+  const valueInput = await findByTestId('label-value-input');
+  const saveButton = await findByTestId('save-label-button');
```
axbarsan, line 520: What do you think about using more accessible queries? We had a discussion about this at some point, I believe before you joined:

These are the rules we usually guide by:
https://testing-library.com/docs/guide-which-query

Also, I see that you're using a lot of async queries. Are all of those really necessary, or can we replace them with sync ones? They have a big impact on the duration of the tests

src/stores/clusterlabels/reducer.ts
ubergesundheit: `return` vs `break` here: Does this change anything since the other cases don't get executed either way?

axbarsan: It does not change the behaviour of the code. It's just a matter of using the correct syntax for the `switch` statement
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch

ubergesundheit:
@axbarsan care to review again? I've taken your advice and changed the tests to use less promise based selectors & also changed all selectors to be by text, role or labeltext

PR #1505, opened 2020-05-28
Enable cluster labelling through feature flag

ubergesundheit:
Enable cluster labelling for production builds & do some last minute changes for firefox


ubergesundheit:
1346284 is currently deployed on gaia, but quay incident makes preview of the newer commits hard.

PR #1486, opened 2020-05-20
Enable AWS HA Masters

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/10831

~This still needs the latest JS client in order to work properly~
All good!

## Preview

**Non-HA, with the master node not ready**

image

**Non-HA, switch to HA**

image

**Non-HA, switching to HA - loading state**

image

**Non-HA, switching to HA - error state**

image

**HA, not all master nodes ready**

image

**HA, all master node ready**

image

**HA and non-HA, no info on number of master nodes ready**

image

PR #1511, opened 2020-05-29
Add HA Masters cluster creation components

axbarsan:
Part of #1486 

## Preview

image

### DIFF
--- a/src/components/Cluster/NewCluster/MasterNodes.tsx
+++ b/src/components/Cluster/NewCluster/MasterNodes.tsx
@@ -1,0 +1,80 @@
+import styled from '@emotion/styled';
+import PropTypes from 'prop-types';
+import React from 'react';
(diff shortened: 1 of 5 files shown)


src/components/Cluster/NewCluster/MasterNodes.tsx
```diff
@@ -0,0 +1,80 @@
+import styled from '@emotion/styled';
```
oponder, line 1: would it make sense to have this in the UI folder given it's a not connected component?

on the other hand, I do like having related things together.

axbarsan: Not sure if we have a rule for this yet, but I consider the `UI` folder suitable for reusable "UI" components. This component is not really reusable outside this context

PR #1512, opened 2020-05-29
Add components for displaying master nodes in the cluster detail page

axbarsan:
Part of #1486

## Preview

**Non-HA, with the master node not ready**

image

**Non-HA, switch to HA**

image

**Non-HA, switching to HA - loading state**

image

**Non-HA, switching to HA - error state**

image

**HA, not all master nodes ready**

image

**HA, all master node ready**

image

**HA, no info on number of master nodes ready**

image

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -11,6 +11,7 @@
     "module": "esnext",
     "moduleResolution": "node",
     "resolveJsonModule": true,
+    "downlevelIteration": true,
     "isolatedModules": true,
     "noEmit": true,
     "jsx": "react",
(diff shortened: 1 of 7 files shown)


tsconfig.json
```diff
@@ -11,6 +11,7 @@
     "module": "esnext",
     "moduleResolution": "node",
     "resolveJsonModule": true,
+    "downlevelIteration": true,
```
axbarsan, line 14: Forgot to mention, this is to be able to iterate over Iterable data structures (Array-like). Without this, we would have to convert them to arrays using `Array.from(...)`

PR #1514, opened 2020-05-29
Integrate Master Nodes components in the cluster details page

axbarsan:
Still part of https://github.com/giantswarm/happa/pull/1486/


axbarsan:
This needs the gs client in order to work. I will keep it in the original implementation

PR #1531, opened 2020-06-03
Input field refactoring 

ubergesundheit:
This refactors some usages of input fields and converts ViewEditName to typescript.

Related to: giantswarm/giantswarm#11004


src/components/AppCatalog/AppDetail/InstallAppForm.js
```diff
@@ -107,43 +106,37 @@ const InstallAppForm = ({
...
-          value={formAbilities.fixedNamespace}
           readOnly={true}
+          value={formAbilities.fixedNamespace}
         />
       ) : (
         <Input
-          key='dedicated-namespace'
           description='We recommend that you create a dedicated namespace. The namespace will be created if it doesn’t exist yet.'
-          hint={<>&nbsp;</>}
+          key='dedicated-namespace'
```
ubergesundheit, line 118: I am unsure why we need the `key` prop

axbarsan: Not sure if this is that case, but I remember we had a problem at some point, where React would believe that it was the same element and it would not re-render it, because it thought it didn't change, and we could force a re-render by using different keys

src/components/Cluster/ClusterDetail/NodePool.tsx
```diff
@@ -260,11 +254,13 @@ class NodePool extends Component<INodePoolsProps, INodePoolsState> {
...
-            onSubmit={this.editNodePoolName}
-            ref={(viewEditName: INPViewAndEditName): void => {
-              this.viewEditNameRef = viewEditName;
+            value={nodePool.name}
+            typeLabel='node pool'
+            onSave={this.editNodePoolName}
+            ref={(viewEditName: HTMLSpanElement | null): void => {
+              if (viewEditName) {
+                this.viewEditNameRef = viewEditName as INPViewAndEditName;
+              }
```
axbarsan, line 263: Why do we need the 'if' statement?

src/components/UI/ViewEditName.tsx
```diff
@@ -109,12 +136,13 @@ class ViewAndEditName extends React.Component {
...
     this.toggleEditMode(false, {
-      value: this.props.name,
+      value: this.props.value,
       errorMessage: '',
+      editing: false,
     });
   };
 
-  handleChange = (e) => {
+  handleChange: React.ChangeEventHandler<ElementRef<'input'>> = (e) => {
```
axbarsan, line 145: The event is actually attached to the input element, not to the ref. Same for the others that use a similar type.

```suggestion
  handleChange: React.ChangeEventHandler<HTMLInputElement>> = (e) => {
```

src/components/AppCatalog/AppDetail/InstallAppForm.js
ubergesundheit: Ok, I'll leave them alone :+1:

src/components/Cluster/ClusterDetail/NodePool.tsx
ubergesundheit: In development I changed the type of the `viewEditNameRef` to be not `| null` so I forgot to revert this.

oponder: I bet it's to make typescript compile, since viewEditName can be null?

ubergesundheit:
I'll deploy on ginger and merge tomorrow if no problems arise

PR #1547, opened 2020-06-10
[CP Auth 5/5]: Add React components and link everything together

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/8622

This PR includes the React components that make the CP Auth usable in the app.

Note: You might notice that after you login/logout of the CP API, happa will do a hard refresh. That is because app catalogs are loaded as part of happa's initial render, and doing a hard refresh is an easy way to trigger that step again.

### Preview

<details>
<summary>User dropdown</summary>

image

</details>

<details>
<summary>Control plane access page - logged out</summary>

image

</details>

<details>
<summary>Control plane access page - logged in</summary>

image

</details>

### How to test

1. Deploy this branch to a testing installation.

<details>
<summary>2. Click on the user drop-down in the top right of the page. Press on the `Control Plane Access` button.</summary>

image

</details>

<details>
<summary>3. Click on the 'Log in to Control Plane via OIDC' button</summary>

image

</details>

4. You will be redirected to GitHub. Authorize the app.

5. You will be redirected back to the Happa homepage.

<details>
<summary>6.. Enter the 'App Catalogs' page (top menu). The app catalogs are fetched using the Control Plane API. If you are doubting that, you can inspect the network tab in DevTools to see that there is still hope in this world.</summary>

image

</details>


src/actions/batchedActions.js
```diff
@@ -14,6 +17,11 @@ import * as userActions from './userActions';
 export const batchedLayout = () => async (dispatch) => {
   try {
     await dispatch(userActions.refreshUserInfo());
+
+    if (FeatureFlags.FEATURE_CP_ACCESS) {
+      await dispatch(loadUser(CPAuth.getInstance()));
```
axbarsan, line 22: Loading existing Control Plane authentication user, if the feature is enabled.

src/components/Auth/CP/CPAuthProvider.tsx
```diff
@@ -0,0 +1,70 @@
...
+    const cpAuth = CPAuth.getInstance();
+    cpAuth.addEventListener(OAuth2Events.UserLoaded, onUserLoaded);
+    cpAuth.addEventListener(OAuth2Events.UserUnloaded, onUserSignedOut);
+    cpAuth.addEventListener(OAuth2Events.UserSignedOut, onUserSignedOut);
+    cpAuth.addEventListener(OAuth2Events.TokenExpiring, onTokenExpiring);
+    cpAuth.addEventListener(OAuth2Events.TokenExpired, onTokenExpired);
+    cpAuth.addEventListener(OAuth2Events.SilentRenewError, onUserSignedOut);
+
+    return () => {
+      onUserSignedOut();
```
axbarsan, line 47: When this component is unmounted, it means that we navigated to a page outside the logged in context, which means that we can safely log out.

src/components/Auth/CP/CPAuthProvider.tsx
```diff
@@ -0,0 +1,70 @@
...
+import {
+  loadUserSuccess,
+  userExpired,
+  userExpiring,
+  userSignedOut,
+} from 'stores/cpauth/actions';
+
+interface ICPAuthProviderProps extends React.PropsWithChildren<{}> {}
+
+const CPAuthProvider: React.FC<ICPAuthProviderProps> = ({ children }) => {
```
axbarsan, line 15: This component sits as a direct child of the `Layout` component, and it acts as a mapping between the Control Plane authentication provider and the Redux store.

src/components/Auth/CP/CPLoginPage.tsx
```diff
@@ -0,0 +1,109 @@
...
+const CPStatus = styled.div`
+  margin-top: 32px;
+  max-width: 1024px;
+  display: flex;
+  align-items: center;
+`;
+
+interface ICPLoginPageProps {}
+
+const CPLoginPage: React.FC<ICPLoginPageProps> = () => {
```
axbarsan, line 22: This is a separate page that we can use to log into the Control Plane API.

src/components/Auth/CP/CPLoginPage.tsx
```diff
@@ -0,0 +1,109 @@
...
+  );
+
+  useEffect(() => {
+    const currentURL = window.location.href;
+
+    if (currentURL.includes(AppRoutes.CPAccessCallback)) {
+      const handleAuthParams = async () => {
+        try {
+          const auth = CPAuth.getInstance();
+          await auth.handleLoginResponse(currentURL);
```
axbarsan, line 65: This is ran after the user was redirected back to the app from the OAuth2 provider. The login token and state are then taken from the URL parameters.

src/components/Auth/CP/CPLoginPage.tsx
```diff
@@ -0,0 +1,109 @@
...
+  const dispatch = useDispatch();
+
+  const handleLogin = useCallback(
+    async (e: React.MouseEvent<HTMLButtonElement>) => {
+      e.preventDefault();
+      e.stopPropagation();
+
+      try {
+        const auth = CPAuth.getInstance();
+        await auth.attemptLogin();
```
axbarsan, line 32: This will redirect the user to the Control Plane OAuth2 provider.

src/components/UI/Navigation/UserDropdown.js
```diff
@@ -111,6 +112,17 @@ class UserDropdown extends React.Component {
...
             </MenuItem>
           ) : undefined}
+
+          {FeatureFlags.FEATURE_CP_ACCESS && (
+            <MenuItem
+              componentClass={NavLink}
+              href={AppRoutes.CPAccess}
+              to={AppRoutes.CPAccess}
+            >
+              Control Plane Access
```
axbarsan, line 122: This will add a control plane API option in the user dropdown.

src/components/index.tsx
```diff
@@ -46,7 +47,7 @@ declare global {
 }
 
 // Configure the redux store.
-const store: Store = configureStore({}, history);
+const store: Store = configureStore({}, history, CPAuth.getInstance());
```
axbarsan, line 50: Passing the Control Plane authentication instance as an argument, to facilitate testing.

src/lib/CPAuth/CPAuth.ts
```diff
@@ -0,0 +1,36 @@
+import { getAuthorityURLFromGSApiURL } from 'lib/CPAuth/util';
+import OAuth2, { IOAuth2Config } from 'lib/OAuth2/OAuth2';
+import { AppRoutes } from 'shared/constants/routes';
+
+const defaultConfig: IOAuth2Config = {
+  authority: getAuthorityURLFromGSApiURL(window.config.apiEndpoint),
+  clientId: 'zQiFLUnrTFQwrybYzeY53hWWfhOKWRAU',
+  clientSecret: 'TVHzVPin2WTiCma6bqp5hdxgKbWZKRbz',
```
axbarsan, line 8: Right now these are the same as the ones from the `kubectl-gigantic` plugin. They will have bespoke values once we merge:
- https://github.com/giantswarm/installations/pull/1298
- https://github.com/giantswarm/dex-app/pull/13

src/lib/CPAuth/CPAuth.ts
```diff
@@ -0,0 +1,36 @@
...
+  automaticSilentRenew: true,
+  includeIDTokenInSilentRenew: true,
+  loadUserInfo: true,
+  revokeAccessTokenOnLogout: true,
+  filterProtocolClaims: true,
+  validateSubOnSilentRenew: true,
+  persistenceMethod: localStorage,
+};
+
+class CPAuth extends OAuth2 {
```
axbarsan, line 24: This inherits from the `OAuth2` class, so we could have this as a singleton, but not limit the number of OAuth2 providers to a single one.

src/model/clients/FetchAdapter.ts
```diff
@@ -0,0 +1,79 @@
+import { GenericResponse } from 'model/clients/GenericResponse';
+import { HttpClient } from 'model/clients/HttpClient';
+
+/**
+ * This is an adapter for using an HttpClient with the same
+ * interface as the global `fetch` function.
+ */
+class FetchAdapter<T extends HttpClient = HttpClient> {
```
axbarsan, line 8: This makes us able to use our classes based on the `HttpClient` base class with the CP Client, as a replacement for the global `fetch` method.

src/model/clients/GenericResponse.js
```diff
@@ -103,4 +103,26 @@ export class GenericResponse {
   get headers() {
     return Object.assign({}, this.config.headers);
   }
+
+  /**
+   * Convert the object to a `Response` object returned by the fetch command.
+   * @returns {Response}
+   */
+  convertToFetchResponse() {
```
axbarsan, line 111: This converts our `GenericResponse` to a `Response` object, which is used by the global `fetch` method.

src/model/services/controlplane/appcatalogs/appcatalogs.ts
```diff
@@ -0,0 +1,57 @@
...
+import { getBaseConfiguration } from 'model/services/controlplane/base';
+
+/**
+ * Get the all the app catalogs in the installation.
+ * @param client - The HTTP client.
+ */
+export async function getAppCatalogs(
+  client: HttpClient
+): Promise<IAppCatalog[]> {
+  const { ApplicationGiantswarmIoV1alpha1Api } = await import(
```
axbarsan, line 17: Importing this dynamically, to save up a couple hundred of KBs from the final bundle size (Which also means we won't have dead Control Plane API code when the feature is not supported).

src/model/services/controlplane/appcatalogs/appcatalogs.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  const catalogs = await appsAPI.listApplicationGiantswarmIoV1alpha1AppCatalog(
+    {}
+  );
+  const result = catalogs.items.map(convertAppCatalogCRToCatalog);
+
+  return result;
+}
+
+function convertAppCatalogCRToCatalog(
```
axbarsan, line 31: This maps the App catalog CR type to the type that we use internally, so we can have control over what we store in memory.

src/model/services/controlplane/appcatalogs/types.ts
```diff
@@ -0,0 +1,21 @@
...
+  title: string;
+  description: string;
+  logoURL: string;
+  storage: IStorage;
+}
+
+export interface IStorage {
+  type: string;
+  URL: string;
+}
```
axbarsan, line 21: Custom App catalog types, same as the ones on the current GS API.

src/model/services/controlplane/base.ts
```diff
@@ -0,0 +1,22 @@
...
+): Promise<Configuration> {
+  const { Configuration } = await import('giantswarm-cp-client');
+  const fetchAdapter = new FetchAdapter(client);
+
+  const configBuilder = new Configuration({
+    basePath: client.requestConfig.baseURL,
+    fetchApi: fetchAdapter.fetch.bind(fetchAdapter),
+  });
+
+  return configBuilder;
```
axbarsan, line 21: This creates the basic configuration that we'll use in all CP API services: 
- setting the Base URL to the CP API URL;
- creating a fetch adapter, to use our internal `HttpClient` objects.

src/shared/constants/routes.js
```diff
@@ -7,6 +7,8 @@ const AppRoutes = {
   SetPassword: '/forgot_password/:token',
   SignUp: '/signup/:token',
   OAuthCallback: '/oauth/callback',
+  CPAccess: '/cp-access',
+  CPAccessCallback: '/cp-access/callback',
```
axbarsan, line 11: This is the callback URL called after authentication.

src/stores/appcatalog/actions.ts
```diff
@@ -1,22 +1,55 @@
...
+      cpAuthUser = getCPAuthUser(currentState);
+    }
+
+    if (cpAuthUser) {
+      const client = new CPClient(
+        cpAuthUser.idToken,
+        cpAuthUser.authorizationType
+      );
+
+      catalogs = await getAppCatalogs(client);
```
axbarsan, line 34: Using the CP API to fetch app catalogs, if the feature is supported and if the user is logged in.

src/stores/cpauth/middleware.ts
```diff
@@ -0,0 +1,39 @@
+import CPAuth from 'lib/CPAuth/CPAuth';
+import OAuth2UserImpl from 'lib/OAuth2/OAuth2User';
+import { AnyAction, Middleware } from 'redux';
+import { loadUserError, userExpired } from 'stores/cpauth/actions';
+import { CPAUTH_USER_EXPIRED, CPAUTH_USER_LOAD } from 'stores/cpauth/constants';
+import { getCPAuthUser } from 'stores/cpauth/selectors';
+
+export function cpAuthMiddleware(cpAuth: CPAuth): Middleware {
```
axbarsan, line 8: This middleware is used to try renew user information, if a user is expired.

In general, the user will be renewed automatically, but the flow will fail if there's a race condition between renewal and a separate request. This prevents that.

This will also prevent actions from getting dispatched, if the user expired.

src/lib/OAuth2/OAuth2.ts
```diff
@@ -0,0 +1,193 @@
...
+  automaticSilentRenew?: boolean;
+  includeIDTokenInSilentRenew?: boolean;
+  loadUserInfo?: boolean;
+  revokeAccessTokenOnLogout?: boolean;
+  filterProtocolClaims?: boolean;
+  validateSubOnSilentRenew?: boolean;
+  persistenceMethod?: Storage;
+}
+
+class OAuth2 {
```
axbarsan, line 45: This is a wrapper around the `oidc-client` module.

This provides support for OAuth2 authentication, including persistence between sessions, token renewal and silent logins.

src/lib/OAuth2/OAuth2.ts
```diff
@@ -0,0 +1,193 @@
...
+  protected userManager: UserManager;
+  protected eventEmitter: EventTarget;
```
axbarsan, line 47: We're using the native `EventTarget`, to benefit from the performance of native events, even if it means using `CustomEvent` objects

src/lib/OAuth2/OAuth2User.ts
```diff
@@ -0,0 +1,71 @@
+import { User } from 'oidc-client';
+import { AuthorizationTypes } from 'shared/constants';
+import { PropertiesOf } from 'shared/types';
+
+export interface IOAuth2User {
```
axbarsan, line 5: This is a wrapper for the `User` class from the `oidc-client` module.
This holds useful information as tokens, groups, and expiration date.

src/lib/OAuth2/OAuth2User.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  public isExpired(): boolean {
+    const sInMs = 1000;
+    const now = Math.trunc(Date.now() / sInMs); // In seconds.
+    const expiresIn = this.expiresAt - now;
+
+    return expiresIn <= 0;
+  }
+
+  public serialize(): IOAuth2User {
```
axbarsan, line 57: This converts the object into an easily storable value (for the Redux store, for example)

src/lib/OAuth2/OAuth2User.ts
```diff
@@ -0,0 +1,71 @@
...
+  constructor(fromObj?: Partial<IOAuth2User>) {
+    if (!fromObj) return;
+
+    for (const [key, value] of Object.entries(fromObj)) {
+      // @ts-ignore
+      this[key] = value;
+    }
+  }
+
+  public isExpired(): boolean {
```
axbarsan, line 49: This calculates if the user is expired or not, based on the current date and the stored expiration date.

webpack.common.js
```diff
@@ -32,6 +32,7 @@ const makeFeatureFlags = () => {
   const defaults = {
     FEATURE_CLUSTER_LABELS_V0: true,
     FEATURE_HA_MASTERS: true,
+    FEATURE_CP_ACCESS: true,
```
axbarsan, line 35: Don't forget this in! It's for testing purposes, only.

src/lib/CPAuth/config.ts
```diff
@@ -0,0 +1,96 @@
+import { getAuthorityURLFromGSApiURL } from 'lib/CPAuth/util';
+import { IOAuth2Config, IOAuth2SigningKey } from 'lib/OAuth2/OAuth2';
+import { AppRoutes } from 'shared/constants/routes';
+
+const authority = getAuthorityURLFromGSApiURL(window.config.apiEndpoint);
+
+const signingKeys: IOAuth2SigningKey[] = [
```
axbarsan, line 7: Unfortunately we have to hard-code all this configuration, because the auth server doesn't allow cross-origin requests.

src/components/Cluster/ClusterDetail/__tests__/UpgradeClusterModal.tsx
```diff
@@ -39,13 +39,13 @@ function createRelease(version: string, active: boolean): IRelease {
...
-) {
+): Partial<IState> {
   return {
-    entities: {
+    entities: ({
       releases: {
         items: releases,
       },
-    },
+    } as unknown) as IState['entities'],
```
axbarsan, line 48: Temporary, since there is no `releases` property in the `IState` interface, but not in the scope of this PR

src/stores/appcatalog/types.ts
```diff
@@ -1,3 +1,21 @@
+import { IAppCatalog } from 'model/services/controlplane/appcatalogs/types';
+
 export interface IApplicationVersion {
   version: string;
 }
+
+export interface IStoredAppCatalog extends IAppCatalog {
+  isFetchingIndex: boolean;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  apps?: any;
```
axbarsan, line 10: Not in the scope of this PR

ubergesundheit:
Will this work in development mode? I tried to run this branch locally against gauss and ginger but clicking the `Log in to Control Plane via OIDC` button tries to fetch `http://localhost:8000/.well-known/openid-configuration` which is not available there?

axbarsan:
> Will this work in development mode? I tried to run this branch locally against gauss and ginger but clicking the `Log in to Control Plane via OIDC` button tries to fetch `http://localhost:8000/.well-known/openid-configuration` which is not available there?

Yeah, it should work. Please make sure that you restart happa after checking out the branch. (CTRL + C and run `yarn start` again), as it leverages environmental variables.

ubergesundheit:
Yeah it was from a clean checkout after installing additional dependencies. Do I need to set the feature flag `FEATURE_CP_ACCESS` to true? I could navigate to the login page without it.

axbarsan:
> Yeah it was from a clean checkout after installing additional dependencies. Do I need to set the feature flag `FEATURE_CP_ACCESS` to true? I could navigate to the login page without it.

The problem was indeed with the proxy configuration. It should work now 👍 

> One question: Why did you decide to use Events for the Oauth client?

Since there are a lot of things that can happen with the OIDC authentication (the token can be refreshed, the token can expire, the user can sign out etc.), it's easy to handle callbacks registered by the redux store, or any other object.

A callback system would also work, but then we'd lose the liberty to subscribe multiple objects (we may not want to be tied to the Redux store at all times).

I can't really see any option besides an event-driven pattern or a callback map (maybe observables, but then things might get too complex). I might also be influenced by the fact that `oidc-client-js` (the library that we're using for OIDC authentication) also uses an event-driven pattern, and it was easy to map it to our custom object.

Have you thought of a another approach? Maybe there's a better way to do it

ubergesundheit:
I thought a little about your choice of Events over the weekend. It's surely the most elegant way for letting more than one subscriber subscribe to its events but it seems a little alien in context to the rest of the application. I like it! I'll now try out the updated PR

ubergesundheit:
Here are my findings: Maybe I was doing something wrong..

On ghost, it opened some Microsoft sign in page. I aborted at this point because I was unsure which credentials I should enter

On gorgoth localhost:9999 did not succeed to load

On ginger, it opened a github sign in page, after entering my credentials, a `POST` request to `/session` was initiated but failed with `404 page not found`.

axbarsan:
> Here are my findings: Maybe I was doing something wrong..
> 
> On ghost, it opened some Microsoft sign in page. I aborted at this point because I was unsure which credentials I should enter
> 
> On gorgoth localhost:9999 did not succeed to load
> 
> On ginger, it opened a github sign in page, after entering my credentials, a `POST` request to `/session` was initiated but failed with `404 page not found`.

Damn it... These are all happening with the local installation proxy, right? I'll try to make it work, but if not, I'll leave it for another PR, as it looks like it's a deeper issue.

Please give it a try with it deployed to an installation, or locally with something like Chrome with the unsafe flag enabled.

axbarsan:
Also, `gorgoth` doesn't have OIDC authentication setup (not sure where we can find a list of the installations that support it).

At the Microsoft login screen you need to pick GitHub login (so login with GitHub, through Microsoft 😬)

ubergesundheit:
> Please give it a try with it deployed to an installation

Will do :)

ubergesundheit:
Currently deployed to gauss: Clicking the login button opens `localhost:9999/auth....`.

axbarsan:
> Currently deployed to gauss: Clicking the login button opens `localhost:9999/auth....`.

Damn, I broke it 😆

axbarsan:
Thank you for helping me test this, @ubergesundheit . Everything should work properly now in all scenarios. Please let me know if anything else breaks

Also, `gauss` was misconfigured after the region change (see fix in https://github.com/giantswarm/installations/pull/1431)

ubergesundheit:
Here is more funny business: When being logged in with OIDC, there is no `App Catalogs` menu for me in the top bar. (Tested on gauss with local proxy)

axbarsan:
> Here is more funny business: When being logged in with OIDC, there is no `App Catalogs` menu for me in the top bar. (Tested on gauss with local proxy)

Please keep in mind that `happa` master needs to be deployed on the installation that you want to test on, so it would have the K8s API proxy ingress in place. I deployed `happa` master on `gauss` and everything works locally, with the proxy enabled. 👍

oponder (approved):
LGTM :D

PR #1550, opened 2020-06-12
Inline release selection

ubergesundheit:
Towards giantswarm/giantswarm#11082

This PR inlines release selection in cluster creation.
<details>
<summary>Screenshots</summary>

image
image
image

</details>


marians:
Just triggered an internal discussion on consistent prefixing of version numbers with `v`. See https://gigantic.slack.com/archives/C13MG9F0W/p1591964805162000

src/components/Cluster/NewCluster/ReleaseSelector/ReleaseRow.tsx
```diff
@@ -0,0 +1,138 @@
...
+          <ClassNames>
+            {({ css }) => (
+              <StyledRadioInput
+                id={`select-${version}`}
+                title={`Select release v${version}`}
+                checked={isSelected}
+                value={isSelected ? 'true' : 'false'}
+                name={`select-${version}`}
+                onChange={() => selectRelease(version)}
+                bulletProps={{ className: css({ margin: 0 }) }}
```
axbarsan, line 70: I wonder if we can declare this outside the component using the regular `css` function.

Example (with `css` imported from `emotion`:
```javascript
const BulletStyle = css`
	margin: 0;
`;
```

And then using it within `bulletProps` like (`css` from render props):
```javascript
...
bulletProps={{ className: css`${BulletStyle}` }}
...
```

That would really improve the readibility.

But would it work?

src/components/Cluster/NewCluster/ReleaseSelector/ReleaseRow.tsx
```diff
@@ -0,0 +1,138 @@
...
+            }}
+          >
+            <i className={`fa fa-${collapsed ? 'eye' : 'eye-with-line'}`} />
+            Show
+          </TableButton>
+        </td>
+        <td>
+          <TableButton
+            data-testid={`open-changelog-${version}`}
+            href={`https://github.com/giantswarm/releases/blob/master/${provider}/v${version}/release-notes.md`}
```
axbarsan, line 94: Can we move the link generation to a separate function?

marians:
On ginger I get the error `FEATURE_CLUSTER_LABELS_V0 is not defined`

src/components/UI/Inputs/RadioInput.tsx
```diff
@@ -73,6 +73,9 @@ interface IRadioInputProps
 
   /* Customization props for the root label element. */
   rootProps?: Omit<ComponentPropsWithRef<'label'>, 'htmlFor'>;
+
+  /* className for applying a class to the root. */
+  className?: string;
```
axbarsan, line 78: This is not really how this was designed to work. You can pass props to the root component by using `rootProps`. The rest of the props are meant to be passed to the internal `input` component. That's why the props type inherits from the base `input` component.

Keeping it like this would be pretty confusing. Imagine setting the `name` prop, which affects the input, but setting the `className` prop affects another component.

ubergesundheit:
> On ginger I get the error `FEATURE_CLUSTER_LABELS_V0 is not defined`

Let me update the branch to newest master. Then try again

src/components/Cluster/NewCluster/ReleaseSelector/ReleaseRow.tsx
ubergesundheit: I'll see if this works :+1:

ubergesundheit: `You have tried to stringify object returned from `css` function. It isn't supposed to be used directly (e.g. as value of the `className` prop), but rather handed to emotion so it can handle it (e.g. as value of `css` prop).`

src/components/Cluster/NewCluster/ReleaseSelector/ReleaseRow.tsx
ubergesundheit: Sure, I'll create a hook for this.

src/components/UI/Inputs/RadioInput.tsx
ubergesundheit: Yeah I see your point, The reason why I've added the `className` prop here to allow style overrides with `styled`. See https://github.com/giantswarm/happa/pull/1550/files#diff-654d39fbe9c6c7c4e51037c47c39a319R38-R40

src/components/Cluster/NewCluster/ReleaseSelector/ReleaseRow.tsx
axbarsan: At least we tried 😓

src/components/UI/Inputs/RadioInput.tsx
axbarsan: I understand why you did it, but it shouldn't be necessary. Why doesn't this work without adding all that?

```javascript
<RadioInput rootProps={{ className }} />
```

ubergesundheit:
Identified Todos through call with @marians & comments of @adrian:

- [ ] Change font-weight to 300 in the releases table
- [ ] Center all table headers & cells
- [ ] Change the cursor to `pointer` for the first four cells in each row
- [ ] Prevent release selection for table cells with buttons & only show pointer for buttons
- [ ] Show `Hide` for the components button if expanded
- [ ] Check the `Available releases` underlining (left)
- [ ] Disable hover effect for the components
- [ ] Remove the `v` in the release version column
- [ ] Display non-active versions in a dark font & show notice to admin users `Dark font indicates a inactive release which is only available to giant swarm staff`
- [ ] Change Button labels to all caps
- [ ] Move release notes url generation to hook
- [ ] Try to fiddle with RadioInput (& Bullet) style overrides

src/components/UI/Inputs/RadioInput.tsx
ubergesundheit: This should be to your linking: 0c8ab74

axbarsan: damn... we need to switch to [this](https://github.com/emotion-js/emotion/tree/master/packages/babel-preset-css-prop) asap

axbarsan: This worked for me:

```javascript
import { ClassNames, css } from '@emotion/core';

const SomeCss = css`
  display: none;
`;

<ClassNames>
  {({ css: css2 }) => (
    <RadioInput
      value='naah'
      id='some-radio'
      name='inputtrickery'
      label='Some text'
      rootProps={{ className: css2(SomeCss) }}
    />
  )}
</ClassNames>
```

ubergesundheit: So do you're fine with the solution in 0c8ab74?

axbarsan: I'm fine either way, it's up to you which one of the 2 solutions fits our codebase best 😄

ubergesundheit:
I've decided to create a separate issue for "Display non-active versions in a dark font & show notice to admin users Dark font indicates a inactive release which is only available to giant swarm staff" https://github.com/giantswarm/giantswarm/issues/11564

ubergesundheit:
The current state is now deployed on ginger

marians:
The highlight color for :hover is a bit too subtle for me

image

Do we have a scale for that in our theme? If yes, I'd just go one step brighter. Otherwise, please try `#2d5369`.

ubergesundheit:
> Do we have a scale for that in our theme?

No scale, but we're defining these colors: https://github.com/giantswarm/happa/blob/master/src/styles/theme.ts#L5-L63

ubergesundheit:
I could use `shade4` `#2F556A`

ubergesundheit:
Changed to  `#2F556A`. We can change it further in another PR.

marians:
Thanks! That's pretty much the scale I was talking about. https://github.com/giantswarm/happa/blob/master/src/styles/theme.ts#L11-L31

PR #1563, opened 2020-06-15
Adapt upgrade information to mention multiple master nodes

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/11563

Related to HA masters, so no changelog entry needed.


marians:
Text looks good. But

I'd expect to see a condition somewhere. If num_masters == 1 then text A, else text B.

axbarsan:
> Text looks good. But
> 
> I'd expect to see a condition somewhere. If num_masters == 1 then text A, else text B.

Oops, I somehow didn't get that from the issue. Updated it now.

PR #1571, opened 2020-06-16
Remove all console.error calls

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/11388

### DIFF
--- a/src/actions/batchedActions.js
+++ b/src/actions/batchedActions.js
@@ -1,4 +1,5 @@
 import { push } from 'connected-react-router';
+import { ErrorReporter } from 'lib/errors';
 import RoutePath from 'lib/routePath';
 import { OrganizationsRoutes } from 'shared/constants/routes';
 import { listCatalogs } from 'stores/appcatalog/actions';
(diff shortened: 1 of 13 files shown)


src/actions/batchedActions.js
```diff
@@ -37,8 +38,7 @@ export const batchedLayout = () => async (dispatch) => {
       })
     );
   } catch (err) {
-    // eslint-disable-next-line no-console
-    console.error('Error in batchedLayout', err);
+    ErrorReporter.getInstance().notify(err);
```
axbarsan, line 41: Replaced these logs with error reports, since otherwise they'd go unnoticed

src/lib/pageVisibilityTracker.js
```diff
@@ -21,18 +21,8 @@ class PageVisibilityTracker {
    * @param {change handler} func
    */
   addEventListener(func) {
-    if (
```
axbarsan, line 24: Removed this since the feature is supported by all modern browsers

PR #1579, opened 2020-06-17
Make precommit hook quicker

axbarsan:
Noticed the pre-commit hook sometimes took a bit too much time.

I tested this by running the hook with 3 files, with multiple changes.
I measured the running time for 5 runs, and then calculated an average.

Before: `23.38s`
After: `15.69s`

I removed some ESLint rules that took too much time to run, and which we can also live without.


ubergesundheit:
I generally like speeding up things. And am fine with these changes. I won't approve because I feel this is a case where it would be good to have consensus from everyone working on this. So approve from @oponder == approve from @ubergesundheit ;)

PR #1580, opened 2020-06-17
Fixing errors in tests

axbarsan:
We currently have 2 major errors in 2 test suites:
* AdminLogin
* ClusterManagement

This fixes those nasty errors


src/components/Auth/AdminLogin.tsx
```diff
@@ -85,7 +85,7 @@ AdminLogin.propTypes = {
   // @ts-ignore
   actions: PropTypes.object.isRequired,
   // @ts-ignore
-  user: PropTypes.object.isRequired,
+  user: PropTypes.object,
```
axbarsan, line 88: This is actually nullable, and `prop-types` was complaining

src/components/Cluster/ClusterDetail/ClusterDetail.js
```diff
@@ -11,37 +11,47 @@ import { AppRoutes } from 'shared/constants/routes';
...
+        `Cluster <code>${clusterId}</code> doesn't exist.`,
+        messageType.INFO,
+        messageTTL.MEDIUM
+      );
 
-    dispatch(push(AppRoutes.Home));
+      dispatch(push(AppRoutes.Home));
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
axbarsan, line 28: Only executing this effect on component mount, to not interfere with the existent logic in `ClusterDetailView`

src/components/Cluster/ClusterDetail/ClusterDetail.js
```diff
@@ -11,37 +11,47 @@ import { AppRoutes } from 'shared/constants/routes';
...
+const ClusterDetail = ({ match, cluster, clusterId, ...rest }) => {
   const dispatch = useDispatch();
 
-  if (!props.cluster) {
-    new FlashMessage(
-      `Cluster <code>${props.clusterId}</code> doesn't exist.`,
-      messageType.INFO,
-      messageTTL.MEDIUM
-    );
+  useEffect(() => {
```
axbarsan, line 17: Using an effect because executing updates during rendering (such as the `dispatch(push(AppRoutes.Home));` we had before) is forbidden by React, and will throw an error.

PR #1581, opened 2020-06-17
Inline instance selection

ubergesundheit:
Towards giantswarm/giantswarm#11082

Replaces the current instance selection modals with an inline instance selection analogue to the new release selection

<details>
<summary>Screenshots Azure</summary>

image

image

</details>

<details>
<summary>Screenshots AWS v4</summary>

image

image

</details>

<details>
<summary>Screenshots AWS node pools</summary>

image

image

</details>

Other changes:

- Removed old instance selection components
- Added helper hooks for InstanceSelector component
- InstanceSelector component now handles both AWS & Azure
- Shuffling around some styles


src/components/Cluster/ClusterDetail/InstanceTypeSelector/InstanceTypeRow.tsx
```diff
@@ -0,0 +1,64 @@
...
+  cpu,
+  ram,
+  name,
+  description,
+  isSelected,
+  selectInstanceType,
+}) => {
+  return (
+    <>
+      <Global styles={BulletStyle} />
```
axbarsan, line 32: Btw, a question about this. Do the styles have to be global? We could have conflicting classNames across components

Would it work to style the `Tr` component and add the bullet styling there?

For example:

```javascript
const StyledTr = styled(Tr)`
	.release-selection-radio {
		...
	}
`;
```

ubergesundheit: I think you're totally right! The current implementation will also inject the styles n times for n items so not really ideal.

I'll change that

axbarsan:
Tested on and it's working great

A little nitpick: What do you think about aligning VM size names to the left?
image

ubergesundheit:
> A little nitpick: What do you think about aligning VM size names to the left?

Yeah, as long as the instance names don't have the same length, I think this is a good idea.

PR #1582, opened 2020-06-17
Prevent duplicate FlashMessages by default

axbarsan:
As seen on https://gigantic.slack.com/archives/C0148R1G12N/p1591879128115000

As some pages are auto-refreshing, there are situations when the user gets spammed with a lot of duplicated flash messages. Since we only display 6 messages at a time, the others will still be in `noty`'s queue, and they will go on until you've exhausted all of them (you have to close them manually, in case they are errors).

In order to solve this, I've created a separate global FlashMessage queue which tracks messages displayed (or prepared to be displayed). New entries will be added when the `noty` `.show()` method is called, and will be removed after they've been closed.

Based on that, new entries will not be added to noty's queue if they are already in our global queue by default. You'd have to call `.show()` manually to get it to display.


src/lib/__tests__/flashMessage.ts
```diff
@@ -0,0 +1,44 @@
...
+      messageType.ERROR,
+      messageTTL.MEDIUM
+    );
+    new FlashMessage(
+      `Yo! Something went wrong.`,
+      messageType.ERROR,
+      messageTTL.MEDIUM
+    );
+    new FlashMessage(
+      `Yo! Something went wrong.`,
```
oponder, line 36: I see the messageType is different, and so that's why we wanna see 2 notifications instead of 1 in this test. But maybe also change the text here so it's a bit clearer what the intent of this test is?

Or did you want to exercise checking that messageType is part of what is checked when checking if a notification is already in the queue?

axbarsan: I added a comment explaining what it's trying to achieve.

Exactly what you said, I want to ensure that the message type is taken into consideration when checking if a message is already in the queue.

PR #1583, opened 2020-06-18
Clean up catalog listing.

oponder:
From: https://gigantic.slack.com/archives/CQ465CFG8/p1591702420005000

When we introduced showing internal catalogs to admins, it made the catalog listing look pretty bad for us, because many internal catalogs are missing their images.

This PR adds a fallback acronym for catalogs that are missing their image.

It also moves all internal catalogs to the bottom of the list.

Also applies a change to the catalog type icons that we discussed during a sig-ux sync (putting the icon and the text in the same box, instead of having them separate. 

Screenshot 2020-06-18 at 3 09 37 PM


src/components/AppCatalog/CatalogList/CatalogRepo.js
```diff
@@ -9,6 +9,13 @@ import CatalogTypeLabel from 'UI/CatalogTypeLabel';
 
 import CatalogExternalLink from './CatalogExternalLink';
 
+function acronymize(text) {
+  var matches = text.match(/\b(\w)/g); // ['J','S','O','N']
+  var acronym = matches.join(''); // JSON
```
ubergesundheit, line 14:
```suggestion
  const matches = text.match(/\b(\w)/g); // ['J','S','O','N']
  const acronym = matches.join(''); // JSON
```

src/components/AppCatalog/CatalogList/CatalogRepo.js
```diff
@@ -47,6 +55,22 @@ const Image = styled.img`
   background-color: ${({ theme }) => theme.colors.white1};
   flex-shrink: 0;
   border-radius: 5px;
+
+  &:before {
+    content: "${(props) => acronymize(props.alt)}";
```
axbarsan, line 60: Can we pass the acronyms as a prop for the `Repo` component instead of having logic in the styled component?

src/components/AppCatalog/CatalogList/CatalogRepo.js
```diff
@@ -47,6 +54,21 @@ const Image = styled.img`
...
+    position: absolute;
+    border-radius: 5px;
+    height: 100px;
+    width: 100px;
+    background-color: #fff;
+    color: #2e556a;
+    font-size: 20px;
+    align-items: center;
+    justify-content: center;
+    display: flex;
```
marians, line 70: VS code just notified me that there are two `display` properties in here.

PR #1586, opened 2020-06-18
Let user install IC from the Cluster Detail Ingress tab

axbarsan:
Fixes https://github.com/giantswarm/giantswarm/issues/10076

This adds the possibility to install an Ingress Controller from the `Ingress` tab in `Cluster details`, if there is none installed.

I refactored the Ingress install button from the `Getting Started / Ingress` guide. It is now using the same view components that we use everywhere, and the appearance slightly changed (paddings, font sizes and weights).

Also followed Gerald's approach and used test queries that are a bit more accessible.

### Preview

<details>

<summary>Cluster details :: Ingress tab</summary>

* **No IC installed**
image

* **IC installed**
image

</details>

<details>

<summary>Refactored Getting Started :: Install Ingress</summary>

* **Default**
image

* **Loading**
image

* **Installed**
image

</details>


src/components/UI/LoadingIndicator.js
```diff
@@ -9,9 +9,20 @@ const ProgressButtonStatusIndicator = styled.div`
 `;
 
 const LoadingIndicator = ({ loading, loadingPosition, timeout, ...rest }) => (
-  <ProgressButtonStatusIndicator {...rest}>
+  <ProgressButtonStatusIndicator
```
axbarsan, line 12: Added a few accessibility attribues, so we'd be able to use more accessible queries in tests

marians:
image

I'd like to change `Click this button to install an ingress controller on <cluster id>` to

    This will install the NGINX Ingress Controller app on cluster <cluster id>

axbarsan:
> I'd like to change `Click this button to install an ingress controller on <cluster id>` to
> 
> ```
> This will install the NGINX Ingress Controller app on cluster <cluster id>
> ```

Updated code and screenshots

PR #1544, opened 2020-06-09
Cluster creation refactor

ubergesundheit:
This PR refactors a huge portion of cluster creation and node pool creation. For now this focues (mostly) on untangling the css shared between some components.

I hope this PR isn't too big. If you want we could go through the changes in a call or I could split this up.

Towards giantswarm/giantswarm#11004 giantswarm/giantswarm#11082

Some notable changes:

- Introduction of several UI components derived from cluster & node pool creation
- Introduction of a "base" spacing which can be used to make margins & paddings more consistent ([Definition](https://github.com/giantswarm/happa/pull/1544/files#diff-ca70e333d313532b406dbb1fb36766c6R65), [Usage example](https://github.com/giantswarm/happa/pull/1544/files#diff-b1d40bbe9324d38f81b881636f301e86R4-R10))
- Usage of the our `<Input>` component for cluster name and node pool name as well as wrapping other inputs like release selector and instance type selector

Here are some screenshots:

## AWS v4 Cluster creation

<details>
<summary>Click here for screenshots</summary>

### Current
aws_legacy_old

### This PR
aws_legacy_new

</details>

## AWS v5 Cluster creation (node pools)

<details>
<summary>Click here for screenshots</summary>

### Current
aws_v5_old

### This PR
aws_v5_new

</details>

## AWS v5 Cluster creation (node pools) \# 2

<details>
<summary>Click here for screenshots</summary>

### Current
aws_v5_2_old

### This PR
aws_v5_2_new

</details>

## AWS add node pool from cluster detail page

<details>
<summary>Click here for screenshots</summary>

### Current
aws_add_np_new

### This PR
aws_add_np_old

</details>

## Azure < 11.1.0

<details>
<summary>Click here for screenshots</summary>

### Current
azure_pre_old

### This PR
azure_pre_new

</details>

## Azure > 11.1.0

<details>
<summary>Click here for screenshots</summary>

### Current
azure_post_old

### This PR
azure_post_new

</details>

## KVM

<details>
<summary>Click here for screenshots</summary>

### Current
kvm_old

### This PR
kvm_new

</details>

PR #1587, opened 2020-06-20
Refactor releases state to use store & let ReleaseSelector directly access the state

ubergesundheit:
In this PR, I follows the store model introduced in #1462 for fetching releases. At the same time I did the following changes:

- Move all cluster creation logic into `NewClusterWrapper.tsx` introduced in #1544 
  - Removal of the `NewCluster.js` which contained a redundant react router route and some logic for document title and breadcrumb config
- Move release massaging into store action
  - Extraction of kubernetes version
  - Extraction of release note url
- The ReleaseSelector directly connects to the redux state
- Release fetching errors don't generate a FlashMessage any more. Instead, the error will be displayed in place of the release selector


src/components/Cluster/NewCluster/NewClusterWrapper.tsx
```diff
@@ -41,21 +43,9 @@ const clusterNameLengthValidator: IValidationFunction = (value) => {
...
-}
-
-const NewClusterWrapper: FC<INewClusterWrapper> = ({
-  activeSortedReleases,
-  breadcrumbPathname,
-  releases,
-  selectableReleases,
-  selectedOrganization,
-}) => {
+interface INewClusterWrapper extends RouteComponentProps<{}> {}
```
axbarsan, line 46:
```suggestion
interface INewClusterWrapperProps extends RouteComponentProps<{}> {}
```

src/components/Cluster/NewCluster/NewClusterWrapper.tsx
```diff
@@ -41,21 +43,9 @@ const clusterNameLengthValidator: IValidationFunction = (value) => {
...
+
+const NewClusterWrapper: FC<INewClusterWrapper> = ({ location, match }) => {
```
axbarsan, line 48:
```suggestion
const NewClusterWrapper: FC<INewClusterWrapperProps> = ({ location, match }) => {
```

src/stores/releases/actions.ts
```diff
@@ -0,0 +1,56 @@
...
+
+import { createAsynchronousAction } from '../asynchronousAction';
+import { IReleaseActionResponse } from './types';
+
+export const loadReleases = createAsynchronousAction<
+  void,
+  IState,
+  IReleaseActionResponse
+>({
+  actionTypePrefix: 'RELEASES',
```
axbarsan, line 14: I believe it would make more sense to have the base action name something like `RELEASES_LOAD`, so we can leave room for other release actions.

ubergesundheit: Ah yeah, `actionTypes.js` also defines these. I forgot to clean this up

PR #1598, opened 2020-06-22
 Make non-active releases better discoverable for Giant Swarm staff 

ubergesundheit:
Solves giantswarm/giantswarm#11564

This PR enables admin users to better see which releases are active for customers.

 
image


src/components/Cluster/NewCluster/ReleaseSelector/ReleaseSelector.tsx
```diff
@@ -93,28 +96,36 @@ const ReleaseSelector: FC<IReleaseSelector> = ({
...
-                isSelected={version === selectedRelease}
-                selectRelease={selectRelease}
-              />
-            ))}
-          </tbody>
-        </Table>
+        <>
+          {isAdmin && (
+            <p>
+              <i className='fa fa-warning' /> Dark font indicates a inactive or
```
ubergesundheit, line 102:
```suggestion
              <i className='fa fa-warning' /> Dark font indicates an inactive or
```

src/components/Cluster/NewCluster/ReleaseSelector/__tests__/ReleaseSelector.ts
```diff
@@ -215,4 +216,23 @@ describe('ReleaseSelector', () => {
...
+        ...defaultStoreState,
+        ...{ main: { loggedInUser: { isAdmin: true } } },
+      }
+    );
+
+    fireEvent.click(screen.getByText(/Available releases/i));
+
+    expect(
+      screen.getByText(
+        /Dark font indicates a inactive or wip release which is only available to Giant Swarm staff/i
```
ubergesundheit, line 234:
```suggestion
        /Dark font indicates an inactive or wip release which is only available to Giant Swarm staff/i
```

axbarsan:
Code looks good, but I'm not sure about the really dark shade of the color. It looks a bit out of place. Can we try something with a little bit more blue?

marians:
+1 on the dark font. The contrast between background also appears too low.

ubergesundheit:
Classical "looks good on my machine" situation ;)

I've changed it to `#020a29` in 31d8087 (Screenshot at the end of this comment)

As marian already wrote, the contrast isn't very high and finding a better color is IMO really difficult. A solution could be to use a lighter color instead of a darker.


Dark variant of 31d8087
image

Light variant (not commited) `#768e9d`

image

marians:
I think the light variant is okay, but...

Let's make sure that at least on mouse over, with the row under the cursor being highlighted, the contrast is ideal. I guess the font should turn full `#fff` to achieve that.

oponder:
Would it be a lot more work to create second dropdown, only shown when logged in as an admin. 

So we'd have like

▼ Available Releases
▼ WIP Releases 

(And then we could use the eye with a slash through it icon, like I did for the internal catalog listing perhaps)

oponder:
But if you wanna move on, I approve the lighter color variant

ubergesundheit:
> Would it be a lot more work to create second dropdown, only shown when logged in as an admin.

Should be doable and not much work and I like the idea! I would do this in another PR though because currenly we're using the `active` property of a release which is deprecated. Ludacris made changes to releases that introduced a `state` property which can take values `active`, `deprecated` and `wip`. So I would take your suggestion and implement it once happa reads the `state` property. Is that ok for you?

ubergesundheit:
Any more change requests here? I created issue giantswarm/giantswarm#11746 for the suggestion in https://github.com/giantswarm/happa/pull/1598#issuecomment-647969420

PR #1599, opened 2020-06-23
Set catalog placeholder image background based on color hash

marians:
The idea here is to give a unique color background to every catalog icon that does not have an image.

image


marians:
@giantswarm/chapter-front-end Any idea how to get the rounded corners right? There is white background appearing in the corners.

image

oponder:
The way I do this fallback is a bit of a hack perhaps, because I found it as a more straightforward way of handling it actually than the onError way we did in the AppListing. If we can get it to work nicely here, we can apply it in the app listing too eventually and remove some code to keep track of which images failed to load.

Anyways, I think what we might be seeing in the top left corner is the missing image icon, peaking through beneath the element that I am adding over it.

Not sure why the other corners aren't perfectly transparent though.. 🤔 

For illustration purposes, I disabled some of the ":before" styles:

Screenshot 2020-06-23 at 3 19 27 PM

There might be a way to adjust the position or look of that missing image icon to at least clean up the top left border.

ubergesundheit:
Checked this out locally and it does not work in either Firefox 77.0.1 nor Chromium  80.0.3987.149 for me:

I suspect setting it needs the `background-color` also being set in the `:before`-style of `Image`.

The white top left corner does not appear in Firefox, I guess oponders suggestion is right, it should be possible to attach an event listener to `onError` of the image and then replace the element with a div

Left Chromium, right Firefox:
image

marians:
This is going beyond me already. Anyone to take this over?

ubergesundheit:
I can do it

src/utils/cachingColorHash.ts
```diff
@@ -0,0 +1,29 @@
+import ColorHash from 'color-hash';
+
+class CachingColorHash {
+  private cache: { [k: string]: string } = {};
```
axbarsan, line 4:
```suggestion
  private cache: Record<string, string> = {};
```

PR #1603, opened 2020-06-25
Refactor cluster load apps

oponder:
Moves the action that loads cluster apps into the new paradigm.

Also moves the cluster reducer into it's own folder, though with practically no changes, aside those needed to support the refactored loadClusterApps action.


src/stores/clusterapps/actions.ts
```diff
@@ -0,0 +1,65 @@
+import GiantSwarm, { V4App } from 'giantswarm';
+import { FlashMessage, messageTTL, messageType } from 'lib/flashMessage';
+
+import { createAsynchronousAction } from '../asynchronousAction';
+
+interface IRequest {
```
oponder, line 6: Is it ok to keep these names super short like this? or will that conflict down the line? my typescript mental still isn't complete.

src/stores/clusterapps/actions.ts
```diff
@@ -0,0 +1,65 @@
...
+}
+/**
+ * Loads apps for a cluster.
+ *
+ * @param {String} clusterId Cluster ID
+ */
+export const loadClusterApps = createAsynchronousAction<
+  IRequest,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  any,
```
oponder, line 22: It might be a monumental undertaking, but if we get a full interface for the State, then we can get rid of this any. I might try it out tonight.

src/stores/clusterapps/actions.ts
axbarsan: As long as they don't contradict with any global interface, then it's fine.

If it was to contradict with another interface, then it would actually extend it.

oponder: Ah ok, so having a more specific name would be better in this case :+1:

PR #1604, opened 2020-06-25
Refactor cluster install app action

oponder:
EDIT: Problem is more of a linting problem than a structure problem.

Moving all the app actions to the `stores` folder. Some slight changes needed to the asynchronous action creator, since some actions dispatch other actions, had to pass in the dispatch function to the perform method, so that it can do that.

Im having some problems though, this new type of action is not awaitable, we don't get the thunk back. 

That enforces something we wanted to do, which was not have dispatches with .then / .catch (basically the same as await) in the UI layer... but that means to sequence things in the UI we need to rely fully on loading flags and their results and... it's feeling like really a lot of effort for something that used to be much simpler in my head before. 

Can someone remind me again of the benefits of this... I'm losing a bit of faith.


src/components/Cluster/ClusterDetail/Ingress/InstallIngressButton.tsx
```diff
@@ -77,11 +77,11 @@ const InstallIngressButton: React.FC<IInstallIngressButtonProps> = ({
     tryToLoadApps();
   }, [dispatch, clusterID]);
 
-  const installIngressController = async () => {
+  const installIngressController = () => {
     try {
       setIsInstalling(true);
 
-      await dispatch(installApp(appToInstall, clusterID));
```
oponder, line 84: A consequence of the new approach, no more thunk is returned here, it is not returned by the callAPIMiddleware.. maybe it could be... but I'm not sure if we wanted that. This enforces something we wanted originally.

But that means this dispatch and the next are fired at pretty much the same moment, which isn't ideal in my mind.

In this case it means we might load cluster apps so fast that we don't see the newly installed app in the list yet.

Screenshot 2020-06-26 at 1 31 02 AM

oponder: Before you would 100% for sure see:

```
INSTALL_APP_REQUEST
INSTALL_APP_SUCCESS
LOAD_CLUSTER_APPS_REQUEST
LOAD_CLUSTER_APPS_SUCCESS
```

oponder:
Actually it seems like it is awaitable, but the linter does not have the right clues to understand that it's ok.

src/stores/asynchronousAction.ts
```diff
@@ -10,13 +14,21 @@ export interface IAsynchronousAction<S, R> {
...
 export interface IAsynchronousActionParams<P, S, R> {
   actionTypePrefix: string;
-  perform: (state: S, payload?: P) => Promise<R>;
+  perform: (
+    state: S,
+    dispatch: ThunkDispatch<IState, void, AnyAction>,
+    payload?: P
+  ) => Promise<R>;
   shouldPerform: (state: S) => boolean;
+  throwOnError: boolean;
```
oponder, line 31: While creating our action we can now define if it should throw or not.

I guess the goal will be to not need throwing, but for now to avoid to many steps in the refactoring at once, I added this as an option. That way when we are done moving everything it will also be easy to identify which actions and bits of the UI need further refactoring.

src/stores/clusterapps/actions.ts
```diff
@@ -91,69 +103,53 @@ export const installApp = createAsynchronousAction<
...
-      );
+    };
+
+    const createApp = createAppV4orV5(state, clusterId);
+
+    try {
+      await createApp(clusterId, name, request);
+    } catch (error) {
+      showAppInstallationErrorFlashMessage(name, clusterId, error);
       throw error;
```
axbarsan, line 142: Hope we can get rid of this at some point

src/components/Cluster/ClusterDetail/Ingress/InstallIngressButton.tsx
axbarsan: I think we can move the call to the app reloading action creator to the app install action creator, since we're really never going to install an app and not fetch the latest app list immediately after.

We can even have a common configuration argument to these type of creational/mutational action creators, whether it should refresh the list after executing the creation/mutation. I think this is a good step in the direction of simplifying the data fetching part of our view components.

src/stores/asynchronousAction.ts
```diff
@@ -10,13 +14,21 @@ export interface IAsynchronousAction<S, R> {
   // and it is not a real "action" that a reducer will ever see.
   type: string;
   types: IActionTypeCollection;
-  doPerform: (state: S) => Promise<R> | undefined;
+  doPerform: (
+    state: S,
+    dispatch: ThunkDispatch<IState, void, AnyAction>
+  ) => Promise<R> | undefined;
```
axbarsan, line 20: Does this make sense with the way you designed it?
```suggestion
  ) => Promise<R> | void;
```

axbarsan:
> Actually it seems like it is awaitable, but the linter does not have the right clues to understand that it's ok.
> 
> Screenshot 2020-06-26 at 2 25 25 AM

A possible solution:

```typescript
// Custom dispatch type in `asynchronousAction.ts`.
export interface IAsynchronousDispatch<S> {
  <R>(action: IAsynchronousAction<S, R>): Promise<R>;
}

// Defining dispatch inside component.
const dispatch: IAsynchronousDispatch<IState> = useDispatch();
```

This seems to work, I haven't tested every case, though.

oponder:
Thanks for the thorough reviews as always! :+1:

PR #1616, opened 2020-06-30
Catalog Icon fixes

ubergesundheit:
- Uses the url for color generation
- Does not set the generated color when image is available


ubergesundheit:
Sorry, I was kind of in a rush.. So here is one screenshot

image

ubergesundheit:
Switching to using just the `name` of the catalog instead of the full path, creates more "pleasing" colors in IMO

image

PR #1617, opened 2020-07-01
Don't rely on bootstrap's col-x classes for layout.

oponder:
Removes all uses of `col-x` classNames throughout the app.

Created `Section` and `TwoInputArea` UI components to reuse two patterns I found while doing this refactoring.

**Responsiveness Changes**

This also changes the larger breakpoint to 1024px instead of 1200px, because it removes the paddings that would get added in certain cases.

- [x] Fix centered footer
- [x] Fix width of getting started guide content
- [x] Fix alignment and margins of navigation items

Towards Happa 2020


src/components/Cluster/ClusterDetail/AddNodePool.js
```diff
@@ -688,12 +688,6 @@ class AddNodePool extends Component {
           <StyledInput
             labelId={`scaling-range-${this.props.id}`}
             label='Scaling range'
-            hint={
```
oponder, line 691: This was duplicated in NodeCountSelector, but hidden when in a certain context (see: https://github.com/giantswarm/happa/pull/1617/files#diff-d8b088a074d65e4a00a05a47c1687d4fL176). 

Instead of doing it twice I just allow the one in NodeCountSelector to show.

ubergesundheit: Good find! We should avoid these "unspecified" css selectors in the future

src/components/Cluster/ClusterDetail/ClusterApps.js
```diff
@@ -142,6 +142,19 @@ const Disclaimer = styled.p`
...
+const PreinstalledApps = styled.div`
+  display: flex;
+
+  & > div {
+    flex: 1;
+    margin-right: 15px;
+
+    &:last-child {
+      margin-right: 0px;
+    }
```
ubergesundheit, line 154: Is it possible to rewrite these selectors to use classes to avoid a problem like you found in https://github.com/giantswarm/happa/pull/1617/files#r448168726

src/components/UI/TwoInputArea.js
```diff
@@ -0,0 +1,16 @@
...
+
+  div {
+    flex: 1;
+    margin-right: 25px;
+
+    &:last-child {
+      margin-right: 0px;
+    }
+  }
+`;
```
ubergesundheit, line 14: I just thought more about this issue with the "nested" element selectors. So writing it like this, it is somewhat hard to change the styles of the inner `div`.

My suggestion is to create an additional styled component in this file like this:

```
const InnerTwoInputArea = styled.div`
  flex: 1;
  margin-right: 25px;

  &:last-child {
    margin-right: 0px;
  }
`;
```

WDYT?

oponder: Good suggestion to make things more explicit

oponder: ok done!

src/components/Cluster/ClusterDetail/ClusterApps.js
ubergesundheit: Is it possible to rewrite these in the same sense as discussed in https://github.com/giantswarm/happa/pull/1617#discussion_r448390365?

I see this is not as severe as the duplicated components like https://github.com/giantswarm/happa/pull/1617/files#diff-d8b088a074d65e4a00a05a47c1687d4fL176 but would be nevertheless good to be resolved.

Giving an approve so this won't block you

oponder:
Yeah maybe in another refactor pass I'll go through and get rid of more of these type of 'div' selectors, and also just solo <div>'s throughout the app, I think they are an antipattern and nice that you identified it. But for now I'd like to move on to other things.

ubergesundheit:
I'll try to come up with a change mentioning this in fmt :+1:

PR #1619, opened 2020-07-02
Clear a delayed value if another one comes in before the current one …

oponder:
I noticed that the loading spinner next to the `install ingress controller` button was always spinning.

After a bit of investigating, it looked like the useDelayedChange effect was firing a 'true' value for the loading flag after a `false` value, so somehow out of sequence. 

My hunch is that it was the setTimeout firing in a sequence we didn't expect.

But still weird that they would happen out of sequence, since there should be some nanoseconds between the loading flag going from false, to true, back to false again. But maybe not enough time for it to matter.

This feels like a regression to me after a refactoring @axbarsan , Im not sure what it is though about the refactoring that made this bug appear though.

### DIFF
--- a/src/lib/effects/useDelayedChange.js
+++ b/src/lib/effects/useDelayedChange.js
@@ -7,6 +7,7 @@ import { useEffect, useRef, useState } from 'react';
  */
 function useDelayedChange(value, delay) {
   const [delayedValue, setDelayedValue] = useState(value);
+  const [timeoutRef, setTimeoutRef] = useState(undefined);
 
   const isComponentMounted = useRef(true);
   const lastUpdateFinishTime = useRef(Date.now());
@@ -29,11 +30,15 @@ function useDelayedChange(value, delay) {
 
     lastUpdateFinishTime.current = currentTime + delay;
 
-    setTimeout(() => {
+    clearTimeout(timeoutRef);
+
+    const ref = setTimeout(() => {
       if (isComponentMounted.current) {
         setDelayedValue(value);
       }
     }, secondsToNextUpdate);
+
+    setTimeoutRef(ref);
   }, [value, delay]);
 
   return delayedValue;


oponder:
oh the tests really don't like this change :P

axbarsan:
ty w

axbarsan:
This is weird, because I don't see an infinite loading spinner, neither on Gauss or Ginger, which use different versions of happa. Can you maybe post a screenshot?

oponder:
We don't auto deploy master anymore, so you'd have to deploy the latest there to see it, or run it locally.

On gorilla `HAPPA_PROXY_INSTALLATION=gorilla yarn start` I see:

Screenshot 2020-07-02 at 7 52 53 PM

axbarsan:
Umm this is a bit weird... Using latest master in happa

image

oponder:
It must be a timing thing or a race condition somehow.

oponder:
Closing because I don't think the timeout is the problem... but investigating further.

oponder:
A little output from what happens on machine at least.

This is logging the isLoading state of the ingress button, and then also logging inside the setTimeout of useDelayedEvent, and showing the 'id' of the timeout (the return value of setTimeout).

So the ingress button goes through false -> true -> true -> false -> false

Three timeouts are created and triggered as a result. What's interesting is that the id's are not in ascending order, assuming setTimeout creates id's in a numerical order, that doesn't seem right to me. 

Screenshot 2020-07-02 at 8 55 40 PM

oponder:
Ok it looks like only the true transitions have a delay value set to them.

Screenshot 2020-07-02 at 8 59 32 PM

oponder:
Ok so interpreting what `useDelayedChange` is supposed to do, it's trying to slow down the transition from `false` to `true` since it happened really quickly in succession. 

But then another transition comes in to bring it back to false. But that comes in sooner than the timeout that it created for the transition to true, but not soon enough for it to get a significant delay of any kind.

I think my idea of clearing the timeout kinda made sense. 

If a new value comes in (with enough of a delay to not need any artificial one), then we can go straight to that value.

In this case though, we could also make this work possibly by making sure that the installingressbutton starts with isLoading set to true.

axbarsan:
What I believe happens:

* Multiple changes are queued, multiple timeouts are created
* Component is unmounted (because of some logic in one of the parent components)
* Component is remounted (the old ref tracking the next update is now set to 0)
* Next update will be scheduled immediately, because the timer has been reset
* Changes scheduled from the previous render will be overlapped with the current changes
* Weird behaviour

Possible solution:

Set up a queue inside the hook, where we would keep all pending timeouts (which we'll then remove from the queue, once they're executed). When the component gets unmounted, we cancel all the pending timeouts.

oponder:
I considered a queue as well, but I felt that the feature doesn't really warrant the implementation complexity. 

In this specific case, we're giving the buttons also a timeout of 0, which seems to indicate we want them to transition right away.

the useDelayChange was always hard coded with a value of 500ms though (which is weird because the timeouts I'm seeing are being set for like 900ms or greater) so there's something odd there as well.

_but_ given the button is saying it wants to transitition with a timeout of 0, I thought that is something that useDelayChange should take into consideration as well, and made this PR: https://github.com/giantswarm/happa/pull/1620

That seems to solve it... but Happa is acting weird for me at the moment because my connection is a bit flaky.

axbarsan:
> I considered a queue as well, but I felt that the feature doesn't really warrant the implementation complexity.

It shouldn't be too hard to do, we already have the basic data structure for this:
https://github.com/giantswarm/happa/blob/master/src/lib/Queue.ts

> I thought that is something that useDelayChange should take into consideration as well, and made this PR: #1620

Not sure I understand. The animation timeout prop is there to configure how long it takes for a transition to appear/enter/exit. http://reactcommunity.org/react-transition-group/transition#Transition-prop-timeout

`useDelayedChange` is meant to postpone certain variable changes, so for example, if I do this:

- value: true;
- value: false;
- value: true;

The result will be:

- wait: 0ms;
- value: true;
- wait: 500ms;
- value: false;
- wait: 500ms;
- value: true;

In order to achieve this, each update will be scheduled by adding extra `500ms` to the time left until the last update scheduled. Better said:

- setTimeout(value: true, 0)
- setTimeout(value: false, 500)
- setTimeout(value: true, 1000)

Not sure why the transition timeouts in/out should influence this.

oponder:
> The animation timeout prop is there to configure how long it takes for a transition to appear/enter/exit. 

I was (wrongly) thinking if you want an animation to appear with 0 delay, then having the `useDelayChange` effect there as well wouldn't be helpful, but I think now with your explanation i get it.

I think the bug is actually that `lastUpdateFinishTime` should be `secondsToNextUpdate` if `secondsToNextUpdate` is not 0.

Right now we think the `lastUpdateFinishTime` is just `currentTime + delay`, but that means that if there was a delayed update already in the pipeline, we wouldn't be considering the offset of _that_ delay.

new PR coming.

oponder:
Ok, here's my latest attempt :P https://github.com/giantswarm/happa/pull/1621

PR #1647, opened 2020-07-13
Add compatibility changes for Azure V5

axbarsan:
Towards https://github.com/giantswarm/roadmap/issues/160

This PR adds support for listing/creating/searching Azure v5 clusters and listing Azure node pools.

This has the `dev-change` label because it shouldn't be in the release notes, as it's not a complete feature yet.


src/shared/constants/index.js
```diff
@@ -21,6 +21,7 @@ export const Constants = {
 
   DEFAULT_NODEPOOL_NAME: 'Unnamed node pool',
   AWS_V5_VERSION: '10.0.0',
+  AZURE_V5_VERSION: '40.0.0',
```
axbarsan, line 24: Don't forget to change this 😬

package.json
```diff
@@ -50,7 +50,7 @@
     "deep-diff": "1.0.2",
     "dotenv": "8.2.0",
     "emotion-theming": "10.0.27",
-    "giantswarm": "git+https://github.com/giantswarm/giantswarm-js-client.git",
+    "giantswarm": "git+https://github.com/giantswarm/giantswarm-js-client.git#enable-azure-v5",
```
axbarsan, line 53: Don't forget to change this 😬

ubergesundheit:
Which installations has the required api changes?

axbarsan:
> Which installations has the required api changes?

Both `ghost` and `godsmack`. This branch is already deployed on both of those 👍

ubergesundheit:
> Both ghost and godsmack. This branch is already deployed on both of those +1

Just tried to create a cluster with node pools on godsmack which resulted in an `Internal server error`. Node pools on existing clusters don't seem to be working either. They're just empty.

axbarsan:
> Just tried to create a cluster with node pools on godsmack which resulted in an `Internal server error`. Node pools on existing clusters don't seem to be working either. They're just empty.

Creating node pools is not supported, yet. That will come in another PR.

I added a `null` check and deployed latest API on `godsmack` and existing v5 clusters work fine. Can you check again, please?

ubergesundheit:
I can confirm creating a node pool cluster without a node pool works on `godsmack`.

PR #1654, opened 2020-07-16
Refactor Organization Dropdown & make it scrollable

ubergesundheit:
Towards giantswarm/giantswarm#11928

This PR refactors the OrganizationSelector 

- to use the DropdownMenu of Happa
- remove lodash & Bootstrap 
- make it scrollable

scroll_organization

### DIFF
--- a/src/components/UI/DropdownMenu.js
+++ b/src/components/UI/DropdownMenu.js
@@ -19,7 +19,6 @@ export const List = styled.ul`
   margin: 2px 0 0;
   padding: 5px 0;
   width: 180px;
-  background: ${(props) => props.theme.colors.shade2};
   z-index: 1;
   background-color: ${({ theme }) => theme.colors.dropdownBackground};
   border: none;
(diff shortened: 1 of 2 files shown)


oponder:
Just noticed that for me the dropdown's behaviour seems to have regressed a bit now. Anyone else?

Selecting an org doesn't seem to work. When clicking on an organization name, the dropdown closes, but Happa doesn't navigate anywhere.

When clicking on "Details for giantswarm" or "Manage Organization", the site navigates correctly, but the dropdown doesn't close. Also the links in the dropdown get bold depending on the page we are on, but "Manage Organizations" stays bold als on the "Details for giantswarm" page. Highlighting the active 'link' in the dropdown isn't needed, I think it came in as an unintended side effect?

I noodled around a bit in the code, I think it's because the DropDown component is not recognising the individual organisation links as a 'child', and there is some logic in place for closing the dropdown if it loses focus to anything that isn't a child of it. I didn't go much further than that, but wanted to write down what I found.

ubergesundheit:
Oh, thanks for letting me know! I'll work on this asap!

PR #1668, opened 2020-07-20
Clean up some unused props in cluster overview (Home)

ubergesundheit:
This PR does a little cleanup of props in the `Home` component and its children.


src/components/Home/Home.js
```diff
@@ -26,28 +23,48 @@ import ClusterDashboardItem from './ClusterDashboardItem';
...
     this.visibilityTracker.removeEventListener(this.handleVisibilityChange);
     this.props.clearInterval(this.refreshInterval);
   }
 
+  componentDidUpdate(prevProps) {
+    const { selectedOrganization } = this.props;
+
+    if (selectedOrganization !== prevProps.selectedOrganization) {
+      this.setState({
+        newClusterPath: Home.newClusterPath(selectedOrganization),
```
axbarsan, line 56: What's the reason for keeping this in the local state? It's triggering an unnecessary component update after the selected organization is changed.

Having it in the `render()` method makes sure the value will be recomputed if the selected organization is changed.

If performance is your concern, as this component will auto-refresh, and the cluster path is recalculated with each update, we can always memoize the value

ubergesundheit: The value depends on `this.props.selectedOrganization`. So my impression is that the implementation before was recomputing the `newClusterPath` value on each render while the changed version will only recompute it when the `selectedOrganization` changes.

axbarsan: That is correct, but as I said, updating the state again after the selected organization changed will actually trigger another update in a component that has many grandchildren, which is expensive.

My suggestion was to keep computing the cluster detail path in the render method, to prevent the second update, but to use memoization, so the value would only be be recomputed when the selected organization is changed

axbarsan: Maybe this is a better explanation
https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#when-to-use-derived-state

ubergesundheit: Ah yeah I see my mistake now. In 35749b48 I changed it back to compute the path on every render.

axbarsan: But your point is valid, it is being computed again with every update. We can use `underscore`'s `memoize()` function to memoize the return value of the `newClusterPath()` static method, so it will only be computed when the path changes.

What do you think?

ubergesundheit: I'll see if it works

ubergesundheit: check out https://github.com/giantswarm/happa/pull/1668/commits/27e4fb0dceb6bf4cd421a729c0426dcc5321d9a1

src/components/Home/Home.js
```diff
@@ -149,13 +151,11 @@ class Home extends React.Component {
 
 Home.propTypes = {
   clearInterval: PropTypes.func,
+  setInterval: PropTypes.func,
```
oponder, line 154: what's actually setting this prop? I can't find it in a quick search...

oponder: Oh its the ReactTimeout() wrapper. ok cool :+1:

PR #1669, opened 2020-07-21
No unused proptypes

oponder:
Inspired by: https://github.com/giantswarm/happa/pull/1668

I thought maybe it's possible to autofix unused props. But that doesn't seem to happen.


axbarsan:
image

ubergesundheit:
What @axbarsan said ;)

src/components/Cluster/NewCluster/NewClusterWrapper.tsx
```diff
@@ -78,7 +78,7 @@ const NewClusterWrapper: FC<INewClusterWrapperProps> = ({
 
   const creationCapabilities = useMemo(
     () => makeCapabilities(selectedRelease, provider),
-    [selectedRelease, provider]
+    [selectedRelease, provider, makeCapabilities]
```
oponder, line 81: Fixes an unrelated warning.

src/components/shared/VirtualizedScrollableGrid.js
```diff
@@ -76,19 +76,20 @@ class VirtualizedScrollableGrid extends React.PureComponent {
   }
 
   updateResponsiveFeatures = debounce(() => {
+    const { theme, columnCount, adaptWidthToElement, width } = this.props;
```
oponder, line 79: destructured it here so the linter can see it being used, also kinda nice for the humans, you can see what `setColumnCount` and `setWidth` actually use from the props directly.

Possibly a `fmt` thing: try to limit (or be more explicit) about the amount of things being passed around, when a function only needs a subset of the object.

src/components/Home/UpgradeNotice.tsx
```diff
@@ -38,11 +38,13 @@ function UpgradeNotice({
   canClusterUpgrade,
   onClick,
   className,
+  clusterId,
 }: IUpgradeNoticeProps) {
   if (!canClusterUpgrade) return null;
 
   return (
     <UpgradeWrapperDiv
+      id={`upgrade-notice-${clusterId}`}
```
oponder, line 47: it warned about clusterId not being used, but it most certainly is being used in `mapStateToProps`

So I added it here as an id to kind of trick the linter...  not too happy about that. Any thoughts?

ubergesundheit: Thats a perfectly valid fix :+1:

PR #1670, opened 2020-07-21
azure v5: nodepool creation

axbarsan:
Towards https://github.com/giantswarm/roadmap/issues/160

⚠️  Asking for review, but will not merge until API changes have been released.
Has `dev-change` label because it shouldn't show up in the release notes, as Azure node pools are not 100% done.

Nothing special, should behave the same as AWS, without node pool scaling and spot instances.

Tested on both Azure and AWS and no issues so far.

## Preview

<details><summary>Cluster creation form on supported release versions</summary>

image

</details>


package.json
```diff
@@ -50,7 +50,7 @@
     "deep-diff": "1.0.2",
     "dotenv": "8.2.0",
     "emotion-theming": "10.0.27",
-    "giantswarm": "git+https://github.com/giantswarm/giantswarm-js-client.git",
+    "giantswarm": "git+https://github.com/giantswarm/giantswarm-js-client.git#add-azure-v5-nodepool-creation",
```
axbarsan, line 53: Use master 😺

PR #1671, opened 2020-07-21
fix non-admins couldn't see release changes

ubergesundheit:
Towards giantswarm/giantswarm#12244

This introduces a new release state property which contains all releases regardless of user status. This is being used for upgrading clusters where non-admins couldn't see changes between releases

This is probably related to https://github.com/giantswarm/happa/pull/1549 https://github.com/giantswarm/happa/pull/1506 and https://github.com/giantswarm/happa/pull/1587

<details>

<summary><b>Before:</b> Screenshot upgrade cluster modal</summary>

image

</details>

<details>

<summary><b>After:</b> Screenshot upgrade cluster modal</summary>

image

</details>


ubergesundheit:
I tried your suggestion in https://github.com/giantswarm/happa/pull/1671/commits/bc868b71528a8e93bcc76da2c1e2bc9ef89a7750

It seems the tests are not playing along but it looks good when running it in the browser (admins seeing non-active releases, non-admins seeing only active releases, all users see changes when upgrading from non-active version)

Anyone able to see my mistake?

ubergesundheit:
fa41c38 seems to fix the issue in the test. But the inline function is exactly the same as the [`getUserIsAdmin` function from `authSelectors`](https://github.com/giantswarm/happa/blob/dfd56b9/src/selectors/authSelectors.js#L47)

PR #1674, opened 2020-07-23
Remove app_catalog.sass

oponder:
Towards UI Refactoring: https://github.com/giantswarm/giantswarm/issues/12347


src/components/AppCatalog/AppList/AppListSearch.js
```diff
@@ -1,18 +1,41 @@
+import styled from '@emotion/styled';
 import PropTypes from 'prop-types';
 import React from 'react';
 
+const Form = styled('form')`
+  width: 300px;
+  font-size: 14px;
+  position: relative;
+
+  input {
```
axbarsan, line 10: I remember us discussing that styled components are preferred over nested selectors. Can't see anything written in `fmt` about it. What's your opinion on that?

oponder: I think it's ok to start off nested and only extract them when it seems necessary. What are the drawbacks of using nested selectors?

axbarsan: In this case there shouldn't be any problem. The problems arrive when we have nested components, and with selectors we'll also apply styles to the elements inside those nested components.
So, if we ever add another component (let's call it `ResetButton`), which would have a nested `<i/>` element inside it, we'll actually apply these styles:
```
color: #fff;
font-size: 13px;
font-weight: 800;
```
to the `<i/>` element inside the `ResetButton` component.

oponder: Ah ok. unless you have strong opinion about it, Im ok with potentially running into that and then fixing it when it's necessary. What do you think?

axbarsan: Yeah, let's move on 👍

ubergesundheit: The `fmt` section in question is stuck in this PR: https://github.com/giantswarm/fmt/pull/118

src/components/AppCatalog/AppList/AppListInner.js
```diff
@@ -161,16 +164,15 @@ class AppListInner extends React.Component {
...
-          <AppListItems
-            apps={filteredApps}
-            catalog={catalog}
-            searchQuery={searchQuery}
-            iconErrors={this.iconErrors}
-            onImgError={this.onImgError}
-            scrollToApp={this.state.scrollToApp}
-          />
-        </div>
+
```
ubergesundheit, line 167:
```suggestion
```

src/components/AppCatalog/CatalogList/CatalogList.js
```diff
@@ -1,9 +1,19 @@
+import styled from '@emotion/styled';
 import DocumentTitle from 'components/shared/DocumentTitle';
 import PropTypes from 'prop-types';
 import React from 'react';
+import Well from 'UI/Well';
 
 import CatalogRepo from './CatalogRepo';
 
+const Wrapper = styled.div`
```
ubergesundheit, line 9: Is the new `div` required here? if its only to get a way to wrap the `h1` styles, I think it would be cleaner to define the `h1` styles in a styled component

PR #1675, opened 2020-07-23
Load app catalog for latest ingress app version

ubergesundheit:
Towards giantswarm/giantswarm#12330

This PR aims to make the `Install Ingress Controller` button in the Ingress tab in cluster details to always install the latest version defined as in its catalog.

I also added a link to the app detail page and included the version to be installed in the description text

<details>

<summary>Screenshot</summary>

image

</details>

While this solution works, this seems to be a very peculiar way to achieve this goal. Do you think this is fine or should I invest the time now to move the catalog loading into the new `store` format?


axbarsan:
How about moving the whole ingress app selection, creation etc. out of the component, and into a new redux action and a few selectors? Do you think that will work?

oponder:
> I also added a link to the app detail page and included the version to be installed in the description text 

Nice!

ubergesundheit:
Yes! I like the idea of creating a redux action from that! But first, I'll fix the test and work from there.

axbarsan:
Just found out that this is a high priority issue. So if everything works properly right now, we can push it forward as it is, and create another issue for refactoring

ubergesundheit:
I tried to add more nock mocks for the offending requests but was not successful. So I chose to skip the failing tests in b642788

Idk if I can fix this until the next release (today or monday?). So if you're fine with skipping the tests for now, this would be the solution for now.

PR #1676, opened 2020-07-23
Fix trying to use apps before loading them in an app catalog

axbarsan:
As seen on https://gigantic.slack.com/archives/C3MCF5EJK/p1595517097001000

If the apps are not loaded yet, then the property is `null`.
Now we're taking that into account.

### DIFF
--- a/src/actions/catalogActions.js
+++ b/src/actions/catalogActions.js
@@ -125,8 +125,6 @@ export function catalogLoadIndex(catalog) {
           catalogName: catalog.metadata.name,
           id: catalog.metadata.name,
         });
-
-        throw error;
       });
   };
 }
--- a/__tests__/AppCatalog.js
+++ b/__tests__/AppCatalog.js
@@ -1,6 +1,4 @@
-import '@testing-library/jest-dom/extend-expect';
-
-import { fireEvent, within } from '@testing-library/react';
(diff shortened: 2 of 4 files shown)


axbarsan:
> Writing a test that fails (without this changed code) in the same way as it did in the exception report would be 💯 bonus points.

Good call! It actually exposed a memory leak in the `AppList` component 👍

src/actions/catalogActions.js
```diff
@@ -125,8 +125,6 @@ export function catalogLoadIndex(catalog) {
           catalogName: catalog.metadata.name,
           id: catalog.metadata.name,
         });
-
-        throw error;
```
oponder, line 129: LGTM, getting rid of these throws makes sense. 

But made me dig a bit into how the error is actually handled in the front end, and there is no user feedback in case of an error.

I think that's because I changed up the way these catalogs get loaded (from loading all catalogs on the CatalogList screen to loading a single one on the AppList screen. ) but I didn't implement the error handling.

So now now on failure we see "no apps matched your query", but it'd be better to show some kind of error message here:

Screenshot 2020-07-24 at 12 15 49 PM

Doesn't have to be tackled in this PR btw, but wanted to mention it

axbarsan: Yep, saw that. Already got another PR on the way 👍

PR #1678, opened 2020-07-23
Remove _breadcrumbs.sass and _gadget.sass

oponder:
Towards UI Refactoring: giantswarm/giantswarm#12347


src/components/UI/Navigation/Navigation.js
```diff
@@ -72,6 +72,38 @@ const OuterNav = styled.nav`
...
+  @media only screen and (max-width: 1024px) {
+    display: none;
+  }
+
+  a {
+    color: ${({ theme }) => theme.colors.darkBlueLighter6};
+  }
+
+  .breadcrumbs {
+    max-width: 1200px;
```
oponder, line 93: I want to move stuff like this, which is more a global layout property, into the theme, but I'm thinking to do this in a separate PR, in a more considered way.

src/components/UI/Navigation/Navigation.js
```diff
@@ -72,6 +72,38 @@ const OuterNav = styled.nav`
...
+const BreadcrumbWrapper = styled.div`
+  line-height: 20px;
+  background-color: ${({ theme }) => theme.colors.shade3};
+  padding: 7px;
+  border-bottom: ${({ theme }) => theme.colors.shade6} 1px solid;
+  border-top: ${({ theme }) => theme.colors.shade1} 1px solid;
+  font-size: 11px;
+  clear: both;
+
+  @media only screen and (max-width: 1024px) {
```
axbarsan, line 84: Can we use the existing utility here?

https://github.com/giantswarm/happa/blob/7341cdfdc44a2319a16fbfa18a2f9f42518b4fe6/src/components/UI/AppDetails/AppDetails.js#L106

src/components/UI/Navigation/Navigation.js
```diff
@@ -97,11 +129,9 @@ class Navigation extends React.Component {
...
         </div>
 
-        <div className='breadcrumb-wrapper'>
-          <div className='main'>
-            <Breadcrumbs />
-          </div>
-        </div>
+        <BreadcrumbWrapper>
+          <Breadcrumbs />
+        </BreadcrumbWrapper>
```
axbarsan, line 134: I'd extract this into a separate component, but not a blocker

PR #1679, opened 2020-07-23
Extract TOS sass, clean up some unused sass I found along the way.

oponder:


src/components/SignUp/TermsOfService.js
```diff
@@ -1,12 +1,42 @@
...
+  margin-bottom: 40px;
+`;
+
+const TOSText = styled.div`
+  overflow: scroll;
+  max-height: 300px;
+  overflow: auto;
+  border: 1px solid #1f3947;
+  border-radius: 4px;
+  background-color: #32526a;
```
axbarsan, line 14:
```suggestion
  background-color: ${({ theme }) => theme.colors.shade5};
```

src/components/SignUp/TermsOfService.js
```diff
@@ -1,12 +1,42 @@
...
 
+const Container = styled.div`
+  margin-bottom: 40px;
+`;
+
+const TOSText = styled.div`
+  overflow: scroll;
+  max-height: 300px;
+  overflow: auto;
+  border: 1px solid #1f3947;
```
axbarsan, line 12: This color is not in the theme. But can we maybe replace it with a similar one that is there?

oponder: Yeah I'll find one! I wanted to remove all these hex color literals when doing these refactorings, somehow forgot these. Thanks for noticing!

PR #1681, opened 2020-07-24
Improve error message when the apps in a catalog failed loading

axbarsan:
This adds a new placeholder that is shown if loading apps from a specific catalog fails.

### Preview

image


axbarsan:
I updated the screenshot with the change 👍

pipo02mix:
thx looks good

PR #1693, opened 2020-07-29
Remove some superfluous tooltips.

oponder:
Reported by @marian here: https://gigantic.slack.com/archives/CQ465CFG8/p1596037745003500

### DIFF
--- a/src/components/Cluster/ClusterDetail/AppDetailsModal/InitialPane.js
+++ b/src/components/Cluster/ClusterDetail/AppDetailsModal/InitialPane.js
@@ -82,7 +82,7 @@ const InitialPane = (props) => {
           <div className='labelvaluepair--label'>RELEASE STATUS</div>
           <div className='labelvaluepair--value code'>
             {props.app.status?.release?.status ? (
-              <Truncated as='span'>{props.app.status.release.status}</Truncated>
+              <span>{props.app.status.release.status}</span>
             ) : (
               <span>Information pending...</span>
             )}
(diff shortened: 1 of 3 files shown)


src/components/Cluster/ClusterDetail/AppDetailsModal/InitialPane.js
```diff
@@ -82,7 +82,7 @@ const InitialPane = (props) => {
           <div className='labelvaluepair--label'>RELEASE STATUS</div>
           <div className='labelvaluepair--value code'>
             {props.app.status?.release?.status ? (
-              <Truncated as='span'>{props.app.status.release.status}</Truncated>
+              <span>{props.app.status.release.status}</span>
```
oponder, line 85: This one will never need to be truncated, so removed the component.

But the app version can be very long in some cases (versions of test apps) so for those we do need some logic in the Truncated component to skip tooltips if they are not needed. (Added below)

src/components/UI/__tests__/Truncated.ts
```diff
@@ -73,6 +73,16 @@ describe('Truncated', () => {
     expect(getByText('ATextThatIsTooLongToBeDisplayed')).toBeInTheDocument();
   });
 
+  it('doesnt display a tooltip for text that wasnt truncated', () => {
```
axbarsan, line 76:
```suggestion
  it(`doesn't display a tooltip for text that wasn't truncated`, () => {
```

src/components/UI/Truncated.tsx
```diff
@@ -34,14 +34,22 @@ const Truncated: React.FC<ITruncatedProps> = ({
   // Safe because we're using defaultProps in case these aren't defined
   str = truncate(str, replacer as string, numStart as number, numEnd as number);
 
+  // Skip the tooltip if the content was not truncated
+  let skipTooltip = false;
+  if (str === children) skipTooltip = true;
```
axbarsan, line 39:
```suggestion
  const skipTooltip = str === children;
```

oponder: 🏌️‍♂️⛳️

PR #1696, opened 2020-07-31
azure v5: hide CAPI cluster name label

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/12431

This PR:
* hides and blocks the creation of the cluster API cluster name label (`cluster.x-k8s.io/cluster-name`) from the label editor - this is used by Azure V5
* centers the `This cluster has no labels` placeholder vertically

### Preview

<details>
<summary>Before - no labels placeholder</summary>

image

</details>

<details>
<summary>Before - with the CAPI label</summary>

image

</details>

<details>
<summary>After</summary>

image

</details>

### DIFF
--- a/src/utils/clusterUtils.js
+++ b/src/utils/clusterUtils.js
@@ -1,6 +1,7 @@
 import { getMinHAMastersVersion } from 'selectors/featureSelectors';
 import cmp from 'semver-compare';
 import { Constants, Providers } from 'shared/constants';
+import { validateLabelKey } from 'utils/labelUtils';
 
 // Here we can store functions that don't return markup/UI and are used in more
 // than one component.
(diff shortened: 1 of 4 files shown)


src/utils/clusterUtils.js
```diff
@@ -221,9 +222,7 @@ export const filterLabels = (labels) => {
   const filteredLabels = {};
 
   for (const key of Object.keys(labels)) {
-    if (
-      key.includes(Constants.RESTRICTED_CLUSTER_LABEL_KEY_SUBSTRING) === false
-    ) {
+    if (validateLabelKey(key).isValid) {
```
axbarsan, line 225: Using the validator function here, as well, instead of duplicating the validation logic

src/utils/labelUtils.ts
```diff
@@ -152,5 +152,12 @@ export const validateLabelKey: IValidationFunction = (key) => {
     };
   }
 
+  if (strKey.toLowerCase() === 'cluster.x-k8s.io/cluster-name') {
```
oponder, line 155: Perhaps this deserves an explanatory comment or a link to this issue: https://github.com/giantswarm/giantswarm/issues/12431

For future code travellers that might come across it and wonder why.

oponder: That said, the commit message will bring us here as well, so not super needed. But it's a nice touch.

PR #1713, opened 2020-08-05
create install ingress actions & reenable tests

ubergesundheit:
Towards giantswarm/giantswarm#12360

I somewhat struggled with reenabling the tests again, so I would really appreciate your input on them and also on the added actions in the appcatalog store.


src/components/Cluster/ClusterDetail/Ingress/__tests__/InstallIngressButton.ts
```diff
@@ -68,16 +76,23 @@ describe.skip('InstallIngressButton', () => {
...
-    ).not.toBeInTheDocument();
-    expect(
-      screen.queryByText(/this will install the nginx ingress controller app/i)
-    ).not.toBeInTheDocument();
+    // expect(
+    //   screen.queryByText(/ingress controller installed/i)
+    // ).not.toBeInTheDocument();
+    // expect(
+    //   screen.queryByText(/this will install the nginx ingress controller app/i)
+    // ).not.toBeInTheDocument();
```
ubergesundheit, line 95: I fear this whole test is somewhat bogus now. But I may have made some mistakes in re-implementing the logic from the original `InstallIngressButton`

testUtils/ingressCatalogStateMocks.ts
```diff
@@ -0,0 +1,15 @@
+import { Constants } from 'shared';
+import { appCatalogsResponse } from 'testUtils/mockHttpCalls';
+
+const ingressCatalog = appCatalogsResponse.find(
+  (c) => c.metadata.name === Constants.INSTALL_INGRESS_TAB_APP_CATALOG_NAME
+);
+export const catalogsState = {
```
ubergesundheit, line 7: Needing to specify this state for the tests seem dirty to me, but I couldn't bring the tests to execute the neccessary catalog loading

src/stores/appcatalog/actions.ts
```diff
@@ -23,3 +29,66 @@ export const listCatalogs = createAsynchronousAction<any, any, any>({
...
+        chartName: name,
+        version,
+        catalog: Constants.INSTALL_INGRESS_TAB_APP_CATALOG_NAME,
+        namespace: 'kube-system',
+        valuesYAML: '',
+        secretsYAML: '',
+      };
+
+      /* eslint-disable @typescript-eslint/await-thenable */
+      await dispatch(
```
ubergesundheit, line 58: any idea what to do here? `await` is required because the following action will fire immediately thus making it somewhat wrong here.

oponder:
@oponder review this!

src/components/Cluster/ClusterDetail/Ingress/__tests__/InstallIngressButton.ts
axbarsan: I think it does what it's supposed to do, even with the lines uncommented. It checks that the flash messages and success message don't appear before the IC app was loaded. It also checks that the button is disabled and that it displays the correct loading animation.

testUtils/ingressCatalogStateMocks.ts
axbarsan: Don't think there's another way, as the component requires the catalog to be already in the store before loading the apps for it

axbarsan:
Code looks good in general, and it works as it's supposed to. I left some suggestions [here](https://github.com/giantswarm/happa/pull/1736):
* fixing the first render test
* improving the `AsynchronousAction` types
* a question about using `dispatch`

Hope you don't mind, it was easier for me to do it this way while testing the code

src/stores/appcatalog/actions.ts
```diff
@@ -23,3 +29,66 @@ export const listCatalogs = createAsynchronousAction<any, any, any>({
...
+        installApp({ app: appToInstall, clusterId: payload.clusterId })
+      );
+      await dispatch(loadClusterApps({ clusterId: payload.clusterId }));
+      /* eslint-enable @typescript-eslint/await-thenable */
```
axbarsan:
```suggestion
      await dispatch(
        installApp({ app: appToInstall, clusterId: payload.clusterId })
      );
      await dispatch(loadClusterApps({ clusterId: payload.clusterId }));
```

src/stores/appcatalog/actions.ts
```diff
@@ -23,3 +29,66 @@ export const listCatalogs = createAsynchronousAction<any, any, any>({
...
+  void
+>({
+  actionTypePrefix: 'PREPARE_INGRESS_TAB_DATA',
+  perform: async (state, dispatch, payload) => {
+    if (payload?.clusterId) {
+      const gsCatalog = selectIngressCatalog(state);
+
+      await Promise.all([
+        dispatch(loadClusterApps({ clusterId: payload.clusterId })),
+        catalogLoadIndex(gsCatalog)(dispatch, () => state),
```
axbarsan:
```suggestion
        dispatch(catalogLoadIndex(gsCatalog)),
```

ubergesundheit: It seems you commented on an outdated revision. The Github UI tells me `Outdated suggestions cannot be applied`

axbarsan: GitHub strikes again 😅

ubergesundheit:
I'm leaving this open in case @oponder wants to take another look

src/components/Cluster/ClusterDetail/Ingress/InstallIngressButton.tsx
```diff
@@ -158,15 +116,15 @@ const InstallIngressButton: React.FC<IInstallIngressButtonProps> = ({
...
         <Text>
-          {ingressApp ? (
+          {installedIngressApp ? (
             '🎉 Ingress controller installed. Please continue to the next step.'
           ) : (
             <>
               This will install the{' '}
               <StyledLink to={ingressAppDetailPath} href={ingressAppDetailPath}>
-                NGINX Ingress Controller app {ingressAppToInstall.version}
+                NGINX Ingress Controller app {ingressAppToInstall?.version}
```
oponder, line 125: Are there things that could go wrong? 

How does the component handle failures to fetch or get the version?

ubergesundheit:
Thank you for your reviews! I've added FlashMessages and Error notifications to the actions.

PR #1714, opened 2020-08-05
Make upgrade version configurable for admins

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/12499

This PR adds support for letting admins pick the cluster version they want to upgrade to.
Please note that some upgrades might be blocked by the backend (such as updating from non-node pools to node pools on AWS, or skipping too many releases).

It re-uses the release version picker from the cluster creation form.

### Preview

<details>
<summary>Inspect changes section, if the user is not an admin</summary>

image

</details>

<details>
<summary>Inspect changes section, if the user is an admin</summary>

image

</details>

<details>
<summary>Version change form</summary>

image

</details>


src/components/Cluster/ClusterDetail/ClusterDetail.js
```diff
@@ -1,24 +1,61 @@
...
+  const region = useSelector((state) => state.main.info.general.datacenter);
+  const isAdmin = useSelector(getUserIsAdmin);
+
+  const defaultTargetRelease = useSelector((state) => {
+    const targetReleaseVersion = selectTargetRelease(state, cluster);
+
+    return state.entities.releases.items[targetReleaseVersion] ?? null;
+  });
+  const [targetRelease, setTargetRelease] = useState(defaultTargetRelease);
+  const handleChangeTargetRelease = (newReleaseVersion) => {
```
axbarsan, line 45: Only needed to move the target release selector inside the component, but moved all selectors, for consistency

src/components/Cluster/ClusterDetail/UpgradeClusterModal.js
```diff
@@ -107,6 +126,7 @@ class UpgradeClusterModal extends React.Component {
                 <ReleaseComponentLabel
                   isAdded
                   name={component.name}
+                  key={component.name}
```
axbarsan, line 129: These missing keys were throwing errors in the console

src/components/Cluster/ClusterDetail/UpgradeClusterModalVersionChanger.tsx
```diff
@@ -0,0 +1,80 @@
...
+  ...rest
+}) => {
+  if (!isAdmin) {
+    onCancel();
+
+    return null;
+  }
+
+  return (
+    <div {...rest}>
```
axbarsan, line 34: Good practice would be to change the modal title and body, instead of changing the whole modal, as we do with the cluster scale modal, but would've been too big of a change for this PR. So I just re-used the initial logic, for now

src/components/Cluster/NewCluster/ReleaseSelector/ReleaseSelector.tsx
```diff
@@ -24,6 +24,8 @@ import ReleaseRow from './ReleaseRow';
 interface IReleaseSelector {
   selectRelease(releaseVersion: string): void;
   selectedRelease: string;
+  collapsible?: boolean;
+  autoSelectLatest?: boolean;
```
axbarsan, line 28: Added support for having the ReleaseSelector component non-collapsible (hide the drop-down caret icon and always have the component open), and for not selecting the latest version by default.

src/selectors/clusterSelectors.js
```diff
@@ -144,36 +145,44 @@ export const selectTargetRelease = (state, cluster) => {
...
 
-  const releases = state.entities.releases.items;
-  let availableVersions = Object.keys(releases).sort(cmp);
-
-  if (!state.main.loggedInUser.isAdmin) {
-    availableVersions = availableVersions.filter(
-      (release) => releases[release].active
-    );
-  }
+  const releases = Object.assign({}, state.entities.releases.items);
```
axbarsan, line 148: The `selectTargetRelease()` function logic has been reworked:
-> It will return the next active version for all user types.
-> If there is no active version and the user is not an admin, return `null`.
-> If there is no active version and the user is an admin, return the next `inactive` version.
-> If there is no next `inactive` version and the user is an admin, return `null`

src/components/Home/UpgradeNotice.tsx
```diff
@@ -44,13 +43,14 @@ interface IUpgradeNoticeProps {
...
-  isClusterUpgrading,
+const UpgradeNotice: React.FC<IUpgradeNoticeProps> = ({
   clusterId,
   className,
   onClick,
-}: IUpgradeNoticeProps) {
+}) => {
+  const canClusterUpgrade = useSelector(selectCanClusterUpgrade(clusterId));
+  const isClusterUpgrading = useSelector(selectIsClusterUpgrading(clusterId));
+
```
axbarsan, line 53: Changed for consistency

src/selectors/clusterSelectors.js
ubergesundheit: Before this, was it possoble for Giant Swarm staff to upgrade cluster through incompatible versions (9->11 on AWS)? 

For a `9.0.9` cluster this implementation would allow me to upgrade to `11.0.1` and above.

Maybe we add a little heads up to the modal text which says something along `Please don't upgrade from 9 to 11 on AWS)?

src/components/Cluster/ClusterDetail/ClusterDetail.js
```diff
@@ -31,24 +68,36 @@ const ClusterDetail = ({ match, cluster, clusterId, ...rest }) => {
...
 
         <Route
           path={`${match.path}`}
           render={() => (
             <ClusterDetailView
-              {...rest}
               match={match}
               cluster={cluster}
-              clusterId={clusterId}
+              isV5Cluster={isV5Cluster}
```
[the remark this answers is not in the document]
axbarsan: There's no benefit, really, since both `ClusterDetail` and `ClusterDetailView` are linked to the store. I didn't want to fiddle with the existing logic too much, but I can move all the data coming from the state to `ClusterDetailView`

src/selectors/clusterSelectors.js
oponder: The API should prevent it,... and I thought we had some logic in Happa that also avoid offering an upgrade past that boundary (for all users), but maybe I am remembering it wrong.

axbarsan: Upgrading from 9.x.x to any newer major is blocked on the API side, so not a big deal.

Please note that for customers the `Upgrade` button will not be shown at all if the newer version has a major larger than `9`.

Here's the logic for showing the `Upgrade` button
https://github.com/giantswarm/happa/blob/0bb388ee730f8a8813e7c6044a694a09d3be9fff/src/selectors/clusterSelectors.js#L186

At the end there's also this check, that doesn't allow upgrading from V4 to V5 on AWS
https://github.com/giantswarm/happa/blob/0bb388ee730f8a8813e7c6044a694a09d3be9fff/src/utils/clusterUtils.js#L10

src/components/Cluster/ClusterDetail/__tests__/UpgradeClusterModal.tsx
```diff
@@ -0,0 +1,261 @@
...
+    renderAndOpen({
+      cluster: cluster,
+    });
+
+    expect(
+      screen.getByText(/Before upgrading please acknowledge the following/i)
+    ).toBeInTheDocument();
+  });
+
+  it(`can't change the upgrade release version if the user is not an admin`, () => {
```
oponder, line 79: More specifically, Is this test something like:

`can't change the upgrade release version if the user is not an admin, and there are no active releases in the state`

axbarsan:
Note from SIG UX sync: we should hide release versions older or equal to the cluster's current release  version

src/components/Cluster/NewCluster/ReleaseSelector/__tests__/ReleaseSelector.ts
```diff
@@ -235,4 +235,16 @@ describe('ReleaseSelector', () => {
...
+
+  it('has the caret hidden if the selector is not collapsible', () => {
+    renderWithStore(
+      ReleaseSelector,
+      { ...defaultProps, collapsible: false },
+      { ...defaultStoreState }
+    );
+
+    expect(
+      screen.getByText(/available releases/i).previousElementSibling
```
oponder, line 247: This is a bit of a brittle / indirect way to check for no caret. 

Maybe we can give it a role and getByRole? https://testing-library.com/docs/dom-testing-library/api-queries#byrole

Not sure what role would be best, but I found 'role=presentation' possibly.


Giving the caret element a test-id and make sure we don't find it would also be acceptable

---- 

Presentation role: https://www.digitala11y.com/presentation-role/
Other roles: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques

src/components/Cluster/ClusterDetail/__tests__/UpgradeClusterModal.tsx
axbarsan: The fact that there are no active releases in the store doesn't make a difference. This just checks that the `Change version` button is not there if the user is not an admin

axbarsan:
Added a few things:
* Made the release selector a bit more accessible, by adding support for using the `Tab` key to navigate around, select versions and toggle the list collapsed state.
* The cluster upgrade modal now only displays the versions that you can actually upgrade to (no more showing old versions, or unsupported versions - such as V5 versions for V4 clusters)

PR #1726, opened 2020-08-11
Always redirect to the cluster overview page after changing organizations

axbarsan:
This PR prevents having a random page reload after selecting a new organization, which prevented redirecting to the cluster overview page.


axbarsan:
Going for the implementation in https://github.com/giantswarm/happa/pull/1728

PR #1730, opened 2020-08-12
Sass refactoring

oponder:

Just finding some low hanging fruit.

There are some nested styles here, against FMT. But I think we need some nuance in that rule. For things where it's kind of a self contained thing, that doesn't take children, I think it's easier to see the nested styles to get a quick picture of what the component is shaped like. Happy for your thoughts on that though.

Removes of 3 more sass files and refactors them into styled components.

- _users_table.sass
- _toggle_alternative.sass
- _text.sass


oponder:
I think the copy hover thing is working as intended, it's only meant to highlight the line it will copy, so the other lines go gray. That might not be super obvious though.

ubergesundheit:
> I think the copy hover thing is working as intended, it's only meant to highlight the line it will copy, so the other lines go gray. That might not be super obvious though.

Makes total sense!

PR #1734, opened 2020-08-13
add ErrorBoundary around Clusterlist to catch Errors there

ubergesundheit:
This PR adds an ErrorBoundary component which can be used to wrap Components and display a fallback.

Its first use is wrapping ClusterDashboardItems and displaying a fallback component like this:

image

### DIFF
--- a/src/components/Home/ClusterDashboardItem.js
+++ b/src/components/Home/ClusterDashboardItem.js
@@ -22,6 +22,7 @@ import { Dot, mq } from 'styles';
 import Button from 'UI/Button';
 import ClusterIDLabel from 'UI/ClusterIDLabel';
 import ErrorFallback from 'UI/ErrorFallback';
+import ErrorText from 'UI/ErrorText';
 import RefreshableLabel from 'UI/RefreshableLabel';
 import { isClusterCreating } from 'utils/clusterUtils';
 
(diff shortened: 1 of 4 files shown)


src/components/Home/ClusterDashboardItem.js
```diff
@@ -202,66 +203,96 @@ function ClusterDashboardItem({
...
-                cluster={cluster}
-                isClusterCreating={isCreating}
-              />
-            )}
-          </ClusterDetailsDiv>
-        </ErrorFallback>
-      </ContentWrapper>
+          <ErrorFallback error={nodePoolsLoadError}>
+            <ClusterDetailsDiv>
+              {/*<ErrorBoundary>*/}
```
oponder, line 264: Commented code can be removed :+1:

src/components/Home/ClusterDashboardItem.js
```diff
@@ -202,66 +203,96 @@ function ClusterDashboardItem({
...
+                  nodePools={nodePools}
+                  isClusterCreating={isCreating}
+                />
+              ) : (
+                <ClusterDashboardResourcesV4
+                  cluster={cluster}
+                  isClusterCreating={isCreating}
+                />
+              )}
+              {/*</ErrorBoundary>*/}
```
oponder, line 277: Here too

src/components/shared/ErrorBoundary.tsx
```diff
@@ -0,0 +1,51 @@
...
+  };
+
+  static getDerivedStateFromError() {
+    return { hasError: true };
+  }
+
+  constructor(props: IErrorBoundaryProps) {
+    super(props);
+    this.state = { hasError: false };
+  }
```
axbarsan, line 36:
```suggestion
  public readonly state: IErrorBoundaryState = {
    hasError: false,
  };
```

PR #1735, opened 2020-08-13
Implement happa universal search

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/10825

This PR implements a search bar at the top of happa (always visible), which will allow searching for different data types. Only clusters are currently supported.

You can search for a cluster using any part of the ID or Name (e.g. a cluster named `An amazing cluster` could be found using the search term `zing`).

The plan in the future, once more data types are supported, to also be able to find certain results by entering the filter name (e.g. find a cluster with the substring `zing` in its name by using the `cluster:zing` search term).

The number of results per query are currently limited to `5`.

### Keyboard interaction

* `Esc`
  * Close the dropdown with search results
* `Arrow Down`
  * Re-open the dropdown with search results, if it was previously closed with the `Esc` key
  * Navigate down through the search results
* `Arrow Up`
  * Navigate up through the search results
* `Enter`
  * Navigate to the search result's page (once you have a result selected with the mouse cursor, or arrow keys).

### Implementation details

* Search filters are designed to only return one data type (e.g. cluster filter will only return clusters).

* Search filters have 4 parts:
  * type (a unique identifier)
  * renderer (a React component factory that will render the result row in the results dropdown)
  * searcher (a generator that will return search results, given a query)
  * URL factory (a function that will return the URL that will be accessed when clicking on a search result in the results dropdown)

* Reasoning behind using generators for filters - it allows filters to have 100% control on stuff such as the order of the search results, and it makes possible having a unique limit for the number of search results from all filters, without unnecessary computation of additional results.

* Search result computation has a debounce rate of 250ms on each key press in the search bar inoput.

### Preview

<details>
<summary>Empty search bar</summary>

image

</details>

<details>
<summary>Search, and receive results</summary>

image

</details>

<details>
<summary>Search, but no results found</summary>

image

</details>

<details>
<summary>Hover on a result (or select with arrow keys)</summary>

image

</details>


src/@types/clusters.d.ts
```diff
@@ -0,0 +1,5 @@
+// FIXME: Create proper cluster type.
```
axbarsan, line 1: Not in the scope of this PR

src/components/UI/Inputs/Input.tsx
```diff
@@ -74,47 +75,65 @@ export interface IInput<T> {
   value?: T;
 }
 
-const Input: FC<IInput<string | FileList>> = (props) => {
-  const onChange = (e: ChangeEvent<ElementRef<'input'>>) => {
-    props.onChange?.(e.target.value);
-  };
+const Input: FC<IInput<string | FileList>> = React.forwardRef(
```
axbarsan, line 78: Extending this a bit, to allow ref forwarding and passing additional props

src/components/UniversalSearch/UniversalSearchProvider.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+interface IUniversalSearchContextValue {
+  searchTerm: string;
+  search(newTerm: string): void;
+  search<T>(newTerm: string, type: string): void;
+  searchResults: IUniversalSearcherResult<unknown>[];
+  filters: UniversalSearchFilterMap;
+}
+
+const SEARCH_ITEMS_LIMIT = 5;
```
axbarsan, line 18: Limit of 5 search results

src/components/UniversalSearch/UniversalSearchProvider.tsx
```diff
@@ -0,0 +1,85 @@
...
+  search: () => {},
+  searchResults: [],
+  filters: {},
+};
+
+const UniversalSearchContext = React.createContext<
+  IUniversalSearchContextValue
+>(initialContextValue);
+
+export function useUniversalSearch() {
```
axbarsan, line 32: Exposing this hook, so we can use the search in any component

src/components/UniversalSearch/UniversalSearch.tsx
```diff
@@ -0,0 +1,164 @@
...
+  padding: 12px;
+  background-color: ${({ theme }) => theme.colors.darkBlueDarker3};
+  position: relative;
+  line-height: initial;
+`;
+
+interface IUniversalSearchProps extends React.ComponentPropsWithoutRef<'div'> {}
+
+const UniversalSearch: React.FC<IUniversalSearchProps> = React.memo(
+  ({ ...rest }) => {
```
ubergesundheit, line 28: Double spread? Here and in 135.

src/components/UniversalSearch/UniversalSearch.tsx
```diff
@@ -0,0 +1,164 @@
...
+    const { searchTerm, search, searchResults, filters } = useUniversalSearch();
+    const debouncedSearchTerm: string = useDebounce(
+      searchTerm,
+      UPDATE_DEBOUNCE_DELAY_MS
+    );
+    const debouncedResults: IUniversalSearcherResult<unknown>[] = useDebounce(
```
ubergesundheit, line 34: Is is really neccessary to debounce both input and output? Shouldn't search input be enough?

src/components/UniversalSearch/UniversalSearchProvider.tsx
ubergesundheit: Could be a fat arrow function since every other function in this file is also one.

```suggestion
export const useUniversalSearch () => {
```

src/components/UniversalSearch/UniversalSearchProvider.tsx
```diff
@@ -0,0 +1,85 @@
...
+}
+
+const UniversalSearchProvider: React.FC<IUniversalSearchProviderProps> = ({
+  children,
+  controller,
+}) => {
+  const [searchTerm, setSearchTerm] = useState(initialContextValue.searchTerm);
+
+  const globalState = useSelector((state) => state);
+  const results = useRef(initialContextValue.searchResults);
```
ubergesundheit, line 48: Why doesn't this use `useState`?

src/components/UniversalSearch/UniversalSearch.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+    const handleResultClick = () => {
+      handleClear();
+      searchInputRef.current?.blur();
+    };
+
+    const handleKeyDown = (e: React.KeyboardEvent<HTMLInputElement>) => {
+      switch (e.key) {
+        case 'Escape':
+          setIsOpened(false);
```
ubergesundheit, line 81: IMO pressing the escape key should also clear the input.

src/lib/UniversalSearcher/filters/ClusterFilter.ts
```diff
@@ -0,0 +1,40 @@
...
+  const termLowerCased = term.toLowerCase();
+  const clusterList = Object.values<ICluster>(state.entities.clusters.items);
+
+  for (const cluster of clusterList) {
+    switch (true) {
+      case typeof cluster.delete_date !== 'undefined':
+        // Ignore, we don't need to show deleted clusters.
+        break;
+      case cluster.id.toLowerCase().includes(termLowerCased):
+      case cluster.name.toLowerCase().includes(termLowerCased):
```
ubergesundheit, line 17: mind blown

ubergesundheit (approved):
Code looks good and works for me.

I'm not a fan of displaying the search as the first thing on top.
My suggestion would be to display it below the breadcrumbs and only on pages related to clusters. It only searches clusters right now after all. Not a blocker for me though.

src/components/UniversalSearch/UniversalSearch.tsx
axbarsan: There were a few problems with my logic: Searching wasn't actually debounced, just displaying the results was debounced. I fixed it now.

Search input doesn't need to be debounced for the search input (otherwise the contents of the input won'
t be updated with the correct text).

Actual searching should be always debounced, to prevent high resource load.

However, there are some checks related to showing or hiding the results dropdown, which need to use a debounced value of the search term. If we don't, then while typing, the user will see `No results found for your query`.

src/components/UniversalSearch/UniversalSearch.tsx
axbarsan: Not sure about that. Every type of search input that I found that works similarly doesn't clear the input (look at GitHub or Google, for example). We might confuse some users by clearing the input, as it might not behave as expected

oponder (commented):
I love it! Just had feedback on the location as discussed earlier today. What Gerald mentioned is nice, also think the way github does it is pretty good.

marians:
My first concern here is the fact that we add a search function based on substring matching, while we already have search framework in use with `lunr` which solves searching in a powerful way, including tokenization (e. g. for cluster names, app descriptions), inverted index, and a flexible query syntax.

I have more concerns about the user experience and would like to collaborate on improving this before we merge.

- Positioning in the top middle
- Placement of the result overlay
- The `cluster` indicator is not needed while we have only clusters as a result
- The typography and spacing in the search result deserves some tweaking
- Highlighted row in the result needs no extra spacing. Could reach from border to border, use the entire overlay width.
- Double border when focused (white, plus -  in my case - blue). I'd like to get rid of one.
- Reaction time: I think 250ms delay is too much. Or does it feel slow for some other reason?
- Lonely positioning of the magnifier glass next to the input field. I'd like to enclose it somehow into a shape, like we do with the `ORG` label in the organization drop-down.
- I'd prefer something more specific than `I'm looking for...` as a prompt. As long as this only finds clusters, we could simply say `Look up a cluster` or `Find a cluster`.

axbarsan:
> My first concern here is the fact that we add a search function based on substring matching, while we already have search framework in use with `lunr` which solves searching in a powerful way, including tokenization (e. g. for cluster names, app descriptions), inverted index, and a flexible query syntax.

I agree that the current implementation does leave room for optimization. We do in fact use `lunr` in the app installation form, for finding the cluster to install the app on.

However, `lunr` is a storage solution itself. All objects have to be stored in it, in order for searching to work. Just placing it into our current implementation would mean that we would have duplicated data in the global state, and in the `lunr` index. As we have limited memory in the browser, this is a problem. It is also very hard to keep both data sources in sync.

A long-term solution would be to change our global state storage (redux store), so it would be search-efficient. This can be done by integrating `lunr` into our custom solution, or using some sort of client-side in-memory database solution. Please note that this is not a straight-forward, nor easy change to make.

Also, keep in mind that indexing might not have as big of an impact as expected, as we have fast access to data (all of it is stored in memory), and a small number of entries to search through. (we're talking dozens of entries, not hundreds, nor thousands).

If a faster search speed is desired (even if it is near instant, right now), there is a lot of room for improvement on the clusters search algorithm. We can gain a lot of speed by writing a more efficient algorithm.

As for the UX improvements, we can experiment with some A/B testing on Slack, until we find a better solution

axbarsan:
Closing for now... we're moving towards using the K8s API for everything, so this implementation no longer makes sense. We can maybe revisit this in 25 years.

PR #1747, opened 2020-08-19
Fix cannot set state of unmounted component warning in test.

oponder:
Fixes a "Can't perform a React state update on an unmounted component" warning that tries to be logged after the tests are done. 

That's because the modal gets unmounted as soon as we click the 'remove expiration' button, because the 'modal.template' state value gets unset, which unmounts the modal before actually getting the result of the asynchronous action later, which then closes the modal properly.


ubergesundheit (approved):
LGTMIIWFY

PR #1748, opened 2020-08-19
Add support for null availability zones on node pools

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/12860

### DIFF
--- a/src/components/Cluster/ClusterDetail/AddNodePool.js
+++ b/src/components/Cluster/ClusterDetail/AddNodePool.js
@@ -22,6 +22,12 @@ import NumberPicker from 'UI/NumberPicker';
 
 import AvailabilityZonesParser from './AvailabilityZonesParser';
 
+const AvailabilityZoneSelection = {
+  Automatic: 0,
+  Manual: 1,
+  None: 3,
+};
+
 // Availability Zones styles
 const Emphasized = css`
   .emphasized {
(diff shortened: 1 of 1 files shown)


ubergesundheit (commented):
Code looks good. Tested on ghost with local proxy.

Though, it would be very nice to explain what happens for selecting `None`. Just adding a explanation or maybe renaming the option would be good. Then after creation, my node pool displays `Availability Zone: 2` which adds to my confusion.

axbarsan:
> Though, it would be very nice to explain what happens for selecting `None`. Just adding a explanation or maybe renaming the option would be good.

Selecting `none` actually means setting the AZs to `null` on Azure. We use the `-1` value to do that, to not break defaulting or change the request data types.

> Then after creation, my node pool displays `Availability Zone: 2` which adds to my confusion.

This shouldn't happen. Did you test with latest API master? `ghost` was reserved yesterday, so it didn't have the API version that allows this change

ubergesundheit:
> This shouldn't happen. Did you test with latest API master? ghost was reserved yesterday, so it didn't have the API version that allows this change

Yes I tested on `ghost`. Tested again on `godsmack` with deployed api `1.7.0-f73f9fe5f08327ad3d2d366546276e418b5c8c1f` and a node pool with instance type `Standard_D4s_v3` and selected availabilty zone `None`.

Resulting node pool shows availabilty zone `3`.

PR #1749, opened 2020-08-20
customer bug: nginx IC is not visible under installed apps, if ICs are optional

axbarsan:
As seen on https://gigantic.slack.com/archives/C13MG9F0W/p1597918009003000

This PR makes the `nginx-ingress-controller` app visible under the `Installed Apps` section, if it's installed and if ingress controllers are optional in the cluster's release version.

### DIFF
--- a/src/components/Cluster/ClusterDetail/ClusterApps.js
+++ b/src/components/Cluster/ClusterDetail/ClusterApps.js
@@ -317,7 +317,7 @@ class ClusterApps extends React.Component {
 
       switch (true) {
         case hasOptionalIngress &&
-          app.metadata.name === 'nginx-ingress-controller':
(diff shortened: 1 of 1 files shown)


oponder:
hmm thats a bit odd. they used to show up.

are they now getting the managed by `cluster-operator` label?

axbarsan:
> are they now getting the managed by `cluster-operator` label?

Looks like it :(

image

oponder:
Ok I think I know what happened, this was an upgraded cluster. the nginx ic app _was_ installed by cluster-operator... so it go that label. Normally if it was installed through happa or manually it wouldn't get that label.

oponder (approved):
LGTM

omegas27 (approved):

PR #1751, opened 2020-08-20
[CP Auth 1/5]: Add environment configuration

axbarsan:
Part of the larger https://github.com/giantswarm/happa/pull/1547

Includes environment configuration, such as env variables, deployment configuration, local proxy settings etc.

### DIFF
--- a/scripts/webpack/proxyPlugin.js
+++ b/scripts/webpack/proxyPlugin.js
@@ -117,7 +117,13 @@ class ProxyPlugin {
 
       req
         .pipe(
-          request(cleanProxyUrl + req.url).on('response', (response) => {
+          request({
+            url: cleanProxyUrl + req.url,
+            rejectUnauthorized: false,
+            followAllRedirects: false,
+            followRedirect: false,
+            followOriginalHttpMethod: true,
+          }).on('response', (response) => {
             // In order to avoid https://github.com/expressjs/cors/issues/134
             const accessControlAllowOriginHeader =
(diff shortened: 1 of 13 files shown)


scripts/webpack/proxyPlugin.js
```diff
@@ -117,7 +117,13 @@ class ProxyPlugin {
 
       req
         .pipe(
-          request(cleanProxyUrl + req.url).on('response', (response) => {
+          request({
+            url: cleanProxyUrl + req.url,
+            rejectUnauthorized: false,
+            followAllRedirects: false,
+            followRedirect: false,
+            followOriginalHttpMethod: true,
```
axbarsan, line 125: Necessary to support login redirects after OIDC login. Doesn't break anything else (so far)

ubergesundheit (approved):
Excellent

webpack.common.js
```diff
@@ -48,30 +48,57 @@ const determineAudienceURL = () => {
   return apiAudienceUrl;
 };
 
+function makeCPApiEndpointURL(currentValue, apiEndpoint) {
```
oponder, line 51: What are these functions for? I don't really follow what they're doing.

Like why do we need to have this logic?

axbarsan: This helps derive the CP API URL from the regular API endpoint URL, if the API endpoint set is pointing to an installation, so you wouldn't have to set the CP API URL manually

PR #1752, opened 2020-08-20
[CP Auth 2/5]: Add OIDC login helpers

axbarsan:
Part of the larger #1547

This includes a wrapper around the `oidc-client-js` library, so we could replace the internal library at some point, and not have to cry about it.

### DIFF
--- a/src/lib/OAuth2/OAuth2.ts
+++ b/src/lib/OAuth2/OAuth2.ts
@@ -1,0 +1,209 @@
+import {
+  convertToOIDCMetadata,
+  IOAuth2CustomMetadata,
+} from 'lib/OAuth2/OAuth2CustomMetadata';
+import OAuth2UserImpl from 'lib/OAuth2/OAuth2User';
+import {
+  User,
+  UserManager,
+  UserManagerSettings,
+  WebStorageStateStore,
+} from 'oidc-client';
+
+export enum OAuth2Events {
+  UserLoaded = 'userLoaded',
+  TokenExpired = 'tokenExpired',
+  TokenExpiring = 'tokenExpiring',
+  UserUnloaded = 'userUnloaded',
+  UserSignedOut = 'userSignedOut',
+  SilentRenewError = 'silentRenewError',
+}
+
+interface IOAuth2EventCallbacks {
+  [OAuth2Events.UserLoaded]: (event: CustomEvent<OAuth2UserImpl>) => void;
+  [OAuth2Events.TokenExpired]: () => void;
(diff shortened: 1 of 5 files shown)


src/lib/OAuth2/OAuth2User.ts
```diff
@@ -0,0 +1,71 @@
+import { User } from 'oidc-client';
+import { AuthorizationTypes } from 'shared/constants';
+import { PropertiesOf } from 'shared/types';
+
+export interface IOAuth2User {
```
axbarsan, line 5: This is a wrapper around the User class from the oidc-client module.
This holds useful information as tokens, groups, and expiration date.

src/lib/OAuth2/OAuth2User.ts
```diff
@@ -0,0 +1,71 @@
...
+  constructor(fromObj?: Partial<IOAuth2User>) {
+    if (!fromObj) return;
+
+    for (const [key, value] of Object.entries(fromObj)) {
+      // @ts-ignore
+      this[key] = value;
+    }
+  }
+
+  public isExpired(): boolean {
```
axbarsan, line 49: This calculates if the user is expired or not, based on the current date and the stored expiration date.

src/lib/OAuth2/OAuth2User.ts
```diff
@@ -0,0 +1,71 @@
...
+    const sInMs = 1000;
+    const now = Math.trunc(Date.now() / sInMs); // In seconds.
+    const expiresIn = this.expiresAt - now;
+
+    return expiresIn <= 0;
+  }
+
+  public serialize(): IOAuth2User {
```
axbarsan, line 57: This converts the object into an easily storable value (for the Redux store, for example)

ubergesundheit (approved):
Looks clean and shiny :stars:

src/lib/OAuth2/OAuth2.ts
```diff
@@ -0,0 +1,209 @@
...
+  public async logout(): Promise<void> {
+    await this.userManager.removeUser();
+    await this.userManager.clearStaleState();
+  }
+
+  public addEventListener<
+    T extends OAuth2Events,
+    U extends IOAuth2EventCallbacks[T]
+  >(event: T, cb: U) {
+    this.eventEmitter.addEventListener(event, cb as EventListener, false);
```
axbarsan, line 143: I forgot to mention that using this approach will auto-suggest the right callback type for each type of event, based on the `IOAuth2EventCallbacks` interface.

PR #1753, opened 2020-08-20
[CP Auth 3/5]: add CP client and catalog service

axbarsan:
Part of the larger #1547

This PR adds the CP client library, client and handlers.
It also links the newly created types throughout the app.

### DIFF
--- a/src/model/clients/FetchAdapter.ts
+++ b/src/model/clients/FetchAdapter.ts
@@ -1,0 +1,79 @@
+import { GenericResponse } from 'model/clients/GenericResponse';
+import { HttpClient } from 'model/clients/HttpClient';
+
+/**
+ * This is an adapter for using an HttpClient with the same
+ * interface as the global `fetch` function.
+ */
+class FetchAdapter<T extends HttpClient = HttpClient> {
+  public readonly baseClient: T;
+
+  constructor(baseClient: T) {
+    this.baseClient = baseClient;
+  }
+
+  private configureFromRequestInfo(reqInfo: RequestInfo) {
+    if (typeof reqInfo === 'string') {
+      this.baseClient.setURL(reqInfo);
+
+      return;
+    }
+
+    // Copy headers.
+    for (const [key, value] of reqInfo.headers.entries()) {
+      this.baseClient.setHeader(key, value);
+    }
+
+    // Copy request method.
+    this.baseClient.setRequestMethod(reqInfo.method);
+
+    // Copy the request body.
+    this.baseClient.setBody(JSON.stringify(reqInfo.body) as never);
+  }
+
+  private configureFromRequestInit(reqInit?: RequestInit) {
+    if (!reqInit) return;
+
+    const { headers, method, body } = reqInit;
+
+    if (headers) {
+      // Copy headers.
+      const newHeaders = new Headers(headers);
+      for (const [key, value] of newHeaders.entries()) {
+        this.baseClient.setHeader(key, value);
+      }
+    }
+
+    if (method) {
+      // Copy request method.
+      this.baseClient.setRequestMethod(method);
+    }
+
+    if (body) {
+      // Copy the request body.
+      this.baseClient.setBody(JSON.stringify(body) as never);
+    }
+  }
+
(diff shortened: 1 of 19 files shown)


src/model/clients/FetchAdapter.ts
```diff
@@ -0,0 +1,79 @@
+import { GenericResponse } from 'model/clients/GenericResponse';
+import { HttpClient } from 'model/clients/HttpClient';
+
+/**
+ * This is an adapter for using an HttpClient with the same
+ * interface as the global `fetch` function.
+ */
+class FetchAdapter<T extends HttpClient = HttpClient> {
```
axbarsan, line 8: This makes us able to use our classes based on the `HttpClient` base class with the CP Client, as a replacement for the global `fetch` method.

src/model/clients/GenericResponse.js
```diff
@@ -103,4 +103,26 @@ export class GenericResponse {
   get headers() {
     return Object.assign({}, this.config.headers);
   }
+
+  /**
+   * Convert the object to a `Response` object returned by the fetch command.
+   * @returns {Response}
+   */
+  convertToFetchResponse() {
```
axbarsan, line 111: This converts our `GenericResponse` to a `Response` object, which is used by the global `fetch` method.

src/model/services/controlplane/appcatalogs/appcatalogs.ts
```diff
@@ -0,0 +1,57 @@
...
+  const catalogs = await appsAPI.listApplicationGiantswarmIoV1alpha1AppCatalog(
+    {}
+  );
+  const result = catalogs.items.map(convertAppCatalogCRToCatalog);
+
+  return result;
+}
+
+function convertAppCatalogCRToCatalog(
+  catalog: import('giantswarm-cp-client').ComGithubGiantswarmApiextensionsPkgApisApplicationV1alpha1AppCatalog
```
axbarsan, line 32: 😬

src/model/services/controlplane/appcatalogs/appcatalogs.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  const catalogs = await appsAPI.listApplicationGiantswarmIoV1alpha1AppCatalog(
+    {}
+  );
+  const result = catalogs.items.map(convertAppCatalogCRToCatalog);
+
+  return result;
+}
+
+function convertAppCatalogCRToCatalog(
```
axbarsan, line 31: This maps the App catalog CR type to the type that we use internally, so we can have control over what we store in memory.

src/model/services/controlplane/appcatalogs/types.ts
```diff
@@ -0,0 +1,21 @@
...
+  title: string;
+  description: string;
+  logoURL: string;
+  storage: IStorage;
+}
+
+export interface IStorage {
+  type: string;
+  URL: string;
+}
```
axbarsan, line 21: Custom App catalog types, same as the ones on the current GS API.

src/model/services/controlplane/appcatalogs/appcatalogs.ts
```diff
@@ -0,0 +1,57 @@
...
+import { getBaseConfiguration } from 'model/services/controlplane/base';
+
+/**
+ * Get the all the app catalogs in the installation.
+ * @param client - The HTTP client.
+ */
+export async function getAppCatalogs(
+  client: HttpClient
+): Promise<IAppCatalog[]> {
+  const { ApplicationGiantswarmIoV1alpha1Api } = await import(
```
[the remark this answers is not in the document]
oponder: Does that mean we're doing code splitting? Is that just automatically enabled?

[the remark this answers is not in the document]
axbarsan: Good call, I was under the impression that we had it enabled. I enabled it now 👍

oponder (approved):
LGTM! it kind feels like a lot... but I think cool that we're pushing the CP API access in Happa story further.

PR #1767, opened 2020-08-24
[CP Auth 4/5]: add the CP Auth object and redux logic

axbarsan:
Part of the larger #1547

This PR adds the dex configuration, and CP Authentication redux logic


src/stores/cpauth/middleware.ts
```diff
@@ -0,0 +1,48 @@
+import CPAuth from 'lib/CPAuth/CPAuth';
+import OAuth2UserImpl from 'lib/OAuth2/OAuth2User';
+import { AnyAction, Middleware } from 'redux';
+import { loadUserError, userExpired } from 'stores/cpauth/actions';
+import { CPAUTH_USER_EXPIRED, CPAUTH_USER_LOAD } from 'stores/cpauth/constants';
+import { getCPAuthUser } from 'stores/cpauth/selectors';
+
+export function cpAuthMiddleware(cpAuth: CPAuth): Middleware {
```
axbarsan, line 8: This middleware is used to renew user information, if a user is expired.

In general, the user will be renewed automatically, but the flow will fail if there's a race condition between renewal and a separate request. This prevents that.
There is also the case where the token expired when the user was not logged in, so it couldn't be renewed.

This can also prevent actions from getting dispatched, if the user expired, if we want to use this auth method as a primary one.

src/lib/CPAuth/CPAuth.ts
```diff
@@ -0,0 +1,16 @@
+import { defaultConfig } from 'lib/CPAuth/config';
+import OAuth2 from 'lib/OAuth2/OAuth2';
+
+class CPAuth extends OAuth2 {
```
axbarsan, line 4: This inherits from the OAuth2 class, so we could have this as a singleton, but not limit the number of OAuth2 providers to a single one.

src/shared/constants/routes.js
```diff
@@ -7,6 +7,8 @@ const AppRoutes = {
   SetPassword: '/forgot_password/:token',
   SignUp: '/signup/:token',
   OAuthCallback: '/oauth/callback',
+  CPAccess: '/cp-access',
+  CPAccessCallback: '/cp-access/callback',
```
axbarsan, line 11: This is the path that the user is redirected to, after logging in

src/lib/CPAuth/config.ts
```diff
@@ -0,0 +1,107 @@
...
+let issuer = authority;
+if (issuer.includes('localhost')) {
+  issuer = window.config.audience.replace('api', 'dex');
+}
+
+export const defaultConfig: IOAuth2Config = {
+  authority,
+  clientId: 'fFlz7lckhWA0kIaW3fLIl8chFSs2wvW6',
+  clientSecret:
+    'PoioOqWKUndxVnbcRzlv59EgvwPVJQIdIlved143Uko0SjGJ7OprnecZQbab3WhH',
```
oponder, line 22: I think we really gotta figure out a way to do this without a client secret...

We kinda need dex to support PKCE: https://github.com/dexidp/dex/pull/1652

That's the flow we're using with auth0 I think

axbarsan: Great find with that PR!
But can we move forward with this PR until we get that sorted?

axbarsan: Btw, I don't think this is a huge problem right now, as users still need to have access to the control plane they try to login on.

Take for example a customer installation. Try authenticating with the kubectl plugin. Dex will allow the authentication request, since the client ID and Secret are correct, but you won't actually be able to log in, since you're not registered in the customer's authentication configuration (e.g. active directory).

oponder (approved):
LGTM but not feeling 100% about the client secret, lets keep an eye on dex and improve that when the opportunity presents itself.

HEllRZA:
Presenting Opportunity ;o)
https://github.com/dexidp/dex/pull/1784

PR #1768, opened 2020-08-24
Fix app name validation in the app installation modal

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/12889

Moved the validation logic into a separate function and wrote tests for that.


src/components/AppCatalog/AppDetail/InstallAppModalUtils.ts
```diff
@@ -0,0 +1,42 @@
...
+const validateCharacters = /^[-.a-z0-9]*$/;
+
+interface IAppValidationResult {
+  valid: boolean;
+  message: string;
+}
+
+export function validateAppName(version: string): IAppValidationResult {
+  switch (true) {
+    case version.length < 1:
```
axbarsan, line 12: Also added a specific error message for an empty name

src/components/AppCatalog/AppDetail/InstallAppModalUtils.ts
```diff
@@ -0,0 +1,42 @@
+const maxLength = 253;
```
axbarsan, line 1: This was `63` before, even though the error message stated `253`

ubergesundheit:
Do you know how this name is being used? Is it something like described in https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/ ?

If so [code from label validation](https://github.com/giantswarm/happa/blob/master/src/utils/labelUtils.ts) could be reused I think

src/components/AppCatalog/AppDetail/InstallAppModalUtils.ts
oponder: Im not sure which is right though... the error message or the validation. I think the 63 has a purpose, because the name of the app get added to something else which then might get longer than 253

oponder (changes requested):
I think it might be best to go with the 63 character limit and change the error message.

I think somewhere along the way this name get used in a resource that has the 'DNS Label Name' requirement: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names

axbarsan:
> Do you know how this name is being used? Is it something like described in [kubernetes.io/docs/concepts/overview/working-with-objects/labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/) ?
> 
> If so [code from label validation](https://github.com/giantswarm/happa/blob/master/src/utils/labelUtils.ts?rgh-link-date=2020-08-24T12%3A22%3A29Z) could be reused I think

I'm not sure where it's used, but let's not overcomplicate it for now

oponder (approved):
LGTM! Thx!

PR #1770, opened 2020-08-24
proxy external app catalog through happa nginx

ubergesundheit:
Resolves giantswarm/giantswarm#12348

This PR replaces the usage an external service for loading app catalogs without cors headers by proxying them through the happa nginx.

@giantswarm/sig-security could you take a look at this and see if this introduces any security issues?


oponder:
Good idea!

giphy

ubergesundheit:
Unfortunately this does not work. Nginx needs something to resolve the proxied url but it seems inside the pod `127.0.0.1` does not seem to resolve names. Using an external resolver seems to be frowned upon because there is a vulnerability. So I have to come up with something else.

oponder:
ah, dang...  Nice that you're thinking of a way to improve that though. 👍

Dockerfile
```diff
@@ -16,7 +16,8 @@ RUN find /www \
   -exec gzip -9 -k '{}' \;
 
 RUN chown -R nginx:nginx /www
-RUN chmod -R u=rwx /www
+RUN chmod u=rwx /www
```
ubergesundheit, line 19: Changed this to apply `rwx` permissions only to the directory, not the files in it. IMO there is no point in an executable `index.html`

src/actions/catalogActions.js
```diff
@@ -5,37 +5,27 @@ import { StatusCodes } from 'shared/constants';
...
-      throw error;
-    });
+  if (indexURL.includes('kubernetes-charts.storage.googleapis.com')) {
+    indexURL = `/catalogs?url=${indexURL}`;
+  }
+
+  const response = await fetch(indexURL, { mode: 'cors' });
+
+  if (response.status !== StatusCodes.Ok) {
+    throw new Error(`Could not fetch index.yaml at ${indexURL}`);
```
stone-z, line 18: Maybe it would be good to preserve the original `url` for debugging, either instead of or in addition to this

scripts/start.sh
```diff
@@ -58,6 +58,10 @@ gzip -f -9 -k /www/index.html
 # gzip metadata.json again because we changed it
 gzip -f -9 -k /www/metadata.json
 
+# Create file to include into nginx config to use
+# the pods nameserver as resolver
+echo resolver $(awk 'BEGIN{ORS=" "} $1=="nameserver" {print $2}' /etc/resolv.conf) ";" > /etc/nginx/config/resolvers.conf
```
stone-z, line 63: Can you elaborate for my puny mind what this is actually doing? Where are the parameters coming from?

webpack.dev.js
```diff
@@ -28,6 +28,20 @@ module.exports = merge(common, {
     port: 7000,
     host: 'localhost',
     historyApiFallback: true,
+    // used for app catalogs proxy /catalogs?url=
+    proxy: {
+      '/catalogs': {
+        target: 'https://cors-anywhere.herokuapp.com/',
```
stone-z, line 34: I guess this would go away?

src/actions/catalogActions.js
ubergesundheit: You're right! I've added this

webpack.dev.js
ubergesundheit: This is only being used in development.

scripts/start.sh
ubergesundheit: Rewritten it easier

src/actions/catalogActions.js
```diff
@@ -5,37 +5,29 @@ import { StatusCodes } from 'shared/constants';
...
-      });
+async function loadCatalogIndex(catalog) {
+  let indexURL = `${catalog.spec.storage.URL}index.yaml`;
 
-      return catalogWithApps;
-    })
-    .catch((error) => {
-      throw error;
-    });
+  if (indexURL.includes('kubernetes-charts.storage.googleapis.com')) {
```
stone-z, line 11: I know it's a longshot but could this be tightened to enforce the root domain? E.g. to prevent things like `kubernetes-charts.storage.googleapis.com.my.evil.site`

scripts/start.sh
```diff
@@ -58,6 +58,10 @@ gzip -f -9 -k /www/index.html
 # gzip metadata.json again because we changed it
 gzip -f -9 -k /www/metadata.json
 
+# Create file to include into nginx config to use
+# the pods nameserver as resolver
+echo resolver $(awk '/^nameserver/{print $2}' /etc/resolv.conf) ";" > /etc/nginx/config/resolvers.conf
```
stone-z, line 63: Just to double check -- it's still valid nginx syntax if there are multiple entries here?

stone-z (approved):
Two sanity checks but otherwise LGTM now

oponder (approved):
LGTMIIWFY

src/actions/catalogActions.js
ubergesundheit: Absolutely! I'll change this to `indexURL === "https://kubernetes-charts.storage.googleapis.com/index.yaml"`

scripts/start.sh
ubergesundheit: Didn't think of that. I'll experiment and modify accordingly if neccessary.

ubergesundheit: Turns out this was working as is, I've disabled ipv6 resolving because I am unsure if our clusters support ipv6 yet.

PR #1771, opened 2020-08-24
Remove NP appear transitions and make enter and exit transitions quicker

axbarsan:
As seen on https://github.com/orgs/giantswarm/projects/85#card-40662167

This PR:
* disables the NP `appear` transition (which is executed when the component is rendered).
* makes the `enter` transition quicker (when a node pool is created)
* adds an `exit` transition (when a node pool is deleted)
* fixes a bug that would make the `exit` transition display a loading spinner fading out, instead of the actual node pool

GIFs don't do this justice. Please take a look 👍

### DIFF
--- a/src/components/Cluster/ClusterDetail/NodePool.tsx
+++ b/src/components/Cluster/ClusterDetail/NodePool.tsx
@@ -270,16 +270,6 @@ class NodePool extends Component<INodePoolsProps, INodePoolsState> {
   }
 }
 
-function mapStateToProps(
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  state: Record<string, any>,
-  ownProps: INodePoolsProps
-) {
-  return {
-    nodePool: state.entities.nodePools.items[ownProps.nodePool.id],
-  };
-}
-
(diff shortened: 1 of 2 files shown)


src/components/Cluster/ClusterDetail/NodePool.tsx
```diff
@@ -270,16 +270,6 @@ class NodePool extends Component<INodePoolsProps, INodePoolsState> {
   }
 }
 
-function mapStateToProps(
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  state: Record<string, any>,
-  ownProps: INodePoolsProps
-) {
-  return {
-    nodePool: state.entities.nodePools.items[ownProps.nodePool.id],
```
oponder, line 279: we were passing in an actual nodePool object and no reason to fetch it from state again?

oponder (approved):
LGTM, thanks!

src/components/Cluster/ClusterDetail/NodePool.tsx
axbarsan: Yeah... not sure what the initial logic behind that was

PR #1774, opened 2020-08-27
Rewrite HTTP clients in TypeScript

axbarsan:
Seems bigger than it is, due to updating imports.

This PR:
* Rewrites HTTP clients and services in TypeScript.
* Removes `index.ts` files from HTTP clients and services, to be in line with fmt.
* Adds punctuation to comments.
* Removes the `LocalStorage` class that I wanted to be used as a mocking solution for browser local storage, because it wasn't used and also, using the existing global `Storage` interface makes more sense.


src/model/clients/HttpClient.ts
```diff
@@ -0,0 +1,276 @@
...
+export interface IHttpClientConfig {
+  timeout: number;
+  headers: Record<string, string>;
+  url: string;
+  method: HttpRequestMethods;
+  data?: Record<string, unknown>;
+  baseURL?: string;
+}
+
+export interface IHttpClient {
```
axbarsan, line 22: Created an interface, so we could mock http clients in tests using plain objects

ubergesundheit (approved):
Looks really massive but everything looks reasonable. Tested on my machine. Works for me

oponder (approved):
Not the best person to review this, but I trust you if you think this is moving the code base in the right direction!

PR #1775, opened 2020-08-27
split container build & update nginx image version

ubergesundheit:
This PR adds a compress stage and updates the nginx image for container builds


oponder (approved):
LGTMIIWFY!

axbarsan:
Does this solve this issue? 
https://github.com/giantswarm/giantswarm/issues/12893

ubergesundheit:
I think the issue is related to the ingress controller maybe? No idea if the ingress controller preserves the downstream server header

PR #1776, opened 2020-08-27
Rewrite React Hooks in TypeScript

axbarsan:

Again, seems big due to import changes.

This PR:
* Refactors React Hooks to TypeScript.
* Updates `prettier` to `2.1.1` to support labeled tuples in TypeScript 4.0
* Rename the `effects` folder to `hooks`


src/lib/hooks/useCopyToClipboard.ts
```diff
@@ -0,0 +1,24 @@
+import copy from 'copy-to-clipboard';
+import { useEffect, useState } from 'react';
+
+/**
+ * Hook for copying text to the user's clipboard.
+ *
+ * Note: Set value to `null` for resetting the clipboard state.
+ */
+function useCopyToClipboard(): [
+  hasContentInClipboard: boolean,
```
axbarsan, line 10: Labeled tuple types in TypeScript 4.0

ubergesundheit (approved):
I like the renaming from effects to hooks :+1: 

Maybe you could also move all hooks into the same place?

Here are some more: https://github.com/giantswarm/happa/tree/master/src/hooks ?

axbarsan:
> Maybe you could also move all hooks into the same place?
> 
> Here are some more: [`master`/src/hooks](https://github.com/giantswarm/happa/tree/master/src/hooks) ?

Yes! Good catch, I'll do that in this PR

ubergesundheit (approved):
:+1:

PR #1778, opened 2020-08-28
Rewrite error reporting logic in TypeScript

axbarsan:

Rewrite `ErrorReporter` and the `CustomError` in TypeScript.


oponder (approved):
:+1:

PR #1779, opened 2020-08-28
Convert constants to TypeScript

axbarsan:

Basically renaming `.js` to `.ts`, and removing the `index.js`.


oponder (approved):
Looking good! thanks!

PR #1780, opened 2020-08-28
Rewrite all src/lib helpers in TypeScript

axbarsan:

This PR:
- Rewrites the rest of the `lib/*.js` files in TypeScript.
- Makes `GenericResponse` and `HttpClient` actually generic, so we could easily use response types.

The `Passage` client has also been adapted to TypeScript. However, I would like to refactor it to fit the rest of our model layer clients (in a separate PR).

Most of the changes are covered by automated tests, but everything was also tested manually.


yarn.lock
```diff
@@ -6081,7 +6086,7 @@ getpass@^0.1.1:
 
 "giantswarm@git+https://github.com/giantswarm/giantswarm-js-client.git":
   version "4.0.1"
-  resolved "git+https://github.com/giantswarm/giantswarm-js-client.git#bf2810afa140487c44a9ef392c65eb211bd416e6"
+  resolved "git+https://github.com/giantswarm/giantswarm-js-client.git#6828532f9de3dae524e2c7f484f1927de5b5c05f"
```
axbarsan, line 6089: Updated to latest commit since I changed some types in the module

package.json
```diff
@@ -133,7 +133,6 @@
     "serialize-javascript": "4.0.0",
     "stacktrace-js": "^2.0.2",
     "superagent": "6.1.0",
-    "superagent-bluebird-promise": "4.2.0",
```
axbarsan, line 136: We no longer need this

oponder (approved):
Scanned it quickly, looks good to me. Thanks for bringing these things in line.

PR #1794, opened 2020-09-02
Supply ClusterDetailView with all releases

ubergesundheit:
This PR lets the ClusterDetailView access all releases instead of just the active ones.

Users with lowered permissions couldn't:

- List preinstalled apps on clusters with inactive releases
- Open the releases information modal on clusters with inactive releases
- Show the kubernetes version in the Giant Swarm release on clusters with inactive releases


axbarsan (approved):
🌮

PR #1795, opened 2020-09-02
Rewrite utils in TypeScript

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/13001

This PR:
* Adds types for V4/V5 Clusters
* Rewrites `utils/*.js` in TypeScript
* Removes all the cluster type replacements and uses the new real types, instead

As usual, tested manually, and also partly covered by tests


src/@types/clusters.d.ts
```diff
@@ -0,0 +1,35 @@
...
+  owner: string;
+  name: string;
+  release_version: string;
+
+  credential_id?: string;
+  delete_date?: string | null;
+
+  // Injected by client-side.
+  capabilities?: IClusterCapabilities;
+  keyPairs?: import('shared/types').IKeyPair[];
```
axbarsan, line 19: A bit weird, but if we would use regular ES module imports, this whole file is treated like an ES module

src/utils/clusterUtils.ts
```diff
@@ -0,0 +1,417 @@
...
+ * Get a cluster's latest condition.
+ * @param cluster
+ */
+export function getClusterLatestCondition(
+  cluster: V4.ICluster | V5.ICluster
+): string | undefined {
+  if (
+    'conditions' in cluster &&
+    cluster.conditions &&
+    cluster.conditions.length > 0
```
axbarsan, line 309: Not as pretty as the way we used optional chaining before, but that way confuses the compiler too much, and it gives a false error.

src/utils/clusterUtils.ts
```diff
@@ -0,0 +1,417 @@
...
+      }
+
+      return JSON.parse(window.config.azureCapabilitiesJSON);
+
+    default:
+      return null;
+  }
+}
+
+export function v4orV5(
```
axbarsan, line 403: Moved from `v4orV5.js`, since it's also related to clusters.

ubergesundheit (approved):
Tested on my machine where everything works. Changes look good.

PR #1797, opened 2020-09-07
Write unit tests for 'lib/*' helpers

axbarsan:
This PR includes:
* Tests for almost all the `lib/*` helpers, except `passageClient` which I want to rewrite in the same way as other model layer functions. It is currently covered almost 100% by integration tests, so no rush there.
* Fixes for a few of the tiny bugs that were found.


ubergesundheit (approved):
:100:

PR #1820, opened 2020-09-14
Add real user monitoring for test installations

marians:
We want to try real user monitoring. See https://github.com/giantswarm/giantswarm/issues/12047

This PR is supposed to add

- Google Analytics
- Datadog

but only for test installations. The plan is to later enable this on a per-installation basis, when customers have given their consent.


src/index.html
```diff
@@ -21,6 +21,7 @@
       href="//fonts.googleapis.com/css?family=Roboto:300,400,700,300italic|Inconsolata:400,700|Roboto+Condensed:400,700"
       type="text/css"
     />
+    <!-- PLACEHOLDER_RUM -->
```
ubergesundheit, line 24: analytic scripts are usually placed at the bottom right before the closing `</body>` tag to prevent blocking the rest of the page

marians: Google explicitly asks users to put it as the first element in the `<head>`. And I can imagine that it makes sens eto do so in order to be able to track events like first contentful paint etc.

ubergesundheit: So we should blame google making our internet slow! :sweat_smile:

scripts/start.sh
```diff
@@ -50,9 +50,28 @@ fi
 # VERSION file.
 VERSION=$(cat VERSION | tr '\n' ' ' | tr -d '[:space:]')
 
+# Set version in config object
 sed -i "s|happaVersion: .*|happaVersion: '${VERSION}',|g" /www/index.html
+# Set version in metadata.json
 sed -i "s|\"version\": .*|\"version\": \"$VERSION\"|" /www/metadata.json
 
+# Add real user monitoring (RUM) scripts to testing installations only
+if [ -n "$ENABLE_RUM"]; then
```
ubergesundheit, line 59:
```suggestion
if [ -n "$ENABLE_RUM" ]; then
```

marians:
With current cookie settings:

image

marians:
With `useSecureSessionCookie: true` the above issue warning disappeared (in Chrome 85).

src/index.html
```diff
@@ -16,6 +16,7 @@
       sizes="128x128"
       href="/images/app_icon_128.png"
     />
+    %%PLACEHOLDER_RUM%%
```
marians, line 19: Currently this string shows up as readable content during development, but it's no longer visible as soon as the actuall app content appears. Any idea how to prevent this? Shall I make it a fake HTML tag?

ubergesundheit: It seems https://en.wikipedia.org/wiki/Conditional_comment are spared from removal. So maybe try `<!--[if true]>PLACEHOLDER_RUM<![endif]-->` or something along the line

marians: I settled for `<placeholder_rum/>`. Uppercase didn't work, because it would be transformed to lowercase during `yarn dist`.

helm/happa/templates/configmap.yaml
```diff
@@ -12,6 +12,13 @@ data:
...
+  installation-name: {{ .Values.Installation.V1.Name }}
   aws-capabilities-json: '{{ .Values.Installation.V1.Provider.AWS.EC2.Instance.Capabilities | toJson }}'
   azure-capabilities-json: '{{ .Values.Installation.V1.Provider.Azure.VM.VmSize.Capabilities | toJson }}'
   default-request-timeout-seconds: '{{ .Values.Installation.V1.GiantSwarm.Happa.DefaultRequestTimeoutSeconds }}'
+  provider: {{ .Values.Installation.V1.Provider.Kind }}
+
+  {{ if .Values.Installation.V1.Infra.TestingEnvironment }}
+  # Enables real user monitoring (RUM)
+  enable-rum: 'TRUE'
+  {{ end }}
```
marians, line 24: After https://github.com/giantswarm/installations/pull/1493 is merged, this is going to change. But I don't want to make it too complicated for now.

axbarsan (approved):

PR #1823, opened 2020-09-16
Allow having no worker nodes in V4 clusters

axbarsan:
As seen on https://gigantic.slack.com/archives/CQ465CFG8/p1600251817026800

In this PR:
* refactoring the V4 worker details utilities, so they all have the same interface, and they all allow empty values for worker nodes
* allow having no worker nodes in the V4 cluster detail table


ubergesundheit (approved):
Didn't test this but looks good!

PR #1824, opened 2020-09-16
Use EnableUserMonitoring setting to enable user monitoring

marians:
Towards https://github.com/giantswarm/giantswarm/issues/12047

This PR makes use of the `.Values.Installation.V1.GiantSwarm.Happa.EnableUserMonitoring` introduced in https://github.com/giantswarm/installations/pull/1493 to enable RUM.

So far, the fact that an installation is a test installation also enabled user monitoring. As we want to roll it out to more non-testing installations, we need a dedicated switch.

### DIFF
--- a/helm/happa/ci/default-values.yaml
+++ b/helm/happa/ci/default-values.yaml
@@ -10,6 +10,7 @@ Installation:
         Letsencrypt: ""
         API:
           Letsencrypt: ""
+        EnableUserMonitoring: false
       Passage:
         Address: ""
       OIDC:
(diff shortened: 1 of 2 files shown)


marians:
Current problem:

image

image

I can't see a problem in index.html.

marians:
Could not reproduce the problem above...

marians:
Confirmed to work as expected in several installations, including non-testing.

axbarsan (approved):
LGTM

PR #1825, opened 2020-09-17
azure v5: Enable node pool autoscaling

axbarsan:
This PR:
* adds support for node pool autoscaling on Azure V5
* displays the actual error message when a node pool name change fails

The behaviour is the same as on AWS, so no logic changes.

### DIFF
--- a/src/actions/nodePoolActions.js
+++ b/src/actions/nodePoolActions.js
@@ -142,13 +142,6 @@ export function nodePoolPatch(clusterId, nodePool, payload) {
           nodePool,
         });
 
-        new FlashMessage(
-          'Something went wrong while trying to update the node pool name',
-          messageType.ERROR,
-          messageTTL.MEDIUM,
-          'Please try again later or contact support: support@giantswarm.io'
-        );
-
         throw error;
       });
   };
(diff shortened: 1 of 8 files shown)


src/actions/nodePoolActions.js
```diff
@@ -142,13 +142,6 @@ export function nodePoolPatch(clusterId, nodePool, payload) {
           nodePool,
         });
 
-        new FlashMessage(
```
axbarsan, line 145: Removed it from here because it was showing up when failing to scale a node pool. Node pool scaling has error displaying built-in, so we don't need the additional flash message

ubergesundheit (approved):
Works for me. Maybe you could distribute columns in the node pool table better?

image

axbarsan:
> image

I made the columns fill the rest of the row 👍🏻 
image

src/components/Cluster/ClusterDetail/V5ClusterDetailTable.js
```diff
@@ -369,6 +373,11 @@ class V5ClusterDetailTable extends React.Component {
       !isClusterCreating(cluster) &&
       !isClusterUpdating(cluster);
 
+    const machineTypeLabel =
+      provider === Providers.AWS ? 'Instance type' : 'VM Size';
```
axbarsan, line 377: This is a little extra. Noticed that on both AWS and Azure this label said `Instance type`, and it should be `VM Size` on Azure

PR #1827, opened 2020-09-17
Fix breaking app catalog layout after clearing the search bar

axbarsan:
If you use a screen that is not very wide, this is what currently happens when clearing the search bar contents quickly after typing:

image


src/components/AppCatalog/AppList/AppListInner.js
```diff
@@ -34,7 +35,7 @@ const TitleAndIcons = styled('div')`
 `;
 
 class AppListInner extends React.Component {
-  static filterApps(searchQuery, allApps) {
+  static filterApps = throttle((searchQuery, allApps) => {
```
axbarsan, line 38: Throttle app filtering, to improve the performance

ubergesundheit (approved):
Confirmed this working. Throttling is a nice addition!

PR #1844, opened 2020-09-21
Refactor organizations state management to a domain-specific way

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/10461

Domain specific folder structure, type definitions, and all the good stuff. I will leave out the refactor to our async action factory for this PR and the next ones, since it sometimes takes a long time to do, and together with so many changes, it can easily break so many stuff. I will create a different issue for that


src/stores/error/reducer.ts
```diff
@@ -18,7 +17,7 @@ const errorActionTypeRegexp = new RegExp(
 
 const initialState: IErrorState = {};
 
-const errorReducer = produce((draft: IState, action: ErrorActions) => {
+const errorReducer = produce((draft: IErrorState, action: ErrorActions) => {
```
axbarsan, line 20: This is a leftover from my previous PR

ubergesundheit (approved):
"Does not crash":tm:

PR #1846, opened 2020-09-22
Re-enable Install Ingress button for non-ready clusters

ubergesundheit:
This PR re-enables the `Install Ingress Controller` button for non-ready clusters, but adds an additional clause to the end indicating a possible delay in installation. (`... once cluster creation|upgrade has finished`)

image


marians (approved):
❤️

PR #1847, opened 2020-09-22
Refactor node pools state management to a domain-specific way

axbarsan:

Domain specific folder structure, type definitions, and all the good stuff


ubergesundheit (approved):
Does not crash, code looking good

PR #1848, opened 2020-09-23
Refactor users state management to a domain-specific way

axbarsan:


ubergesundheit (commented):
Both signing in with auth0 and email/password works.

After sign in there is a brief period of time where the organization selector displays "no organization", then reloading everything when signing in with email/password. This happens only at the first sign in.

Is this maybe related to #1844 ?

axbarsan:
> After sign in there is a brief period of time where the organization selector displays "no organization", then reloading everything when signing in with email/password. This happens only at the first sign in.

That's a good catch. That seems to be a bug that existed even before this refactor. I was able to reproduce it on current master:
* Log in
* Log out
* Log in again
* Observe organization error

The problem was that we were leveraging the clusters loading flag, which [was set to `true` by default](https://github.com/giantswarm/happa/blob/84e77a76a5cc4c8233e6f4cda17713495b37b2f9/src/reducers/loadingReducer.js#L4), but was not set again to `true` on log out.

I made it work by creating a new action just for the initial load request. Seems to work great

axbarsan:
> I made it work by creating a new action just for the initial load request. Seems to work great

If only it was that easy... There were also problems with flashing and multiple reloads, which were breaking tests and are all fixed now

ubergesundheit (approved):
As smooth as butter :butter:

PR #1849, opened 2020-09-24
Move all password reset actions into the user state management domain

axbarsan:

There's no point to have these in a different place


ubergesundheit (approved):
LGTM

PR #1853, opened 2020-09-25
Move the remaining app catalog logic to the app catalog state domain

axbarsan:

Pretty much what the title says.


src/components/Auth/Login.tsx
```diff
@@ -19,8 +19,7 @@ import { parseErrorMessages } from './parseErrorMessages';
 
 // The props coming from the global state (AKA: `mapStateToProps`)
 interface IStateProps {
-  // For now, until we have a type for it
-  user: Record<string, never>;
+  user: IUser;
```
axbarsan, line 22: Unrelated 😅

src/components/Cluster/ClusterDetail/KeypairCreateModal/KeyPairCreateModal.tsx
```diff
@@ -35,7 +35,7 @@ const StyledModal = styled(BootstrapModal)`
 `;
 
 interface IKeyPairCreateModalProps {
-  user: Record<string, never>;
+  user: IUser;
```
axbarsan, line 38: Unrelated 😅

src/stores/callAPIMiddleware.js
```diff
@@ -19,30 +19,29 @@ export function callAPIMiddleware({ dispatch, getState }) {
...
-        if (throwOnError) {
-          throw error;
-        }
-
-        // TODO: Remove this, this supports a pattern we want to factor out
-        // eventually.
-        return {
-          error: error.message,
-        };
+    try {
```
axbarsan, line 22: The problem here was when the `shouldPerform()` check returned `false`. The result would've been `undefined`, which has no `then()` method, and then resulted in an error.

src/stores/callAPIMiddleware.js
```diff
@@ -19,7 +19,10 @@ export function callAPIMiddleware({ dispatch, getState }) {
       type: types.request,
     });
 
-    return doPerform(getState(), dispatch)
+    const request = doPerform(getState(), dispatch);
+    if (!request) return next(action);
```
axbarsan, line 23: This is needed because if we use `shouldPerform()`, and it returns `false`, then the result of `doPerform()` is `undefined`, which is not a promise, so we can't use `.then()` on it.

ubergesundheit (approved):
Works like a charm

PR #1864, opened 2020-09-28
Move all credential-specific state management logic into the organization domain

axbarsan:

Seems like a good place for these


ubergesundheit (approved):
:+1:

PR #1865, opened 2020-09-28
Skip code refactors from release notes

axbarsan:
These should not be included in the release notes, as stated here: https://intranet.giantswarm.io/docs/dev-and-releng/happa-release/


.github/workflows/release-notes.yml
```diff
@@ -7,7 +7,7 @@ jobs:
   create-release-notes:
     runs-on: ubuntu-latest
     steps:
-      - uses: release-drafter/release-drafter@v5
+      - uses: release-drafter/release-drafter@v5.11
```
axbarsan, line 10: Let's also update to the latest version while we're here

ubergesundheit (approved):
:+1:

PR #1867, opened 2020-09-28
Show kubernetes version in cluster overview

marians:
Closes https://github.com/giantswarm/giantswarm/issues/13478

This PR adds the k8s version number to the cluster overview.

## Before

image

## After

image

### DIFF
--- a/src/components/Cluster/ClusterDetail/RegionAndVersions.js
+++ b/src/components/Cluster/ClusterDetail/RegionAndVersions.js
@@ -66,14 +66,7 @@ function RegionAndVersions({
             <>
               <Dot />
               <i className='fa fa-kubernetes' />
-              {(() => {
-                const kubernetes = release.components.find(
-                  (component) => component.name === 'kubernetes'
-                );
-                if (kubernetes) return kubernetes.version;
-
-                return null;
-              })()}
+              {release.kubernetesVersion}
             </>
           )}
           {!release &&
--- a/src/components/Cluster/ClusterDetail/RegionAndVersions.js
+++ b/src/components/Cluster/ClusterDetail/RegionAndVersions.js
@@ -66,7 +66,14 @@ function RegionAndVersions({
             <>
               <Dot />
               <i className='fa fa-kubernetes' />
-              {release.kubernetesVersion}
+              {(() => {
+                const kubernetes = release.components.find(
+                  (component) => component.name === 'kubernetes'
+                );
+                if (kubernetes) return kubernetes.version;
+
+                return null;
+              })()}
             </>
           )}
           {!release &&
--- a/src/components/Home/ClusterDashboardItem.js
+++ b/src/components/Home/ClusterDashboardItem.js
@@ -254,10 +254,10 @@ function ClusterDashboardItem({
               </span>
             </RefreshableLabel>
             <Dot style={{ paddingLeft: 0 }} />
-            <RefreshableLabel value={release.kubernetesVersion}>
+            <RefreshableLabel value={release?.kubernetesVersion}>
               <span>
                 <i className='fa fa-kubernetes' title='Kubernetes version' />{' '}
-                {release.kubernetesVersion}
+                {release?.kubernetesVersion ?? 'n/a'}
               </span>
             </RefreshableLabel>
             <Dot style={{ paddingLeft: 0 }} />
--- a/src/components/Home/ClusterDashboardItem.js
+++ b/src/components/Home/ClusterDashboardItem.js
@@ -254,19 +254,10 @@ function ClusterDashboardItem({
               </span>
             </RefreshableLabel>
             <Dot style={{ paddingLeft: 0 }} />
-            <RefreshableLabel value={cluster.kubernetes_version}>
+            <RefreshableLabel value={release.kubernetesVersion}>
               <span>
                 <i className='fa fa-kubernetes' title='Kubernetes version' />{' '}
-                <span>
-                  {(() => {
-                    const kubernetes = release.components.find(
-                      (component) => component.name === 'kubernetes'
-                    );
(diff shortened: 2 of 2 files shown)


src/components/Home/ClusterDashboardItem.js
```diff
@@ -251,6 +254,22 @@ function ClusterDashboardItem({
               </span>
             </RefreshableLabel>
             <Dot style={{ paddingLeft: 0 }} />
+            <RefreshableLabel value={cluster.kubernetes_version}>
```
ubergesundheit, line 257: there is no `kubernetes_version` field in clusters

src/components/Home/ClusterDashboardItem.js
```diff
@@ -251,6 +254,22 @@ function ClusterDashboardItem({
...
+                <i className='fa fa-kubernetes' title='Kubernetes version' />{' '}
+                <span>
+                  {(() => {
+                    const kubernetes = release.components.find(
+                      (component) => component.name === 'kubernetes'
+                    );
+                    if (kubernetes) return kubernetes.version;
+
+                    return null;
+                  })()}
```
ubergesundheit, line 268: The `release` contains the field `kubernetesVersion` which should be used instead of this.

src/components/Home/ClusterDashboardItem.js
marians: Right, changing

src/components/Home/ClusterDashboardItem.js
marians: Oh. Then I copied a bad example :)

axbarsan, line 268: Can you please extract this into a separate function, this is a pattern that we're trying to get rid of?

ubergesundheit (changes requested):
Nice addition. Implementation is a little rough around the edges

src/components/Home/ClusterDashboardItem.js
marians: Gerald already mentioned that it can be replaced simply with `release.kubernetesVersion`.

marians:
There is a lot going on in the tests which I don't understand. Maybe somebody could help me here?

src/components/Home/ClusterDashboardItem.js
```diff
@@ -251,6 +254,13 @@ function ClusterDashboardItem({
               </span>
             </RefreshableLabel>
             <Dot style={{ paddingLeft: 0 }} />
+            <RefreshableLabel value={release.kubernetesVersion}>
```
axbarsan, line 257: The problem is possibly because if the release version is not found, then the value of `release` is `null`, which doesn't have a `kubernetesVersion` property.

This should be enough here.

```suggestion
            <RefreshableLabel value={release?.kubernetesVersion}>
```

marians:
There is also this problem:

image

This remains with my branch, while it's not happening in current master.

src/components/Home/ClusterDashboardItem.js
```diff
@@ -251,6 +254,13 @@ function ClusterDashboardItem({
...
+              <span>
+                <i className='fa fa-kubernetes' title='Kubernetes version' />{' '}
+                {release.kubernetesVersion}
```
axbarsan, line 260: However, inside the `span` element, you might want to add a placeholder

```suggestion
                <i className='fa fa-kubernetes' title='Kubernetes version' />{' '}
                {release?.kubernetesVersion ?? 'N/A'}
```

src/components/Home/ClusterDashboardItem.js
```diff
@@ -143,10 +144,12 @@ function ClusterDashboardItem({
   dispatch,
   nodePoolsLoadError,
   clusterId,
+  releases,
 }) {
   // If the cluster has been deleted using gsctl, Happa doesn't know yet.
   if (!cluster) return null;
 
+  const release = releases[cluster.release_version] ?? null;
```
ubergesundheit, line 152: Maybe the errors in the tests are related to this line. `null` does not support being accessed (Line 257 - 263)

marians: So what should I do?

src/components/Home/ClusterDashboardItem.js
marians: Done

src/components/Home/ClusterDashboardItem.js
marians: Done

marians:
Tests are green now, but this keeps showing up as zero:

image

src/components/Cluster/ClusterDetail/RegionAndVersions.js
```diff
@@ -66,14 +66,7 @@ function RegionAndVersions({
...
               <i className='fa fa-kubernetes' />
-              {(() => {
-                const kubernetes = release.components.find(
-                  (component) => component.name === 'kubernetes'
-                );
-                if (kubernetes) return kubernetes.version;
-
-                return null;
-              })()}
+              {release.kubernetesVersion}
```
ubergesundheit, line 69: Could you please undo this change? I'm preparing a PR which rewrites this component in typescript lined up and would like to avoid having merge conflicts

marians:
Ack, thanks!

src/components/Home/ClusterDashboardItem.js
ubergesundheit: You already solved this by adding `?` to the mentioned lines

src/components/Cluster/ClusterDetail/RegionAndVersions.js
marians: If that makes you happy :) Done

ubergesundheit (approved):
:bow:

PR #1870, opened 2020-09-28
Move all the app related state interaction into the app catalog state domain

axbarsan:

What the title says, with:
* App actions
* App config actions
* App secret actions
* Cluster app actions


ubergesundheit (approved):
I like that you expanded the appcatalog store tests :+1:

axbarsan:
> I like that you expanded the appcatalog store tests 👍

I actually didn't 😬 just moved the other app action related tests from everywhere else

PR #1871, opened 2020-09-29
Add data-dd-action-name attribute to clickable items for better event names

marians:
For https://github.com/giantswarm/giantswarm/issues/13497

This PR adds named click events to components in the cluster creation form.

In the case of the `NumberPicker` component we have a specialty: We want to distinguish clicks on the plus and minus buttons. However, we also want to name each click specifically, which prevents hard-coding the event name inside the component. So I suggest to pass down an `eventNameSuffix` property.

image


marians:
Added the attribute to the main menu items, but can't see it work.

axbarsan (approved):
Not blocking for now, but we should think of a more straightforward interface for registering these custom events

PR #1873, opened 2020-09-29
change dropdown trigger & menu width

ubergesundheit:
Fixes giantswarm/giantswarm#13502

This PR:

- restricts the maximum length of the organization dropdown trigger
- allows maximum length of the organization dropdown menu

<details>

<summary>Screenshots</summary>

image

image

</details>

### DIFF
--- a/src/components/UI/Navigation/OrganizationDropdown.js
+++ b/src/components/UI/Navigation/OrganizationDropdown.js
@@ -6,6 +6,8 @@ import { NavLink } from 'react-router-dom';
 import { AppRoutes, OrganizationsRoutes } from 'shared/constants/routes';
 import DropdownMenu, { DropdownTrigger, List } from 'UI/DropdownMenu';
 
+const TriggerLabelMaxLength = 15;
+
 const OrganizationDropdownTrigger = styled(DropdownTrigger)`
   width: unset;
   height: unset;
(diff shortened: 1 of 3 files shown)


src/components/UI/Navigation/OrganizationDropdown.js
```diff
@@ -145,6 +147,18 @@ const OrganizationDropdown = ({
...
+
+  if (sortedOrganizations.length !== 0) {
+    if (selectedOrganization.length > TriggerLabelMaxLength) {
+      triggerLabel = (
+        <>{selectedOrganization.slice(0, TriggerLabelMaxLength)}&hellip;</>
+      );
+    } else {
+      triggerLabel = selectedOrganization;
+    }
+  }
```
ubergesundheit, line 160: Tried to do this with css overflow and ellipsis but couldn't make it working. The component uses many inline divs. This was quick and easy

axbarsan: We actually have an element made specifically for this
https://github.com/giantswarm/happa/blob/a939dc81043fab10d0ee676f9cfc4927fd25c355/src/components/UI/Truncated.tsx

ubergesundheit: :facepalm: I was searchting for this.. I tried to look for overflow or ellipsis :sweat_smile:

axbarsan (approved):
LGTM

marians (approved):
Looking good!

We have a limit on org name length, right? So what we see there is pretty much the worst case.

ubergesundheit:
> We have a limit on org name length, right? So what we see there is pretty much the worst case.

Yes 63 characters is the upper limit

PR #1874, opened 2020-09-29
Move all cluster state management into the cluster state management domain

axbarsan:

This one is huge, and it is also the last big one. This is the last `entity` from the state. We only have to refactor the `main` part and some minor tweaks here and there and we're done.

I left the `clusterLabels` state domain in a separate folder, as it was before, because it was added recently and I figured it has a reason to be separate.

There's a lot of dark magic going on in the cluster actions (a lot of data is injected into the clusters fetched from the API). I will make a separate issue for cleaning that up.


ubergesundheit (approved):
Absolutely massive :O

Took me a long time to look at the code. Everything looks good. Tests are green. :+1:

axbarsan:
> Absolutely massive :O
> 
> Took me a long time to look at the code. Everything looks good. Tests are green. 👍

Thank you for taking the time to review it 🤗

PR #1875, opened 2020-09-29
Create new semver compare utility, which can compare pre-release versions

axbarsan:
The old `semver-compare` package did not know how to compare pre-release version numbers (e.g. `13.0.0-alpha`, which we currently use for Azure)


src/shared/constants/index.ts
```diff
@@ -22,7 +22,7 @@ export const Constants = {
   DEFAULT_NODEPOOL_NAME: 'Unnamed node pool',
   AZURE_NODEPOOL_DEFAULT_VM_SIZE: 'Standard_D4_v3',
   AWS_V5_VERSION: '10.0.0',
-  AZURE_V5_VERSION: '40.0.0',
+  AZURE_V5_VERSION: '13.0.0',
```
axbarsan, line 25: While we're here, also change the default Azure v5 version to the one that we'll use in production (the version here only gets used if it is not present in the API response.

src/shared/constants/index.ts
```diff
@@ -22,7 +22,7 @@ export const Constants = {
   DEFAULT_NODEPOOL_NAME: 'Unnamed node pool',
   AZURE_NODEPOOL_DEFAULT_VM_SIZE: 'Standard_D4_v3',
   AWS_V5_VERSION: '10.0.0',
-  AZURE_V5_VERSION: '40.0.0',
+  AZURE_V5_VERSION: '13.0.0-alpha',
```
axbarsan, line 25: While we're here, also change the default Azure v5 version to the one that we'll use in production (the version here only gets used if it is not present in the API response.

ubergesundheit (approved):
LGTM.

Not for this PR but behaviour when supplying non semver formatted strings into the `compare` function is pretty much undefined right? I've looked at the original library. They also don't handle non semver formatted strings so this worked fine until now, but nevertheless something to keep in mind.

axbarsan:
> Not for this PR but behaviour when supplying non semver formatted strings into the `compare` function is pretty much undefined right? I've looked at the original library. They also don't handle non semver formatted strings so this worked fine until now, but nevertheless something to keep in mind.

Comparison between strings that don't respect the format and a string that does is handled here:
- https://github.com/giantswarm/happa/blob/66dafa2fd88b0d182ce3aa30d6f51071210a782f/src/lib/semver.ts#L25
- https://github.com/giantswarm/happa/blob/66dafa2fd88b0d182ce3aa30d6f51071210a782f/src/lib/semver.ts#L29

Comparing between 2 strings that don't respect the format is not handled, indeed.

I added test inputs for regular strings, to keep track of their behaviour.

PR #1876, opened 2020-09-30
Move all main actions to the main state management domain

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/10461

Moving all the `main` related stuff to the main state management domain.
The PR seems big because I also moved the login stuff from the user state domain to the main state domain (they make changes in the `main` state, using the `main` reducer, so it makes sense to have them here).


testUtils/modelMocks.js
```diff
@@ -1,11 +1,16 @@
...
-// eslint-disable-next-line no-import-assign
-userSelectors.selectAuthToken = jest.fn(() => (state) => {
-  return [state.main.loggedInUser.auth.token, AuthorizationTypes.GS];
-});
+function mockMainSelectors() {
+  const original = jest.requireActual('stores/main/selectors');
+
+  return {
+    ...original,
+    selectAuthToken: jest.fn(() => (state) => {
```
axbarsan, line 12: Only mocking the `selectAuthToken` function in all tests, so no more random tests failing because of some mocked selector 👍🏻

ubergesundheit (approved):
To be honest here's what I did:

- checked out locally 
- ran tests
- started a development server and browsed a minute through happa

Nothing particular came up.. While scrolling through all changes I notice you've added additional checks here and there.

Tests are green, lets go with it

PR #1877, opened 2020-09-30
Move all loading/error state management to their own domains

axbarsan:


ubergesundheit (approved):
Everything loaded as expected

PR #1879, opened 2020-10-01
Allow having apps defined without labels

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/12568

This PR accounts for having a `null` label map in the app spec. It also defines other label maps as nullable, since that's how they behave in Go, and how they are serialized and passed around.


ubergesundheit (approved):
Implementation looks straight forward

PR #1894, opened 2020-10-05
Extract k8s version into component, remove patch level

marians:
Extracts the Kubernetes version label into a component. Also reduces the amount of information given to the minor version, cuts off the patch version (in the cluster overview only).

Preparation for https://github.com/giantswarm/giantswarm/issues/13439

### Preview

image

image

### DIFF
--- a/src/components/UI/KubernetesVersionLabel.tsx
+++ b/src/components/UI/KubernetesVersionLabel.tsx
@@ -1,14 +1,11 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 
-
 interface IKubernetesVersionLabelProps {
   version: string;
 }
 
-const KubernetesVersionLabel = ({
-  version,
-}: IKubernetesVersionLabelProps) => {
+const KubernetesVersionLabel = ({ version }: IKubernetesVersionLabelProps) => {
   let versionLabel = 'n/a';
   if (version) {
     // only show major and minor k8s version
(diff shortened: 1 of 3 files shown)


src/components/UI/KubernetesVersionLabel.tsx
```diff
@@ -0,0 +1,32 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+interface IKubernetesVersionLabelProps {
+  version: string;
+}
+
+const KubernetesVersionLabel = ({ version }: IKubernetesVersionLabelProps) => {
```
axbarsan, line 8: We usually use this pattern

```suggestion
const KubernetesVersionLabel: React.FC<IKubernetesVersionLabelProps> = ({ version }) => {
```

axbarsan (commented):
LGTM, just this small thing

src/components/UI/KubernetesVersionLabel.tsx
marians: Look, I have no idea what I'm doing there. Just copied and modified `ValueLabel.tsx` 😜

marians:
typecheck is failing. Don't know what this means:

image

src/components/UI/KubernetesVersionLabel.tsx
```diff
@@ -0,0 +1,34 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+interface IKubernetesVersionLabelProps {
+  version: string;
```
axbarsan, line 5: Typechecking fails because this isn't marked as optional in the props interface, but the prop-types say that it is optional (`PropTypes.string.isRequired` would make it not optional).

When using default props, the easiest way is to just do

```suggestion
  version?: string;
```

axbarsan (approved):

PR #1896, opened 2020-10-06
Don't let user execute requests if they are not logged in

axbarsan:
As seen on https://gigantic.slack.com/archives/C03CPNRTS/p1601930766351800

### DIFF
--- a/src/stores/main/selectors.ts
+++ b/src/stores/main/selectors.ts
@@ -39,7 +39,7 @@ export const selectAuthToken = (
   > => {
     const auth = state.main.loggedInUser?.auth ?? null;
     if (!auth) {
-      return Promise.resolve(['', AuthorizationTypes.GS]);
+      return Promise.reject(new Error('You are not logged in.'));
     }
     const scheme = auth.scheme;
     let currentToken = auth.token;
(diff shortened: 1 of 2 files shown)


src/stores/main/selectors.ts
```diff
@@ -39,7 +39,7 @@ export const selectAuthToken = (
   > => {
     const auth = state.main.loggedInUser?.auth ?? null;
     if (!auth) {
-      return Promise.resolve(['', AuthorizationTypes.GS]);
+      return Promise.reject(new Error('You are not logged in.'));
```
MarcelMue, line 42: Going to level with you: I have very little idea about typescript or what is happening here. How does this solve the issue we are seeing?

axbarsan: The purpose of this function is to fetch the currently logged in user authentication token, or renew it via Auth0 if it's expired.

It is only used as part of the installation info GET request, since this is a new pattern that we were trying out.

Leaving promises aside, the problem before is that this function was returning an empty authentication token if there was no user logged in, instead of throwing an error.

Throwing an error here means that the user will be redirected to the login page, instead of executing the info request with no authentication header.

I tested this on both `viking` and `ghost`, and everything worked perfectly.

MarcelMue (approved):
Okay got it - works for me!

PR #1898, opened 2020-10-07
Display full cluster update error details

axbarsan:
As seen on https://gigantic.slack.com/archives/CNUDS3AMU/p1602056757165900

This makes sure that the error received from the API is visible in the UI. Especially useful when a cluster upgrade is blocked by a webhook, and you have no idea why.

I also wrote an utility function that we can reuse for getting the error message from other requests.


whites11 (approved):
Seems klingon but seems reasonable :p

PR #1923, opened 2020-10-12
Automatically enable the latest active version, even for admins

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/13589

Listed this as a `dev-change`, because it is only relevant for GS Staff, as this behaviour was already present for non-GS staff.

This PR makes the latest active version be selected by default in the cluster creation form, regardless if you are an admin or not.


marians (approved):
The code changes don't speak to me, however I trust in your abilities :)

PR #1925, opened 2020-10-12
Add basic dependabot configuration

marians:
Adds a basic configuration for dependabot, which should prevent that we get multiple open PRs hammering our CI system.


axbarsan (approved):
Nice 👍🏻

PR #1929, opened 2020-10-13
Worker node counter is broken when there are no node pools

axbarsan:
When you have a cluster ready, with no node pools (and therefore no nodes), the node count display is broken.

### Before

image

### After

image

### DIFF
--- a/src/components/Cluster/ClusterDetail/NodesRunning.tsx
+++ b/src/components/Cluster/ClusterDetail/NodesRunning.tsx
@@ -18,7 +18,7 @@ const NodesRunning: FC<INodesRunningProps> = ({
   numNodePools,
   workerNodesRunning,
 }) => {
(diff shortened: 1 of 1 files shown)


src/components/Cluster/ClusterDetail/NodesRunning.tsx
```diff
@@ -18,7 +18,7 @@ const NodesRunning: FC<INodesRunningProps> = ({
   numNodePools,
   workerNodesRunning,
 }) => {
-  if (workerNodesRunning === 0 && isClusterCreating) {
+  if (isClusterCreating) {
```
axbarsan, line 21: Also small change here, we should always display the `Nodes are not ready` while a cluster is creating.

marians (approved):
LGTM if it works for you

pipo02mix (approved):
nice

PR #1935, opened 2020-10-15
Improve RUM event binding

axbarsan:
In order to provide an easier to use interface for custom RUM events, this PR includes these changes:

- Added the `DD_RUM` property to the window object
- Move all custom RUM action names into a single enum data structure
- Create the `mergeActionNames` utility, which can be used for merging action names (for example when you want to add a certain suffix to a more generic action name)
- Create the `RUMActionTarget` component, which acts like a decorator for other elements, adding the `data-dd-action-name` property to the wrapped element. It only allows a single child.

### DIFF
--- a/src/components/RUM/RUMActionTarget.tsx
+++ b/src/components/RUM/RUMActionTarget.tsx
@@ -1,0 +1,34 @@
+import PropTypes from 'prop-types';
+import React, { ReactElement } from 'react';
+
+interface IRUMActionTargetProps {
+  name: string;
+  children: ReactElement;
+}
+
+/**
+ * Decorator that adds the RUM element properties to the wrapped
+ * child element, for dispatching custom RUM actions.
+ */
+const RUMActionTarget: React.FC<IRUMActionTargetProps> = ({
+  name,
+  children,
+}) => {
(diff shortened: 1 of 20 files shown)


src/components/RUM/RUMActionTarget.tsx
```diff
@@ -0,0 +1,34 @@
+import PropTypes from 'prop-types';
+import React, { ReactElement } from 'react';
+
+interface IRUMActionTargetProps {
+  name: string;
+  children: ReactElement;
```
axbarsan, line 6: Enforces a single child in the compiler

src/components/RUM/RUMActionTarget.tsx
```diff
@@ -0,0 +1,34 @@
...
+  if (!children.props) return children;
+
+  const newChildren = Object.assign({}, children, {
+    props: {
+      ...children.props,
+      'data-dd-action-name': name.toUpperCase(),
+    },
+  });
+
+  return React.Children.only(newChildren);
```
axbarsan, line 26: Enforces a single child at run time

oponder (approved):
LGTM! Nice work!

PR #1937, opened 2020-10-16
Set background color early

marians:
This PR attempts to change the loading behaviour so that the final background color of the viewport is set as early as possible, preventing a white flash during loading.

Note that this uses the background color of the login page, which is darker than the app's normal background

## Before

image

Timing here based on webpagetest.org with "cable" as network bandwith (5 MBit/Sec).

Results: https://www.webpagetest.org/result/201016_DiYT_ae9f28b21731b39e2c4c287dfa1455ba/

## After

image

Results: https://www.webpagetest.org/result/201019_DiDF_5a79b726d23f0ad1c5c95755a8cebb8c/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -16,6 +16,11 @@
       sizes="128x128"
       href="/images/app_icon_128.png"
     />
+    <style>
+    html {
+      background-color: #234a61;
+    }
+    </style>
     <placeholder_rum/>
     <link
       rel="stylesheet"
@@ -51,13 +56,12 @@
     </script>
 
     <style>
-      body {
-        background: #234a61;
-      }
-
-      .loading footer {
-        display: none;
-      }
+    html, body {
+      background-color: #234a61;
+    }
+    .loading footer {
+      display: none;
+    }
     </style>
   </head>
   <body class="loading">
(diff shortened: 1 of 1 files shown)


src/index.html
```diff
@@ -51,13 +56,12 @@
...
     <style>
-      body {
-        background: #234a61;
-      }
-
-      .loading footer {
-        display: none;
-      }
+    html, body {
+      background-color: #152531;
```
oponder, line 60: do we still need this here if we also moved it up now?

oponder (changes requested):

src/index.html
marians: Not sure, really. I left it in, since it could be that the CSS we load in-between resets the color again, no?

oponder: only if we explicitly have something in there that sets it to white or some other color again.. other wise, it's 'cascading style sheets', and it cascades into eachother.

oponder (approved):
approving so you can keep movin 👍 

I think the background-color at the bottom again shouldn't be needed, but not a big deal.

marians:
Removed the redundant style definition and tested again. Works. Thanks!

PR #1940, opened 2020-10-19
Show upgrade in progress immediately after upgrade has been started

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/13349

The problem is that when an user initiates an upgrade, it takes 1-2 minutes before the cluster conditions are actually updated and the upgrade is started. This means that the user can not tell that an upgrade has been started for the first 1-2 minutes.

With this change, we make sure that the user who initiated an upgrade sees that an upgrade has actually been requested, by putting the cluster in an `awaiting upgrade` state, where everything behaves like it would during an upgrade (app installation blocked etc.). The only difference between this state and an `updating` state is that we display a different label in the cluster status, and that the status is only visible for the user that initiated the upgrade, since the change is only done on the client side.

To achieve this, we create a new map in the cluster state domain called `idsAwaitingUpgrade`, where we store all the cluster IDs that are awaiting the `Updating` condition in the CR. On every operation that changes clusters that are currently stored (such as refreshing the cluster list or deleting a cluster), we also reconcile the list of cluster IDs awaiting upgrade, to see whether one of the IDs belongs to a cluster that has been deleted, for example.

### Preview

<details>
<summary>Cluster details</summary>

image

</details>

<details>
<summary>Cluster details (hover)</summary>

image

</details>


oponder (approved):
UX wise I think this is great. Giving users immediate feedback is so important 👍 💯 
Code wise, also great!

Architecturally, I left a note in the parent issue about possibly tackling this in our k8s API resource design as well.

marians (approved):
Looks good to me

PR #1941, opened 2020-10-20
azure v5: change 'None' node pool AZ selection to 'Not specified'

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/13843

In this PR, the `None` NP AZ selection option is changed to `Not specified` and a longer explanation is now provided. Also, before, the explanation was shown in a tooltip, which was different from other options. That is now aligned, as well.

Unfortunately, to be able to make the option look like the others, it was easier to rewrite the whole AZ selector, than to add the option to the old implementation. In the previous implementation, we were relying on fixed heights, and a lot of CSS class magic. It would've required a lot of hacks to convert the new option and make everything behave correctly on all providers.

In the new implementation, I'm leveraging the `Collapse` component from `react-bootstrap`. Since we discussed about switching to a UI library, I didn't bother writing custom code for that.

The new implementation looks almost 100% the same as the old one, but the transitions are now way quicker, and there are no glitches when switching between `Automatic` and `Manual`, anymore.

### Preview

image


src/components/Cluster/ClusterDetail/AddNodePool/AddNodePoolAZSelectionCheckbox.tsx
```diff
@@ -0,0 +1,58 @@
...
+  value,
+  npID,
+  ...rest
+}) => {
+  const typeName = AvailabilityZoneSelection[type as AvailabilityZoneSelection];
+  const id = `np-${npID}-az-${typeName.toLowerCase()}`;
+
+  return (
+    <RUMActionTarget
+      name={mergeActionNames(RUMActions.SelectAZSelection, typeName)}
```
axbarsan, line 32: The RUM action name is now computed here, by appending the selection type to the end of the base action name (e.g. `SELECT_AZ_SELECTION_AUTOMATIC`)

T-Kukawka:
i would say: "By not specifying the availability zone, Azure will select the zone by itself where the VM size is most available."
rest looks good 👍

T-Kukawka (approved):

axbarsan:
> i would say: "By not specifying the availability zone, Azure will select the zone by itself where the VM size is most available."
> rest looks good 👍

Updated the screenshot 👍🏻

src/components/Cluster/ClusterDetail/AddNodePool/AddNodePoolAZSelection.tsx
```diff
@@ -0,0 +1,255 @@
...
+            npID={npID}
+            label='Not specified'
+            type={AvailabilityZoneSelection.NotSpecified}
+          />
+          <StyledPanelCollapse>
+            <p>
+              By not specifying an availability zone, Azure will select a zone
+              by itself, where the requested VM size has the best availability.
+              This is especially useful for GPU instances, which are not
+              available in all regions.
```
marians, line 220: I think this should be about AZ, not region

```suggestion
              available in all availability zones.
```

src/components/Cluster/ClusterDetail/AddNodePool/AddNodePoolAZSelection.tsx
```diff
@@ -0,0 +1,255 @@
...
+            onChange={onChange}
+            value={value}
+            npID={npID}
+            label='Not specified'
+            type={AvailabilityZoneSelection.NotSpecified}
+          />
+          <StyledPanelCollapse>
+            <p>
+              By not specifying an availability zone, Azure will select a zone
+              by itself, where the requested VM size has the best availability.
```
marians, line 218: Avoiding abbreviations where ever we can

```suggestion
              by itself, where the requested virtual machine size has the best availability.
```

src/components/Cluster/ClusterDetail/AddNodePool/AddNodePoolAZSelection.tsx
```diff
@@ -0,0 +1,255 @@
...
+              This is especially useful for GPU instances, which are not
```
marians, line 219: Sticking with Azure terminology

```suggestion
              This is especially useful for virtual machine sizes with GPU, which are not
```

marians:
LGTM! Gave some input regarding wording.

PR #1943, opened 2020-10-22
Select the organization ID from storage before comparing it to the cluster owner reference

axbarsan:
Related to https://github.com/giantswarm/api/pull/1159

Because organization IDs are now normalized in the API cluster lister, some calls and checks no longer work like before (e.g. we filter clusters using the `owner` reference field, to see if any of them belong to a specific organization).

To achieve this, we now get the original organization ID from the global state, based on the cluster `owner` reference, in order to get the ID with the proper casing.


oponder (approved):
LGTM. 

The normalizing in the API feels a bit weird.. but I guess there was a good reason for it?

tuommaki:
> The normalizing in the API feels a bit weird.. but I guess there was a good reason for it?

It was kind of a first mitigation to the problem. I'm writing currently a change to admission controller to enforce normalized form in CRs, but this won't change the problem completely still because `api` organization endpoint would still return mixed-case form which might or might not cause problems for client.

PR #1944, opened 2020-10-22
Fix relative URLs in READMEs

oponder:
towards: https://github.com/giantswarm/giantswarm/issues/13624

This adds a link renderer and some logic to generate a valid base URL for relative links in our App README's.

### DIFF
--- a/src/components/AppCatalog/AppDetail/AppDetail.js
+++ b/src/components/AppCatalog/AppDetail/AppDetail.js
@@ -4,6 +4,7 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import { Breadcrumb } from 'react-breadcrumbs';
 import { connect } from 'react-redux';
+import { Constants } from 'shared/constants';
 import { AppCatalogRoutes } from 'shared/constants/routes';
 import { listCatalogs, loadAppReadme } from 'stores/appcatalog/actions';
 import {
@@ -28,7 +29,7 @@ function hasReadmeSource(appVersion) {
     return false;
   }
 
-  return appVersion.sources.some((url) => url.endsWith('README.md'));
+  return appVersion.sources.some((url) => url.endsWith(Constants.README_FILE));
 }
 
 class AppDetail extends React.Component {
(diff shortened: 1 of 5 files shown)


src/components/AppCatalog/AppDetail/AppDetail.js
```diff
@@ -90,6 +91,34 @@ class AppDetail extends React.Component {
...
+
+    if (!readmeURL) return undefined;
+
+    // https://github.com/giantswarm/efk-stack-app/v0.3.2/README.md
+    readmeURL = readmeURL.replace(
+      'https://raw.githubusercontent.com/',
+      'https://github.com/'
+    );
+
+    const l = 'README.md'.length;
```
axbarsan, line 109: I see that we're using the `README.md` string in multiple places, can we move it to the constants module (`src/shared/constants`)?

oponder: I kinda think this makes it more "factored" than it needs to be.  But implemented it! Let me know what you think :+1:

src/stores/appcatalog/actions.ts
```diff
@@ -313,8 +316,15 @@ function fixTestAppReadmeURLs(readmeURL: string): string {
    * doesn't match, then the string is returned as is.
    * https://regex101.com/r/K2dxdN/1
    */
-  const regexMatcher = /^(.*)\/v?[0-9]+\.[0-9]+\.[0-9]+-(.*)\/README\.md$/;
-  const fixedReadmeURL = readmeURL.replace(regexMatcher, '$1/$2/README.md');
+
+  const escapedReadmeFile = Constants.README_FILE.replace('.', '.');
```
axbarsan, line 320: How does this change the value? 😬

oponder: Lol I think I was testing to make sure the tests would fail, but didn't put it back to the right value.

in this case the test still pass because of the meaning of '.' in regex.

Fixed :) thanks for spotting it

oponder: Ohh no.. it's prettier that's doing it... hmm

oponder: Ah ok I wasn't escaping the escape character correctly.

axbarsan: What I meant was that I'm not sure why we are replacing dots with dots. Am I missing something?

axbarsan (approved):
LGTM

PR #1947, opened 2020-10-27
Write RUM data to custom analytics endpoint

marians:
Towards https://github.com/giantswarm/giantswarm/issues/12047

Uses https://github.com/giantswarm/api/pull/1164

Also adds `bowser` to collect basic client/browser information.

Suggestions on how to tweak are highly welcome.

### DIFF
--- a/src/components/index.tsx
+++ b/src/components/index.tsx
@@ -38,16 +38,17 @@ enum GlobalEnvironment {
 
 interface IGlobalConfig {
   apiEndpoint: string;
-  cpApiEndpoint: string;
   audience: string;
-  cpAudience: string;
-  passageEndpoint: string;
-  environment: GlobalEnvironment;
-  ingressBaseDomain: string;
   awsCapabilitiesJSON: string;
   azureCapabilitiesJSON: string;
-  happaVersion: string;
+  cpApiEndpoint: string;
+  cpAudience: string;
   defaultRequestTimeoutSeconds: number;
+  enableRealUserMonitoring: boolean;
+  environment: GlobalEnvironment;
+  happaVersion: string;
+  ingressBaseDomain: string;
+  passageEndpoint: string;
 }
 
 declare global {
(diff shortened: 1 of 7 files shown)


marians:
Snyk complains:

image

However that dependency has not been introduced by this PR.

src/index.html
```diff
@@ -39,6 +39,7 @@
         audience: '<%= audience %>',
         cpAudience: '<%= cpAudience %>',
         passageEndpoint: '<%= passageEndpoint %>',
+        enableRealUserMonitoring: '<%= enableRealUserMonitoring %>',
```
marians, line 42: @giantswarm/chapter-front-end Is it necessary to update the webpack config with `enableRealUserMonitoring` ? If yes, how? Currently the build fails with `ERROR in Template execution failed: ReferenceError: enableRealUserMonitoring is not defined`

marians: Seems solved.

marians:
@giantswarm/chapter-front-end This is ready for review. I welcome any suggestions regarding structuring the code that is currently in `index.tsx`.

src/components/index.tsx
```diff
@@ -126,18 +133,58 @@ const getSizes = () => {
   };
 };
 
+async function submitCustomRUM(
+  payloadType: string,
+  payloadSchemaVersion: number,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  payload: any
```
ubergesundheit, line 140: Just a nitpick, works perfectly fine as is..

But if you would put the `payloadSchemaVersion` as last argument here and set a default, calls to this function would look nicer, (as in omitting the `payloadSchemaVersion`)

Also IMO it should be possible to not use `any` but `Record<string, never>`. Best would be to create an interface for this.

ubergesundheit: Also, why have already 2 versions? I've seen both `1` and `2`

marians: > Also IMO it should be possible to not use any but Record<string, never>. Best would be to create an interface for this.

That gives me some trouble:

image

ubergesundheit: This is also only guessed. Its possible it could work with `Record<string, string | number>`.

But: This strengthens my opinion that the payload should have a dedicated interface.

marians: I'll do `Record<string, string | number | object>` then.

marians:
@giantswarm/chapter-front-end I'd like to move one here and release. Anything that needs to be taken care of quickly? If not, I'd appreciate an approval. Thanks!

oponder (approved):
LGTM!

PR #1951, opened 2020-10-30
Attempt to make CI faster

marians:
This PR tests some ways to make CI finish faster.

### Baseline

image

### Attempts

1. Run `prettier` and `lint` directly in `checkout` job, since they are fast compared to the duration of starting an extra job.
2. Combine `build` and `build-container`, as it can save a ton of persisting and restoring of data between the two steps.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,14 +33,6 @@ workflows:
           requires:
             - checkout
           <<: *job_filters
-      - lint:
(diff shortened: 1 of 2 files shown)


marians:
With the tweaked setup now, for three runs:

image

image

image

marians:
Closing, as it didn't bring a notable effect

PR #1954, opened 2020-10-30
Hack the catalog urls for helm stable special case.

oponder:
The Helm Stable catalog is going EOL, but a mirror will be kept up at a new url.

Unfortunately that new URL doesn't really play nice with our assumptions, so have to add a special case here in Happa. 

Towards: https://github.com/giantswarm/giantswarm/issues/9985
See also: https://gigantic.slack.com/archives/C76JX6YLQ/p1604058274081800

Fyi @rossf7

### DIFF
--- a/src/stores/appcatalog/actions.ts
+++ b/src/stores/appcatalog/actions.ts
@@ -203,11 +203,21 @@ export function catalogLoadIndex(
 async function loadIndexForCatalog(catalog: IAppCatalog): Promise<IAppCatalog> {
   let indexURL = `${catalog.spec.storage.URL}index.yaml`;
 
+  // If we are trying to reach the Helm Stable catalog at it's old location,
+  // this URL hack helps us get it properly.
   if (
     catalog.spec.storage.URL ===
     'https://kubernetes-charts.storage.googleapis.com/'
   ) {
     indexURL = `/catalogs?url=${indexURL}`;
+  }
+
+  // If we are trying to reach the Helm Stable catalog at it's new location,
+  // it's url structure is a bit different.
+  // We need to remove /packages/ from the URL before adding /index.yaml.
+  if (catalog.spec.storage.URL === 'https://charts.helm.sh/stable/packages/') {
+    console.log('bingo bongo');
+    indexURL = 'https://charts.helm.sh/stable/index.yaml';
   }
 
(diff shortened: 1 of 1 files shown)


src/stores/appcatalog/actions.ts
```diff
@@ -203,13 +203,23 @@ export function catalogLoadIndex(
 async function loadIndexForCatalog(catalog: IAppCatalog): Promise<IAppCatalog> {
   let indexURL = `${catalog.spec.storage.URL}index.yaml`;
 
+  // If we are trying to reach the Helm Stable catalog at it's old location,
+  // route the request through a proxy in Happa's container which adds necessary
+  // CORS headers.
```
oponder, line 208: This comment is a boyscout commit to clarify and technically unrelated to this PR.

axbarsan (approved):

oponder:
@rossf7 We ran into a bug with `kubectl gs validate apps` because it would need the same kind of logic to detect this special case.

But now I'm kind of questioning why this change was really necessary. I can't remember why we did it that the `index.yaml` part has to be 'derived' from the storage URL.

The URLs to the chart tarballs in the `index.yaml` are absolute right, so it's not like we need to combine some info from the `index.yaml` with the storage URL..

Right now we have: 

```
 # AppCatalog CR:
 ...
  storage:
    URL: https://charts.helm.sh/stable/packages/
    type: helm
```

How does `URL` get used in other components? Why don't we have just an absolute reference to the `index.yaml` there?

rossf7:
@oponder app-operator needs to generate the tarball URLs and set them in the chart CRs.

The tarball URLs are in the index.yaml so we should probably add them to the appcatalogentry CRs. That would mean we're more flexible with the URL format.

But the helm stable catalog is deprecated and will be deleted. So I don't really want to make changes in app-operator for it.

oponder:
Ok I think I get it. We don't want app-operator to download the index.yaml. But in the future where we have the appcatalogentry CRs, then we could use that as the source of truth for the URLs instead of deriving them based on the storage url.

@webwurst that means the problem we're seeing with `kubectl gs validate apps` only applys to  apps in the `helm-stable` catalog, and I think we'll have to work around it by teaching `kubectl gs validate apps` the same thing we taught happa here.

PR #1960, opened 2020-11-03
Relax a test by using a regex instead of a single string match

marians:
To fix this problem, which apparently can happen in CI.

image


oponder (approved):
LGTM

axbarsan (changes requested):
With this change, the test no longer asserts for what the test case name suggests (`returns the correct status for a version that just reached its EOL now`).

`now != a few seconds ago`

Since we're testing a function that is making calculations based on time, and the tests are ran asynchronously, it is normal that the result is not right 1/10 times. The test plan can just be ran again, and I see no problem with that.

marians:
I find it a bit annoying and time-consuming to wait for a release or some checks and then find it's some random weirdness blocking progress. Today it blocked a release from being rolled out.

Shall I rename the test to make it fit better?

oponder:
`returns the correct status for a version that just reached its EOL now`

"A few seconds ago" and "today" are both close enough to "now" in the meaning of this test, so I don't see a blocker here.

In fact we're adding support for a case that is closer to 'now' than it was before..

----

Sidenote: That's a crazy sentence ain't it?

marians:
@axbarsan How strong are your feelings here? Do I have to change something, or can you approve like this?

axbarsan (approved):
disagree and commit :)

PR #1970, opened 2020-11-12
Refactor clusterId to id in some actions so that they show up in the 'loading flags by entity' state

oponder:
Fixes a bug reported in this slack thread: https://gigantic.slack.com/archives/C02EVLE9W/p1605183699371200


oponder:
Contains some boyscout refactoring of 'clusterId' to 'id' so that these requests will also show up in the loading flags by entity.

It ended up not being necessary to fix the bug. But I though it was an ok change.

axbarsan:
A question: Before, `isClusterCreating` would return `true` if the cluster had no conditions (I added this logic because I thought it made sense back then), however, I removed this yesterday [in a PR](https://github.com/giantswarm/happa/pull/1977), so it shouldn't be an issue anymore.

Does the bug still happen with the latest master?

oponder:
Nope! bug doesn't happen on master anymore! Was kinda puzzling over my PR as I rolled back some non working things and I couldn't get the bug to happen anymore.. 

This PR doesn't really do much besides refactor some things then 😅 

Not sure if you like the refactorings or not, I can close the PR, but I feel like the changes make sense on their own?

src/components/Home/ClusterStatus.tsx
```diff
@@ -74,8 +74,14 @@ const ClusterStatus: React.FC<IClusterStatusProps> = ({
...
   switch (true) {
     case typeof cluster === 'undefined':
+      return null;
+
     case typeof cluster.delete_date !== 'undefined':
+      return null;
+
     case cluster.delete_date === null:
+      return null;
+
```
axbarsan, line 84: Why is this necessary? It does the same thing

axbarsan (commented):
LGTM in general, just a small thing

src/components/Home/ClusterStatus.tsx
oponder: It was a prettier thing if I recall correctly.

oponder: Can't recreate it though.. but something complained.. let me see if I can find what

oponder: Not sure now.. ¯\_(ツ)_/¯ I can't make anything complain about those lines as they were originally.. strange.

I'll revert changes to this file.

axbarsan (approved):

PR #1972, opened 2020-11-16
Display that NP creation failed when trying to create a cluster and NPs at the same time

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/14226

With this PR:
* When creating a cluster with node pools at the same time, and node pool creation fails, you are redirected to the cluster detail page (because that was created successfully), and a flash message with the node pool creation error from the API is shown to you.


marians (approved):
LGTM from a logical standpoint. Didn't look at the code.

PR #1974, opened 2020-11-16
Let users retry creating a cluster, if there was an error on the server side the first time around

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/14223

With this PR:
* We rely on flash messages to display all cluster creation errors. We no longer show an error in the cluster creation form footer, and the cluster creation button is still there after an error occurs while creating a cluster.

### DIFF
--- a/src/stores/cluster/actions.ts
+++ b/src/stores/cluster/actions.ts
@@ -385,7 +385,7 @@ export function clusterCreate(
       new FlashMessage(
         'Something went wrong while trying to create the cluster.',
         messageType.ERROR,
-        messageTTL.MEDIUM,
+        messageTTL.LONG,
         errorBody
       );
     }
--- a/src/stores/cluster/actions.ts
+++ b/src/stores/cluster/actions.ts
@@ -377,11 +377,16 @@ export function clusterCreate(
     } catch (error) {
       dispatch({ type: CLUSTER_CREATE_ERROR, error: error.message });
 
-      new FlashMessage(
-        'An error occurred when trying to create the cluster.',
-        messageType.ERROR,
-        messageTTL.LONG,
+      const errorBody = extractMessageFromError(
+        error,
         'Please try again later or contact support: support@giantswarm.io'
+      );
+
+      new FlashMessage(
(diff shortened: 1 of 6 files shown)


src/stores/cluster/actions.ts
```diff
@@ -377,11 +377,16 @@ export function clusterCreate(
     } catch (error) {
       dispatch({ type: CLUSTER_CREATE_ERROR, error: error.message });
 
+      const errorBody = extractMessageFromError(
```
axbarsan, line 380: Display full error output in the flash message

oponder:
This `isComponentMounted` part feels a bit wrong to me... maybe we refactor `batchedClusterCreate` so it doesn't dispatch the push, and we do that instead in the component, after (or in the callback to) `this.setState({ submitting: false });`,

oponder:
Or maybe more like, if it succeeded we `dispatch the push to the cluster detail page`, and if it failed we do the `this.setState({ submitting: false });`

otherwise there will be a tiny moment where the submit button might flash green again before we get redirected to the cluster detail page

but the point would be to refactor batchedClusterCreate so we can correctly decide what to do in each case from within the component.

axbarsan:
> This `isComponentMounted` part feels a bit wrong to me... maybe we refactor `batchedClusterCreate` so it doesn't dispatch the push, and we do that instead in the component, after (or in the callback to) `this.setState({ submitting: false });`,

That's actually valuable feedback... I ended up using the loading flag from the cluster creation action, since it was the easiest solution, and it seems to get the job done.

marians (approved):
Not much to review for me, guess it's fine.

Do we do any limitation on the amount of content we try to cram into a Flash message?

axbarsan:
> Do we do any limitation on the amount of content we try to cram into a Flash message?

We don't do any truncation, it can have any length

oponder:
> That's actually valuable feedback... I ended up using the loading flag from the cluster creation action, since it was the easiest solution, and it seems to get the job done.

Nice! That's indeed the best way!

oponder (approved):
LGTM! 👍

PR #1975, opened 2020-11-16
Remove minimum width from the release selector component

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/14201

This doesn't seem to have any bad impact on how the release version selector on the cluster creation page looks, so it's safe.

**Before**

image

**After**

image


oponder (approved):
LGTM!

axbarsan:
Changing this to `dev-change`, since it isn't visible to customers. It's only visible to GS staff.

PR #1976, opened 2020-11-17
Create utility class for parsing semver-formatted versions

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/14345

This will come in use while dealing with different types of pre-release versions.

### DIFF
--- a/src/lib/Version.ts
+++ b/src/lib/Version.ts
@@ -105,11 +105,13 @@ export class VersionImpl implements IVersion {
 
   protected parseFromString(from: string): void {
     const groups = semverRegexp.exec(from)?.groups;
-    if (!groups) return;
+    if (!groups || !groups.major || !groups.minor || !groups.patch) {
+      throw new Error('Invalid version provided.');
+    }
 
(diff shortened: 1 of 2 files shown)


src/lib/Version.ts
```diff
@@ -0,0 +1,116 @@
...
+  protected preRelease = '';
+  protected metadata = '';
+
+  protected parseFromString(from: string): void {
+    const groups = semverRegexp.exec(from)?.groups;
+    if (!groups) return;
+
+    this.major = groups.major ?? '0';
+    this.minor = groups.minor ?? '0';
+    this.patch = groups.patch ?? '0';
```
ubergesundheit, line 112: Shouldn't this fail to parse a version string missing the `major`, `minor` or `patch` number? Is this supported from the semver spec? If yes do we want to allow such versions?

axbarsan: You're right, it should throw an error if the parsing fails or if any of the 3 groups is missing. I've updated the implementation and the tests.

ubergesundheit (approved):
Lookin good!

PR #1977, opened 2020-11-17
Improve upgrade UI interaction during Azure v4 to v5 cluster upgrade

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/14122

This PR improves the UI experience on the Azure upgrade from V4 to V5:
* The cluster details UI is automatically changed to the one that includes node pools during the upgrade
* The cluster information in the cluster overview is no longer lost during the upgrade (e.g. the cluster name is no longer set to the cluster ID)

Bonus: Clusters that don't have any conditions are no longer considered in creation.

### DIFF
--- a/src/@types/clusters.d.ts
+++ b/src/@types/clusters.d.ts
@@ -11,6 +11,7 @@ interface IBaseCluster {
   name: string;
   release_version: string;
 
+  path?: string;
   credential_id?: string;
   delete_date?: string | null;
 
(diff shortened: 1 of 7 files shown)


whites11 (approved):
Looks good as far as I can tell.
It tested properly let's do it

fiunchinho (approved):
I don't have the knowledge to review this properly, so I believe it's your call :smile:

tuommaki (approved):
LGTMIIWFY

PR #1982, opened 2020-11-19
Added support for selecting no availability zones for azure masters

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/12101 and https://github.com/giantswarm/giantswarm/issues/13842

This PR has 2 goals:

1. Remove the usage of the deprecated `masters` field in the v5/cluster/create API call and replace it with the new `master_nodes` field.
2. Allow more fine grained control over the availability zone that master nodes should be put in for an azure cluster (the same we already do for node pools).

The AZ selection before my change was:

Before

My proposed change makes it look like this:

after
after2
after3

I tested this as best as I could but I please bear in mind I am not fluent with TypeScript

### DIFF
--- a/src/components/Cluster/NewCluster/CreateNodePoolsCluster.js
+++ b/src/components/Cluster/NewCluster/CreateNodePoolsCluster.js
@@ -145,14 +145,29 @@ class CreateNodePoolsCluster extends Component {
       release_version: this.props.selectedRelease,
     };
 
-    if (this.state.masterAZMode === MASTER_AZ_MODE_MANUAL) {
-      createPayload.master = {
-        availability_zone: this.state.availabilityZonesLabels.zonesArray[0],
-      };
-    } else {
-      createPayload.master = {
-        availability_zone: this.state.masterAZMode,
-      };
+    switch (this.state.masterAZMode) {
+      case MASTER_AZ_MODE_MANUAL:
+        createPayload.master_nodes = {
+          availability_zones: this.state.availabilityZonesLabels.zonesArray,
+          azure: {
+            availability_zones_unspecified: false
+          }
+        };
+        break;
+      case MASTER_AZ_MODE_AUTO:
+        createPayload.master_nodes = {
+          azure: {
+            availability_zones_unspecified: false
+          }
+        };
+        break;
+      case MASTER_AZ_MODE_NOT_SPECIFIED:
+        createPayload.master_nodes = {
+          azure: {
+            availability_zones_unspecified: true
+          }
+        };
+        break;
     }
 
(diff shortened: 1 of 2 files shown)


oponder (commented):
This is looking good, thanks for driving this and making it happen!

I could use still some 'reason' for why a user would want to pick automatic vs not specified in the ui still though. Is that easy enough to explain in the helper text for "automatic" or "not specified"?

whites11:
> This is looking good, thanks for driving this and making it happen!
> 
> I could use still some 'reason' for why a user would want to pick automatic vs not specified in the ui still though. Is that easy enough to explain in the helper text for "automatic" or "not specified"?

This was explained quite well by Tobiasz here: https://github.com/giantswarm/giantswarm/issues/13842#issue-724509125

```
As a user i want to be able to set the Masters AZ to 'None' or 'Not Specified' same as we do for workers now.

This will ensure no costs for cross-zones data transfer if workers are set to None as well.
```

Furthermore, some regions in Azure do not support availability zones at all so "not specified" is actually the only option possible.

src/components/Cluster/NewCluster/CreateNodePoolsCluster.js
```diff
@@ -234,49 +262,82 @@ class CreateNodePoolsCluster extends Component {
...
+                      )}
+                      {zonesArray.length > maxAZ && (
+                        <p className='danger'>
+                          {maxAZ} is the maximum you can have. Please uncheck{' '}
+                          {zonesArray.length - maxAZ} of them.
+                        </p>
+                      )}
+                    </AZWrapperDiv>
+                  )}
+                  <RUMActionTarget
```
axbarsan, line 317: I guess the new option only makes sense on Azure, right?

Can we only render it when the provider is `azure`, similar to what we in the node pool creation form?
https://github.com/giantswarm/happa/blob/c1db97f7b6da7d017da430e69c6d5e829fedcbc9/src/components/Cluster/ClusterDetail/AddNodePool/AddNodePoolAZSelection.tsx#L203

oponder:
> Not sure if it's worth adding more info to this already complicated form. I am afraid about information overloading. Customer can always ask us for help if they don't understand the difference and we can iterate later if we notice confusion about this subject WDYT?

Ok it's not a blocker for me, though I do think we should add maybe a more details link or something like that. but that can come later

axbarsan (approved):
LGTM after fixing the CI stuff 😬

src/components/Cluster/NewCluster/CreateNodePoolsCluster.js
whites11: done

PR #1985, opened 2020-11-19
Add support for testing beta versions

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/14345

As the spec in the issue above states, regular users can now upgrade a cluster to one of the supported versions, right from the `Release details` modal.
Also, the modal suffered a few small cosmetic changes. Try to spot the differences:

<details>
<summary>Release details modal before this PR</summary>

image

</details>

<details>
<summary>Release details modal with no upgrade available (there's no scroll bar)</summary>

image

</details>

<details>
<summary>Release details modal with single production version available (scrolled to bottom of modal)</summary>

image

</details>

<details>
<summary>Release details modal with single beta version available (scrolled to bottom of modal)</summary>

image

</details>

<details>
<summary>Release details modal with multiple production versions available (scrolled to bottom of modal)</summary>

image

</details>

<details>
<summary>Release details modal with multiple beta versions available (scrolled to bottom of modal)</summary>

image

</details>

After clicking on any of the upgrade version numbers, the `Release details` modal will close, and the `Upgrade cluster` modal will open, with the version that the user clicked on pre-selected.

> ⚠️ Admins (GS Staff) also see inactive releases.


src/lib/__tests__/ReleaseHelper.ts
```diff
@@ -0,0 +1,153 @@
...
+    ${'9.2.4-beta3'}   | ${'aws'}   | ${true}           | ${false} | ${'9.3.0'}         | ${true}
+    ${'10.1.0'}        | ${'aws'}   | ${true}           | ${false} | ${'10.0.23'}       | ${false}
+    ${'9.3.0'}         | ${'aws'}   | ${true}           | ${false} | ${'10.0.23'}       | ${false}
+    ${'9.3.0'}         | ${'aws'}   | ${true}           | ${true}  | ${'10.0.23'}       | ${false}
+    ${'9.3.0'}         | ${'azure'} | ${true}           | ${false} | ${'10.0.23'}       | ${true}
+    ${'9.3.0'}         | ${'azure'} | ${true}           | ${true}  | ${'10.0.23'}       | ${true}
+    ${'8.0.0'}         | ${'azure'} | ${true}           | ${false} | ${'10.0.23'}       | ${false}
+    ${'8.0.0'}         | ${'azure'} | ${true}           | ${true}  | ${'10.0.23'}       | ${false}
+    ${'9.1.0'}         | ${'azure'} | ${true}           | ${false} | ${'9.3.0'}         | ${false}
+    ${'9.1.0'}         | ${'azure'} | ${true}           | ${true}  | ${'9.3.0'}         | ${false}
```
axbarsan, line 44: Today I learned...

src/lib/__tests__/ReleaseHelper.ts
```diff
@@ -0,0 +1,153 @@
+import { ReleaseHelper } from 'lib/ReleaseHelper';
+
+describe('ReleaseHelper', () => {
```
axbarsan, line 3: For validating if the release logic is correct, please look through these tests.

src/lib/ReleaseHelper.ts
```diff
@@ -0,0 +1,182 @@
...
+
+export interface ReleaseHelperConfig {
+  currentReleaseVersion: string;
+  availableReleases: IReleases;
+  provider: PropertiesOf<typeof Providers>;
+  ignorePreReleases?: boolean;
+  isAdmin?: boolean;
+}
+
+export class ReleaseHelper {
```
axbarsan, line 14: This class now contains all the release-specific logic (e.g. which versions are not upgradable, which version are skipped etc.). In a separate PR, I will migrate all the existing version checks to use this class.

marians:
Looking good! I would

- add the `v` prefix to version numbers
- Reduce slightly the vertical distance for line items in the list of target versions to upgrade to
- Reduce slightly the line height of the BETA foot note

axbarsan:
> Looking good! I would
> 
> * add the `v` prefix to version numbers

We should probably make this more consistent at some point... right now we don't use `v`s in some places, but we do in others.

> * Reduce slightly the vertical distance for line items in the list of target versions to upgrade to
> * Reduce slightly the line height of the BETA foot note

axbarsan:
I updated the screenshots, @marians . Please take another look when you can 👍🏻

src/components/Modals/ReleaseDetailsModal/ReleaseDetailsModalUpgradeOptions.tsx
```diff
@@ -0,0 +1,139 @@
...
+const VersionWrapper = styled.div`
+  & + & {
+    margin-top: ${({ theme }) => theme.spacingPx * 2}px;
+  }
+`;
+
+const StyledBetaDisclaimer = styled(
+  ReleaseDetailsModalUpgradeOptionsBetaDisclaimer
+)`
+  margin-top: ${({ theme }) => theme.spacingPx * 4}px;
```
whites11, line 21: Is this something we do? Put CSS in the typescript files?

whites11: Oh it's a dynamic generated rule, now I see

axbarsan: We actually use CSS-in-JS (aka JSS) for styling, so all of our styles are in the component files 😬 
We use [this module](https://github.com/emotion-js/emotion) for that.

This makes stuff like switching between different color themes really easy

src/components/Modals/ReleaseDetailsModal/ReleaseDetailsModalUpgradeOptionsVersion.tsx
```diff
@@ -0,0 +1,50 @@
+import styled from '@emotion/styled';
+import PropTypes from 'prop-types';
+import React from 'react';
+
+import ReleaseDetailsModalUpgradeOptionsBetaLabel from './ReleaseDetailsModalUpgradeOptionsBetaLabel';
```
oponder, line 5: Do we need to repeat the folder path in these components? Could they be just "BetaLabel.tsx" and "Version.tsx"?

marians (approved):

src/lib/__tests__/ReleaseHelper.ts
```diff
@@ -0,0 +1,156 @@
...
+    ${'9.0.0'}         | ${'aws'}   | ${true}           | ${true}  | ${'9.0.1'}
+    ${'9.1.0'}         | ${'aws'}   | ${false}          | ${false} | ${'9.1.5'}
+    ${'9.1.0'}         | ${'aws'}   | ${false}          | ${true}  | ${'9.1.5'}
+    ${'9.1.3-alpha.1'} | ${'aws'}   | ${true}           | ${false} | ${null}
+    ${'9.1.3-alpha.1'} | ${'aws'}   | ${true}           | ${true}  | ${null}
+    ${'9.2.3'}         | ${'aws'}   | ${true}           | ${false} | ${'9.3.0'}
+    ${'9.2.3'}         | ${'aws'}   | ${false}          | ${false} | ${'9.2.4-beta3'}
+    ${'9.2.4-beta3'}   | ${'aws'}   | ${true}           | ${false} | ${'9.3.0'}
+    ${'9.2.4-beta3'}   | ${'aws'}   | ${false}          | ${false} | ${'9.2.4-beta4'}
+    ${'9.2.4-beta3'}   | ${'aws'}   | ${true}           | ${false} | ${'9.3.0'}
```
oponder, line 79: I think this is the same test case as on line 77

src/lib/__tests__/ReleaseHelper.ts
```diff
@@ -0,0 +1,156 @@
...
+    ${'9.0.0'}         | ${'aws'}   | ${true}           | ${false} | ${['9.0.3']}
+    ${'9.1.0'}         | ${'aws'}   | ${false}          | ${false} | ${['9.1.5']}
+    ${'9.1.0'}         | ${'aws'}   | ${false}          | ${true}  | ${['9.1.5']}
+    ${'9.1.3-alpha.1'} | ${'aws'}   | ${true}           | ${false} | ${[]}
+    ${'9.1.3-alpha.1'} | ${'aws'}   | ${true}           | ${true}  | ${[]}
+    ${'9.2.3'}         | ${'aws'}   | ${true}           | ${false} | ${['9.3.0']}
+    ${'9.2.3'}         | ${'aws'}   | ${false}          | ${false} | ${['9.2.4-beta3', '9.2.4-beta4']}
+    ${'9.2.4-beta3'}   | ${'aws'}   | ${true}           | ${false} | ${['9.3.0']}
+    ${'9.2.4-beta3'}   | ${'aws'}   | ${false}          | ${false} | ${['9.2.4-beta4']}
+    ${'9.2.4-beta3'}   | ${'aws'}   | ${true}           | ${false} | ${['9.3.0']}
```
oponder, line 117: Same as on line 115

oponder (changes requested):
Notice a few small things.

src/components/Modals/ReleaseDetailsModal/ReleaseDetailsModalUpgradeOptionsVersion.tsx
axbarsan: I know that it's a bit long, but the problem with having a name similar to `Version` for the component and the file is that those names will be used by multiple components. And when you try to use one of these components, 90% of the time TypeScript will automatically import the wrong one 😅

oponder (approved):
LGTM! :shipit: 

(The release logic is a bit too dense for me to verify, but the tests seem thorough 👍 )

whites11 (approved):
I cannot say much about the code (but you have great reviewers already) but I tested it and I can say I like it very much. Thanks @axbarsan

PR #1987, opened 2020-11-23
Fix cluster not being selected for app installation

axbarsan:
As seen in https://github.com/orgs/giantswarm/projects/85#card-49438733

Nothing too special, just me getting punched by JS.
With this occasion, I also changed the type of the selected cluster ID in the global store.

### DIFF
--- a/src/components/AppCatalog/AppDetail/AppDetail.js
+++ b/src/components/AppCatalog/AppDetail/AppDetail.js
@@ -218,7 +218,7 @@ function mapStateToProps(state, ownProps) {
     !selectedCluster ||
     isClusterCreating(selectedCluster) ||
     isClusterUpdating(selectedCluster) ||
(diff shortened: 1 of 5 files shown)


src/components/AppCatalog/AppDetail/AppDetail.js
```diff
@@ -218,7 +218,7 @@ function mapStateToProps(state, ownProps) {
     !selectedCluster ||
     isClusterCreating(selectedCluster) ||
     isClusterUpdating(selectedCluster) ||
-    selectIsClusterAwaitingUpgrade(selectedClusterID)
+    selectIsClusterAwaitingUpgrade(selectedClusterID)(state)
```
axbarsan, line 221: 😅

oponder (approved):
LGTMIIWFY! Thanks for fixing this!

PR #1988, opened 2020-11-24
Align release skipping logic with docs

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/14310

This PR:
* Aligns release skipping logic with docs -> we can now skip an infinite number of minors, but we can not skip majors.
* Fix a bug that would allow upgrades from V4 to V5 on AWS
* Remove the version changing capabilities in the upgrade modal (previously available for admins), since we can now do the same from the version details modal.


marians:
> we can only skip a single major.

In my books, we must not skip any major version. If the cluster is on v12, it can only go to v13 next, not v14. Is it a misunderstanding?

Great to get the admin-only widget removed! Hadn't thought of that.

axbarsan:
> > we can only skip a single major.
> 
> In my books, we must not skip any major version. If the cluster is on v12, it can only go to v13 next, not v14. Is it a misunderstanding?

That was just bad wording from my side... I had something in my head about skipping to the next major version. I updated the description and the comment in the code.

marians (approved):
Approving based on conversation, not based on testing or code review.

PR #1989, opened 2020-11-24
Fix nginx limitation removal

oponder:
Hi @pipo02mix , here's what would need to change to make sure the useEffect block isn't being called all the time, and thus resetting the state of namespace back to `kube-system`

### DIFF
--- a/src/components/AppCatalog/AppDetail/InstallAppModal.js
+++ b/src/components/AppCatalog/AppDetail/InstallAppModal.js
@@ -4,7 +4,7 @@ import yaml from 'js-yaml';
 import RoutePath from 'lib/routePath';
 import lunr from 'lunr';
 import PropTypes from 'prop-types';
-import React, { useState } from 'react';
+import React, { useCallback, useState } from 'react';
 import { connect } from 'react-redux';
 import { OrganizationsRoutes } from 'shared/constants/routes';
 import { installApp } from 'stores/appcatalog/actions';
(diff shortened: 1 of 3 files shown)


src/components/AppCatalog/AppDetail/__tests__/InstallAppForm.js
```diff
@@ -5,21 +5,24 @@ import { renderWithTheme } from 'testUtils/renderUtils';
 import InstallAppForm from '../InstallAppForm';
 
 it('renders without crashing', () => {
-  renderWithTheme(InstallAppForm);
+  renderWithTheme(InstallAppForm, { version: '1' });
```
oponder, line 8: adding version here fixes a console error about children being undefined.

src/components/AppCatalog/AppDetail/InstallAppModal.js
```diff
@@ -104,10 +104,13 @@ const InstallAppModal = (props) => {
       });
   }
 
-  const updateNamespace = (ns) => {
-    setNamespace(ns);
-    setNamespaceError(validateAppName(ns).message);
-  };
+  const updateNamespace = useCallback(
```
oponder, line 107: This is the core of the fix. It makes sure `updateNamespace` does not get 'remade' every render, which causes the `useEffect` in the form to also fire again.

pipo02mix: ok so if I understand well `useCallback` you pass a function and an array of functions/variables from this context. If the argument of the function does not change the body of the function is not called, right?

pipo02mix (approved):

PR #1991, opened 2020-11-27
Hide availability zone options when location does not support AZs

whites11:
When creating node pools in regions that do not support AZs (on azure) there is only one option for the availability zone choice: "Not Specified".
This PR hides the other options in this scenario.

### DIFF
--- a/src/components/Cluster/ClusterDetail/AddNodePool/AddNodePoolAZSelection.tsx
+++ b/src/components/Cluster/ClusterDetail/AddNodePool/AddNodePoolAZSelection.tsx
@@ -135,70 +135,74 @@ const AddNodePoolAZSelection: React.FC<IAddNodePoolAZSelectionProps> = ({
   return (
     <div {...rest}>
       <AZSelectionLabel>Availability Zones selection</AZSelectionLabel>
-      <StyledPanel
-        expanded={value === AvailabilityZoneSelection.Automatic}
-        onToggle={onToggleFakeCallback}
-      >
-        <AddNodePoolAZSelectionCheckbox
-          onChange={onChange}
-          value={value}
-          npID={npID}
-          label='Automatic'
-          type={AvailabilityZoneSelection.Automatic}
-        />
-        <StyledPanelCollapse>
-          <AZSelectorWrapper>
-            <p>Number of availability zones to use:</p>
-            <AvailabilityZonesParser
-              min={minNumOfZones}
-              max={maxNumOfZones}
-              defaultValue={defaultNumOfZones}
-              zones={allZones}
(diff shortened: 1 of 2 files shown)


src/components/Cluster/ClusterDetail/AddNodePool/AddNodePoolAZSelection.tsx
```diff
@@ -135,70 +135,74 @@ const AddNodePoolAZSelection: React.FC<IAddNodePoolAZSelectionProps> = ({
...
-          type={AvailabilityZoneSelection.Manual}
-        />
-        <StyledPanelCollapse>
-          <p>
-            You can select up to {maxNumOfZones} availability zones to make use
-            of.
-          </p>
-          <AZSelectorWrapper>
-            <ManualAZSelector>
+      {maxNumOfZones !== undefined && maxNumOfZones > 0 && (
```
axbarsan, line 138: We have a default value set up for this
https://github.com/giantswarm/happa/blob/53c22fd108c74cf2f1aa3fef35821c6f03908bc2/src/components/Cluster/ClusterDetail/AddNodePool/AddNodePoolAZSelection.tsx#L241

So when the `maxNumOfZones` prop is not assigned any value, it takes the default value declared above. This means that the value is never `undefined`. However, we have to define it as optional in the props type definition in order to not have to specify it when using the component.

Long story short, we don't have to check for `undefined` here, and it's safe to use type casting.

```suggestion
      {(maxNumOfZones as number) > 0 && (
```

src/components/Cluster/ClusterDetail/AddNodePool/AddNodePoolAZSelection.tsx
```diff
@@ -135,70 +135,74 @@ const AddNodePoolAZSelection: React.FC<IAddNodePoolAZSelectionProps> = ({
...
-            )}
-          </AZSelectorWrapper>
-        </StyledPanelCollapse>
-      </StyledPanel>
+            </AZSelectorWrapper>
+            <p>{automaticAZSelectionMessage}</p>
+          </StyledPanelCollapse>
+        </StyledPanel>
+      )}
+      {maxNumOfZones !== undefined && maxNumOfZones > 0 && (
```
axbarsan, line 168:
```suggestion
      {(maxNumOfZones as number) > 0 && (
```

axbarsan (commented):
Looks good in general, I'll do some testing and follow up

axbarsan (approved):
Works like a charm 👍🏻 Looks like we have a new FE on the team 😄

whites11:
> Works like a charm 👍🏻 Looks like we have a new FE on the team

Thanks for testing it

PR #1992, opened 2020-12-01
Fix a crash when labels is undefined.

oponder:
Stops happa from crashing when cluster details cannot be fetched for whatever reason.

Fixes exception report:
```
TypeError: Cannot convert undefined or null to object
 at keys (webpack:///./src/components/Cluster/ClusterDetail/ClusterLabels/ClusterLabels.tsx:73:26)
```

### DIFF
--- a/src/components/Cluster/ClusterDetail/ClusterLabels/ClusterLabels.tsx
+++ b/src/components/Cluster/ClusterDetail/ClusterLabels/ClusterLabels.tsx
@@ -70,7 +70,7 @@ const ClusterLabels: FC<IClusterLabelsProps> = ({
 }) => {
   const [allowEditing, setAllowEditing] = useState(true);
 
-  const noLabels = Object.keys(labels).length === 0;
+  const noLabels = !labels || Object.keys(labels).length === 0;
 
   const dispatch = useDispatch();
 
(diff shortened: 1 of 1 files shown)


axbarsan:
Can you also make them optional in the props interface/prop-types? They are optional in the cluster interface, as well.
https://github.com/giantswarm/happa/blob/470817d5d2e87b8928d650470dfbb3180f062cba/src/%40types/v5clusters.d.ts#L26

oponder:
I actually kinda want labels to always be there, even if it is just the empty object, by the time we're trying to render this component.. let me see if I can move the clean up somewhere earlier in the chain.

ubergesundheit (approved):

oponder:
> I actually kinda want labels to always be there

I'll save that for another day.

src/components/Cluster/ClusterDetail/ClusterLabels/ClusterLabels.tsx
```diff
@@ -98,28 +98,29 @@ const ClusterLabels: FC<IClusterLabelsProps> = ({
...
-                    save({ key: label, value: null });
-                  }}
-                  role='button'
-                  aria-label={`Delete '${label}' label`}
-                >
-                  &times;
-                </DeleteLabelButton>
-              </LabelWrapper>
-            ))}
+            {labels &&
```
oponder, line 101: Narrows the type of `labels` and allows `Object.entries(labels).map` to pass the typechecker.

axbarsan (approved):

PR #1993, opened 2020-12-02
Support readme urls in annotations as well as sources of an app's Chart.yml.

oponder:
We've changed the location of the app README url to the 'annotations' block in Chart.yaml instead of the 'sources' array.

Happa should be backwards compatible for a while though. This PR lets Happa also check for README's in the "application.giantswarm.io/readme" annotation, with a fallback to the original 'sources' location.

https://github.com/giantswarm/giantswarm/pull/13982

fyi @cokiengchiara

### DIFF
--- a/src/stores/appcatalog/selectors.ts
+++ b/src/stores/appcatalog/selectors.ts
@@ -28,5 +28,8 @@ export function selectIngressAppFromCluster(cluster: Cluster) {
 export function selectReadmeURL(
   appVersion: IAppCatalogApp
 ): string | undefined {
-  return appVersion.sources?.find((url) => url.endsWith(Constants.README_FILE));
+  return (
+    appVersion.annotations?.['application.giantswarm.io/readme'] ||
+    appVersion.sources?.find((url) => url.endsWith(Constants.README_FILE))
+  );
 }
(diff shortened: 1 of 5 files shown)


src/stores/appcatalog/selectors.ts
```diff
@@ -28,5 +28,8 @@ export function selectIngressAppFromCluster(cluster: Cluster) {
 export function selectReadmeURL(
   appVersion: IAppCatalogApp
 ): string | undefined {
-  return appVersion.sources?.find((url) => url.endsWith(Constants.README_FILE));
+  return (
+    appVersion.annotations?.['application.giantswarm.io/readme'] ||
+    appVersion.sources?.find((url) => url.endsWith(Constants.README_FILE))
```
axbarsan, line 33: Neither one of `annotations` or `sources` are optional in the interface.
https://github.com/giantswarm/happa/blob/f58409e8798101eb8d712d43f2da71ad2914e9cd/src/%40types/appcatalogs.d.ts#L18

Should they be?

oponder: Yup! They could be there or not there depending on the app. Our own apps should always have it, but it's still something that we could forget on accident. Fixing 👍

axbarsan (approved):

PR #1996, opened 2020-12-04
Unify AZ selection styling

axbarsan:
This PR refactors the Node pools `Availability Zones selection` section, so it could be used for both Master nodes and Node pools. The reason for that is so we could have a consistent UX between both sections.

So now, the look and feel is the same, but the descriptions are slightly different.

### Preview

<details>
<summary>Automatic master AZs</summary>

image

</details>

<details>
<summary>Manual master AZs</summary>

image

</details>

<details>
<summary>Not specified master AZs</summary>

image

</details>


whites11 (commented):
Sounds like a great idea.
I don't feel like reviewing the code, though.
Will leave that to sig ux.

Thanks for this Adrian

.eslintrc.js
```diff
@@ -232,6 +232,7 @@ module.exports = {
     '@typescript-eslint/ban-types': 0,
     '@typescript-eslint/restrict-template-expressions': 0,
     '@typescript-eslint/no-shadow': 2,
+    '@typescript-eslint/no-non-null-assertion': 0,
```
axbarsan, line 235: Finally giving in... This greatly simplifies the usage of props that have default values, by using `someOptionalProp!.length` instead of `(someOptionalProp as string[]).length`

oponder (approved):
LGTM! 👍

PR #2000, opened 2020-12-09
Rewrite user dropdown using our custom dropdown component

axbarsan:
I noticed that with some of the latest dependency upgrades, the user dropdown stopped working. This is a good occasion to refactor it using our custom dropdown component.

The new implementation mostly consists of copying the existing `OrganizationDropdown` component. The look and feel is almost identical as before.


oponder (approved):
LGTM!

PR #2002, opened 2020-12-10
Fix V5 cluster creation loading flag being stuck after completing the request

axbarsan:
When creating node pools clusters, you get a loading icon near the submit button, and at the end you get redirected to the cluster details page. If you navigate again to the cluster creation page, the loading state will still be active.

The fix: Trying not to bork anything, so just adding another separate loading flag for batched cluster creation.


src/stores/cluster/constants.ts
```diff
@@ -23,6 +23,13 @@ export const CLUSTER_CREATE_SUCCESS = 'CLUSTER_CREATE_SUCCESS';
 export const V5_CLUSTER_CREATE_SUCCESS = 'V5_CLUSTER_CREATE_SUCCESS';
 export const CLUSTER_CREATE_ERROR = 'CLUSTER_CREATE_ERROR';
 
+export const CLUSTER_COMPLETE_CREATION_REQUEST =
```
oponder, line 26: I feel like "COMPLETE" is a newish domain term and overlaps a bit with "batched".

Could we call these actions "BATCHED_CLUSTER_CREATION_*******" ?

Then it's kinda clear what it's about and we don't introduce the new term ("COMPLETE")

oponder (changes requested):

oponder (approved):
LGTM 👍

PR #2003, opened 2020-12-10
Fix cluster stuck on 'Awaiting upgrade' after the upgrade request failed

axbarsan:
It seems that if you try to upgrade a cluster, even if the cluster upgrade request fails, the cluster is still stuck with the `Awaiting upgrade` status. This PR fixes that.


oponder (approved):
LGTMIIWFY

PR #2005, opened 2020-12-15
Rename AppRoutes to MainRoutes

oponder:
Because I will be introducing `AppsRoutes` in another PR towards https://github.com/giantswarm/giantswarm/issues/13652

Having `AppRoutes` and `AppsRoutes` is a bit confusing to me. I'd like to reserve the App name for things specifically in the domain of the App Catalog / Managed Apps world.


axbarsan (approved):
`Other` is a bit vague in my opinion, and maybe not that suitable for the main routes of the application (such as the homepage or other login stuff), but not blocking

oponder:
I could call it "MainRoutes"... that would be ok with me too

PR #2006, opened 2020-12-16
Add support for Azure node pool autoscaling

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/14795

This PR enables node pool autoscaling for release versions newer than (or equal to) `13.1.0`.

It also improves the logic of node pool capabilities (such as autoscaling or spot instances), so they don't have to be unnecessarily calculated if they already have been calculated by parent components.


scripts/webpack/getServiceURL.js
```diff
@@ -36,9 +36,9 @@ const getInstallationInfo = function (installation) {
 const getServiceURL = (installation, service) => {
   const installationInfo = getInstallationInfo(installation);
 
-  const { scheme, host } = installationInfo.services[service];
-  const hostWithoutPort = host.split(':')[0];
-  return scheme + '://' + hostWithoutPort;
+  const { Scheme, Host } = installationInfo.Services[service];
+  const hostWithoutPort = Host.split(':')[0];
+  return Scheme + '://' + hostWithoutPort;
 };
```
axbarsan, line 42: The development API proxy was broken for me with the latest `opsctl`. This fixes it

whites11:
I am confused by this UI:

nps

Why do we have min, max and desired nodes? what does "desired" indicate here?

axbarsan:
> I am confused by this UI:
> 
> nps
> 
> Why do we have min, max and desired nodes? what does "desired" indicate here?

From what I can understand (this was implemented in the same way for AWS):
- `MIN` -> Minimum number of nodes out of the autoscaling range
- `MAX` -> Maximum number of nodes out of the autoscaling range
- `DESIRED` -> The number of nodes that the autoscaler decided to use (within the min-max range)
- `CURRENT` -> The number of nodes that are currently ready (will be wrapped in a yellow box if the value is not the same as `DESIRED` yet)

whites11:
Alright, LGTM then

whites11 (approved):

whites11 (approved):

PR #2007, opened 2020-12-18
Refresh the app catalog - Browse apps without having to select a catalog first

oponder:
This is a feature branch that will collect merges from other PR's to try and keep the other PR's small.

- [x] Add routes for /apps and render a placeholder #2008
- [x] Add new app component and app listing area (#2041)
- [x] Add a search query input, sorting dropdown, and an area for facets (#2075)
- [x] Create the app detail page (#2083)
- [x] Enable searching (#2137)
- [x] Wire up the data to the app listing page (#2123)
- [x] Wire up the data to the app detail page (#2133)
- [x] Enable sorting (#2138)
- [x] Reset search button (#2139)
- [x] Create a version picker component on the app detail page (#2147)
- [x] Make version of catalog logos with transparent backgrounds (#2139)
- [x] Create fallback icons for catalogs too or ensure all catalogs have some icon (#2139)
- [x] Add the managed badges & Catalog tooltip showing the description (#2148)
- [x] make sort dropdown actually work (#2151)
- [x] Show app icons on apps list too. Fix look and feel, padding, etc. (#2153)
- [ ] ~keep scroll position on back~
- [ ] ~virtualised list or pagination?~

Towards https://github.com/giantswarm/giantswarm/issues/13652

PR #2008, opened 2020-12-18
Add routes for /apps and render a placeholder

oponder:
This sets up the `/apps` routes and adds `Apps` to the navigation.

I keep the `/app-catalog` routes in this PR and will remove them all and the old components in a separate PR, probably the last one.

This gets merged into a feature collection branch here: https://github.com/giantswarm/happa/pull/2007
Towards: giantswarm/giantswarm#13652


src/components/Apps/Apps.js
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import { Breadcrumb } from 'react-breadcrumbs';
+import { connect } from 'react-redux';
+import { Route, Switch } from 'react-router-dom';
+import { AppsRoutes } from 'shared/constants/routes';
+
+class Apps extends React.Component {
```
axbarsan, line 7: Not sure what your plans are for this file, but does it have to be JavaScript?

oponder: We can make it typescript, but it'd be the first of these routing components I think.

Wasn't sure if I should break the mold already.

axbarsan: Let's do it! 👍🏻

src/components/Apps/Apps.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import { Breadcrumb } from 'react-breadcrumbs';
+import { Route, Switch } from 'react-router-dom';
+import { AppsRoutes } from 'shared/constants/routes';
+
+const Apps: React.FC = () => {
+  // eslint-disable-next-line class-methods-use-this
```
axbarsan:
```suggestion
```

axbarsan (approved):
LGTM

PR #2014, opened 2021-01-05
Clarify that the EOL tag is associated with the Kubernetes release

marians:
Closes https://github.com/giantswarm/giantswarm/issues/14987

This PR adds clarification regarding the `EOL`( end of life) marker displayed next to a Kubernetes version. This should help avoid the impression that the Giant Swarm tenant cluster release is EOL.

The wording change makes the function specific to Kubernetes, so I renamed it from `getReleaseEOLStatus` to `getKubernetesReleaseEOLStatus`.

### Before

image

### After

image


oponder (approved):
There's still a test that needs to be adjusted, but other than that LGTM

PR #2015, opened 2021-01-05
Change variables to uppercase

marians:
This updates happa to the latest behaviour of `opsctl show installation`, which prints some variables we need with mixed case names.


oponder (approved):
LGTM

PR #2031, opened 2021-01-13
Improve number picker UX to allow the user enter the value they want

axbarsan:
Reproduce the problem:
- Enter the cluster creation page
- Select a release version that supports node pool autoscaling
- Try to create a new node pool, set the minimum scaling to `3`, and select the text in the maximum input, and try to enter `20`
- Notice that the value is `30` instead of `20`

With this PR, we let the user enter the value they want via keyboard. But don't worry, if the user tries to enter an invalid value, the input will be marked as invalid, and a validation error will be shown. This improves UX by letting the user enter the value they want.

cc @whites11 for extra details about the issue

Affected components:
- Node pool scaling (both autoscaling on or off)
- Node pool availability zone count
- V4 cluster creation availability zone counter

### DIFF
--- a/__tests__/V4AzureClusterManagement.js
+++ b/__tests__/V4AzureClusterManagement.js
@@ -166,14 +166,14 @@ describe('V4AzureClusterManagement', () => {
 
     fireEvent.change(azInput, {
       target: {
-        value: defaultAZCount - 1,
+        value: defaultAZCount,
       },
     });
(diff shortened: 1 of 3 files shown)


pipo02mix (approved):
LGTM

oponder (approved):
LGTMIIWFY

whites11 (approved):
You rock

PR #2036, opened 2021-01-14
Add grommet and storybook

oponder:
This adds the grommet and storybook dependencies and gets it working without changing anything more in Happa.

Also kept the sample storybook components for now so we can all take a look at how it works.

You can run storybook with `yarn run storybook`

Out of this I got a couple of todos which I might turn into tickets:

- Hook up the theme through grommet instead of with ThemeProvider?
- Remove sample storybook components, write our own little intro


package.json
```diff
@@ -202,7 +209,7 @@
     "express": "^4.16.3",
     "file-loader": "6.2.0",
     "fstream": "1.0.12",
-    "html-webpack-plugin": "4.5.1",
+    "html-webpack-plugin": "^5.0.0-beta.5",
```
oponder, line 212: Storybook wouldn't start without using this beta version of `html-webpack-plugin`

src/components/App.tsx
```diff
@@ -9,9 +10,20 @@ import { Provider } from 'react-redux';
 import { Store } from 'redux';
 import FeatureFlags from 'shared/FeatureFlags';
 import { DefaultTheme, ThemeProvider } from 'styled-components';
+import styled from 'styled-components';
 
 import Routes from './Routes';
 
+const StyledGrommet = styled(Grommet)`
```
oponder, line 17: These are the styles that used to be on the <div id="app">

axbarsan (approved):

PR #2037, opened 2021-01-14
Show correct node pool size on Azure autoscaling clusters

axbarsan:
Fixes https://github.com/giantswarm/giantswarm/issues/15187

Really don't know what I had in mind when I wrote this


oponder (approved):
LGTM I think? Do we actually know what the autoscaler desires? this seems to be just the number of nodes and doesn't give the insight that the column "should" give.. but I do grant that it'll be closer than `scaling.max` at least

axbarsan:
Well that's what the spec says anyway 😬 
https://docs.giantswarm.io/api/#operation/getNodePool

PR #2041, opened 2021-01-15
Add new app component and app listing area

oponder:
This PR is focused mainly on the styling for the app component, and the fallback outlined letter icon style.

Towards: https://github.com/giantswarm/giantswarm/issues/13652

Screenshot 2021-01-15 at 6 24 51 PM

### DIFF
--- a/src/components/Apps/AppsList/AppsList.tsx
+++ b/src/components/Apps/AppsList/AppsList.tsx
@@ -50,6 +50,7 @@ const AppsList: React.FC = () => {
             'prometheus-mongodb-exporter',
             'prometheus-pushgateway',
             'prometheus-postgres-exporter',
+            ...Array(100).fill('other app'),
           ].map((name) => (
             <App
               key={name}
--- a/src/components/Apps/AppsList/AppsList.tsx
+++ b/src/components/Apps/AppsList/AppsList.tsx
@@ -19,7 +19,6 @@ const List = styled.div`
   width: 100%;
   padding-left: 20px;
 `;
-const ONE_HUNDRED = 100;
 
 const AppsList: React.FC = () => {
   return (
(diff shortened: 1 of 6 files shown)


src/components/shared/ImgWithFallback.tsx
```diff
@@ -4,32 +4,58 @@ import React, { FC, ImgHTMLAttributes, useState } from 'react';
...
   backgroundColor: string;
+  textColor: string;
 }
 
 interface IImgWithFallback extends ImgHTMLAttributes<HTMLImageElement> {
   fallback: IFallback;
 }
 
 const ImgWithFallback: FC<IImgWithFallback> = (props) => {
-  const [loadError, setLoadError] = useState(false);
```
oponder, line 14: I flipped the logic of this component a bit to start with a loading state, and to 'prove itself' that it is not in an error instead of waiting for a failure to happen and then hiding the image.

This prevents a flash of the alt text and the broken image icon.

marians (approved):

src/components/Apps/AppsList/AppsList.tsx
```diff
@@ -0,0 +1,68 @@
...
+  gap: 20px;
+  border-left: 1px solid ${({ theme }) => theme.colors.darkBlueLighter1};
+  width: 100%;
+  padding-left: 20px;
+`;
+
+const AppsList: React.FC = () => {
+  return (
+    <div>
+      <h1 data-testid='apps-browser'>Apps</h1>
```
axbarsan, line 26:
```suggestion
      <h1>Apps</h1>
```

yarn.lock
```diff
@@ -21,14 +21,7 @@
   dependencies:
```
axbarsan, line 21: Pushed by accident?

axbarsan (approved):
LGTM

yarn.lock
oponder: Hmm no.. I did `yarn install` a couple of times because I was a bit confused about `emotion` still being there (it's a dependency of storybook), that must've changed some things in yarn.lock, but I think its ok?

PR #2043, opened 2021-01-16
Fix safari token refresh

oponder:
It seems that auth0 introduced a method for silently refreshing tokens even on browsers that do not allow third party cookies  like Safari (and probably Brave). Related: https://github.com/giantswarm/giantswarm/issues/7230

For a while this was simply not possible (unless the user disabled `prevent cross site tracking`), but I recently checked the docs again and they offered an alternative called "Refresh Token Rotation"

Since I was using Safari a bit more these days, I thought I'd give implementing it a go. 

This is really only "a thing" for Giant Swarm staff that for now still log in through Auth0. 

- [ ] There's a broken test which doesn't break when I run it on it's own.. will dig a bit deeper on that later.

https://auth0.com/docs/authorization/renew-tokens-when-using-safari

 https://auth0.com/docs/tokens/refresh-tokens/refresh-token-rotation


oponder:
Closing this, it's pretty specific to me and I don't think the drawbacks are worth it (storing refresh token in the browser, even though it gets rotated automatically)

PR #2047, opened 2021-01-18
Allow passing date objects to date formatting utilities

axbarsan:
This prevents the functions from breaking when passing `Date` objects to them (which occasionally happened in the cluster overview, after switching to `date-fns`)


oponder (approved):
LGTM

PR #2048, opened 2021-01-18
Move cluster capability computation into single place

axbarsan:
This PR moves all the stray capability computation (such as node pool autoscaling, node pool spot instances etc.) into the common capability computation function, so it would be easier to understand from where everything comes from.


oponder (approved):
LGTM 👍

PR #2053, opened 2021-01-19
Fence in the Button, consolidate the styles, replace 'large' buttons

oponder:

This refactoring takes various uses of Buttons in Happa and makes them all an instance of Button.js

See this as a first step towards turning them into a Grommet button

It also consolidates the styles, removes `_button.sass` and tries to get rid of conflicting or confusing class names.

Looking throughout the code base I found 5 button colour stylings: green, blue, red, transparent, transparent with outline
(as well as disabled styling)

here are some examples:
Screenshot 2021-01-19 at 1 16 03 PM
Screenshot 2021-01-19 at 1 15 56 PM
Screenshot 2021-01-19 at 1 15 49 PM
Screenshot 2021-01-19 at 1 15 44 PM
Screenshot 2021-01-19 at 1 16 47 PM

However in the code base we had 7 classNames used for button styling. `success`, `primary`, `secondary`, `danger`, `default`, `outlined`, `link`

`default` and `outlined` were doing the same thing, so I replaced references to `default` with `outlined`
`primary` and `success` were also doing the same thing, so I replaced `success` with `primary`

I found the large buttons feel a bit out of place, so I replaced them with normal sized buttons. For example the buttons in the getting started guide, or the cluster creation form were 'large'.. but it kinda looks goofy.

Before:
Screenshot 2021-01-19 at 3 47 02 AM

After:
Screenshot 2021-01-19 at 3 47 11 AM

The small sizing is used when buttons appear in table rows and was just tidied up a bit in terms of padding and font size:

Before:
Screenshot 2021-01-19 at 1 20 06 PM

After:
Screenshot 2021-01-19 at 1 20 14 PM

I also I tried to make sure we are not styling the `button` element, and instead style the `btn` class. 

These changes should make it so that we can rewrite/refactor Button.js to remove its dependency on Bootstrap and simplify the styles further. When doing that we can also rename the props `bsStyle` and `bsSize` since `bs` stands for `bootstrap`


oponder:
There are a few components that style the `.btn` class. So the abstraction is not fully contained yet.

In another PR I'll tackle that by styling the Button component instead of trying to style the `.btn` class.

marians (approved):
Awesome! LGTM if it works :)

src/components/UI/Button.js
```diff
@@ -1,14 +1,15 @@
...
 
 // Button
 //
 // <Button
 //   type='button|submit|reset'
-//   bsStyle='primary|success|danger|info|warning|link'
+//   bsStyle='primary|secondary|danger|outline|link'
 //   loading=true|false
 //   disabled=true|false
 //   onClick=function>
```
axbarsan, line 15: Can we have a props interface instead of documenting the props here?

axbarsan (commented):
LGTM in general. A very nice improvement. I just have a quick question.

src/components/UI/Button.js
oponder: Yeah for sure. plan to do that in another PR

axbarsan (approved):

PR #2058, opened 2021-01-20
Move CodeBlock & FileBlock to UI Folder.  Set up our first couple of stories.

oponder:

This extracts CodeBlock & FileBlock to UI folder and cheats a bit with the Sass removal by moving all the sass into a `Styles` component and using that as a wrapper.

I don't want to refactor all the classNames into individual styled components just yet on this one. The behaviour of the component is a bit special and I used classNames to toggle the look of things based on various states.

Refactoring this further can remain a TODO in my opinion, since I don't think this component deserves that much more of my time.

Screenshot 2021-01-20 at 10 51 12 PM


axbarsan (approved):
Nice 👍🏻

PR #2064, opened 2021-01-21
Stop displaying the 'Get started' button in the cluster overview for old clusters

axbarsan:
When switching from `moment` to `date-fns`, I got the order for comparison arguments wrong, and all date differences were borked 😬


marians (approved):
Should that be caught in a unit test?

LGTM if it works for you

axbarsan:
> Should that be caught in a unit test?

It is covered already, but the test was using some broken logic that was only working on `moment`.

oponder (approved):
LGTM IIWFY

PR #2068, opened 2021-01-21
Set main nav logo width and height

marians:
Mitigating a tiny recommendation in Lighthouse.

image

As this element uses a data URI in the current implementation, this will not have a noticable effect. But I'd like to get the complaint gone. ;-)


oponder (approved):
LGTM!

PR #2074, opened 2021-01-22
Create currency input for handling money

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/15310

This PR adds a new input for dealing with currencies. It leverages the browser validation for numbers (input=`number`).

This PR also adds some configuration for storybook, to have the same global styles as our production app.

## Preview

<details>
<summary>With good ol' random integer</summary>

image

</details>

<details>
<summary>With float value</summary>

image

</details>

<details>
<summary>With label</summary>

image

</details>


.storybook/preview.js
```diff
@@ -1,3 +1,16 @@
...
+import 'bootstrap/dist/css/bootstrap.min.css';
+import 'noty/lib/noty.css';
+import 'react-datepicker/dist/react-datepicker.css';
+import 'styles/app.sass';
+
 export const parameters = {
   actions: { argTypesRegex: '^on[A-Z].*' },
 };
+
+addDecorator((story) => <ThemeProvider theme={theme}>{story()}</ThemeProvider>);
```
axbarsan, line 16: Wrap all stories with the theme provider, so every component will have the theme context available.

oponder (approved):
LGTM!

src/components/UI/Inputs/CurrencyInput/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+    onChange?.(valueConvertedToPrecision);
+  };
+
+  const visibleValue = formatNumberWithPrecision(value ?? 0, precision!);
+
+  return (
+    <div {...rootProps}>
+      {label && (
+        <LabelText {...labelTextProps} htmlFor={id}>
```
oponder, line 89: This might be a TODO or out of scope for now, but I think the base Input element also has a label ability.. do we need to implement it twice ?

src/components/UI/Inputs/CurrencyInput/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+          {label}
+        </LabelText>
+      )}
+      <InputWrapper>
+        <CurrencyLabel>{currencyLabel}</CurrencyLabel>
```
oponder, line 94: Could we instead add this CurrencyLabel as a feature of the base input component? not sure what to call it, maybe  "InputIcon"

So that we don't have to unset the border of the Input here to only rewrap it again..

and so I can reuse it for search input component 😅

axbarsan:
> This might be a TODO or out of scope for now, but I think the base Input element also has a label ability.. do we need to implement it twice ?

> Could we instead add this CurrencyLabel as a feature of the base input component? not sure what to call it, maybe "InputIcon"

Answering both of these, the `Input` component allows a label and an icon (only FontAwesome icons are supported at the moment), but it doesn't allow the level of customizability needed for this component, therefore this component only uses the `input` element of the `Input` component.

The `Input` component needs to be refactored in order to support this kind of customizability. Since we also need to refactor it when we switch it to `grommet`, I'd rather do that only once, instead of refactoring it now, and refactoring it again later.

marians:
image

Is that comma the decimal separator? If yes, that should be a period in the english locale, no?

axbarsan:
> image
> 
> Is that comma the decimal separator? If yes, that should be a period in the english locale, no?

The number is formatted using your browser's locale. You probably have a German locale, which is why you see a comma.

marians:
I just used your screenshot ;-) Thanks

PR #2075, opened 2021-01-24
Add a search query input, sorting dropdown, and an area for facets

oponder:
This adds the remaining UI elements and controls to the new App listing page.

Not hooked up to any actual data yet. Just the UI elements, and a bit of the presentational logic.

Icons will have to have transparent backgrounds, but used what we have for now.

Screenshot 2021-01-24 at 10 03 59 PM

### DIFF
--- a/src/components/UI/AppList/Icon.tsx
+++ b/src/components/UI/AppList/Icon.tsx
@@ -26,22 +26,16 @@ function acronymize(text: string) {
   return acronym;
 }
 
-interface IStyledImageWithFallbackProps {
-  outlineColor: string;
-}
-
-const StyledImageWithFallback = styled(
-  ImgWithFallback
-)<IStyledImageWithFallbackProps>`
-  font-size: 40px;
-  align-items: center;
-  font-weight: 800;
-  justify-content: center;
-  text-shadow: -1px -1px 0 ${({ outlineColor }) => outlineColor},
-    1px -1px 0 ${({ outlineColor }) => outlineColor},
-    -1px 1px 0 ${({ outlineColor }) => outlineColor},
-    1px 1px 0 ${({ outlineColor }) => outlineColor};
-`;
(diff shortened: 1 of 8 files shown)


src/components/UI/AppList/Icon.tsx
```diff
@@ -26,34 +26,29 @@ function acronymize(text: string) {
...
+  `;
+};
 
 interface IIconProps {
   name: string;
   src: string;
 }
 
 const Icon: React.FC<IIconProps> = ({ name, src }) => {
+  const Image = StyledImageWithFallback(colorHash.calculateColor(name));
```
axbarsan, line 46: Wouldn't this create a new component with every single re-render?

oponder: Hmm It didn't really sit right with me, but I couldn't figure out the right way to pass `outlineColor` in as a prop and avoid an error about outlineColor not being a valid attribute.

axbarsan: The previous way should have worked. That is what we do in other components, as well
https://github.com/giantswarm/happa/blob/273cb78609fb34a4286c7a171d5d0bb2f1294713/src/components/UI/Controls/VersionPicker/VersionPicker.tsx#L106

oponder: Yeah the problem I was having is that the underlying component applies all the props it gets to the `img` tag, which results in: 

```
Warning: React does not recognize the `outlineColor` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `outlinecolor` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
```

but I'll revert and use the lowercase `outlinecolor` instead and be fine with it appearing in the Dom

axbarsan (approved):

PR #2083, opened 2021-01-26
App detail page

oponder:
This adds a UI display component for the app detail page.

Couple things missing which I'll address in subsequent PRs

- The breadcrumbs aren't right yet
- The version selector dropdown is missing
- The catalog label is missing an icon and special treatment if is a managed catalog.

The readme and props being passed in to the AppDetails UI component are placeholders and will be replaced when I wire it up to the datastore.

Screenshot 2021-01-27 at 2 44 29 AM

Screenshot 2021-01-27 at 2 44 40 AM

towards: https://github.com/giantswarm/giantswarm/issues/13652


src/components/UI/Display/Apps/AppDetailsNew/utils.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
```
axbarsan, line 1: Can this also be a TypeScript file?

axbarsan (approved):

src/components/UI/Display/Apps/AppDetailsNew/AppDetails.tsx
```diff
@@ -0,0 +1,247 @@
...
+  appTitle: string;
+  appIconURL: string;
+  catalogName: string;
+  chartVersion: string;
+  createDate: Date;
+  includesVersion: string;
+  description: string;
+  website: string;
+  keywords: string[];
+  readme?: string;
```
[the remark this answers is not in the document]
oponder: I was thinking about this from a general point of view, also possibly something for `fmt`, but two related things:

1. I have a strong preference for 'flat' props, so not passing in a complex object

   I think that leads to more maintainable code in the long run, passing in a complex object means coupling the  component to the shape of that object.
   
2. UI components should be as dumb as possible about Giant Swarm specific domain things.

  So by the time we get to the UI display layer, we shouldn't be entangled in the details of what an App is and what its   structure looks like.  That way, if the data structure of an app changes, we're not digging into UI display code to make it work again, and instead we create a clear boundary between presentation and business logic.

src/components/UI/Display/Apps/AppDetailsNew/utils.js
oponder: Yes! but can you perhaps help me figure out the type for headingProps in Headingrenderer.

ubergesundheit (approved):
Looks good!

src/components/UI/Display/Apps/AppDetailsNew/utils.js
axbarsan: `headingProps` is supposed to contain props for a `h1(2,3,4,5,6)` element, right? As well as the `level` prop.

How about this?
```typescript
interface IHeadingProps extends React.ComponentPropsWithoutRef<'h1'> {
	level: 1 | 2 | 3 | 4 | 5 | 6;
}
```

PR #2087, opened 2021-01-27
Make app detail utils a typescript file

oponder:
The types are pretty complex for ReactNodes. I had to make sure a `props` property was there, and ended up googling for a way to narrow a type based on a property on the object.. that worked, but I also don't fully understand it.


src/components/UI/Display/Apps/AppDetailsNew/utils.tsx
```diff
@@ -0,0 +1,57 @@
...
+}
+
+// https://github.com/remarkjs/react-markdown/issues/69
+export function HeadingRenderer(headingProps: IHeadingProps) {
+  const children = React.Children.toArray(headingProps.children);
+  const text = children.reduce(flatten, '');
+  const slug = text.toLowerCase().replace(/\W/g, '-');
+
+  console.log(slug);
+
```
ubergesundheit, line 14:
```suggestion

```

axbarsan (approved):

PR #2088, opened 2021-01-27
Fix release selector buttons being cut off

axbarsan:
This probably occured as a result of the button refactor

<details>
<summary>Before</summary>

image

</details>

<details>
<summary>After</summary>

image

</details>


oponder (approved):
Thanks!

PR #2089, opened 2021-01-27
Expose storybook in production app

axbarsan:
This PR adds support for visiting storybook in the deployed application, in the `/storybook` path.


.storybook/preview-head.html
```diff
@@ -1,9 +1,6 @@
...
-href="//fonts.googleapis.com/css?family=Roboto:300,400,700,300italic|Inconsolata:400,700|Roboto+Condensed:400,700"
-type="text/css"
+  rel="stylesheet"
+  href="//fonts.googleapis.com/css?family=Roboto:300,400,700,300italic|Inconsolata:400,700|Roboto+Condensed:400,700"
+  type="text/css"
 />
 <script src="https://use.fortawesome.com/b5c7b73e.js"></script>
-
-<!-- Include modernizr. Requires Happa dev server to be running.-->
-<script type="text/javascript" src="http://localhost:7000/vendor/modernizr.js"></script>
```
axbarsan, line 9: We only use this for the `download` attribute of the `a` element.

I believe we don't actually need it in storybook
https://caniuse.com/download

oponder: https://github.com/giantswarm/happa/blob/72af9395d68fd4889fb22f29ca5050e9e8497242/src/components/UI/Display/Documentation/FileBlock.js#L102

Will need a bit of a change in order for it to work in the storybook then. Like to check if moderniser is there first.

oponder:
Thx for picking this up! Will make showing off changes and communicating about UX things a lot easier <3

.storybook/preview-head.html
```diff
@@ -1,9 +1,12 @@
...
+  type="text/css"
 />
 <script src="https://use.fortawesome.com/b5c7b73e.js"></script>
 
-<!-- Include modernizr. Requires Happa dev server to be running.-->
-<script type="text/javascript" src="http://localhost:7000/vendor/modernizr.js"></script>
+<script>
+  window.Modernizr = {
+    adownload: true,
+  };
```
axbarsan, line 11: I will remove modernizr altogether from happa in a separate PR

helm/happa/templates/ingress.yaml
```diff
@@ -14,7 +14,7 @@ metadata:
     nginx.ingress.kubernetes.io/whitelist-source-range: {{ template "whitelistCIDR" . }}
     {{- end }}
     nginx.ingress.kubernetes.io/configuration-snippet: |
-      add_header X-Frame-Options deny always;
+      add_header X-Frame-Options sameorigin always;
```
axbarsan, line 17: Storybook uses an iframe to load components, and needs to be able to do so

oponder (approved):
LGTM!

oponder:
This got me thinking a bit though about some next steps for this.

Because now we're deploying storybook to every installation, and it doesn't need to be _everywhere_.

Instead it's more like internal documentation at the moment, like maybe: intranet.giantswarm.io/happa-storybook ?

Eventually it could be external documentation as well to encourage non-giantswarm people to contribute and then it'd live at `docs.giantswarm.io/happa-storybook`.

Also finding a way to build PR's and show them as a version in the same storybook would be super duper cool. (https://panosvoudouris.medium.com/versioned-style-guides-with-react-storybook-aed6eebc585d) 

I don't think keeping historical versions of the storybook is that important, but being able to compare a PR to the latest production version would be pretty sweet.

That's a LOT more complicated. And not the priority when it comes to further storybook dev now.. but just wanted to put the idea out there already and see what everyone thinks.

PR #2093, opened 2021-01-28
Modernizr straight to the trash

axbarsan:
We were only using it for the `adownload` property for `a` elements.

That already has pretty good browser support, so we can get rid of it.
https://caniuse.com/download


oponder (approved):
LGTM!

PR #2094, opened 2021-01-28
Extract appslist into ui component. Add empty state. Fix breadcrumb i…

oponder:
This adds the empty state to the app list. 

Also adds it as a story for it to the storybook.

And adds a breadcrumb to the App Detail page.

I also started a convention to try and avoid having the same name for the business logic container component and the UI component. So I called the container component 'AppDetail' and the UI component is called 'AppDetailPage'. 

Im considering creating a storybook category called "Pages" for these larger Page UI components. But for now I just put it in the Display category.

Screenshot 2021-01-29 at 1 27 18 AM


src/components/UI/Display/Apps/AppList/AppsListPage.tsx
```diff
@@ -0,0 +1,108 @@
...
+    margin-bottom: 0px;
+    font-weight: 400;
+  }
+`;
+
+export interface IAppsListPageProps {
+  matchCount: number;
+  onChangeFacets: (value: string, checked: boolean) => void;
+  facetOptions: IFacetOption[];
+  apps: IAppProps[];
```
axbarsan, line 45: Why not use the actual app type? That's what will be coming here from the global state
https://github.com/giantswarm/happa/blob/389e0d4418cc418860968b4a219d8108b8f51742/src/%40types/appcatalogs.d.ts#L18

oponder: I think it's to avoid that tight coupling between the state and the presentational layer. Makes it really obvious to see what properties of an App this component really cares about. 

Quicker to write a storybook about without having to create 'half representations' of an app?

I'm kinda justifying it after the fact, but it feels right to have this really clear boundary between presentation layer and business logic layer.. what do you think?

axbarsan (approved):

PR #2098, opened 2021-01-29
Allow setting node counter values in any order

axbarsan:
This was a problem found by the app destroyer (@whites11).

Let's say you wanted to create a node pool, and you wanted to set up the autoscaling configuration. Let's say you don't want any autoscaling, and you want `2` nodes set to both minimum and maximum.
If you set the `MIN` value first, and the `MAX` value second, everything's perfect. However, if you set the `MAX` value first, and then the `MIN` value, you'll still see the validation error near the `MAX` input, and you wouldn't be able to create your node pool, even though the values are valid (theoretically).

This PR addresses the issue by re-validating the number pickers if their minimum or maximum limits change.

### DIFF
--- a/src/components/UI/Inputs/NumberPicker.js
+++ b/src/components/UI/Inputs/NumberPicker.js
@@ -141,6 +141,14 @@ class NumberPicker extends React.Component {
     validationError: '',
   };
 
+  componentDidUpdate(prevProps) {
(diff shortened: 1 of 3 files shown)


oponder (approved):
LGTMIIWFY

PR #2106, opened 2021-02-02
Catalog facets

oponder:
This wires up the left side of the Apps page with real data, and adds actions and state management for toggling those checkboxes. It had to be in the store because we want to preserve the users chosen catalogs.

The icons not being transparent will be addressed in another PR.

Towards giantswarm/giantswarm#13652

Screenshot 2021-02-02 at 6 18 15 PM


src/components/UI/Display/Apps/AppDetailsNew/utils.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+
+// https://github.com/remarkjs/react-markdown/issues/69
+export function HeadingRenderer(headingProps) {
```
axbarsan, line 4: I have a deja vu 😄 
https://github.com/giantswarm/happa/blob/app-catalog-refresh/src/components/UI/Display/Apps/AppDetailNew/utils.tsx

src/components/UI/Display/Apps/AppList/Icon.tsx
```diff
@@ -0,0 +1,72 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import ImgWithFallback from 'shared/ImgWithFallback';
+import styled from 'styled-components';
+import theme from 'styles/theme';
+import CachingColorHash from 'utils/cachingColorHash';
+
+const colorHash = new CachingColorHash();
+
+function acronymize(text: string) {
```
axbarsan, line 10: Would be awesome to memoize this. Maybe in a future PR

oponder: I did something wrong in this PR :P these files were renamed, but when I did a rebase I committed both versions of them. 😅 (Also applied to:  src/components/UI/Display/Apps/AppDetailsNew/utils.js)

but yeah, that can totally be memoized :+1: will note it down.

will delete these two files though since they should not be getting referenced anywhere. thx for noticing!!

oponder:
sorry about the extra files in this PR originally.. messed up my git flow.

src/components/Apps/AppsList/AppsList.tsx
```diff
@@ -1,17 +1,131 @@
...
+  };
+}
+
+AppsList.propTypes = {
+  catalogs: (PropTypes.object as PropTypes.Requireable<IAppCatalogsState>)
+    .isRequired,
+  isAdmin: PropTypes.bool.isRequired,
+};
+
+export default connect(mapStateToProps)(AppsList);
```
axbarsan, line 131: This should not change functionality, but in regards to consistency to other files: Why not use `useSelector` here, since we're using a functional component?

https://github.com/giantswarm/happa/blob/273cb78609fb34a4286c7a171d5d0bb2f1294713/src/components/Cluster/ClusterDetail/Ingress/InstallIngressButton.tsx#L62

axbarsan (approved):

src/components/Apps/AppsList/AppsList.tsx
oponder: Not used to it yet.. kinda went with what I remembered and saw. But I think it makes sense to use `useSelector` I'll apply that before merging.

src/components/Apps/AppsList/AppsList.tsx
```diff
@@ -1,15 +1,116 @@
...
 import AppsListPage from 'UI/Display/Apps/AppList/AppsListPage';
+import CatalogLabel from 'UI/Display/Apps/AppList/CatalogLabel';
+import { IFacetOption } from 'UI/Inputs/Facets';
+
+interface IStateProps {
+  catalogs: IAppCatalogsState;
+  isAdmin: boolean;
+}
+
+interface IAppsListProps extends IStateProps {}
```
axbarsan, line 16:
```suggestion
interface IAppsListProps {}
```

PR #2110, opened 2021-02-03
Upgrade webpack-cli to latest version

oponder:
There's a bug which prevents us from starting the local dev server in v4.4.0


axbarsan (approved):
Dependoliver 👍🏻

PR #2112, opened 2021-02-03
Show an error and reduce opacity of facet if catalog index failed to …

oponder:
This adds some error visibility to the catalog load index action. If a catalog fails to load, we reduce the opacity of its facet and add a red warning icon with a tooltip explaining something went wrong.

Screenshot 2021-02-03 at 11 48 45 PM

### DIFF
--- a/src/components/UI/Display/Apps/AppList/CatalogLabel.tsx
+++ b/src/components/UI/Display/Apps/AppList/CatalogLabel.tsx
@@ -4,10 +4,11 @@ import OverlayTrigger from 'react-bootstrap/lib/OverlayTrigger';
 import Tooltip from 'react-bootstrap/lib/Tooltip';
 import styled from 'styled-components';
 
-const Wrapper = styled.div`
-  &.hasError {
-    opacity: 0.4;
-  }
+const LOW_OPACITY = 0.4;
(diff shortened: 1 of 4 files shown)


src/components/UI/Display/Apps/AppList/CatalogLabel.tsx
```diff
@@ -19,18 +25,44 @@ const CatalogType = styled.span`
...
+};
+
+ErrorIcon.propTypes = {
+  name: PropTypes.string.isRequired,
+};
+
 const CatalogLabel: React.FC<ICatalogLabelProps> = (props) => {
   return (
-    <Wrapper {...props}>
+    <Wrapper {...props} className={props.error ? 'hasError' : ''}>
```
axbarsan, line 61: This can be done easier with a custom prop. If `className` would be left alone, we would be able to style this component from other components

axbarsan (approved):

src/components/UI/Display/Apps/AppList/CatalogLabel.tsx
oponder: Fixed

axbarsan (approved):

PR #2122, opened 2021-02-05
Upgrade to yarn v2

axbarsan:


axbarsan:
No-no. Snyk doesn't support Yarn 2, even 1 year after its release date 😄

PR #2123, opened 2021-02-05
Wire apps into the applist

oponder:
Trying to keep PR's small, this one focuses on the selector used to get all the apps that belong to the catalogs that the user selected in the facets.

Screenshot 2021-02-05 at 10 06 40 PM

I renamed `IAppCatalogApp` to `IAppCatalogAppVersion`, and created a new `IAppCatalogApp` interface.

What used to be `IAppCatalogApp` was actually just a single version of an app. Catalogs contain all known versions of an app, so I created a new interface which kind of wraps all the known versions and surfaces the name and other details relevant for rendering in the AppList component.

### DIFF
--- a/src/stores/appcatalog/selectors.ts
+++ b/src/stores/appcatalog/selectors.ts
@@ -1,3 +1,4 @@
+import { createSelector } from 'reselect';
 import { Constants } from 'shared/constants';
 import { IState } from 'stores/state';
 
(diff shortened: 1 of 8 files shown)


src/stores/appcatalog/selectors.ts
```diff
@@ -13,9 +14,40 @@ export function selectCatalogs(state: IState): IAppCatalogsState {
   return state.entities.catalogs;
 }
 
+export function selectSelectedCatalogs(state: IState): string[] {
+  return Object.entries(state.entities.catalogs.ui.selectedCatalogs)
+    .filter(([_, enabled]) => enabled)
+    .map(([catalogName]) => catalogName);
```
axbarsan, line 20: Not a big deal, but we can do this in a single loop using `.reduce()` or just a regular `for` loop

axbarsan (approved):

PR #2132, opened 2021-02-09
Remove the warning regarding runnin minimum 3 nodes in a pool

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/15557

This PR the warning saying `We recommend that you run clusters with at least 3 workers`, when trying to set a node pool's minimum scale to a value lower than `3`.


oponder (approved):
LGTM

PR #2137, opened 2021-02-10
Enable searching 

oponder:
The search query used to be 'stored' in the query params, because I thought maybe someone wants to share their search result with someone else by sharing a link.

However now the search result depends also on which facets are selected, which I moved into the state to preserve it there. That means a 'q' param in the URL is not so useful anymore, because the result of that query depends on the state that the user is in (which facets are selected)

So the search query is now part of the app state as well.


axbarsan (approved):
Noice

PR #2139, opened 2021-02-10
Enable the "Reset Search". Clean up catalog icons 

oponder:
This PR enables the reset search button when there are no search results and also cleans up and add some missing catalog repo icons.

Screenshot 2021-02-10 at 8 05 56 PM


ubergesundheit (approved):
Looks good

PR #2141, opened 2021-02-10
Fix a crash when delete dates are somehow invalid

oponder:
Fixes an exception I've noticed recently.

It looks like the delete_date of a cluster can be "Invalid Date" for a while.

### DIFF
--- a/src/components/Home/ClusterDashboardItem.js
+++ b/src/components/Home/ClusterDashboardItem.js
@@ -201,7 +201,8 @@ function ClusterDashboardItem({
             <NameWrapper>{cluster.name}</NameWrapper>
           </TitleWrapper>
           <DeleteDateWrapper>
-            Deleted {relativeDate(cluster.delete_date)}
+            Deleted{' '}
+            {!isNaN(cluster.delete_date) && relativeDate(cluster.delete_date)}
           </DeleteDateWrapper>
         </ContentWrapper>
       </WrapperDeleted>
--- a/src/@types/clusters.d.ts
+++ b/src/@types/clusters.d.ts
@@ -16,7 +16,7 @@ interface IBaseCluster {
 
   path?: string;
   credential_id?: string;
-  delete_date?: string | null;
+  delete_date?: Date | null;
 
   // Injected by client-side.
   capabilities?: IClusterCapabilities;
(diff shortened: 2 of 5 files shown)


src/components/Home/ClusterDashboardItem.js
```diff
@@ -201,7 +201,8 @@ function ClusterDashboardItem({
             <NameWrapper>{cluster.name}</NameWrapper>
           </TitleWrapper>
           <DeleteDateWrapper>
-            Deleted {relativeDate(cluster.delete_date)}
+            Deleted{' '}
+            {!isNaN(cluster.delete_date) && relativeDate(cluster.delete_date)}
```
axbarsan, line 205: Does this solve the issue? If the date was `NaN` (or a nullish value), it would be caught by `relativeDate()`.
https://github.com/giantswarm/happa/blob/ed00b918fcbc21cdeb7164f1e0e9729358268c42/src/lib/helpers.tsx#L207

yasn77 (approved):

src/components/Home/ClusterDashboardItem.js
oponder: It does somehow fix it. The actual object that is coming through is "Invalid Date" when I log `cluster.delete_date`

I think we should investigate why this happens.. it's only on Safari. 

The dates in the API response for that cluster looks like

```
    "create_date": "2020-11-27T13:12:08Z",
    "delete_date": "2021-02-10T16:39:27Z",
```

Which look fine to me. I'm also kind of confused at what point this string becomes a date object. Probably the api client is doing this.

But for now this does fix the immediate crash.

axbarsan: Also, if the date is present in the response, it is a `string`, so `isNaN()` will always be `true`.

image

oponder: https://stackoverflow.com/questions/1353684/detecting-an-invalid-date-date-instance-in-javascript

axbarsan: To me the fact that it becomes a `Date` object is a bug, and we should solve that instead

oponder: I agree.. can I make it an issue and call this a temporary workaround

oponder: Issue to fix this properly is here: https://github.com/giantswarm/giantswarm/issues/15315

axbarsan (approved):

PR #2148, opened 2021-02-11
Managed app label & Catalog description tooltip

oponder:
This adds a little tooltip with the description of the catalog for each facet. 

(I don't really like the UX of this btw, ping @marians, this was specced but now that I see it in action im not sure it feels that great. if you want me to deploy it somewhere so you can see, let me know)

It also adds the managed label to the giant swarm catalog, currently the logic for determining that is just based on the name of the catalog. Eventually it will be a specific label on the AppCatalogCR, but we haven't specced that and agreed on it yet.

Screenshot 2021-02-12 at 1 12 54 AM
&

### DIFF
--- a/src/components/UI/Display/Apps/AppList/CatalogLabel.tsx
+++ b/src/components/UI/Display/Apps/AppList/CatalogLabel.tsx
@@ -93,7 +93,7 @@ const CatalogLabel: React.FC<ICatalogLabelProps> = (props) => {
           </OverlayTrigger>
         )}
 
-        {(!props.description || props.description === '') && text}
+        {!props.description && text}
 
         {props.error && <ErrorIcon name={props.catalogName} />}
       </Text>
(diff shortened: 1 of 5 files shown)


src/components/UI/Display/Apps/AppList/CatalogLabel.tsx
```diff
@@ -63,14 +64,37 @@ ErrorIcon.propTypes = {
...
+              <Tooltip id={`app-catalog-description-${props.catalogName}`}>
+                {props.description}
+              </Tooltip>
+            }
+          >
+            {text}
+          </OverlayTrigger>
+        )}
+
+        {(!props.description || props.description === '') && text}
```
axbarsan, line 96: Doesn't `!props.description` also equal to `true` when it's an empty string?

axbarsan (approved):

src/components/UI/Display/Apps/AppList/CatalogLabel.tsx
oponder: Yes! but you'd think I'd have learned that by now right. :P

PR #2151, opened 2021-02-12
Sorting

oponder:
Implement sorting functions and fixes a few visual glitches with the version picker.

@axbarsan my memoizing attempts aren't really working out.. could we maybe pair ?  I feel like this AppList component can be a lot better in terms of performance and avoiding expensive calls all the time but not 100% sure how to achieve it.

### DIFF
--- a/src/components/Apps/AppsList/AppsList.tsx
+++ b/src/components/Apps/AppsList/AppsList.tsx
@@ -73,14 +73,18 @@ const AppsList: React.FC = () => {
   const allApps = useSelector(selectApps);
   const selectedCatalogs = useSelector(selectSelectedCatalogs);
   const memoSelectedCatalogs = selectedCatalogs.join('');
+  const sortOrder = useSelector(selectAppSortOrder);
 
-  const apps = useMemo(() => searchApps(debouncedSearchQuery, allApps), [
+  const apps = useMemo(() => {
+    const appCollection = searchApps(debouncedSearchQuery, allApps);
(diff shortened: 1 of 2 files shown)


src/components/Apps/AppsList/AppsList.tsx
```diff
@@ -46,13 +73,15 @@ const AppsList: React.FC = () => {
...
-  const sortOrder = useSelector(selectAppSortOrder);
 
   const apps = useMemo(() => searchApps(debouncedSearchQuery, allApps), [
     debouncedSearchQuery,
     memoSelectedCatalogs,
   ]);
 
+  const sortOrder = useSelector(selectAppSortOrder);
+  apps.sort(sortFuncs[sortOrder]);
+
```
axbarsan, line 84: But why not do the sorting inside of the `useMemo` function argument?

I believe this should do it:

```suggestion
  const sortOrder = useSelector(selectAppSortOrder);

  const apps = useMemo(() => {
    const appCollection = searchApps(debouncedSearchQuery, allApps);
    appCollection.sort(sortFuncs[sortOrder]);
    
    return appCollection;
  }, [
    debouncedSearchQuery,
    memoSelectedCatalogs,
    sortOrder,
  ]);

```

oponder:
ok I'll go with that for now, but I think there's still some stuff running without need. I'll show ya what I mean though in front end sync

axbarsan (approved):

PR #2153, opened 2021-02-12
App icons on app List page

oponder:
Enables the app icons on the app listing page.

Doesn't look so great right now, but I have icon sized correctly and made for the dark background ready to go.

Screenshot 2021-02-12 at 5 45 20 PM


src/stores/appcatalog/selectors.ts
```diff
@@ -32,10 +32,10 @@ export const selectApps = createSelector(
         allApps.push(
           ...Object.entries(catalog.apps).map(([appName, appVersions]) => ({
             name: appName,
+            appIconURL: appVersions[0].icon,
```
axbarsan, line 35: This (`appVersions[0]`) is the latest version, right?

axbarsan (approved):

src/stores/appcatalog/selectors.ts
oponder: It should be yeah. We could actually ensure that by sorting all the versions early in the action that fetches the catalog, but it should come with the latest at the first position.

PR #2157, opened 2021-02-15
Use dark bg app icons if possible

oponder:
[When I made all the app icons myself](https://github.com/giantswarm/web-assets/tree/master/assets/app-icons/1/png), I made two versions, for dark and light backgrounds.

Now that we're switching, I'd like it to already look good, instead of forcing all apps to change the URL and release a new version. This way we let Happa detect apps which would have a dark-bg version of their logo, and use that instead


ubergesundheit (approved):
The mechanism to display a fallback if the image couldn't be loaded is still in place, right?

Code looks good

oponder:
> The mechanism to display a fallback if the image couldn't be loaded is still in place, right?

yup!

PR #2161, opened 2021-02-16
Apply sig-ux feedback

oponder:
This applies various points of feedback on the app catalog from sig-ux feedback round.

- remove “Giantswarm” at the start of internal catalog names from the facets list
- link website in app detail page
- alphabetise the catalogs in the facet list
- add select all / select none to facet list
- fix table rendering in readme's

Screenshot 2021-02-16 at 10 47 10 PM


__tests__/Apps.js
```diff
@@ -98,8 +104,19 @@ describe('Apps and App Catalog', () => {
 
         expect(introText).toBeInTheDocument();
 
+        const main = await findByTestId('main');
```
oponder, line 107: I did this so the debug output would be a bit more constrained when this test fails.

src/components/Apps/AppsList/utils.tsx
```diff
@@ -7,7 +7,7 @@ import { IFacetOption } from 'UI/Inputs/Facets';
 // We changed the location of this information at some point, so happa currently
 // supports both. TODO: Check if there are any catalogs left with the old label
 // if not, simplify this code.
-const isInternal = (catalog: IAppCatalog): boolean => {
+export const isInternal = (catalog: IAppCatalog): boolean => {
```
oponder, line 10: exporting it so I can use this logic in a test. which feels a bit weird

axbarsan (approved):

PR #2168, opened 2021-02-17
Create new select input and replace all the existing ones with it

axbarsan:
Towards giantswarm/giantswarm#11004 and towards the grommet migration

This PR includes the new `Select` input, snapshot tests and Storybook examples. It also includes a new `MultiSelect` input based on it, for being able to select multiple values.

The new `Select` component uses virtualization for rendering the options, and it also has handy functionalities, such as searching for a specific option.

This PR also replaces the multi-value select in the `InviteUserForm` component with the new `MultiSelect` component. Unfortunately, I couldn't get the search functionality to work in the modal, the search input was just unable to be selected, no matter what I did.

### Preview

<details>
<summary>Simple</summary>

image

</details>

<details>
<summary>Simple - open</summary>

image

</details>

<details>
<summary>Disabled</summary>

image

</details>

<details>
<summary>With info and help messages</summary>

image

</details>

<details>
<summary>With support for searching for a specific option</summary>

image

</details>

<details>
<summary>MultiSelect with no value selected</summary>

image

</details>

<details>
<summary>MultiSelect with values selected</summary>

image

</details>

<details>
<summary>MultiSelect with values selected - open</summary>

image

</details>


src/components/UI/Inputs/CheckBoxInput/index.tsx
```diff
@@ -83,14 +83,18 @@ const CheckBoxInput = React.forwardRef<HTMLInputElement, ICheckBoxInputProps>(
...
+        pad: {
+          horizontal: 'none',
+          vertical: 'xsmall',
+        },
+        background: 'none',
       },
-      background: 'none',
-    });
+      contentProps
+    );
```
axbarsan, line 97: Changed the order to allow users to customize the padding. (I also did the same for the `RadioInput` component, since it works in the same way)

src/components/UI/Inputs/Select/index.tsx
```diff
@@ -0,0 +1,171 @@
...
+  ThemeContext,
+  ThemeType,
+} from 'grommet';
+import PropTypes from 'prop-types';
+import * as React from 'react';
+import { css } from 'styled-components';
+
+import SearchInput from './SearchInput';
+
+const customTheme: ThemeType = {
```
axbarsan, line 14: Handy little trick to be able to customize nested components that can't be customized through props. These styles only apply to this component and its children.

src/components/UI/Inputs/Select/index.tsx
```diff
@@ -0,0 +1,171 @@
...
+  select: {
+    background: 'none',
+    icons: {
+      color: { dark: 'text', light: 'text' },
+    },
+    // @ts-expect-error
+    searchInput: SearchInput,
+    container: {
+      // @ts-expect-error
+      extend: css`
```
axbarsan, line 34: Just learned that we can use `styled-components`'s `css` function instead of a regular arrow function. The type definitions don't like it, but it works.

marians:
Great work!

Tested the user invite dialog. When the input changes position, the overlay does not move with the input. Would be nice if we could make that stick together. Recording:

https://user-images.githubusercontent.com/273727/108424340-53d36f80-7239-11eb-879b-b7ed1a04d250.mov

marians (approved):
The glitch I found should not block this

axbarsan:
> Tested the user invite dialog. When the input changes position, the overlay does not move with the input. Would be nice if we could make that stick together.

Fixed ✅

PR #2181, opened 2021-02-19
Move docs URLs into library

marians:
This PR

- moves all docs.giantswarm.io URL strings into a lib, where the expectations we have for each URL is documented, and where we can maintain them all on a single place and avoid duplication.
- Updates changed URLs for docs and blog
- Updates some text in the Getting Started sections, as we can no longer point to a Guides sections, and to avoid uppercase


marians:
@giantswarm/chapter-front-end Any conventions I should be aware of when it comes to naming? I wasn't sure whether the file name should be `docs.ts` or `Docs.ts` and whether these constants are fine in lowercase. RFR

oponder:
Nice! The casing is ok. My heuristic was always that files that export a React Component should get the upper case in their filename. So in this case, this is like a utility file that exports multiple docs related things, instead of the one Component.

oponder:
See also: https://github.com/giantswarm/fmt/blob/master/javascript/general_style_guide.md

```
Naming

Regular JS files (not components) and non-component folders: camelCase.js
Component files and folders: PascalCase
Props, state and methods: camelCase
We don't use index.js for component files.
```


Although we recently started using `index.js` again, which I am actually ok with tbh because it makes the import statements stutter less.

marians:
OK. So of this is fine, I'd welcome an approval :)

oponder (approved):
:+1:

PR #2183, opened 2021-02-19
Create new date input and replace all the existing ones with it

axbarsan:
Towards giantswarm/giantswarm#11004 and towards the grommet migration

This PR includes the new date input, snapshot tests and Storybook examples. It also replaces the date input in the keypair creation modal with the new one, getting rid of another dependency.

If specified a `format` (e.g. `yyyy-mm-dd`), you can also enter the date via text, with a masked input (meaning you don't have to type the `-` between the years, months and days - similar to a phone field on some websits). Otherwise, the form acts as a button, which you can customize as you wish.

The current button display is the `grommet` default, so it will change once we will customize buttons. That's why the date button in the keypair creation modal currently uses a custom style.

### Preview

<details>
<summary>With label</summary>

image

</details>

<details>
<summary>With label - open</summary>

image

</details>

<details>
<summary>With multiple other configurations</summary>

image

</details>

<details>
<summary>With multiple other configurations - open</summary>

image

</details>

<details>
<summary>With a button instead of input</summary>

image

</details>

<details>
<summary>With a button instead of input - with value</summary>

image

</details>

<details>
<summary>With a button instead of input - open</summary>

image

</details>

<details>
<summary>Keypair creation modal</summary>

image

</details>

<details>
<summary>Keypair creation modal - date input open</summary>

image

</details>


src/components/UI/Inputs/DateInput/index.tsx
```diff
@@ -0,0 +1,205 @@
...
+          error={error}
+          info={info}
+          help={help}
+          margin={margin}
+          pad={pad}
+          {...formFieldProps}
+        >
+          <Input
+            // @ts-expect-error
+            ref={ref}
```
axbarsan, line 141: The props interface of the date input doesn't allow setting a ref, but it's actually supported 🧇

oponder (approved):
Beautiful! LGTM, didn't test run it though.

PR #2188, opened 2021-02-22
Refactor NumberPicker to use the new input and structure

axbarsan:
Towards the `grommet` migration

The new component keeps the same functionality as the old one, but adds consistency to the visual and programmatical interface with the other inputs.

### Preview

image


oponder (approved):
LGTM 🤘

PR #2230, opened 2021-03-04
MAPI Playground

axbarsan:


oponder (approved):
LGTM!

It's not mounted anywhere though right? Maybe we could add a route like /mapi-cluster-list ?

axbarsan:
It's not mounted anywhere, it's just as an example for the first MAPI feature that we'll implement

PR #2231, opened 2021-03-04
Add experimental support for OAuth2 login

axbarsan:
Towards [logging in with OAuth2 via Dex](https://github.com/giantswarm/roadmap/issues/218).

This PR adds support for logging in with OAuth2 for regular users and admins. The old `Auth0` admin login was removed, as we now use `GitHub` instead.

The OAuth2 login for regular users is behind a feature flag (`feature-mapi-auth` key in the configmap). When it is disabled, the regular user and password login form is displayed. When it's enabled, you also have a button that will redirect you to the OAuth2 provider.

:warning: OAuth2 login for regular users is enabled by default on testing installations.


src/components/Auth/parseErrorMessages.js
```diff
@@ -1,5 +1,5 @@
 const parseErrorMessages = (error) => {
-  let heading = 'Could not log in';
+  let heading = 'Could not login';
```
oponder, line 2: I think it should be `log in` still, like verb vs noun thing : https://www.grammar.com/log_in_vs._login

src/lib/MapiAuth/makeDefaultConfig.ts
```diff
@@ -0,0 +1,114 @@
...
+        'email_verified',
+        'exp',
+        'iat',
+        'iss',
+        'locale',
+        'name',
+        'sub',
+      ],
+    },
+    signingKeys: [
```
oponder, line 66: I think the right way to get at least some of this info are the 'well-known' and 'keys' endpoints of the issuer: 

https://dex.g8s.ghost.westeurope.azure.gigantic.io/.well-known/ and
https://dex.g8s.ghost.westeurope.azure.gigantic.io/keys

src/lib/OAuth2/TestOAuth2.ts
```diff
@@ -0,0 +1,163 @@
...
+
+      return asBase64.replace(/-/g, '+').replace(/_/g, '/');
+    };
+
+    const idToken = `${encode(header)}.${encode(tokenPayload)}.${encode(
+      signature
+    )}`;
+
+    return {
+      refreshToken: 'some-random-key',
```
oponder, line 152: I guess we're storing the refresh token in the local storage then? That's a bit risky since I think it never expires.

we should look into if Dex supports the PKCE flow, that'll also make it so we don't need a clientSecret here

oponder (commented):
really awesome, thanks for figuring out a way forward with this, making it all work, and keeping the test coverage in mind. 👍 

The reason I'm not approving yet is while I'm not a OIDC/OAuth expert, I know some best practices that have been echoed around. I think if we can figure out how to do the PKCE flow that would be good. (it was only recently added: https://github.com/dexidp/dex/pull/1784)

I think otherwise we'll trigger some flags for security.

src/components/Auth/parseErrorMessages.js
oponder: ping @marians

__tests__/AdminLogin.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+    renderRouteWithStore(
+      MainRoutes.AdminLogin,
+      undefined,
+      undefined,
+      testAuth,
+      history
+    );
+
+    expect(attemptLoginMockFn).toHaveBeenCalledWith(
```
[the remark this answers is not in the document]
axbarsan: Since we are using a different OAuth2 provider for testing (`TestOAuth2`, instead of `MapiAuth`, to not have to test the `oidc-client-js` library), there's no way to tell if we are using the admin login (which is executed by adding an extra parameter in the URL), without actually replicating the logic in `TestOAuth2`.

If we would actually replicate the `MapiAuth` logic in `TestOAuth2` and test the "behaviour" - if the browser redirects to the correct URL, then we would just be testing the test.

By checking if the correct connector is used, we know that we call `MapiAuth` correctly. It's its job to do the right thing starting then.

oponder (approved):
Approving so we can keep moving. 

Might be good to get a glance from @giantswarm/sig-security too.

nice that dex support refresh token rotation!

src/lib/OAuth2/TestOAuth2.ts
oponder: Sweet. Does that mean there might be a issue with running happa in multiple tabs?

src/lib/MapiAuth/makeDefaultConfig.ts
axbarsan: it works now 👍🏻

PR #2248, opened 2021-03-09
Fix dimensions of Button style "danger"

marians:
The Button with style `danger` was 32 pixels high instead of the 34 pixels the other styles use. This PR fixes that.


oponder (approved):
lgtm , thx

PR #2255, opened 2021-03-10
Add storybook item for Tabs

marians:
## Preview

image

One thing that I don't understand: why is it showing both Tab's content? I saw an example in `AppDetails.js`.


marians:
Ping @oponder 

> One thing that I don't understand: why is it showing both Tab's content?

Can you explain this?

marians:
Ouch!

oponder (approved):

PR #2260, opened 2021-03-11
Add storybook on colors

marians:
Creating an overview for our color palettes

Includes a group named "Deprecated" which I think should be replaced with other color values at some point.

image


oponder:
cool! makes me think we can probably consolidate some of these colors and start getting rid of them. like white5 isn't really needed, and can be replaced with white2 maybe.

oponder (approved):

PR #2261, opened 2021-03-11
Stop pre-commit hook if compilation fails

axbarsan:
Right now, if compilation fails and linting succeeds, then the commit passes through.


marians:
`yarn typecheck` takes 5.3 seconds for me. Adding that to every pre-commit call would slow things down significantly. Is there some way around this?

axbarsan:
> `yarn typecheck` takes 5.3 seconds for me. Adding that to every pre-commit call would slow things down significantly. Is there some way around this?

The TypeScript compilation has already been running on every commit since we started using git hooks, last year.
What happened was that when we migrated to version `5` of `husky`, which introduced a different way of configuring the webhook, we forgot those `&&` in between the commands. Without those `&&`, linting will still run, even if compilation failed. This change doesn't slow down the git hook in any way.

And also, we use incremental TypeScript compilation, which only re-compiles the parts of the code that changed since the last compilation.

If running any of these commands slows you down, you can always use the `--no-verify` flag with `git commit`, which doesn't run any hooks. CI will check all of your code once you push it to GitHub.

oponder (approved):

PR #2284, opened 2021-03-19
Disable airbrake remoteConfig to prevent requests to notifier-configs.airbrake.io

marians:
Closes https://github.com/giantswarm/giantswarm/issues/16339

This PR turns off the airbrake remote configuration feature. Having this on by default caused requests to an airbrake.io server, which resulted in 403 responses.


ubergesundheit (approved):
LGTM

PR #2286, opened 2021-03-23
Organization management using the MAPI

oponder:
This is a collection PR of various other approved PRs towards organization management using the MAPI in Happa.

related: https://github.com/giantswarm/giantswarm/issues/16300
related: https://github.com/giantswarm/giantswarm/issues/16303


oponder:
When logged in as an admin, we now get the new org list with the new org creation button  that creates actual Org CRs below it.

axbarsan (approved):

PR #2287, opened 2021-03-23
Add new organization listing and creating ui elements

oponder:
This PR implements the UI behaviour specced for creating and listing organisations. https://github.com/giantswarm/giantswarm/issues/16300

There is a story at `Display/Organizations/OrganizationListPage` so you can check out the behaviour if you'd like.


marians (approved):
Oliver just showed me around and it looks great!

axbarsan (approved):
✅

PR #2289, opened 2021-03-24
Implement MAPI requests for fetching and creating organisations

oponder:


axbarsan (approved):
LGTM ✅  

Maybe we can declare some of the functions at the top of the component, instead of directly in the JSX, but other than that, awesome job 👍🏻

PR #2291, opened 2021-03-25
Add mapi service for performing selfSubjectRulesReview

oponder:
@axbarsan I'm looking to make the `getHasAccessToResources` function check if the user can LIST Organizations by actually checking for the permission instead of checking the organization count. This would help in the case of a new cluster like @whites11 ran into yesterday on `flamingo`

I started with setting up the MAPI call that does the selfSubjectRulesReview, and am getting back a good response. Basically from this I can figure out if they have access to listing orgs. 

(verbs: *, apiGroups: *, resources: *) would be an instant yes, and I can write something that figures out if they specifically have the list orgs permission too.

but I'm wondering how to integrate this into the app... right now Ive added it to `batchActions.ts`, and will probably write an action to add it to the global state. This seems to make sense to me.. we're not going to completely remove redux I think? We still need some global state for things like the user.

Also am I initialising the http client correctly in this case?

### DIFF
--- a/src/model/services/mapi/authorization/selfSubjectRulesReview.ts
+++ b/src/model/services/mapi/authorization/selfSubjectRulesReview.ts
@@ -1,0 +1,36 @@
+import { HttpRequestMethods, IHttpClient } from 'model/clients/HttpClient';
+import * as k8sUrl from 'model/services/mapi/k8sUrl';
+
+import { ISelfSubjectRulesReview } from './';
+
+export function selfSubjectRulesReview(
+  client: IHttpClient,
+  user: ILoggedInUser
+) {
+  return async () => {
+    const url = k8sUrl.create({
+      baseUrl: window.config.mapiEndpoint,
+      apiVersion: 'authorization.k8s.io/v1',
+      kind: 'SelfSubjectRulesReviews',
+    });
+
+    const requestBody: ISelfSubjectRulesReview = {
+      apiVersion: 'authorization.k8s.io/v1',
+      kind: 'SelfSubjectRulesReview',
+      spec: {
+        namespace: 'default',
+      },
+    };
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    client.setBody(requestBody as Record<string, any>);
+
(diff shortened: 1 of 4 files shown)


src/model/services/mapi/authorization/selfSubjectRulesReview.ts
```diff
@@ -0,0 +1,36 @@
+import { HttpRequestMethods, IHttpClient } from 'model/clients/HttpClient';
```
axbarsan, line 1: The name of this folder should be `authorizationv1`. Other than that, LGTM

axbarsan (approved):

src/stores/batchActions.ts
```diff
@@ -4,6 +4,8 @@ import ErrorReporter from 'lib/errors/ErrorReporter';
 import { FlashMessage, messageTTL, messageType } from 'lib/flashMessage';
 import { IOAuth2Provider } from 'lib/OAuth2/OAuth2';
 import RoutePath from 'lib/routePath';
+import { HttpClientImpl } from 'model/clients/HttpClient';
+import { selfSubjectRulesReview } from 'model/services/mapi/authorization';
```
oponder, line 8: @axbarsan could you help me figure out how to mock this in a test? Since it's part of the main batchActions, I need a way to avoid having this try to make an actual API call, since we don't have a nock for the MAPI requests and want to avoid doing that right

axbarsan: For now I think we can mock the function in all tests, like we do with the info and metadata requests.

- https://github.com/giantswarm/happa/blob/master/testUtils/modelMocks.js
- https://github.com/giantswarm/happa/blob/master/testUtils/mockHttpCalls/metadata.ts
- https://github.com/giantswarm/happa/blob/779eac7d32679c7a64fbaf54a9cb6ab435a6d910/__tests__/AdminLogin.tsx#L24

Also, you might want to only do this request for MAPI users, and not for oldschool GS users

oponder:
This got incorporated into another PR.

PR #2293, opened 2021-03-25
Add types and functions required for access control

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/16306

There are also types and functions for `ClusterRoleBinding` resources, even though we don't use them in the actual implementation. I wrote them as part of an (incorrect) assumption, but maybe we need them at some point.


oponder (approved):
LGTM 👍

PR #2297, opened 2021-03-26
Orgs listing using MAPI, also for non admins

oponder:
This transitions organization fetching fully to the MAPI when logged in as a MAPI user.


axbarsan (approved):
Noice 😬

PR #2298, opened 2021-03-26
access control: add components needed for displaying the role list sidebar

axbarsan:
Towards giantswarm/giantswarm#16306

This PR adds the UI components, UI types, and business logic required to render the role list sidebar in the `Access control` page.


marians (approved):
I won't be in the way here, but I can't review this.

oponder (approved):
LGTM!

PR #2305, opened 2021-03-29
access control: add other components required to render the whole thing

axbarsan:
Towards giantswarm/giantswarm#16306

This PR adds the extra components required to render Access Control. This is still missing a bigger integration test with all the request/parsing/etc. logic. I have a plan for that, don't worry 😜

### DIFF
--- a/src/components/MAPI/organizations/AccessControl/index.tsx
+++ b/src/components/MAPI/organizations/AccessControl/index.tsx
@@ -1,0 +1,178 @@
+import { Box } from 'grommet';
+import produce from 'immer';
+import { useHttpClient } from 'lib/hooks/useHttpClient';
+import { GenericResponse } from 'model/clients/GenericResponse';
+import PropTypes from 'prop-types';
+import React, { useLayoutEffect, useMemo, useState } from 'react';
+import { useSelector } from 'react-redux';
+import DocumentTitle from 'shared/DocumentTitle';
+import { getLoggedInUser } from 'stores/main/selectors';
+import useSWR from 'swr';
+import AccessControlRoleDescription from 'UI/Display/MAPI/AccessControl/AccessControlDescription';
+import AccessControlRoleDetail from 'UI/Display/MAPI/AccessControl/AccessControlRoleDetail';
+import AccessControlRoleList from 'UI/Display/MAPI/AccessControl/AccessControlRoleList';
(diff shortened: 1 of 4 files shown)


src/components/UI/Display/MAPI/AccessControl/AccessControlRoleType.tsx
```diff
@@ -18,7 +18,11 @@ const AccessControlRoleType: React.FC<IAccessControlRoleTypeProps> = ({
     <Box gap='xxsmall' direction='row' {...props}>
       {inCluster && (
         <Text>
-          <i className='fa fa-globe' aria-label='Cluster role' />
+          <i
+            className='fa fa-globe'
+            aria-label='Cluster role'
+            role='presentation'
```
axbarsan, line 24: Forgot to push this in my previous PR

src/components/MAPI/organizations/AccessControl/index.tsx
```diff
@@ -0,0 +1,178 @@
...
+    getRoleItems(client, user!, orgNamespace)
+  );
+
+  const [activeRoleName, setActiveRoleName] = useState('');
+  const activeRole = useMemo(
+    () => data?.find((role) => role.name === activeRoleName),
+    [data, activeRoleName]
+  );
+
+  useLayoutEffect(() => {
```
axbarsan, line 53: Using `useLayoutEffect` instead of `useEffect` gets rid of displaying the search animation for a split second in this scenario:
- Navigate to the `Access control` tab in Organization details to view a role's details
- Switch to the `General` tab in Organization details
- Switch back to the `Access control` tab
- See the loader for a split second

src/components/MAPI/organizations/AccessControl/utils.ts
```diff
@@ -1,3 +1,8 @@
+import { GenericResponse } from 'model/clients/GenericResponse';
+import { IHttpClient } from 'model/clients/HttpClient';
+import * as corev1 from 'model/services/mapi/corev1';
+import * as metav1 from 'model/services/mapi/metav1';
+import * as rbacv1 from 'model/services/mapi/rbacv1';
```
axbarsan, line 5: Importing these as namespaces lets us use multiple versions of the same api group in the same file. It also decreases confusion, since the types in there might be similar in naming to the types here.

oponder (approved):
LGTMIIWFY

PR #2307, opened 2021-03-30
Use what's in the ui annotation for org name if it's there.

oponder:
For users logged in through a SSO, we have transitioned to relying on the MAPI for the source of truth for the organisations list.

This adjusts that logic to support the new `ui.giantswarm.io/original-organization-name`

### DIFF
--- a/src/model/services/mapi/k8sUrl.ts
+++ b/src/model/services/mapi/k8sUrl.ts
@@ -43,7 +43,7 @@ export interface IK8sGetOptions extends IK8sBaseOptions {
   /**
    * The namespace that the resource lives in.
    */
-  namespace: string;
+  namespace?: string;
 }
 
 export interface IK8sListOptions extends IK8sBaseOptions {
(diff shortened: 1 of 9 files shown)


src/model/services/mapi/k8sUrl.ts
```diff
@@ -43,7 +43,7 @@ export interface IK8sGetOptions extends IK8sBaseOptions {
   /**
    * The namespace that the resource lives in.
    */
-  namespace: string;
+  namespace?: string;
```
oponder, line 46: We could be getting ClusterScoped resources, which don't have a namespace.

src/stores/organization/actions.ts
```diff
@@ -155,6 +157,17 @@ export function organizationsLoadMAPI(): ThunkAction<
             orgs.push(...rule.resourceNames);
           }
         });
+
+        // We now know what orgs they can get. We still need to fetch them
+        // to check if the orgs have a 'ui.giantswarm.io/original-organization-name'
+        // annotation.
+        const orgGetRequests = orgs.map((orgName) =>
+          getOrganization(client, user, orgName)()
```
oponder, line 165: Users without ability to list OrgCRs now need to do 1 request for every org, to check if the orgCR has the annotation.

axbarsan (approved):
LGTM

PR #2308, opened 2021-03-31
Truncate cluster ID label if longer than 5 characters

marians:
In order to allow for longer cluster IDs without breaking the UI, I want to make the ClusterIDLabel component truncated.

This is not working as it is right now.


marians:
Current problem: there should be a tooltip when placing the mouse pointer on the truncated ID string, but it doesn't appear.

image

Any idea what's wrong here?

marians:
Added a storybook in Display / Clusters / ClusterIDLabel to simplify testing this.

marians:
Now I have a test to fix. I got rid of the title attribute, as

- A tooltip appears faster and IMO is more readable.
- there is no point in having both a tooltip and a title attribute.

This breaks a test which relies on the presence of the title attribute (`getByTitle`). Can this be done using a data attribute instead?

axbarsan:
> This breaks a test which relies on the presence of the title attribute (`getByTitle`). Can this be done using a data attribute instead?

You can use `aria-label` instead, and `getByLabelText`. It's visible to screen readers, which tooltips aren't.

marians:
Ah, that rings a bell. I might have known that some time. Thanks!

src/components/UI/Display/Cluster/ClusterIDLabel/index.js
```diff
@@ -48,20 +48,38 @@ const ClusterIDLabel = ({ clusterID, copyEnabled }) => {
...
+    <Label clusterID={realClusterID}>
+      <OverlayTrigger
+        overlay={
+          <Tooltip id='idtooltip'>
+            <>Cluster ID: {realClusterID}</>
+          </Tooltip>
+        }
+        placement='top'
+      >
+        <span aria-label={realClusterID}>{label}</span>
```
axbarsan, line 68: There's no need for a new variable here 👍🏻 

```suggestion
  let label = clusterID.substring(0, 5);
  if (label !== clusterID) {
    label = `${clusterID.substring(0, 4)}…`;
  }

  const labelComponent = (
    <Label clusterID={clusterID}>
      <OverlayTrigger
        overlay={
          <Tooltip id='idtooltip'>
            <>Cluster ID: {clusterID}</>
          </Tooltip>
        }
        placement='top'
      >
        <span aria-label={clusterID}>{label}</span>
```

axbarsan (commented):
Looking good 👍🏻 a small nit

axbarsan (approved):
LGTM

PR #2311, opened 2021-03-31
Bring back cluster count in new MAPI based org list

oponder:
Screenshot 2021-03-31 at 7 43 47 PM


src/stores/cluster/selectors.ts
```diff
@@ -14,6 +14,12 @@ import { isPreRelease } from 'stores/releases/utils';
 import { IState } from 'stores/state';
 import { createDeepEqualSelector } from 'stores/utils';
 
+export const selectClusters = () => (
+  state: IState
+): Record<string, Cluster> | undefined => {
```
axbarsan, line 19: Can this actually be undefined?

https://github.com/giantswarm/happa/blob/041e0f77ef627ee42bf3a9e544e63e189da5b1be/src/stores/cluster/reducer.ts#L32

oponder: Nope! fixed.

axbarsan (approved):

PR #2321, opened 2021-04-01
Ensure that MAPI requests always use a fresh, not expired token

axbarsan:
This PR refactors all the MAPI client functions, so we could renew tokens before executing requests. If the token is still expired after trying to renew it, we would just throw an error.


src/components/MAPI/organizations/AccessControl/index.tsx
```diff
@@ -34,14 +33,13 @@ const AccessControl: React.FC<IAccessControlProps> = ({
 }) => {
   const orgNamespace = getOrgNamespaceFromOrgName(organizationName);
 
-  const client = useHttpClient();
-  const user = useSelector(getLoggedInUser);
+  const clientFactory = useHttpClientFactory();
```
axbarsan, line 36: Now using a client factory instead of a single client, since multiple requests that run in parallel have different request configurations. Somehow this wasn't a problem before, but it now is 😬

__tests__/AdminLogin.tsx
```diff
@@ -21,13 +21,9 @@ import {
 
 describe('AdminLogin', () => {
   it('performs the admin login flow via OAuth2', async () => {
-    (getOrganizationList as jest.Mock).mockReturnValue(() =>
```
oponder, line 24: thx!.. for some reason the other way didn't work for me but it was probably due to something unrelated.

oponder (approved):
LGTMIIWFY! Thx for cleaning up a few bits and bobs there from my previous PRs too.

__tests__/AdminLogin.tsx
axbarsan: It didn't work before because client functions were higher order functions (they returned other functions). Now they're regular async functions, so `mockResolvedValue()` works

PR #2329, opened 2021-04-06
Access control UI tweaks

marians:
Some attempts to improve the access control UI.

- Change of column widths: Narrower role list, wider right hand side. Less truncation. No more resource count. (https://github.com/giantswarm/giantswarm/issues/16462)
- Use monospaced font in permissions table
- Use unicode character for ellipsis in truncation and in permission verbs

## Preview

image


oponder (approved):
LGTMIIWFY

marians:
Actually I just noticed that I don't like summary terms like "All" showing up in monospaced. It's not a literal value. Will tweak that.

marians:
Changed and also updated the preview image.

As a side effect, we now only use the `<Truncated>` component if there is a string to be truncated, but not for the `All` case.

marians:
@axbarsan Could you please review this? Since it's mostly your work I am modifying.

axbarsan (approved):
LGTM

PR #578, opened 2019-05-08
App Catalog Feature

oponder:
This is a collection of two accepted app catalog related PRs for Happa

- Showing the app catalog: https://github.com/giantswarm/happa/pull/526
- Showing installed apps in the cluster detail page: https://github.com/giantswarm/happa/pull/574

Installing Apps & Deleting Apps will come next. But I'd like to avoid stacking too much stuff on top of each other and already get feedback on this.

Towards: giantswarm/giantswarm#4848

PR #2336, opened 2021-04-07
update some deps and stuff

oponder:
Updates various dependencies and fixes some errors / warnings in the tests.

Also removes some unused dependencies.


oponder:
Hmm ok storybook doesn't like one of the changed web pack deps. will check it out tmw.

package.json
```diff
@@ -85,10 +85,8 @@
     "@airbrake/browser": "2.1.3",
     "@hot-loader/react-dom": "17.0.1",
     "axios": "0.21.1",
-    "bluebird": "3.7.2",
```
axbarsan, line 88: Wasn't this required as a peer dependency of `giantswarm-js-client`?

oponder: I don't think so... 🕵️ it was introduced here: https://github.com/giantswarm/happa/pull/578/files which was about the original app catalog, but I probably didn't end up using it? Not sure...but as far as I can tell Happa works fine without it.

marians:
ICYM, pushing to catalogs currently fails. https://gigantic.slack.com/archives/C3C7ZQXC1/p1617811021034800

marians (approved):

PR #2337, opened 2021-04-07
Use the right config key for happaapi letsencrypt

marians:
This configuration change uses `.Values.Installation.V1.GiantSwarm.Happa.API.Letsencrypt` instead of `.Values.Installation.V1.GiantSwarm.Happa.Letsencrypt` to decide whether or not to activate Let's encrypt certificates.


whites11 (approved):
Seems legit

PR #2345, opened 2021-04-09
Add props to mitigate warnings in storybook

marians:
This PR adds some required props to components in a storybook story, to mitigate some warnings.


oponder (approved):
LGTM from my mobile. But so we need that button import?

marians:
The `installAppModal` is a required prop, so I decided to make its value somehow meaningful and placed a Button there.

PR #2348, opened 2021-04-13
Add corev1.ConfigMap client functions and types

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/16717

This PR adds client-side types and functions for dealing with MAPI `ConfigMap` resources.


src/model/services/mapi/corev1/types.ts
```diff
@@ -23,3 +23,14 @@ export interface IServiceAccount {
...
   automountServiceAccountToken?: boolean;
 }
+
+export const ConfigMap = 'ConfigMap';
+
+export interface IConfigMap {
+  apiVersion: 'v1';
+  kind: typeof ConfigMap;
+  metadata: metav1.IObjectMeta;
+  immutable?: boolean;
```
kopiczko, line 33: TIL CM has immutable field

src/model/services/mapi/corev1/createConfigMap.ts
```diff
@@ -0,0 +1,26 @@
...
+    isCore: true,
+    kind: 'configmaps',
+    namespace: configMap.metadata.namespace!,
+  });
+
+  return createResource<IConfigMap>(
+    client,
+    auth,
+    url.toString(),
+    (configMap as unknown) as Record<string, unknown>
```
kopiczko, line 24: This is a bit unclear to me.

kopiczko (approved):
Looks sane

src/model/services/mapi/corev1/createConfigMap.ts
```diff
@@ -0,0 +1,26 @@
+import { IOAuth2Provider } from 'lib/OAuth2/OAuth2';
+import { IHttpClient } from 'model/clients/HttpClient';
+import * as k8sUrl from 'model/services/mapi/k8sUrl';
```
kopiczko, line 3: I guess there are reasons to not use some external k8s libs?

src/model/services/mapi/corev1/types.ts
kubasobon: **+1**
> Starting from v1.19, you can add an immutable field to a ConfigMap definition to create an immutable ConfigMap.

https://kubernetes.io/docs/concepts/configuration/configmap/#configmap-immutable

src/model/services/mapi/corev1/createConfigMap.ts
axbarsan: The type I used was for trying to make sure that we use a data structure derived from a map here, and avoid using `any`, at the same time. But seems like `any` was actually the way to go

src/model/services/mapi/corev1/createConfigMap.ts
axbarsan: The library scene for client-side k8s is pretty much unexistent at this time. Pretty much everything there is out there is for running on the server side.

We had some automation setup for generating all the client-side types and methods from our `apiextensions` repo [in here](https://github.com/giantswarm/giantswarm-cp-client). However, it was very brittle (something would always crash when adding new types), and the generated code was a bit clunky. Due to how it was built, we couldn't strip unused code at build time, so we had a huge bundle.

We reverted to using the K8s API as a regular REST API for now, and created a few utilities, such as `k8sUrl` to make it a bit easier

kopiczko: How about this https://github.com/kubernetes-client/javascript ?

axbarsan: That's only for the server-side.
https://github.com/kubernetes-client/javascript/issues/165

I tried porting it to the browser at some point, but there were so many Node.js native libraries to create, that I gave up eventually

kopiczko: Alright I see what you mean by "client-side" now :)

PR #2359, opened 2021-04-15
Move happaapi resources from namespace 'default' to 'giantswarm'

marians:
This moves two resources related to the `happaapi` proxy into the `giantswarm` namespace, instead of the `default` namespace.


marians:
Closing, as this was BS. happaapi has to be in the default namespace, as `kubernetes` also is.

PR #2360, opened 2021-04-15
access control: reuse existing service accounts and display existing ones as suggestions

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/16468

This PR includes a bunch of improvements for service account creation (some also apply to subject creation in general):
* you can bind existing service accounts (a new one will only get created if it doesn't exist)
* you get suggestions for existing service account names
* when you remove a service account from a role binding, the `ServiceAccount` resource no longer gets deleted
* you are no longer able to create multiple subjects with the same name at the same time (it resulted in an error)
* fix a subject name parsing bug that would try to bind subjects with empty names (e.g. `test1, ` would bind both `test1`, and ` `)

### Preview

<details>
<summary>Full suggestion list</summary>

image

</details>

<details>
<summary>Suggestions based on the text you entered</summary>

image

</details>

<details>
<summary>No suggestions for a new account</summary>

image

</details>

You can navigate between suggestions with the `Up` and `Down` arrow keys. Also, if you closed the suggestion box by pressing `Esc`, for example, you can re-open it by pressing the `Down` arrow key.

When you click on one of the suggestions, it will get appended to your text input.

### DIFF
--- a/src/components/MAPI/organizations/AccessControl/AccessControlRoleSubjects.tsx
+++ b/src/components/MAPI/organizations/AccessControl/AccessControlRoleSubjects.tsx
@@ -1,7 +1,11 @@
+import { useAuthProvider } from 'Auth/MAPI/MapiAuthProvider';
 import { Box, Text } from 'grommet';
 import { FlashMessage, messageTTL, messageType } from 'lib/flashMessage';
+import { useHttpClient } from 'lib/hooks/useHttpClient';
+import { GenericResponse } from 'model/clients/GenericResponse';
 import PropTypes from 'prop-types';
 import React, { useEffect, useReducer } from 'react';
+import useSWR from 'swr';
 import AccessControlSubjectSet, {
   IAccessControlSubjectSetItem,
 } from 'UI/Display/MAPI/AccessControl/AccessControlSubjectSet';
@@ -12,7 +16,11 @@ import {
   IAccessControlRoleItem,
   IAccessControlRoleSubjectItem,
 } from '../../../UI/Display/MAPI/AccessControl/types';
-import { getUserNameParts } from './utils';
+import {
+  fetchServiceAccountSuggestions,
+  fetchServiceAccountSuggestionsKey,
+  getUserNameParts,
+} from './utils';
 
 interface IStateValue {
   isAdding: boolean;
@@ -141,7 +149,10 @@ const mapValueToSetItem = (stateValue: IStateValue) => (
 };
 
 interface IAccessControlRoleSubjectsProps
-  extends Pick<IAccessControlRoleItem, 'groups' | 'users' | 'serviceAccounts'>,
+  extends Pick<
+      IAccessControlRoleItem,
+      'groups' | 'users' | 'serviceAccounts' | 'namespace'
+    >,
     React.ComponentPropsWithoutRef<typeof Box> {
   roleName: string;
   onAdd: (type: AccessControlSubjectTypes, names: string[]) => Promise<void>;
@@ -149,6 +160,7 @@ interface IAccessControlRoleSubjectsProps
 }
 
 const AccessControlRoleSubjects: React.FC<IAccessControlRoleSubjectsProps> = ({
+  namespace,
   roleName,
   groups,
   users,
@@ -245,6 +257,14 @@ const AccessControlRoleSubjects: React.FC<IAccessControlRoleSubjectsProps> = ({
       });
     };
   }, [roleName]);
+
+  const client = useHttpClient();
+  const auth = useAuthProvider();
+
+  const { data: serviceAccountSuggestions } = useSWR<string[], GenericResponse>(
+    fetchServiceAccountSuggestionsKey(namespace),
+    () => fetchServiceAccountSuggestions(client, auth, namespace)
+  );
 
   const groupType = state[AccessControlSubjectTypes.Group];
   const userType = state[AccessControlSubjectTypes.User];
@@ -348,6 +368,7 @@ const AccessControlRoleSubjects: React.FC<IAccessControlRoleSubjectsProps> = ({
           )}
           isAdding={serviceAccountType.isAdding}
           isLoading={serviceAccountType.isLoading}
+          inputSuggestions={serviceAccountSuggestions}
         />
         {serviceAccountType.isAdding && (
           <Box>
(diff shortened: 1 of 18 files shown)


src/components/MAPI/organizations/AccessControl/AccessControlRoleSubjects.tsx
```diff
@@ -246,6 +258,14 @@ const AccessControlRoleSubjects: React.FC<IAccessControlRoleSubjectsProps> = ({
     };
   }, [roleName]);
 
+  const client = useHttpClient();
+  const auth = useAuthProvider();
+
+  const { data: serviceAccountSuggestions } = useSWR<string[], GenericResponse>(
```
axbarsan, line 264: We ignore the error. We just won't display any suggestions if fetching service accounts fails.

pipo02mix (approved):
LGTM from UX point of view, did not check code in detail :)

src/components/MAPI/organizations/AccessControl/AccessControlRoleSubjects.tsx
kopiczko: Shouldn't we have some sort metrics/logging for that?

BTW where is that error?

axbarsan: This line fetches the service accounts from the K8s API, so they could be shown as suggestions in the service account binding form. An error might only occur if the K8s API is unreachable (which might be a problem on the user's side, or on the ingress side). I don't believe that `happa` should report this error, as it isn't really its fault.

If we were to display the error to the user, we could destructure it from our request controller object, and place it somewhere in the UI (in a flash message, for example).

```suggestion
  const { data: serviceAccountSuggestions, error } = useSWR<string[], GenericResponse>(
```

I've decided to ignore the error if it occured because the only difference in UX would be that the user would get no suggestions for existing service accounts. Everything else would work as expected.

However, if you think it's important to display the error, I see no problem in doing it 👍🏻

kopiczko: Jeez I keep forgetting the request comes from the browser.

kopiczko (approved):
I tried to do a time-boxed review and it's fine as far as I'm concerned.

marians:
Testing now

marians:
I have a service account with a single letter name `f`. I can see the item in the drop-down when I enter "f" in the input field, but it does not get added as a selection.

marians:
That wasn't correct.

I can see the `f` item when I focus the input field:

image

But when I type "f" it is no longer there for selection.

image

axbarsan:
> That wasn't correct.
> 
> I can see the `f` item when I focus the input field:
> 
> image
> 
> But when I type "f" it is no longer there for selection.
> 
> image

Yes, I actually intended to do that. If the user has already entered the value, why do they need a suggestion anymore?

marians:
> Yes, I actually intended to do that. If the user has already entered the value, why do they need a suggestion anymore?

I would argue it's expected behaviour. I understand the list as "List of existing items matching my input". If the item is not visible, I am assuming that it does not exist. (Don't expect the user to see that it exists and then disappears while making an input)

axbarsan:
> > Yes, I actually intended to do that. If the user has already entered the value, why do they need a suggestion anymore?
> 
> I would argue it's expected behaviour. I understand the list as "List of existing items matching my input". If the item is not visible, I am assuming that it does not exist. (Don't expect the user to see that it exists and then disappears while making an input)

Ok, I fixed that

marians:
### Confirmation prompt

We _create_ service accounts "on the fly" through this UI of the user enters a name that does not yet exist. Then, when removing the binding, we simply ask "Are you sure"?

image

We do have the overlay when hovering the X, but that is easy to miss.

I would like to clarify what gets deleted after confirming, so I'd like to replace "Are you sure" with this text:

> Remove the binding between service account `name` and role `role`?
> The service account will stay.

Ideally the first sentence, which is the question to be answered, is separated from the second one, which is additional info.

### Conformation buttons

- Confirmation button label: `Yes, delete it` -> `Remove`

I know it's not related to the change, but I noticed it in this context and it's minor: The button heights for **Yes** and **Cancel** differ. Would be nice if they could have the same height.

marians:
I think here we should try to swap the z index order, so that the prompt appears on top of the tooltip.

image

marians:
When I enter a service account name, it seems as if my browser want to spell-check it. We should turn that off.

marians:
### Flash messages

Currently:

- `Subject added successfully`
- `Subject <name> deleted successfully`

We should have specific flash messages for

1. Binding a service account to a role
2. Creating a new service account and binding it to a role
3. Removing a service account binding

#### 1. Binding a service account to a role

> Service account `name` has been bound to the role.

#### 2. Creating a new service account and binding it to a role

> The service account `name` has been created and bound to the role.

#### 3. Removing a service account binding

> The binding for service account `name` has been removed.

Please use `<code></code>` to make the name literal stand out in all flash messages.

marians:
There is a problem after creating a new role binding. To reproduce:

1. Add a subject to groups, users, or service accounts.
2. Click the "Permissions" tab.

The page is empty and errors appear in the console:

image

axbarsan:
> There is a problem after creating a new role binding. To reproduce:
> 
> 1. Add a subject to groups, users, or service accounts.
> 2. Click the "Permissions" tab.
> 
> The page is empty and errors appear in the console:
> 
> image

I will take a look.

Also, you noted some cool UX improvements, but please create separate issues for them, as they are unrelated to the issue that this PR addresses, and this PR is already big enough.

marians:
Order of subjects: I'm thinking that (natural) sorting will help us more than keeping the order of entry.

image

Can we make it so that newly added subjects get appended at the end of the list? I'd like to make it so that the new subject appears in the place where the user has entered the name, which is at the end. However, after leaving the page or reloading it, I think it would be fine if the ordering would be systematic.

axbarsan:
> There is a problem after creating a new role binding. To reproduce:
> 
> 1. Add a subject to groups, users, or service accounts.
> 2. Click the "Permissions" tab.
> 
> The page is empty and errors appear in the console:
> 
> image

Ok, this is fixed. TIL that `Object.keys()` can return a frozen array.

@marians, is this ready to be merged?

marians:
When entering a bad service account name , this error occurs:

image

The first line (`Could not add subject`) could be improved by being more specific:

> Could not create service account `name`.

The literal error message (second paragraph) we could display in monospaced (Inconsolata) font in order to highlight the fact that it is the verbatim error the system returned.

marians (approved):
> is this ready to be merged?

Yes

marians:
Issues are created for the improvements

PR #2361, opened 2021-04-16
access control: display 'All (*)' where all verbs are supported

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/16464

With this PR, we now display `All (*)`, instead of `All`, where all verbs are supported for a role, in the `Permissions` tab.

### Preview 

image


pipo02mix (approved):
LGTM

PR #2362, opened 2021-04-16
access control: extend search to look for queries EVERYWHERE

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/16466

With this PR, we now look for the role search query in:
- the role names
- role subjects
  - group names
  - user names
  - service account names
- role permissions
  - api groups
  - resources
  - resource names

Performance is fine at the moment, but as datasets get larger, we will probably have to execute this in a web worker.


pipo02mix (approved):
LGTM that is cool

marians:
This is so awesome! Thank you!

PR #2364, opened 2021-04-16
access control: aggregate similar permission rules

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/16467

With this PR we aggregate similar permission verbs (if they have the same API Groups, Resources and Resource Names, the verbs will be shown into a single table row).

Also, as a bonus, we now sort permissions by these fields:
* API Groups
* Resources
* Resource Names
(e.g. if 2 permissions have the same API Group, they will be ordered by `Resources`, and so on)


oponder (approved):
LGTM

marians (approved):
LGTM too.

The cluster-admin role permissions look so much better with this aggregation!

image

PR #2366, opened 2021-04-16
access control: various ux improvements and bug fixes

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/16463

This PR includes the following:
* Display permission verbs in tooltips in the same order they are displayed in the table
* Fix an error that would be displayed when a `Role`'s `rules` property was `null`.
* Fix a bug that was causing horizontal scroll in the role list when a role had a longer name
* Make the role search input and role items have the same width.

### DIFF
--- a/src/components/UI/Display/MAPI/AccessControl/AccessControlRoleVerbs.tsx
+++ b/src/components/UI/Display/MAPI/AccessControl/AccessControlRoleVerbs.tsx
@@ -8,7 +8,7 @@ import {
   IAccessControlRoleItemPermission,
 } from './types';
 
-function formatVerbs(verbs: string[]): string {
+function formatVerbs(verbs: string[], verbMap: IVerbMap): string {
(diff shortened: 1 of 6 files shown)


src/components/UI/Display/MAPI/AccessControl/AccessControlRoleVerbs.tsx
```diff
@@ -17,7 +17,22 @@ function formatVerbs(verbs: string[]): string {
...
+    }
+    let bIdx = verbOrder.indexOf(b);
+    if (bIdx < 0) {
+      bIdx = verbOrder.length - 1;
+    }
+
+    return aIdx - bIdx;
+  });
+
+  return orderedVerbs.join(', ');
```
axbarsan, line 35: TIL

oponder (approved):
LGTM 🚀

PR #2367, opened 2021-04-19
Add chart and configuration for FE Docs

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/16817

With this change, the front end docs (or `storybook`) get pushed to the `giantswarm-operations-platform-catalog` as a separate app. This allows us to only have storybook running on a single cluster, and expose it to the world from the FE cluster.

You can check it out at https://fe-docs.giantswarm.io/


oponder (approved):
LGTM!

PR #2373, opened 2021-04-20
Add cluster and release info to org details page

marians:
Attempt to add some meaningful information to the MAPI organization details page.

Not sure if pulling in info from redux is the right attempt here, but this is what we already do in the org list (see `OrganizationListPage`).

I removed most of the rows we don't have information for yet, to get the page to a valid state.

### Preview

image

### TODO

- Add kubernetes version info to oldest and newest release
- Add unit tests


marians:
@axbarsan Is this the right direction here for what I'm trying to accomplish?

axbarsan (commented):
This works, and thank you for the effort, but the plan was to not add functionality based on the old REST API to these new MAPI features. It would make refactoring them to use MAPI 3x harder

marians:
ACK. Then let's keep `OrganizationListPage` the exception to the rule. I'll close this.

PR #2376, opened 2021-04-21
authentication: make happa aware of permissions

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/16821 and towards https://github.com/giantswarm/giantswarm/issues/16823

With this change, we actually check if the user has the required permissions or not before showing the `Insufficient permissions` screen.

In order to have access to happa, it is required for a user to have access to the `list` or `get` verb for the `clusters` resource, in the `cluster.x-k8s.io` group in at least one namespace (or the `default` namespace) that belongs to an organization that they're a part of. 

For example:
* Lucy has access to `happa` because she is a part of the `giantswarm` organization, and has permission to `list` `clusters.cluster.x-k8s.io` resources in the `org-giantswarm` namespace.
* Steve doesn't have access to `happa` because he doesn't have access to `clusters.cluster.x-k8s.io` resources in any of the namespaces he has access to.
* Mary has access to `happa` because she is a part of the GS staff, and even though there are no organizations in the installation, she has full permissions in the `default` namespace.

To achieve this functionality, this PR adds a permission system to `happa`, which is computed by creating `SelfSubjectRulesReview` resources in each organization's dedicated namespace (+ in the `default` namespace). This is later stored in the global state, as a hash map, where the hash keys are created based on the resource group, resource and resource name.

This approach is a bit heavier on the CPU side, but lighter on the memory side. I believe this is fine for us, since we can always delegate the CPU intensive tasks to a worker thread if performance becomes a problem, but we can't grow more memory 😬


src/@types/users.d.ts
```diff
@@ -34,3 +34,18 @@ interface IInvitation {
...
+  | 'create'
+  | 'update'
+  | 'patch'
+  | 'delete'
+  | '*'
+  | string;
+
+interface INamespacePermissions extends Record<string, PermissionVerb[]> {}
+
+interface IPermissionMap extends Record<string, INamespacePermissions> {}
```
axbarsan, line 51: Keeping these here for now

src/components/MAPI/organizations/AccessControl/utils.ts
```diff
@@ -214,14 +214,6 @@ export function sortPermissions(
   return ruleResult;
 }
 
-/**
- * Compute an organization namespace from the given organization name.
- * @param name
- */
-export function getOrgNamespaceFromOrgName(name: string): string {
```
axbarsan, line 221: Moved this into a common place

src/stores/batchActions.ts
```diff
@@ -65,22 +66,12 @@ export function batchedLayout(
...
-      } catch (err) {
-        const hasAccessToResources = getHasAccessToResources(getState());
-        if (!hasAccessToResources) {
-          dispatch(globalLoadFinish());
-
-          return;
-        }
-
-        throw err;
+      const user = getLoggedInUser(getState())!;
```
axbarsan, line 69: This can't be `null`. `resumeLogin()` would fail if there wasn't any user available

src/stores/main/selectors.ts
```diff
@@ -86,15 +87,60 @@ export function getLoggedInUser(state: IState): ILoggedInUser | null {
...
+        'list',
+        'cluster.x-k8s.io',
+        'clusters'
+      ):
+      case hasNamespacePermission(
+        permissions[namespace],
+        'get',
+        'cluster.x-k8s.io',
+        'clusters'
+      ):
```
axbarsan, line 125: The user needs one of these permissions to use the app

src/stores/main/utils.ts
```diff
@@ -18,3 +19,250 @@ export function mapOAuth2UserToUser(user: IOAuth2User): ILoggedInUser {
...
+}
+
+function computeResourceRules(
+  rules: authorizationv1.IResourceRule[]
+): INamespacePermissions {
+  let permissions: INamespacePermissions = {};
+
+  for (const rule of rules) {
+    for (const group of rule.apiGroups) {
+      for (const resource of rule.resources) {
```
axbarsan, line 57: Don't @ me

src/stores/main/utils.ts
```diff
@@ -18,3 +19,250 @@ export function mapOAuth2UserToUser(user: IOAuth2User): ILoggedInUser {
...
+    resourceName
+  );
+}
+
+/**
+ * Compute an organization namespace from the given organization name.
+ * This also makes the org name follow the [DNS label standard](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names).
+ * @param name
+ */
+export function getNamespaceFromOrgName(name: string): string {
```
axbarsan, line 240: This is smarter now, by transforming the organization name into a DNS friendly format

src/model/services/mapi/authorizationv1/types.ts
```diff
@@ -18,7 +18,7 @@ export interface IResourceRule {
   //  "*/foo" represents the subresource 'foo' for all resources in the specified apiGroups.
   resources: string[];
   // ResourceNames is an optional white list of names that the rule applies to.  An empty set means that everything is allowed.  "*" means all.
```
marians, line 20:
```suggestion
  // ResourceNames is an optional list of allowed names that the rule applies to.  An empty set means that everything is allowed.  "*" means all.
```

We'll try to avoid the terms "white list" and "black list" in the product. Might as well avoid it in code.

marians:
Looks sane, but it's beyond my understanding. Want to ask SIG security?

axbarsan:
> Want to ask SIG security?

We can, but why?

src/stores/main/utils.ts
```diff
@@ -18,3 +19,250 @@ export function mapOAuth2UserToUser(user: IOAuth2User): ILoggedInUser {
...
+    namespace: string,
+    review: authorizationv1.ISelfSubjectRulesReview
+  ][]
+): IPermissionMap {
+  const permissions: IPermissionMap = {};
+
+  for (const [namespace, review] of ruleReviews) {
+    if (!review.status || review.status.incomplete) {
+      continue;
+    }
```
kopiczko, line 34: I wonder if that isn't a race if that status is not created yet.

src/stores/main/selectors.ts
```diff
@@ -86,15 +87,60 @@ export function getLoggedInUser(state: IState): ILoggedInUser | null {
   return state.main.loggedInUser;
 }
 
-export function getHasAccessToResources(state: IState): boolean {
+export function selectHasPermission(
```
kopiczko, line 90: I can't find where `selectHasPermission` used. Am I missing something?

src/stores/main/utils.ts
axbarsan: Actually it seems that our type is incorrect on that one. The type in the API is not nullable. I'll update it.

https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/pkg/apis/authorization/types.go#L166

src/stores/main/selectors.ts
axbarsan: It's not actively used in this PR, but it's a must for any other feature. We need to be able to easily check if a user has the required permission to use a feature on the UI side.

Here's an example of how it can be used (with `selectHasPermission` instead of `getUserIsAdmin`):
https://github.com/giantswarm/happa/blob/dbe7906d471fe0add5eb90083e1e0503e666825d/src/components/Apps/AppsList/AppsList.tsx#L60

It's just a wrapper over `hasPermissions` that can be used easier as a Redux (global state) selector.

marians:
> We can, but why?

Your call. To get some more 👁️  on a security-relevant bit of code.

src/stores/main/selectors.ts
kopiczko: We talked, I know a bit more now :)

kopiczko (approved):

PR #2379, opened 2021-04-22
organizations: populate cluster summary stats with real data

axbarsan:
In this PR:
* The organization details `General` tab gets extracted into a separate component
* Adding 93723 requests to populate the `Cluster summary` part of the organization detail `General` tab

Note: The implementation only looks for CAPI resources, which means that on AWS some fields will have the `n/a` value.

Note 2: If you test locally, please remember that the local version of `happa` doesn't have all the instance types/vm sizes, so stuff like memory and CPU will be `n/a`.

This PR doesn't include tests, as it is already quite large. Those will be added later, after we populate the rest of the data on the page.

##### AWS

* CAPI `v1alpha3.Cluster` (there's conversion set up, so we still get `v1alpha2` resources)
* CAPI `v1alpha3.MachineDeploymentList` (there's conversion set up, so we still get `v1alpha2` resources) that belongs to each cluster

##### Azure 

* CAPI `v1alpha3.Cluster`
* CAPI `v1alpha3.AzureMachineList` that belongs to each cluster
* CAPI `v1alpha3.MachinePoolList` that belongs to each cluster
* CAPZ `v1alpha3.AzureMachinePool` referenced from the infrastructure ref for each `MachinePool`

### Preview

<details>
<summary>Before</summary>

image

</details>

<details>
<summary>After (on Azure)</summary>

image

</details>

<details>
<summary>After (on AWS)</summary>

image

</details>


src/components/MAPI/organizations/OrganizationDetailGeneral/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+      );
+    }
+  }, [clusterListError]);
+
+  const handleDelete = async () => {
+    try {
+      const client = clientFactory();
+
+      const org = await securityv1alpha1.getOrganization(
+        clientFactory(),
```
kopiczko, line 64: FMI: why not `client,`?

kopiczko (approved):

src/components/MAPI/organizations/OrganizationDetailGeneral/index.tsx
axbarsan: Nice catch 👍🏻

PR #2381, opened 2021-04-23
Remove error notification for expired credentials

marians:
This removes an error notification for the case that the user is no longer authenticated, which happens frequently when credentials expire.


axbarsan (approved):
Nice

marians:
> Nice

I gave my absolute best!

PR #2383, opened 2021-04-23
Convert signup page to typescript.

oponder:
Fixes the issue with not being signed in after creating your account too.


src/components/SignUp/SignUp.tsx
```diff
@@ -171,14 +206,15 @@ export class SignUp extends React.Component {
           password: this.state.passwordField.value,
         })
         .then((data) => {
-          const userData = {
-            username: data.username,
+          const userData: ILoggedInUser = {
```
oponder, line 209: I thought fixing the shape of this and calling loginSuccess would've fixed the problem, but unfortunately not. Will try to dig a bit more on Monday.

oponder:
This should work now!

Though I made it log in for itself, instead of trying to use the auth token that passage gives it. So we can remove the login on behalf of the user part in passage now.

oponder:
Fixes: https://github.com/giantswarm/giantswarm/issues/16698

ubergesundheit (approved):
LGTM

PR #2387, opened 2021-04-26
organizations: show stats for releases in the organization details page

axbarsan:
This also fixes https://github.com/giantswarm/giantswarm/issues/16923, since it was a matter of changing a few lines of code.

### Preview

image


src/components/MAPI/organizations/OrganizationDetailGeneral/utils.ts
```diff
@@ -331,3 +333,80 @@ function mergeClusterSummaries(
...
+  if (summary.releasesInUseCount < 1) return Promise.resolve(summary);
+
+  summary.oldestReleaseVersion = releases[0];
+  summary.newestReleaseVersion = releases[releases.length - 1];
+
+  try {
+    const [
+      oldestReleasePromise,
+      newestReleasePromise,
+    ] = await Promise.allSettled([
```
axbarsan, line 365: Using `Promise.allSettled()` instead of `Promise.all()`, so I could run the requests in parallel, but still be able to use each value if one of them decides to error out.

oponder (approved):
LGTM

PR #2389, opened 2021-04-27
organizations: display stats about apps installed in clusters

axbarsan:
This PR adds stats about apps that are installed in the clusters that belong to the selected organization:
* `Apps in use` -> the total number of apps installed in the cluster (if an app is installed multiple times, it's only counted once)
* `App deployments` -> the total number of apps installed in the cluster (also counting apps installed more than once)

The PR adds a couple very happy path integration tests for the whole stats thingy, and a few fixes for the K8s API types.

Don't get discouraged by the PR size, there are 12312958 mock responses for K8s API requests.

### Preview

image


__tests__/ClusterManagement.js
```diff
@@ -47,6 +47,7 @@ describe('ClusterManagement', () => {
 
   /************ TESTS ************/
   it('creates a v5 cluster and redirect to details view', async () => {
+    jest.setTimeout(20000);
```
axbarsan, line 50: This is a flaky test that fails 2/5 times. I increased its timeout so it won't bother us, for now

marians:
FYI @cokiengchiara - This adds a little bit of app usage summary to the new organization details page for SSO/Management API users.

marians (approved):
Tested and LGTM

src/components/MAPI/organizations/OrganizationDetailGeneral/utils.ts
```diff
@@ -410,3 +416,64 @@ export function fetchReleasesSummaryKey(
...
+ * @param clusters
+ */
+export function fetchAppsSummaryKey(
+  clusters?: capiv1alpha3.ICluster[]
+): string | null {
+  if (!clusters) return null;
+
+  const clusterKeys = clusters.map(
+    (c) => `${c.metadata.namespace}/${c.metadata.name}`
+  );
```
kopiczko, line 476: Shouldn't clusters be sorted first?

axbarsan: That would be the intelligent thing to do 😬

PR #2393, opened 2021-04-28
Use Sentry for monitoring

axbarsan:
This PR adds monitoring via Sentry. With this we can track errors, performance, and the environments that access us (OS/browser/etc.). The old way via the REST API was deprecated, but Sentry alerts are now routed via Slack, as well. In the future maybe we can set up some OpsGenie magic.

This PR also adds a default error page for when the app crashes completely and cannot be recovered (currently we get an empty blue page), and a slight change in how we display pre-releases in the footer (we now display pre-releases as `1.7.3-as3fs` instead of `as3fs`).

The reason this PR is so big is that there is now a custom `<Route />` component for dealing with `react-router` routes. With this we can make Sentry aware of our path templates (with this it displays `/organizations/:orgId` instead of `/organizations/adrian`).

Also, now as part of the release process, we also upload the sourcemaps to Sentry, so we could get spot-on stack traces.

You can log in on Sentry with your GitHub account and try it out. Keep in mind that there is domain filtering set up, so you won't be able to dispatch errors from your local machine.

<details>
<summary>When the app crashes</summary>

image

</details>


oponder:
giphy

ubergesundheit (approved):
Very nice!

marians (approved):

kopiczko (approved):

PR #2413, opened 2021-05-05
apps: install apps via MAPI if the user is logged in via SSO

axbarsan:

This PR adds functionality for installing apps via MAPI (only if the user is logged in via Dex).
The logic is copied from the REST API.

This does not include quick ingress installation (from the `Getting Started` guide). That will be done in a separate PR.


src/components/MAPI/apps/AppInstallModal.tsx
```diff
@@ -0,0 +1,393 @@
...
+  name: string;
+  versions: IAppCatalogAppVersion[];
+}
+
+interface IAppInstallModalProps {
+  app: IAppInstallModalApp;
+  selectedClusterID: string | null;
+}
+
+const AppInstallModal: React.FC<IAppInstallModalProps> = (props) => {
```
axbarsan, line 55: I duplicated the old `InstallAppModal` component, and replaced the REST API functionality with the MAPI one.

src/components/Apps/AppDetail/AppDetail.tsx
```diff
@@ -65,6 +68,12 @@ const AppDetail: React.FC = () => {
     }
   };
 
+  const user = useSelector(getLoggedInUser);
+  let AppInstallComponent = InstallAppModal;
+  if (user?.type === LoggedInUserTypes.MAPI) {
+    AppInstallComponent = AppInstallModal;
+  }
```
axbarsan, line 75: We're mounting one or the other app installation components based on if the user is logged in via SSO or not.

src/components/MAPI/apps/AppInstallModal.tsx
```diff
@@ -0,0 +1,393 @@
...
+      return;
+    }
+
+    const reader = new FileReader();
+
+    reader.onload = (e) => {
+      try {
+        const input = (e.target?.result as string) ?? '';
+        const parsedYAML = yaml.load(input) as string;
+        setValuesYAML(yaml.dump(parsedYAML));
```
axbarsan, line 186: Parsing the YAML first to make sure that it is valid

src/model/services/mapi/generic/updateResource.ts
```diff
@@ -3,12 +3,10 @@ import { HttpRequestMethods, IHttpClient } from 'model/clients/HttpClient';
...
-export function updateResource<T>(
-  client: IHttpClient,
-  auth: IOAuth2Provider,
-  url: string,
-  data: Record<string, unknown>
-) {
+export function updateResource<
+  T, // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  U extends Record<string, any> = Record<string, any>
+>(client: IHttpClient, auth: IOAuth2Provider, url: string, data: U) {
```
axbarsan, line 9: Made the same change that I previously made to the `createResource` function

src/components/MAPI/apps/utils.ts
```diff
@@ -0,0 +1,270 @@
...
+  const kubeConfigSecretName = getKubeConfigSecretName(clusterID);
+
+  const app: applicationv1alpha1.IApp = {
+    apiVersion: 'application.giantswarm.io/v1alpha1',
+    kind: 'App',
+    metadata: {
+      name: appConfig.name,
+      namespace: clusterID,
+      labels: {
+        [applicationv1alpha1.labelAppOperator]: '1.0.0',
```
kopiczko, line 177: Are you sure this is a fixed version? I thought that depends on the release.

src/components/MAPI/apps/AppInstallModal.tsx
```diff
@@ -0,0 +1,393 @@
...
+  const [namespaceError, setNamespaceError] = useState('');
+
+  const [valuesYAML, setValuesYAML] = useState<string | null>(null);
+  const [valuesYAMLError, setValuesYAMLError] = useState('');
+
+  const [secretsYAML, setSecretsYAML] = useState<string | null>(null);
+  const [secretsYAMLError, setSecretsYAMLError] = useState('');
+
+  const [loading, setLoading] = useState(false);
+  const [query, setQuery] = useState('');
```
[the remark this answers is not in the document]
axbarsan: This one is used for storing the search query for the cluster search bar.

The search functionality is visible when you navigate to the `Apps` page directly, and if you want to install an app, you will need to pick a cluster first.

image

The first part of the tuple, `query` is the current value of the input, and `setQuery` is used for changing the value. Every time the value is changed, the component is re-rendered.

Here's a more in depth explanation:
https://dmitripavlutin.com/controlled-inputs-using-react-hooks/

src/components/MAPI/apps/utils.ts
axbarsan: We have this in `kubectl gs` and the REST API, as well. After talking to Batman, seems that the reason behind it is that apps running on clusters that have older release versions (with app-operator older than `v3.0.0`) don't go through the `app-admission-controller`, so they need this label pre-set.

More on that:
https://docs.giantswarm.io/app-platform/defaulting-validation/

kopiczko: Sure, if Batman says so.

kopiczko (approved):

src/components/MAPI/apps/AppInstallModal.tsx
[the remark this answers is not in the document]
kopiczko: Thanks for the article it explains more than I expected to learn :D 

Another mistery now is `useEffect`.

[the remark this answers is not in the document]
axbarsan: With `useEffect` you can make "side effects" to components (updates bound to other changes). For example, you can execute some code:
- after the component is mounted or before it's unmounted
- whenever the component is re-rendered
- when a variable changes its value

Let's say we wanted to log something to the console whenever the `query` variable changes its value:
```typescript
useEffect(() => {
	console.log('something');
}, [query]);
```

You can find more here:
https://reactjs.org/docs/hooks-effect.html

PR #2416, opened 2021-05-07
apps: update app configuration via MAPI

axbarsan:

As with the previous PR with app installation, this one duplicates the existing necessary components, and replaces the REST API logic with the MAPI one.

### DIFF
--- a/src/components/MAPI/apps/utils.ts
+++ b/src/components/MAPI/apps/utils.ts
@@ -268,3 +268,163 @@ export function filterClusters(
     }
   });
 }
+
+export async function ensureConfigMapForApp(
+  client: IHttpClient,
+  auth: IOAuth2Provider,
+  namespace: string,
+  appName: string,
+  contents: string
+) {
+  const userConfigMapName = getUserConfigMapName(appName);
+  const configMap = await ensureAppUserConfigMap(
+    client,
+    auth,
+    namespace,
+    userConfigMapName,
+    contents
+  );
+
+  if (!configMap) return null;
+
+  const app = await applicationv1alpha1.getApp(
+    client,
+    auth,
+    namespace,
+    appName
+  );
+  app.spec.userConfig = {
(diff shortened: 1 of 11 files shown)


src/components/MAPI/apps/utils.ts
```diff
@@ -268,3 +268,209 @@ export function filterClusters(
...
+    );
+    await corev1.deleteConfigMap(client, auth, configMap);
+  } catch (err) {
+    if (
+      !metav1.isStatusError(
+        (err as GenericResponse).data,
+        metav1.K8sStatusErrorReasons.NotFound
+      )
+    ) {
+      return Promise.reject(err);
```
axbarsan, line 380: Maybe the `ConfigMap` doesn't even exist, so we don't need to delete it ⭐

src/components/MAPI/apps/utils.ts
```diff
@@ -268,3 +268,209 @@ export function filterClusters(
...
+  const userConfigMapName = getUserConfigMapName(appName);
+  const configMap = await ensureAppUserConfigMap(
+    client,
+    auth,
+    namespace,
+    userConfigMapName,
+    contents
+  );
+
+  if (!configMap) return null;
```
kopiczko, line 288: When `!configMap` is true? what happens when `null` is returned?

kopiczko (approved):

src/components/MAPI/apps/utils.ts
axbarsan: In this case, `configMap` can be either `corev1.IConfigMap` or `null`, so `!configMap` is `true` when `configMap` is `null`.

`null` is returned when creating a `ConfigMap` resource is not necessary: If the given contents is an empty string and an existing `ConfigMap` does not exist, then we skip creating one.

If an empty string is supplied but a `ConfigMap` resource already exists, then we'll update it to have empty contents, and we won't return `null`

PR #2421, opened 2021-05-11
apps: fetch apps in cluster via MAPI if the user is logged in via SSO

axbarsan:

This PR duplicates the existing `ClusterApps` component and adds MAPI-specific logic to the new one. It behaves the same as the REST API implementation, with a couple differences:

- We display if an app was deleted
image

Hover over the `deleted about 5 hours ago` label
image

- Preinstalled apps now get a cheeky loading animation
image


src/components/Cluster/ClusterDetail/ClusterApps.tsx
```diff
@@ -322,25 +317,16 @@ const ClusterApps: React.FC<IClusterAppsProps> = ({
...
-          <AppDetailsModal
-            // Instead of just assigning the selected app to the state of this component,
-            // this ensures any updates to the apps continue to flow down into the modal.
-            app={appToDisplay}
-            clusterId={clusterId}
-            onClose={hideAppModal}
-            visible={detailsModalIsVisible}
-          />
-        ))}
+      {appToDisplay && (
```
axbarsan, line 320: No longer need for this logic. `<ClusterDetailApps />` will be rendered instead of `<ClusterApps />` when the user is logged in via SSO

src/components/MAPI/apps/AppDetailsModal/index.tsx
```diff
@@ -167,7 +167,7 @@ const AppDetailsModal: React.FC<IAppDetailsModalProps> = ({
       );
 
       mutateApp(updatedApp);
-      // TODO(axbarsan): Mutate apps list, once that request exists.
+      mutate(applicationv1alpha1.getAppListKey({ namespace: clusterId }));
```
axbarsan, line 170: This makes sure that we re-execute the app fetching request after we delete an app, to get the updated list of apps

src/components/MAPI/apps/ClusterDetailApps.tsx
```diff
@@ -0,0 +1,314 @@
...
+
+import ClusterDetailAppLoadingPlaceholder from './ClusterDetailAppLoadingPlaceholder';
+import { filterUserInstalledApps, mapDefaultApps } from './utils';
+
+const LOADING_COMPONENTS = new Array(6).fill(0).map((_, idx) => idx);
+
+function formatAppVersion(appMeta: AppConstants.IAppMetaApp) {
+  const { version } = appMeta;
+
+  // TODO(axbarsan): Handle K8s version EOL dates once available.
```
axbarsan, line 36: There is no way to get them in the MAPI world yet

src/components/MAPI/apps/utils.ts
```diff
@@ -474,3 +476,72 @@ export async function updateAppVersion(
...
   return applicationv1alpha1.updateApp(client, auth, app);
 }
+
+export function filterUserInstalledApps(
+  apps: applicationv1alpha1.IApp[],
+  supportsOptionalIngress: boolean
+): applicationv1alpha1.IApp[] {
+  return apps.filter((app) => {
+    switch (true) {
+      case typeof app.metadata.deletionTimestamp !== 'undefined':
```
axbarsan, line 486: Don't show deleting apps, so once we delete an app, it doesn't just hang around

src/stores/nodepool/utils.ts
```diff
@@ -47,3 +47,18 @@ export function supportsAlikeInstances(
       return false;
   }
 }
+
+export function supportsOptionalIngress(
```
axbarsan, line 51: Moved this logic into a separate function, so I could reuse it here

src/stores/cluster/utils.ts
```diff
@@ -471,3 +459,18 @@ export function reconcileClustersAwaitingUpgrade(
 
   return awaitingUpgrade;
 }
+
+export function supportsOptionalIngress(
```
axbarsan, line 463: Moved this logic into a separate function, so I could reuse in this PR

src/components/MAPI/apps/utils.ts
kopiczko: But it may be too much into debugging :)

kopiczko (approved):

marians:
Taking a look now

marians:
I see that the REST API endpoint `/v5/clusters/{clusterid}/apps/` is still being called when accessing a cluster details page. Are you aware? Is that for another PR?

src/components/MAPI/apps/ClusterDetailApps.tsx
marians: There is https://github.com/giantswarm/roadmap/issues/336 for that now.

marians:
Since showing apps that are being deleted is new, let's take a moment to think about this.

- When I delete an app, I see it disappear immediately from the list of installed apps. I don't see a state saying "deleted X ago". Is that an optimistic update, or are we in fact waiting for the confirmation?
- If we display an app that's being deleted, we should make sure that it cannot be edited any more. (Clicking should have no effect IMO). Also it should be visually distinc from a normal installed app. For clusters on the home page, we use a different background color and also adapt the font color. I'd like to re-use that pattern here.

axbarsan:
> * When I delete an app, I see it disappear immediately from the list of installed apps. I don't see a state saying "deleted X ago". Is that an optimistic update, or are we in fact waiting for the confirmation?

If it disappears immediatelly it means it was deleted right away. You only see the `deleted ...` label if the resource exists and it has a `deletionTimestamp` property

> * If we display an app that's being deleted, we should make sure that it cannot be edited any more. (Clicking should have no effect IMO). Also it should be visually distinct from a normal installed app. For clusters on the home page, we use a different background color and also adapt the font color. I'd like to re-use that pattern here.

👍🏻

axbarsan:
> If we display an app that's being deleted, we should make sure that it cannot be edited any more. (Clicking should have no effect IMO). Also it should be visually distinct from a normal installed app. For clusters on the home page, we use a different background color and also adapt the font color. I'd like to re-use that pattern here.

I updated the screenshots, please take a look

marians (approved):
Nice!

PR #2422, opened 2021-05-11
Faster pre commit hook

marians:
This PR modifies the pre-commit hook so that it only executes the time-consuming actions if there are modifications to .ts(x) or .js files.


axbarsan (approved):
LGTM if it works for you

PR #2430, opened 2021-05-17
auth: fix memory leak when renewing the auth token fails

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/16830

This PR fixes a memory leak when renewing the auth token fails. It now also displays a flash message when that happens.

### Preview

image

### DIFF
--- a/src/lib/OAuth2/OAuth2.ts
+++ b/src/lib/OAuth2/OAuth2.ts
@@ -135,12 +135,11 @@ class OAuth2 implements IOAuth2Provider {
     let origUser = await this.userManager.getUser();
     if (!origUser) return null;
 
-    // If user is already expired, renew his authentication.
+    // If user is already expired, renew their authentication.
     if (origUser.expired) {
       origUser = await this.userManager.signinSilent();
     }
 
-    this.userManager.events.load(origUser);
     const newUser = getUserFromOIDCUser(origUser);
 
     return newUser;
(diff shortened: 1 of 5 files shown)


src/lib/OAuth2/OAuth2.ts
```diff
@@ -135,12 +135,11 @@ class OAuth2 implements IOAuth2Provider {
     let origUser = await this.userManager.getUser();
     if (!origUser) return null;
 
-    // If user is already expired, renew his authentication.
+    // If user is already expired, renew their authentication.
     if (origUser.expired) {
       origUser = await this.userManager.signinSilent();
     }
 
-    this.userManager.events.load(origUser);
```
axbarsan, line 143: These are actually called internally, there's no need to call them again

src/stores/main/middleware.ts
```diff
@@ -1,14 +1,18 @@
...
 import { getLoggedInUser } from 'stores/main/selectors';
 
 import { LoggedInUserTypes } from './types';
 
 export function mainAuthMiddleware(auth: IOAuth2Provider): Middleware {
   // eslint-disable-next-line consistent-return
   return (store) => (next) => async (action) => {
+    if (action.type?.startsWith('LOGOUT_')) {
+      return next(action);
+    }
```
axbarsan, line 14: If we logged the user out after failing to renew the token, then the logout actions would be dispatched. If we don't ignore them, then we'll try to renew the user again, which would fail, and we'll dispatch the logout actions again, resulting in a memory leak.

oponder (approved):
LGTM!

PR #2436, opened 2021-05-19
Revert copy path for images

marians:
Fixes a problem with missing images, as they were placed in the wrog path under `/www`.

Also consolidates the Sentry ignore configuration in the webpack config.


marians:
Current state (as of commit https://github.com/giantswarm/happa/pull/2436/commits/5d61a058d9aac0bc11ca6249c402001de37c392f) is definitely not right

image

axbarsan (approved):

PR #2441, opened 2021-05-20
apps: implement app browsing via MAPI for SSO users

axbarsan:


src/components/Apps/Apps.tsx
```diff
@@ -17,7 +25,12 @@ const Apps: React.FC = () => {
     >
       <Switch>
         <Route exact path={AppsRoutes.AppDetail} component={AppDetail} />
-        <Route path={AppsRoutes.Home} component={AppsList} />
+
+        {user?.type === LoggedInUserTypes.MAPI && provider !== Providers.KVM ? (
```
axbarsan, line 29: KVM can't sit with us

src/components/MAPI/apps/AppList/index.tsx
```diff
@@ -0,0 +1,210 @@
...
+  const appCatalogListIsLoading =
+    typeof appCatalogList === 'undefined' && appCatalogListIsValidating;
+
+  const [searchQuery, setSearchQuery] = useState('');
+  const debouncedSearchQuery = useDebounce(
+    searchQuery,
+    SEARCH_THROTTLE_RATE_MS
+  );
+
+  const [selectedCatalogs, dispatchSelectedCatalogs] = useReducer(
```
axbarsan, line 116: Fun fact: This functionality didn't work using `useState()`, because that batched together multiple state updates (such as what happens when you select or deselect all catalogs), which resulted in the map not being updated correctly.

src/components/MAPI/apps/AppList/index.tsx
```diff
@@ -0,0 +1,210 @@
...
+    if (isAdmin) return {};
+
+    return {
+      labelSelector: {
+        matchingLabels: {
+          [applicationv1alpha1.labelCatalogVisibility]: 'public',
+          [applicationv1alpha1.labelCatalogType]: 'stable',
+        },
+      },
+    };
```
axbarsan, line 76: Admins can see everything, non-admins can only see public and stable catalogs

src/components/MAPI/apps/AppList/index.tsx
```diff
@@ -0,0 +1,210 @@
...
+      Object.keys(selectedCatalogs).length > 0 ||
+      typeof prevAppCatalogList !== 'undefined' ||
+      typeof appCatalogList === 'undefined'
+    ) {
+      return;
+    }
+
+    // Pre-select all catalogs except `helm-stable`.
+    for (const catalog of appCatalogList.items) {
+      if (catalog.metadata.name === 'helm-stable') continue;
```
axbarsan, line 133: Pre-select all catalogs except `helm-stable`, as that is HYOOOOGE

src/components/MAPI/apps/AppList/index.tsx
```diff
@@ -0,0 +1,210 @@
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [appCatalogList?.items, prevAppCatalogList]);
+
+  const {
+    data: appCatalogIndexAppList,
+    isValidating: appCatalogIndexAppListIsValidating,
+  } = useSWR<IAppCatalogIndexAppList, GenericResponse>(
+    getAppCatalogsIndexAppListKey(appCatalogList?.items),
+    // TODO(axbarsan): Find a more elegant solution for passing `fetch` here.
+    () => getAppCatalogsIndexAppList(fetch, auth, appCatalogList!.items)
```
axbarsan, line 151: Had to use `fetch` instead of our HTTP client, since our client doesn't support enforcing CORS from the client-side, which we need for fetching catalogs from GitHub.

src/components/MAPI/apps/AppList/utils.tsx
```diff
@@ -0,0 +1,327 @@
...
+) {
+  const normalizedQuery = searchQuery.trim().toLowerCase();
+
+  return indexApps.filter((indexApp) => {
+    if (!selectedAppCatalogs.hasOwnProperty(indexApp.catalogName)) return false;
+
+    if (normalizedQuery.length < 1) return true;
+
+    if (indexApp.versions.length < 1) return false;
+    const version = indexApp.versions[0];
```
axbarsan, line 200: This is slightly different from the GS API implementation: Here we only search in the latest version, not in all the versions, to improve performance.

src/components/MAPI/apps/AppList/utils.tsx
```diff
@@ -0,0 +1,327 @@
...
+ */
+async function fetchAppCatalogIndex(
+  fetchFunc: (input: RequestInfo, init?: RequestInit) => Promise<Response>,
+  _auth: IOAuth2Provider,
+  catalog: applicationv1alpha1.IAppCatalog
+): Promise<IAppCatalogIndex> {
+  const url = normalizeAppCatalogIndexURL(catalog.spec.storage.URL);
+
+  // Enforce client-side CORS.
+  const response = await fetchFunc(url, { mode: 'cors' });
```
axbarsan, line 254: We need to fetch the catalog index from GitHub, as apps include more info than via the K8s API (e.g. images)

src/components/MAPI/apps/AppList/utils.tsx
```diff
@@ -0,0 +1,327 @@
...
+export async function getAppCatalogsIndexAppList(
+  fetchFunc: (input: RequestInfo, init?: RequestInit) => Promise<Response>,
+  auth: IOAuth2Provider,
+  catalogs: applicationv1alpha1.IAppCatalog[]
+): Promise<IAppCatalogIndexAppList> {
+  const requests = catalogs.map((catalog) =>
+    fetchAppCatalogIndex(fetchFunc, auth, catalog)
+  );
+
+  const responses = await Promise.allSettled(requests);
```
axbarsan, line 297: We don't want the whole thing to throw if one of the catalogs fails to load.

src/components/MAPI/apps/AppList/utils.tsx
```diff
@@ -0,0 +1,327 @@
...
+  for (let i = 0; i < responses.length; i++) {
+    const response = responses[i];
+
+    if (response.status === 'rejected') {
+      const catalogName = catalogs[i].metadata.name;
+      indexList.errors[catalogName] = extractErrorMessage(response.reason);
+      continue;
+    }
+
+    indexList.items.push(...Object.values(response.value.entries));
```
axbarsan, line 312: Get error messages from requests that failed and get data from requests that didn't

src/stores/appcatalog/utils.ts
```diff
@@ -0,0 +1,19 @@
+export function normalizeAppCatalogIndexURL(url: string): string {
```
axbarsan, line 1: Extracted this out of the action, so I could re-use it

marians:
I like the fact that we show that loading animation!

Here is something regarding loading catalog YAML: I see that we are loading catalogs already once when opening happa. And we are requesting them again each time when opening the Apps section. With a functioning cache, that is fast. However I wonder if this is intentional or not.

Also there are some exceptions logged:

```
Warning: Failed prop type: Invalid prop `appsIsLoading` of type `string` supplied to `AppsList`, expected `boolean`.
```

```
Warning: Failed prop type: Invalid prop `isLoading` of type `string` supplied to `Toolbar`, expected `boolean`.
```

axbarsan:
> Here is something regarding loading catalog YAML: I see that we are loading catalogs already once when opening happa. And we are requesting them again each time when opening the Apps section. With a functioning cache, that is fast. However I wonder if this is intentional or not.

Removing requests will be done after all the app features are migrated to MAPI, since there are still features left that depend on them. It's a different checkbox in [the issue](https://github.com/giantswarm/roadmap/issues/328)

> Also there are some exceptions logged:
> 
> ```
> Warning: Failed prop type: Invalid prop `appsIsLoading` of type `string` supplied to `AppsList`, expected `boolean`.
> ```
> 
> ```
> Warning: Failed prop type: Invalid prop `isLoading` of type `string` supplied to `Toolbar`, expected `boolean`.
> ```

This is a bug in the `swr` library, we also have it on the org details page. I'm still trying to find a way to fix it in the library.

marians:
Found an unexpected behaviour:

1. Select only one catalog to show in the apps list
2. Navigate to one app from the list
3. Use the back button

In the current REST API implementation the previous catalog selection is persisted, which is the intended behaviour.

Here, the catalog selection is reset.

marians:
When I select an option from the sort drop down, the menu stays visble, when it should disappear.

https://user-images.githubusercontent.com/273727/119119552-9456ab00-ba2b-11eb-87bc-56110e7fccb2.mov

axbarsan:
> Found an unexpected behaviour:
> 
> 1. Select only one catalog to show in the apps list
> 2. Navigate to one app from the list
> 3. Use the back button
> 
> In the current REST API implementation the previous catalog selection is persisted, which is the intended behaviour.
> 
> Here, the catalog selection is reset.

That is because previously the selected catalogs were stored in the global state, which means that they were the same even if the route changed. Right now we handle those in the actual view component.

We might be able to bring that back once app details are also transitioned to MAPI, but until then, we have to live with it

axbarsan:
> When I select an option from the sort drop down, the menu stays visble, when it should disappear.
> 
>  Screen.Recording.2021-05-21.at.11.54.30.mov

That is a bug in the original implementation. Please open a separate issue for it.

marians (approved):

PR #2442, opened 2021-05-21
Fix uncaught error when fetching user invitations fails

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/17359


src/stores/user/actions.ts
```diff
@@ -192,9 +190,9 @@ export function invitationsLoad(): ThunkAction<
       dispatch({
         type: INVITATIONS_LOAD_ERROR,
       });
-
-      return Promise.reject(err);
```
axbarsan, line 196: No need to throw this again, we already display an error message.

corest (approved):
This size of PR in typescript I can read :)

PR #2443, opened 2021-05-21
Close app sorting dropdown after picking a value

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/17360

### DIFF
--- a/src/components/UI/Controls/DropdownMenu.js
+++ b/src/components/UI/Controls/DropdownMenu.js
@@ -1,5 +1,5 @@
 import PropTypes from 'prop-types';
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import styled from 'styled-components';
 
 /*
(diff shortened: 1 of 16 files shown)


src/components/UI/Controls/DropdownMenu.js
```diff
@@ -64,23 +64,23 @@ const MenuWrapper = styled.div`
...
 function DropdownMenu(props) {
   const [isOpen, setIsOpen] = useState(false);
-  let timeOutId = null;
+  const timeOutId = useRef(null);
 
   useEffect(() => {
-    return function cleanup() {
-      clearTimeout(timeOutId);
+    return () => {
+      clearTimeout(timeOutId.current);
```
axbarsan, line 71: The problem was that our timeout was cleared before it managed to fire `onBlurHandler`, because the component was re-rendered

kopiczko (approved):

PR #2449, opened 2021-05-25
apps: take into consideration that the ingress app can be in another catalog

axbarsan:
Fixes https://github.com/giantswarm/giantswarm/issues/17323

Turns out the ingress app can be coming from other catalogs, as well (not only `giantswarm`). For example, on `c68pn`, the ingress app comes from the `default` catalog.


marians (approved):
LGTM if it works for you

PR #2457, opened 2021-05-26
apps: persist selected catalogs when navigating between app pages

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/17361

With this PR, the selected catalogs in the sidebar of the `Apps` page will be persisted between route changes (disclaimer: only if we change routes between the app-specific routes - app list and app details, if we navigate outside of that, the selection will be reset).

### DIFF
--- a/src/components/MAPI/apps/AppsProvider.tsx
+++ b/src/components/MAPI/apps/AppsProvider.tsx
@@ -14,6 +14,8 @@ export interface IAppsContext {
   selectedCatalogs: SelectedCatalogs;
   selectCatalog: (catalogName: string) => void;
   deselectCatalog: (catalogName: string) => void;
+  searchQuery: string;
+  setSearchQuery: (value: string) => void;
 }
 
 const appsContext = createContext<IAppsContext | null>(null);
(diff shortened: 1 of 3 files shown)


src/components/MAPI/apps/AppsProvider.tsx
```diff
@@ -0,0 +1,103 @@
...
+  createContext,
+  useCallback,
+  useContext,
+  useMemo,
+  useReducer,
+} from 'react';
+
+export type SelectedCatalogs = Record<string, boolean>;
+
+export interface IAppsContext {
```
axbarsan, line 13: Made this a little more generic, so we could add some more common app functionality later

marians:
Seeing some exceptions which are supposedly coming from React Dev tools:

```
Warning: Failed prop type: Invalid prop `appsIsLoading` of type `string` supplied to `AppsList`, expected `boolean`.
Warning: Failed prop type: Invalid prop `isLoading` of type `string` supplied to `Toolbar`, expected `boolean`.
```

axbarsan:
> Seeing some exceptions which are supposedly coming from React Dev tools:
> 
> ```
> Warning: Failed prop type: Invalid prop `appsIsLoading` of type `string` supplied to `AppsList`, expected `boolean`.
> Warning: Failed prop type: Invalid prop `isLoading` of type `string` supplied to `Toolbar`, expected `boolean`.
> ```

I already answered that here https://github.com/giantswarm/happa/pull/2441#issuecomment-845823182

marians:
Regarding persisting the list page state

1. The search term should be persisted, too. When I search for "calico" and then navigate into the calico app, then hit Back, I should get back to the search result for "calico".
2. The list sorting selection should be persisted, too.

axbarsan:
> Regarding persisting the list page state
> 
> 1. The search term should be persisted, too. When I search for "calico" and then navigate into the calico app, then hit Back, I should get back to the search result for "calico".
> 2. The list sorting selection should be persisted, too.

Please take a look 👍🏻

marians (approved):
Much appreciated!

PR #2459, opened 2021-05-27
apps: prevent infinite loading spinners for apps that don't have an app version or homepage set

axbarsan:


marians:
Shouldn't we use the `NotAvailable` UI component instead of an `n/a` string?

marians (approved):

axbarsan:
Actually, this doesn't behave properly. `n/a` is displayed instead of loading flags. I'll try to come up with a smarter solution

PR #2461, opened 2021-05-27
apps: use the correct catalog name when installing app in workload cluster

axbarsan:
As part of the: `can you break happa every day of the week?` challenge


ubergesundheit (approved):
Nice, thanks for the quick fix!

PR #2462, opened 2021-05-27
Prevent throwing an error if the app metadata is not loaded yet and the footer version is invalid

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/17418

While the app is loading `/metadata.json`, this version will be invalid


marians (approved):
LGTM if it kills the exception

PR #2464, opened 2021-05-27
Refactor HTTP client to use the browser's `fetch` function

axbarsan:
Doing this for 2 reasons:
- Removing a large dependency
- `axios` doesn't support enforcing CORS on the client side (which is what we need for fetching app catalog indexes from GitHub)


package.json
```diff
@@ -213,6 +212,7 @@
     "html-webpack-plugin": "^5.3.1",
     "husky": "6.0.0",
     "imports-loader": "^2.0.0",
+    "isomorphic-fetch": "^3.0.0",
```
axbarsan, line 215: Necessary in tests, since `node-fetch` doesn't have the `Request` and `Response` browser objects.

src/model/clients/GenericResponseError.ts
```diff
@@ -0,0 +1,90 @@
+import { StatusCodes } from 'shared/constants';
+
+import { IHttpResponseConfig } from './GenericResponse';
+import { HttpBody } from './HttpClient';
+
+export class GenericResponseError<T = HttpBody> extends Error {
```
axbarsan, line 6: Necessary so we could throw actual `Error` objects when requests fail. Otherwise we get pretty bad stack traces in Sentry.

marians (approved):
Seems to work as expected 👍

PR #2473, opened 2021-05-28
Add support for experiments

axbarsan:

As I will start working on the clusters migration to MAPI soon, I want to hide all those features behind a feature flag until they're 100% ready to use. However, the problem with feature flags is that if you want to experiment with the new functionalities, you have to enable the feature for the whole installation. This could negatively impact the experience for other users. Also, with that approach, you can't try out features on customer installations (for obvious reasons).

### Introducing `Experiments`

This is a new page only visible to admins, which you can access from the user dropdown menu.

<details>
<summary>Preview</summary>

image

</details>

On this page you can toggle specific features on or off.

<details>
<summary>Preview</summary>

image

</details>

As the disclaimer in the screenshot says, these features will only be enabled for yourself, and only on the current installation.

The values of these flags are stored using the `localStorage` Browser API, so they are be persistent between sessions.
Currently, only `Customer Single Sign-On` is visible in the `Experiments` section, but clusters will follow up shortly.


src/shared/featureFlags.ts
```diff
@@ -0,0 +1,95 @@
+export interface IFeatureFlag {
```
axbarsan, line 1: This is how you can configure a flag

marians (approved):
Awesome!

PR #2477, opened 2021-05-31
apps: use the correct chart name when installing an app

axbarsan:
As seen on https://gigantic.slack.com/archives/CRZN9GLJW/p1622460313091900


pipo02mix (approved):
LGTM

PR #2478, opened 2021-06-01
Set JSON content type header for Passage requests

axbarsan:
This bug was created during the HTTP Client refactor. As seen on https://gigantic.slack.com/archives/CLPMFRVU6/p1622535930048700

### DIFF
--- a/src/lib/passageClient.ts
+++ b/src/lib/passageClient.ts
@@ -118,9 +118,6 @@ class Passage {
         invite_token: params.inviteToken,
         password: params.password,
       },
-      headers: {
-        ContentType: 'application/json',
-      },
     });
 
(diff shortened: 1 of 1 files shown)


src/lib/passageClient.ts
```diff
@@ -118,9 +118,6 @@ class Passage {
         invite_token: params.inviteToken,
         password: params.password,
       },
-      headers: {
-        ContentType: 'application/json',
-      },
```
axbarsan, line 123: `Content-Type: application/json` is the default, if no headers are specified. If there are any headers specified, then it will get overriden

marians (approved):
LGTM if it works for you

PR #2480, opened 2021-06-02
Switch to the `swc` compiler

axbarsan:
This PR switches the compiler used in building for development/production to `swc`, a performant alternative to `babel`, written in Rust.

### Benchmarks for building (`yarn dist`)

* Before: 50.211 seconds
* After: 31.331 seconds

All benchmarks were ran using `hyperfine`, and the displayed values are averages calculated after 10 test runs.

### Bundle

Also with this PR, the bundle shipped to browsers gets slimmer, because we no longer ship polyfills for ancient browsers.

* Before: 3.58 MB
* After: 2.75 MB

Unfortunately, tests and `storybook` still need `babel` in order to function properly 😭 `storybook` is 100% dependent on `babel`, and tests were working really well on my machine, but were broken in CI. This would require extra investigation, but I already spent too much time playing with this, so I'll leave it for later.


webpack.common.js
```diff
@@ -131,14 +131,33 @@ module.exports = {
...
+                legacyDecorator: true,
+                decoratorMetadata: true,
+                react: {
+                  runtime: 'automatic',
+                },
+              },
+              externalHelpers: true,
+            },
+            env: {
+              targets: '> 0.25%, not dead',
```
axbarsan, line 157: Now only including features and polyfills necessary for browser versions used by more than 0.25% of the global population

webpack.common.js
```diff
@@ -131,14 +131,33 @@ module.exports = {
...
-      },
-      {
-        test: /\.ts(x?)$/,
-        exclude: /node_modules/,
-        use: ['babel-loader'],
+        use: {
+          loader: 'swc-loader',
+          options: {
+            jsc: {
+              target: 'es2015',
```
axbarsan, line 140: It's 2021, we can graduate from es5

package.json
```diff
@@ -72,12 +72,6 @@
     "jsxSingleQuote": true,
     "trailingComma": "es5"
   },
-  "jest": {
-    "testEnvironment": "node",
-    "setupFilesAfterEnv": [
-      "./src/setupTests.js"
-    ]
-  },
```
axbarsan, line 80: This was overwritten by `jest.config.js`

webpack.common.js
marians: How does this effect the JS size?

marians (approved):
Speedups and resource saving is always welcome!

We are still using a fairly large worker or the build step in CircleCI. 
image

Can this be reduced, without making the build slower?

webpack.common.js
axbarsan: Less polyfills = less code necessary to be shipped

Older browsers don't have support for all the recent JS features, so they need to have polyfills (code that implements features that aren't supported). This adds weight to the shipped bundle

axbarsan:
> Speedups and resource saving is always welcome!
> 
> We are still using a fairly large worker or the build step in CircleCI.
> image
> 
> Can this be reduced, without making the build slower?

I believe that's still necessary for tests, but I will try a less powerful one

axbarsan:
I switched to the default resource class for building, and to the `large` class for tests, and everything in bearable 👍🏻

PR #2485, opened 2021-06-08
clusters: add cluster widgets for worker node, app, key pair stats

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/9077

This PR adds 3 new widgets to the `Cluster detail` page:
* Worker nodes
* Apps
* Key pairs

### Preview

<details>
<summary>With some data</summary>

image

</details>

<details>
<summary>Loading state</summary>

image

</details>

<details>
<summary>Values not available</summary>

image

</details>

<details>
<summary>Nothing to display</summary>

image

This looks a bit goofy, but it's not that bad, so I'd rather not add the extra complexity for making it look 100% like the spec.

</details>


src/components/UI/Display/MAPI/clusters/ClusterDetail/ClusterDetailWidget/index.tsx
```diff
@@ -7,14 +7,14 @@ const Title = styled(Text)`
   text-transform: uppercase;
 `;
 
-interface IClusterDetailOverviewProps
+interface IClusterDetailWidgetProps
```
axbarsan, line 10: Typo 😬

marians:
The switch from the loading state, where we show placeholders for stats, to then showing no stats and instead showing the widget with their empty state content, creates a somewhat strange effect.

marians (approved):

marians:
BTW, didn't intend to block this.

For c68pn on gollum, I get this display:

image

What's the reason that there are no CPU and RAM figures?

axbarsan:
> BTW, didn't intend to block this.
> 
> For c68pn on gollum, I get this display:
> 
> image
> 
> What's the reason that there are no CPU and RAM figures?

if you are testing it locally, you don't have all the VM Sizes defined in `index.ejs`, so the app doesn't know how many CPUs or how much RAM is in the worker nodes

PR #2487, opened 2021-06-10
clusters: add Control Plane nodes widget to the MAPI cluster detail page

axbarsan:
Towards giantswarm/giantswarm#9077

This PR adds a new widget for displaying info about a cluster's Control Plane nodes, via the Management API. In the spec it was originally called `Master nodes`, but from my understanding, we changed that wording to `Control Plane nodes` everywhere.

## Preview

<details>
<summary>1/1 nodes ready</summary>

image

</details>

<details>
<summary>0/3 nodes ready</summary>

image

</details>

<details>
<summary>1/3 nodes ready</summary>

image

</details>

<details>
<summary>3/3 nodes ready</summary>

image

</details>

<details>
<summary>loading state</summary>

image

</details>

<details>
<summary>nodes not available (e.g. an error occured)</summary>

image

</details>


src/components/MAPI/types.ts
```diff
@@ -0,0 +1,5 @@
+import * as capzv1alpha3 from 'model/services/mapi/capzv1alpha3';
+
+export type ControlPlaneNode = capzv1alpha3.IAzureMachine;
+
+export type ControlPlaneNodeList = capzv1alpha3.IAzureMachineList;
```
axbarsan, line 5: Created a type union for all control plane node types

src/components/MAPI/utils.ts
```diff
@@ -181,6 +183,7 @@ export async function fetchMasterListForCluster(
         labelSelector: {
           matchingLabels: {
             [capiv1alpha3.labelCluster]: cluster.metadata.name,
+            [capzv1alpha3.labelControlPlane]: 'true',
```
axbarsan, line 186: In the current REST API logic, this is also necessary in order for an `AzureMachine` to be a Control Plane node.

src/model/services/mapi/capiv1alpha3/key.ts
```diff
@@ -41,26 +42,32 @@ export function getKubernetesAPIEndpointURL(
   return `https://${hostname}`;
 }
 
+interface IConditionGetter {
+  status?: {
+    conditions?: ICondition[];
+  };
+}
+
 export function getCondition(
```
axbarsan, line 51: Thanks to duck typing we can make these condition helpers more generic, and use them for multiple resources

marians:
> In the spec it was originally called `Master nodes`, but from my understanding, we changed that wording to `Control Plane nodes` everywhere.

That's right, there is https://github.com/giantswarm/giantswarm/issues/14964

I'm fine with starting here. We still might have to do a bit more in happa.

I noticed that I would like to change the spec a bit. How about using a `<code>Ready</code>` literal for the state name, to indicate that it's about the Kubernetes node state `Ready` and not any other readiness definition.

marians (approved):

PR #2488, opened 2021-06-10
build: give the build process a bit of love

axbarsan:
With this PR:
- We compile code in tests using `swc` instead of `babel`, so they run faster
- We switch to a smarter caching approach for built assets
- We get rid of warnings during the build process

With the new caching approach we split the whole app into 3 parts, that are all loaded separately:
* runtime (`webpack` runtime)
* vendors (dependencies from `node_modules`: 1x `.js` and 1x `.css` file)
* main (happa's own code: 1x `.js` and 1x `.css` file)

The nice thing about that is that all files will have the same names between builds, unless some of that code changed:
- runtime file would change if we update `webpack` or change some important configuration
- vendor files would change if we change our dependencies
- main files would change if we change some logic in our app

This means that when we deploy a new release, users will only need to fetch the files that have changed, and not the full bundle.

For example:
> We release a new version with a bug fix in the Cluster Overview page. Users will only need to fetch the main JS file, and not the runtime or vendors, or even the main CSS file. They will fetch a 700 KB file instead of a 2.5 MB bundle.
 
In the future, if we implement code splitting and load specific parts of the app dynamically, we can improve this even more, by only loading code necessary for a specific page.


webpack.prod.js
```diff
@@ -39,17 +41,23 @@ const plugins = [
   new CopyPlugin({
     patterns: [
       { from: 'src/metadata.json', to: 'metadata.json' },
-      { from: 'src/images', to: 'images' },
+      {
+        from: 'src/images',
+        globOptions: {
+          ignore: ['**/*.sketch'],
```
axbarsan, line 45: The app icon template ended up in the build assets

webpack.prod.js
```diff
@@ -66,13 +74,23 @@ module.exports = merge(common, {
     reasons: true,
   },
   optimization: {
-    // Terser is a substitution for AgressiveMergingPlugin
     minimizer: [
       new TerserPlugin({
         extractComments: 'some',
       }),
-      new OptimizeCSSAssetsPlugin({}),
+      new CssMinimizerPlugin(),
```
axbarsan, line 79: To fix this note https://github.com/orgs/giantswarm/projects/85#card-62021202

.circleci/config.yml
```diff
@@ -5,7 +5,7 @@ orbs:
 defaults: &defaults
   working_directory: ~/happa
   docker:
-    - image: circleci/node:14-buster
+    - image: circleci/node:16-buster
```
axbarsan, line 8: To support running tests with `swc`

src/components/UI/Display/MAPI/clusters/ClusterDetail/ClusterDetailWidget/__tests__/__snapshots__/index.tsx.snap
```diff
@@ -5,7 +5,7 @@ exports[`ClusterDetailWidget renders a inline widget 1`] = `
   class="StyledGrommet-sc-19lkkz7-0 hSiATR"
 >
   <div
-    class="ThemeProvider__AppWrapper-sc-1qc9t9l-0 bLmXgB"
+    class="sc-gKckTs izRacH"
```
axbarsan, line 8: We no longer have the compiler plugin for these, not a big deal

webpack.common.js
```diff
@@ -120,13 +120,13 @@ const makeFeatureFlags = () => {
 
 module.exports = {
   amd: false,
-  entry: ['react-hot-loader/patch', './src/components/index.tsx'],
```
axbarsan, line 123: This isn't necessary with recent versions of `react-hot-loader`

webpack.common.js
```diff
@@ -256,6 +241,8 @@ module.exports = {
       fs: false,
       path: false,
     },
+    symlinks: false,
+    cacheWithContext: false,
```
axbarsan, line 245: speed++

webpack.prod.js
```diff
@@ -81,6 +97,10 @@ module.exports = merge(common, {
         test: /\.css$/,
         use: [MiniCssExtractPlugin.loader, 'css-loader'],
       },
+      {
+        test: /\.(scss|sass)$/,
+        use: [MiniCssExtractPlugin.loader, 'css-loader', 'sass-loader'],
```
axbarsan, line 102: `.sass` files deserve some love, too

marians (approved):
Choremaster! LGTM, but didn't test.

PR #2489, opened 2021-06-11
Only run pre-commit hook for staged files

axbarsan:
With the changes in the PR, the pre-commit hook runs miles faster than before. The linter and formatter now only run for the staged files, and not for the whole codebase, as they did before.


.husky/pre-commit
```diff
@@ -1,8 +1,3 @@
 # .husky/pre-commit (v5)
 
-git diff --cached --quiet -- "./src/**.js" "./src/**.ts" "./src/**.tsx"
-
-if [ "$?" -eq "1" ]
-then
-   yarn typecheck && yarn precommit
-fi
```
axbarsan, line 8: No longer necessary, `lint-staged` only runs for `.js`/`.ts`/`.tsx` files

package.json
```diff
@@ -247,8 +249,9 @@
   },
   "lint-staged": {
     "*.{js,ts,tsx}": [
-      "yarn lint:fix",
-      "yarn prettier"
+      "bash -c \"yarn typecheck\"",
```
axbarsan, line 252: Necessary instead of just `yarn typecheck` because `tsc` loses its configuration if ran by `lint-staged`

package.json
```diff
@@ -6,6 +6,7 @@
   "scripts": {
     "lint": "./node_modules/.bin/eslint --ext .js,.ts,.tsx ./src/",
     "lint:fix": "yarn lint --fix",
+    "lint:staged": "./node_modules/.bin/eslint --fix",
```
axbarsan, line 9: These 2 new commands take the files that they must run on from STDIN

kopiczko (approved):

marians (approved):

oponder:
Super nice

PR #2500, opened 2021-06-15
clusters: add the Apps and Ingress tabs to the MAPI cluster detail page

axbarsan:

This PR adds the `Apps` and `Ingress` tabs to the MAPI cluster detail page. Luckily, these have already been refactored to use the MAPI, so it's just a matter of a few small tweaks

### DIFF
--- a/src/components/MAPI/apps/ClusterDetailApps.tsx
+++ b/src/components/MAPI/apps/ClusterDetailApps.tsx
@@ -19,6 +19,7 @@ import React, {
   useState,
 } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
+import { useParams } from 'react-router';
 import { AppConstants } from 'shared/constants';
 import { AppsRoutes } from 'shared/constants/routes';
 import { supportsOptionalIngress } from 'stores/cluster/utils';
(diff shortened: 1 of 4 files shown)


src/components/MAPI/apps/ClusterDetailApps.tsx
```diff
@@ -86,15 +87,16 @@ const PreinstalledApps = styled.div`
...
 interface IClusterDetailApps {
-  clusterId: string;
   releaseVersion: string;
 }
 
 const ClusterDetailApps: React.FC<IClusterDetailApps> = ({
-  clusterId,
   releaseVersion,
 }) => {
+  const { clusterId } = useParams<{ clusterId: string }>();
```
kopiczko, line 98: > // FIXME(axbarsan): Fetch cluster by ID in URL params once AWS has namespaced clusters.

Isn't that happening here?

marians (approved):
Works for me

src/components/MAPI/apps/ClusterDetailApps.tsx
axbarsan: Here we only get the cluster ID from the URL, but we get the cluster's release version as a prop (to maintain compatibility with the GS API world.

What I want to do later on is to fetch the cluster from MAPI here, and get the release version from the cluster itself, and not as a prop.

PR #2501, opened 2021-06-15
clusters: make cluster detail components and cluster list components more flexible

axbarsan:

This PR refactors all the MAPI cluster components stuff, so there's no custom UI type in between. For cluster details, the business logic now lives in each widget component. This allows greater flexibility when implementing complex features, and gets rid of all the mess in the `ClusterDetailOverview` component.


kopiczko:
Too big for me to review. I'm fine if you simply merge it as from my POV there was a code and now there is a different code. If you think it makes sense maybe splitting it is an option.

axbarsan:
> Too big for me to review. I'm fine if you simply merge it as from my POV there was a code and now there is a different code. If you think it makes sense maybe splitting it is an option.

The PR just moves code around, from the main cluster overview component to each widget. There's not much to review IMO, as it is the same code that we had before, but in different places.

marians (approved):

kopiczko (approved):

PR #2504, opened 2021-06-16
Switch to using const enums, for less runtime code

axbarsan:
This PR switches to using `const` enums for almost all enums (except for `AvailabilityZoneSelection`, because we need its keys [here](https://github.com/giantswarm/happa/blob/8c4aebefc21a884d0e2f5b88cb54552bdfc86b44/src/components/Cluster/AZSelection/AZSelectionCheckbox.tsx#L29))

The difference between regular enums and `const` enums is that for each regular enum, a new object gets created. For example, for the `AZSelectionVariants` enum, which looks like this:
```typescript
export enum AZSelectionVariants {
	Master,
	NodePool
}
```

The compiled code for it looks similar to this:
```typescript
const AZSelectionVariants = {
	Master: 0,
	NodePool: 1,
	0: "Master",
	1: "NodePool"
}
```

When it's used, it looks like this:
```typescript
const azVariant = AZSelectionVariants.Master;
```

In the compiled code, its usage looks the same.

However, if we use `const` enums, the value of the key in the enum will be used in place:

```typescript
export const enum AZSelectionVariants {
	Master,
	NodePool
}
```

There is no compiled code for it.

When it's used, it looks like this:
```typescript
const azVariant = AZSelectionVariants.Master;
```

But when it's compiled, its usage will look like this:
```typescript
const azVariant = 0;
```

Using this, we can output less code, and have less property accesses, so MORE SPEED ⚡ .

### DIFF
--- a/src/components/Cluster/AZSelection/AZSelectionUtils.tsx
+++ b/src/components/Cluster/AZSelection/AZSelectionUtils.tsx
@@ -6,7 +6,7 @@ export enum AvailabilityZoneSelection {
   NotSpecified,
 }
 
-export enum AZSelectionVariants {
+export const enum AZSelectionVariants {
   Master,
   NodePool,
 }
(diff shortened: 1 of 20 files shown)


axbarsan:
Ah shit, `babel` doesn't support them, so storybook crashes 🥳

oponder:
TIL! too bad about babel though

PR #2511, opened 2021-06-17
Create swc plugin that removes prop types in production builds

axbarsan:
This PR adds a plugin for the `swc` compiler, which strips react component `propTypes` and removes the `prop-types` import when building the production app. We don't need them because they don't run in production anyway, so they just take up extra code.

The savings are not noticeable (~40 KB), but it was an occasion for me to have some fun with `swc`.

Not all possible scenarios are handled (you can see what's handled in the `PropTypesStripper` test), but all the patterns that we use right now work.
In order to make my job easier, I've changed the code that was not following our current pattern, to have everything working smoothly. 👍🏻

### DIFF
--- a/scripts/swc/PropTypesStripper.ts
+++ b/scripts/swc/PropTypesStripper.ts
@@ -1,0 +1,107 @@
+import {
+  AssignmentExpression,
+  ClassDeclaration,
+  Declaration,
+  Expression,
+  ExpressionStatement,
+  ImportDeclaration,
+  Module,
+  Program,
+} from '@swc/core';
+
+class PropTypesStripper {
+  public visitProgram(n: Program): Program {
+    switch (n.type) {
+      case 'Module':
+        return this.visitModule(n);
+      default:
+        return n;
+    }
+  }
+
+  public visitModule(n: Module): Module {
+    n.body = n.body.map((member) => {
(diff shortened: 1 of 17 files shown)


scripts/swc/__tests__/PropTypesStripper.ts
```diff
@@ -0,0 +1,55 @@
+import { parse, parseSync, transformSync } from '@swc/core';
+import PropTypesStripper from '../PropTypesStripper';
+
+describe('PropTypesStripper', () => {
```
axbarsan, line 4: Plugin test cases

scripts/swc/PropTypesStripper.ts
```diff
@@ -0,0 +1,107 @@
...
+  ClassDeclaration,
+  Declaration,
+  Expression,
+  ExpressionStatement,
+  ImportDeclaration,
+  Module,
+  Program,
+} from '@swc/core';
+
+class PropTypesStripper {
```
axbarsan, line 12: Plugin

axbarsan: Visits certain AST nodes and replaces them with other AST nodes

marians (approved):
Every byte counts :)

PR #2517, opened 2021-06-24
clusters: add the worker nodes tab and the list of node pools

axbarsan:

This PR adds the `Worker nodes` tab to the new MAPI Cluster detail UI. It behaves mostly like the old one, with an exception: the node pool name can now be copied to clipboard.

There is some stuff missing currently:
- the `Actions` button on the right of each node pool row. This will be added together with node pool renaming.
- a placeholder for when are no node pools. This will be added with node pool creation.

## Preview

<details>
<summary>Some node pools sitting in a tree</summary>

image

</details>

### DIFF
--- a/src/components/MAPI/workernodes/WorkerNodesAzureMachinePoolSpotInstances.tsx
+++ b/src/components/MAPI/workernodes/WorkerNodesAzureMachinePoolSpotInstances.tsx
@@ -19,13 +19,13 @@ const WorkerNodesAzureMachinePoolSpotInstances: React.FC<IWorkerNodesAzureMachin
     headline = 'Spot instances enabled';
   }
 
-  let maxPriceRow = '';
+  let maxPriceText = '';
   if (featureEnabled) {
     const maxPrice = providerNodePool?.spec?.template.spotVMOptions?.maxPrice;
     if (maxPrice && maxPrice > 0) {
-      maxPriceRow = `Using maximum price: $${maxPrice}`;
+      maxPriceText = `Using maximum price: $${maxPrice}`;
     } else {
-      maxPriceRow = `Using current on-demand pricing as maximum`;
+      maxPriceText = `Using current on-demand pricing as maximum`;
     }
   }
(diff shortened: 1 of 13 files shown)


marians:
Looking at it this way, I think we should name the tab "Node pools" and get rid of the extra "Node pools" headline. WDYT?

axbarsan:
> Looking at it this way, I think we should name the tab "Node pools" and get rid of the extra "Node pools" headline. WDYT?

Then we should rename the widget on the `Cluster overview` tab, and one of the items on the `Organization detail` page, as well

marians:
> > Looking at it this way, I think we should name the tab "Node pools" and get rid of the extra "Node pools" headline. WDYT?
> 
> Then we should rename the widget on the `Cluster overview` tab, and one of the items on the `Organization detail` page, as well

You're absolutely right. Let's keep it as it is here for now.

Just noticed a tiny font size difference in overlays.

This one seems right:

image

This one seems too large:

image

The rest looks good!

marians (approved):

src/components/MAPI/workernodes/WorkerNodesAzureMachinePoolSpotInstances.tsx
```diff
@@ -44,10 +44,10 @@ const WorkerNodesAzureMachinePoolSpotInstances: React.FC<IWorkerNodesAzureMachin
...
             >
               <Box width='small'>
-                <Text size='small'>{headline}</Text>
-                {maxPriceRow && (
-                  <Text size='small' margin={{ top: 'xsmall' }}>
-                    {maxPriceRow}
+                <Text size='xsmall'>{headline}</Text>
+                {maxPriceText && (
+                  <Text size='xsmall' margin={{ top: 'xsmall' }}>
+                    {maxPriceText}
```
marians: Looks like something we shouldn't have to adjust each time. I'd expect all tooltips to have their text in the same size.

PR #2523, opened 2021-06-28
clusters: add support for node pool scaling from the MAPI cluster detail page

axbarsan:

This PR adds support for scaling node pools in the MAPI cluster detail page. The implementation is slightly different compared to deletion, as the logic for scaling is not included in the `WorkerNodesNodePoolItem` component. I will update the deletion component, for consistency, in a separate PR.

## Preview

<details>
<summary>no value changed</summary>

image

</details>

<details>
<summary>increased min by 2</summary>

image

</details>

<details>
<summary>made max lower than current number of nodes</summary>

image

</details>

### DIFF
--- a/src/components/MAPI/workernodes/WorkerNodesNodePoolItem.tsx
+++ b/src/components/MAPI/workernodes/WorkerNodesNodePoolItem.tsx
@@ -268,9 +268,11 @@ const WorkerNodesNodePoolItem: React.FC<IWorkerNodesNodePoolItemProps> = ({
                 loaderWidth={30}
               >
                 {(value) => (
-                  <Text aria-label='Node pool autoscaler minimum node count'>
-                    {value}
-                  </Text>
+                  <Box pad={{ horizontal: 'xsmall', vertical: 'xxsmall' }}>
+                    <Text aria-label='Node pool autoscaler minimum node count'>
+                      {value}
+                    </Text>
+                  </Box>
                 )}
               </ClusterDetailWidgetOptionalValue>
             </Box>
@@ -280,9 +282,11 @@ const WorkerNodesNodePoolItem: React.FC<IWorkerNodesNodePoolItemProps> = ({
                 loaderWidth={30}
               >
                 {(value) => (
-                  <Text aria-label='Node pool autoscaler maximum node count'>
-                    {value}
-                  </Text>
+                  <Box pad={{ horizontal: 'xsmall', vertical: 'xxsmall' }}>
+                    <Text aria-label='Node pool autoscaler maximum node count'>
+                      {value}
+                    </Text>
+                  </Box>
                 )}
               </ClusterDetailWidgetOptionalValue>
             </Box>
@@ -292,9 +296,11 @@ const WorkerNodesNodePoolItem: React.FC<IWorkerNodesNodePoolItemProps> = ({
                 loaderWidth={30}
               >
                 {(value) => (
-                  <Text aria-label='Node pool autoscaler target node count'>
-                    {value}
-                  </Text>
+                  <Box pad={{ horizontal: 'xsmall', vertical: 'xxsmall' }}>
+                    <Text aria-label='Node pool autoscaler target node count'>
+                      {value}
+                    </Text>
+                  </Box>
                 )}
               </ClusterDetailWidgetOptionalValue>
             </Box>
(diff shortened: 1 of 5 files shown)


src/components/MAPI/workernodes/WorkerNodesNodePoolItem.tsx
```diff
@@ -338,6 +350,17 @@ const WorkerNodesNodePoolItem: React.FC<IWorkerNodesNodePoolItemProps> = ({
           />
         </Box>
       )}
+
+      {nodePool && (
+        <Box margin={{ top: isScaleConfirmOpen ? 'small' : undefined }}>
+          <WorkerNodesNodePoolItemScale
+            nodePool={nodePool}
+            onConfirm={onCancelScale}
```
axbarsan, line 358: On purpose

src/components/MAPI/workernodes/utils.ts
```diff
@@ -119,3 +119,82 @@ export async function deleteNodePool(
...
+
+    if (
+      nodePool.metadata.annotations?.[
+        capiexpv1alpha3.annotationMachinePoolMinSize
+      ] === min.toString() &&
+      nodePool.metadata.annotations?.[
+        capiexpv1alpha3.annotationMachinePoolMaxSize
+      ] === max.toString()
+    ) {
+      return machinePool;
```
axbarsan, line 146: If scaling is the same, we don't execute the request

src/components/UI/Display/MAPI/workernodes/WorkerNodesNodePoolActions.tsx
```diff
@@ -69,7 +71,22 @@ const WorkerNodesNodePoolActions: React.FC<IWorkerNodesNodePoolActionsProps> = (
                         onBlurHandler();
                       }}
                     >
-                      Rename
+                      <Text>Rename</Text>
```
axbarsan, line 74: Also wrapped the rename button in a `<Text />` component, since its font size was slightly different compared to the delete button

src/components/shared/NodeCountSelector.js
```diff
@@ -89,11 +90,11 @@ class NodeCountSelector extends React.Component {
               />
             </InnerTwoInputArea>
           </TwoInputArea>
-          <p data-testid='node-count-selector-autoscaling-label'>
+          <Text data-testid='node-count-selector-autoscaling-label'>
```
axbarsan, line 93: To keep stuff looking the same way

marians:
image

Let's please limit the length of the last paragraph (the yellow warning). That should happen automatically when using the grommet `Text` component, I think.

marians:
On my previous comment, I see that the text is already wrapped in `Text`. Can't remember which component limits the width. Is it `Paragraph`? Anyway, it's something we should do in general and might as well apply to Text.

More wishes:

- The "edit scaling range" panel should have a visible "Apply" button by default, which is disabled. So there should always be two buttons visible, including the Cancel button.
- The panel headline could be simplified change from `Edit scaling settings for <cluster_id>` to `Edit scaling limits`.
- After applying a change, the NODES MIN/MAX values get highlighted briefly. The span this applies to is very narrow (looks to be exactly the width of the number). There should be a horizontal padding in that span.

Also I'd like to make a visual connection between the node pool being edited and the edit panel. Some ways I can think of:

- Reduce the vertical space between the node pool row and the edit panel

plus one of

- Highlight the node pool row being edited, by using a background color 1 step lighter for the entire row.
- Highlight the currently edited node pool's NODES MIN/MAX values using a different background color

Let me know if you want to talk about details.

marians:
Just realized that the suggestions I mentioned regarding "make a visual connection between the node pool being edited and the edit panel" would also require adaption of the removal panel for consistency.

axbarsan:
> * Let's please limit the length of the last paragraph (the yellow warning). That should happen automatically when using the grommet Text component, I think.
> * The "edit scaling range" panel should have a visible "Apply" button by default, which is disabled. So there should always be two buttons visible, including the Cancel button.
> * The panel headline could be simplified change from `Edit scaling settings for <cluster_id>` to `Edit scaling limits`.
> * After applying a change, the NODES MIN/MAX values get highlighted briefly. The span this applies to is very narrow (looks to be exactly the width of the number). There should be a horizontal padding in that span.

These are done.


> Also I'd like to make a visual connection between the node pool being edited and the edit panel. Some ways I can think of:
> 
> * Reduce the vertical space between the node pool row and the edit panel
> 
> plus one of
> 
> * Highlight the node pool row being edited, by using a background color 1 step lighter for the entire row.
> * Highlight the currently edited node pool's NODES MIN/MAX values using a different background color
> 
> Let me know if you want to talk about details.

As for this one, please create a separate note/issue on the board, since it's also related to the deletion component.

marians (approved):
All right then, thanks!

PR #2525, opened 2021-06-30
clusters: add basic node pool creation from the MAPI cluster details page

axbarsan:

This PR adds a very basic node pool creation form, and all the code required to add further configuration options.
In this PR the only configurable option is the node pool description. Everything else is defaulted (NP creation works as it is). All the other options will come in separate PR, to prevent this PR from becoming too big.

The resource name (previously: node pool ID) is generated on the client side (99% the same logic as the generator we use right now). There's a disabled input showing the name that the node pool will have. Don't worry, if you manage to change the value of the input, the resource's name will not change. It is just for show right now, but we could make it configurable in the future.

For implementation details, check out the comments below.

## Preview

<details>
<summary>NP creation button just chilling</summary>

image

</details>

<details>
<summary>NP creation form open</summary>

image

</details>

<details>
<summary>No NPs placeholder</summary>

image

</details>


src/components/MAPI/types.ts
```diff
@@ -2,11 +2,14 @@ import * as capiv1alpha3 from 'model/services/mapi/capiv1alpha3';
 import * as capiexpv1alpha3 from 'model/services/mapi/capiv1alpha3/exp';
 import * as capzv1alpha3 from 'model/services/mapi/capzv1alpha3';
 import * as capzexpv1alpha3 from 'model/services/mapi/capzv1alpha3/exp';
+import * as gscorev1alpha1 from 'model/services/mapi/gscorev1alpha1';
 
 export type ControlPlaneNode = capzv1alpha3.IAzureMachine;
 
 export type ControlPlaneNodeList = capzv1alpha3.IAzureMachineList;
 
+export type Cluster = capiv1alpha3.ICluster;
```
axbarsan, line 11: This needs to be used where we currently use `capiv1alpha3.ICluster` resources

src/components/MAPI/utils.ts
```diff
@@ -380,3 +381,53 @@ export function getNodePoolAvailabilityZones(nodePool: NodePool): string[] {
...
+}
+
+const uidRegexp = /^[a-z]([a-z][0-9]|[0-9][a-z])+$/;
+const supportedUIDChars = '023456789abcdefghijkmnopqrstuvwxyz';
+
+/**
+ * Generate unique resource names, that can be used for node pool or cluster names.
+ * @param length
+ */
+export function generateUID(length: number): string {
```
axbarsan, line 414: 99% the same as the current ID generation function in `kubectl gs` and `api`

src/components/MAPI/workernodes/ClusterDetailWorkerNodes.tsx
```diff
@@ -173,6 +178,22 @@ const ClusterDetailWorkerNodes: React.FC<IClusterDetailWorkerNodesProps> = () =>
       capiv1alpha3.getCluster(clusterClient.current, auth, namespace, clusterId)
   );
 
+  const providerClusterKey = cluster
+    ? fetchProviderClusterForClusterKey(cluster)
+    : null;
+  const { data: providerCluster, error: providerClusterError } = useSWR<
```
axbarsan, line 184: We need the provider-specific cluster for the region that the cluster is in

src/components/MAPI/workernodes/patches.ts
```diff
@@ -0,0 +1,32 @@
+import { NodePool, ProviderNodePool } from 'MAPI/types';
+import * as capiexpv1alpha3 from 'model/services/mapi/capiv1alpha3/exp';
+
+export type NodePoolPatch = (
```
axbarsan, line 4: Tried to make the node pool creation code a bit lighter by having all feature-specific code in separate components (e.g. a spot instances component is in charge of setting up spot instances). By using patches, we make sure that we don't randomly mutate the node pool data structures from feature-specific components.

Stuff changes in feature components, they return patches to the parent node pool creation form, which applies those patches onto the node pool data structures stored in memory.

So far this pattern has been working pretty well. If I like how it evolves, I'll also use it for cluster creation, and maybe refactor cluster updates/node pool updates to use it, as well.

src/components/MAPI/workernodes/patches.ts
```diff
@@ -0,0 +1,32 @@
...
+export interface INodePoolPropertyProps {
+  id: string;
+  nodePool: NodePool;
+  providerNodePool: ProviderNodePool;
+  onChange: (value: INodePoolPropertyValue) => void;
+  disabled?: boolean;
+  readOnly?: boolean;
+}
+
+export function withNodePoolDescription(newDescription: string): NodePoolPatch {
```
axbarsan, line 23: Every feature has its own patch(es)

src/components/UI/Display/MAPI/workernodes/WorkerNodesNodePoolListPlaceholder.tsx
```diff
@@ -0,0 +1,46 @@
...
+import PropTypes from 'prop-types';
+import React from 'react';
+import Button from 'UI/Controls/Button';
+
+interface IWorkerNodesNodePoolListPlaceholderProps
+  extends React.ComponentPropsWithoutRef<typeof Box> {
+  onCreateButtonClick?: () => void;
+}
+
+const WorkerNodesNodePoolListPlaceholder: React.FC<IWorkerNodesNodePoolListPlaceholderProps> = ({
```
axbarsan, line 11: Placeholder component that is displayed when there are no node pools

marians:
I created a cluster and found that the `worker nodes` link (bottom line in the screenshot) in the overview panel misses the cluster ID:

`http://localhost:7000/organizations/:orgId/clusters/:clusterId/worker-nodes`

image

marians:
image

Something more related to this PR:

- I would like to change the label from `Name` to `ID`. That's also the label we use for the column in the node pools table, right? It also conveys better the fact that this is the unique identifier.
- The contrast of the text in the disabled name input field is a bit low.
- The sentence `Each node pool also gets a unique identifier.` can be removed in this design, as the identifier is visible above.

axbarsan:
> I created a cluster and found that the `worker nodes` link (bottom line in the screenshot) in the overview panel misses the cluster ID

Done ✅ 

> * The contrast of the text in the disabled name input field is a bit low.
> * The sentence `Each node pool also gets a unique identifier.` can be removed in this design, as the identifier is visible above.

Done and done ✅ Updated screenshots

> * I would like to change the label from `Name` to `ID`. That's also the label we use for the column in the node pools table, right? It also conveys better the fact that this is the unique identifier.

I'm not sure about this one. In the Kubernetes lingo, there's no such thing as ID. Resources have unique identifiers, which are the resource names. Since we want to lower the abstraction level over the K8s API in happa, I believe we should also use the K8s terminologies (where applicable, and not when the feature is invented by us, e.g. cluster/node pool descriptions).

The column label that we use in the new node pools table UI is also called `Name`.

image

marians:
> I'm not sure about this one. In the Kubernetes lingo, there's no such thing as ID. Resources have unique identifiers, which are the resource names. Since we want to lower the abstraction level over the K8s API in happa, I believe we should also use the K8s terminologies (where applicable, and not when the feature is invented by us, e.g. cluster/node pool descriptions).
> 
> The column label that we use in the new node pools table UI is also called `Name`.

Makes sense

marians (approved):
Tested and LGTM

marians:
One little thing occurred to me: The default `Unnamed node pool` makes less sense now that we renamed the ID to be the name. What should we change this to?

- Please add description
- Node pool without description

axbarsan:
> One little thing occurred to me: The default `Unnamed node pool` makes less sense now that we renamed the ID to be the name. What should we change this to?
> 
> * Please add description
> * Node pool without description

Hmm, good call. I would vote for `Node pool without description`

axbarsan:
> > One little thing occurred to me: The default `Unnamed node pool` makes less sense now that we renamed the ID to be the name. What should we change this to?
> > 
> > * Please add description
> > * Node pool without description
> 
> Hmm, good call. I would vote for `Node pool without description`

Nah, that's too long and it looks weird. I'll revert to `Please add description`

PR #2526, opened 2021-07-01
Refactor MAPI resource deletion

axbarsan:
It occurred to me that recent versions of the K8s API return a `Status` response, as opposed to the deleted resource.

This PR takes that into account, and changes the implementations that depended on the previous behaviour.


src/components/MAPI/clusters/ClusterDetail/utils.ts
```diff
@@ -44,7 +44,11 @@ export async function deleteCluster(
     name
   );
 
-  return capiv1alpha3.deleteCluster(httpClient, auth, cluster);
+  await capiv1alpha3.deleteCluster(httpClient, auth, cluster);
+
+  cluster.metadata.deletionTimestamp = new Date().toISOString();
```
axbarsan, line 49: Quick and dirty

marians (approved):
Just unblocking this, as I can't really review it.

oponder (approved):
LGTM

PR #2532, opened 2021-07-02
clusters: add support for configuring the machine type when creating node pools via the MAPI cluster detail page

axbarsan:

This PR adds support for configuring the Machine Type (i.e. VM Size on Azure) when creating node pools. It uses the exact same underlying component as the current GS API implementation, so no big change there.

The PR also makes the machine type selector close when pressing the `Esc` key, and it slightly increases the font weight, to match the rest of the MAPI-specific UIs.

## Preview

<details>
<summary>The creation form with the new option</summary>

image

</details>

### DIFF
--- a/package.json
+++ b/package.json
@@ -219,7 +219,7 @@
     "ts-node": "^10.0.0",
     "typescript": "4.2.4",
     "url-loader": "4.1.1",
-    "webpack": "5.36.2",
+    "webpack": "5.42.0",
     "webpack-cli": "4.7.0",
     "webpack-dev-server": "3.11.2",
     "webpack-merge": "5.7.3"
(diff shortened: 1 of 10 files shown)


marians (approved):
Works for me!

If we want to use the occasion for an improvement, I'd like to suggest:

- prevent line wraps with the memory table column
- Set the VM size identifier as a literal using Inconsolata

image

axbarsan:
> Works for me!
> 
> If we want to use the occasion for an improvement, I'd like to suggest:
> 
> * prevent line wraps in the memory table column
> * Set the VM size identifier as a literal using Inconsolata
> 
> image

✅ Done

image

marians (approved):

PR #2536, opened 2021-07-02
clusters: add support for setting AZs when creating NPs via the MAPI cluster detail page

axbarsan:

This PR adds support for setting availability zones when creating node pools. It uses the same underlying components as the current GS API implementation. However, I've made some improvements in how those components behave (consistent font sizes, improved keyboard navigation).

## Preview

<details>
<summary>Automatic AZ selection</summary>

image

</details>

<details>
<summary>Manual AZ selection</summary>

image

</details>

<details>
<summary>Not specified selection</summary>

image

</details>

### DIFF
--- a/src/components/MAPI/__tests__/utils.ts
+++ b/src/components/MAPI/__tests__/utils.ts
@@ -1,4 +1,4 @@
-import { generateUID } from 'MAPI/utils';
+import { determineRandomAZs, generateUID } from 'MAPI/utils';
 
 describe('mapi::utils', () => {
   describe('generateUID', () => {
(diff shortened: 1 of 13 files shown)


src/components/MAPI/__tests__/utils.ts
```diff
@@ -15,4 +15,70 @@ describe('mapi::utils', () => {
       }
     });
   });
+
+  describe('determineRandomAZs', () => {
+    it('returns a collection of random availability zones', () => {
```
axbarsan, line 20: This test tries to get `3` unique arrays of `2` AZs from the available `['1', '2', '3', '4', '5']` list of AZs. It runs that 50 times, and expects that at least 50% of the time the returned AZs are random. 

This is so we wouldn't have a flaky test that makes us re-run CI every 5 mins.

src/components/MAPI/utils.ts
```diff
@@ -405,6 +406,71 @@ export function getProviderNodePoolLocation(
...
+
+/**
+ * Determine a random list list of availability zones,
+ * based on a given collection of available zones and
+ * a required number of elements.
+ * @param count - The maximum number of zones.
+ * @param fromList - The available zones.
+ * @param existing - Some zones that must be included in the output.
+ */
+export function determineRandomAZs(
```
axbarsan, line 436: This function generates a random list of AZs, taking into account the desired number of zones, and the list of supported zones.

If the AZs of the control plane nodes are supplied, then the list is not that random, as those zones would be picked first.

For example, if the cluster has 1 control plane node in the `2` AZ, and we want to have a node pool with 3 AZs, and all the supported zones are `['1', '2', '3', '4']`, then the result would probably be `['1', '2', '3']`, `['2', '3', '4']` or other combination of the rest.

If you want to have a more detailed explanation of the behaviour, please check the tests for this function.

src/components/MAPI/workernodes/WorkerNodesCreateNodePool.tsx
```diff
@@ -114,6 +116,7 @@ function makeInitialState(
       [NodePoolPropertyField.Name]: true,
       [NodePoolPropertyField.Description]: true,
       [NodePoolPropertyField.MachineType]: true,
+      [NodePoolPropertyField.AvailabilityZones]: false,
```
axbarsan, line 119: AZs are invalid by default, since the control plane nodes have to be loaded first. This will be set to `true` after the control plane nodes have been loaded successfully.

src/components/MAPI/workernodes/WorkerNodesCreateNodePoolAvailabilityZones.tsx
```diff
@@ -0,0 +1,184 @@
...
+        break;
+
+      case AvailabilityZoneSelection.NotSpecified:
+        onChange({
+          isValid: true,
+          patch: withNodePoolAvailabilityZones(undefined),
+        });
+        break;
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
axbarsan, line 125: Disabling the rule because it requires that we add `onChange` to the list of dependencies, but that results in a memory leak, since `onChange` has a different pointer in every re-render.

marians (approved):
LGTM - didn't test

PR #2537, opened 2021-07-02
Avoid error notification for invalid app values file

marians:
With this PR we turn off the error notification (submission to Sentry) for errors happening if a user tries uploading an invalid values.yaml file.


axbarsan (approved):
Nice

PR #2538, opened 2021-07-05
clusters: add support for configuring scaling when creating NPs via the MAPI cluster detail page

axbarsan:

No fancy stuff here, this allows us to set node pool scaling when creating node pools via the MAPI cluster details page.

## Preview

<details>
<summary>The form with the new option</summary>

image

</details>

### DIFF
--- a/src/components/UI/Inputs/NumberPicker/index.tsx
+++ b/src/components/UI/Inputs/NumberPicker/index.tsx
@@ -187,7 +187,7 @@ const NumberPicker = React.forwardRef<HTMLInputElement, INumberPickerProps>(
     };
 
     useEffect(() => {
-      if (prevMax !== max || prevMin !== min) {
+      if ((prevMax && prevMax !== max) || (prevMin && prevMin !== min)) {
         const error = validateInput(currValue, min, max);
         updateValue(currValue, error);
       }
--- a/src/components/shared/NodeCountSelector.js
+++ b/src/components/shared/NodeCountSelector.js
@@ -16,10 +16,6 @@ const DEFAULT_VALUE_CONSTRAINTS = {
  * max values for scaling limits.
  */
 class NodeCountSelector extends React.Component {
-  handleFormSubmit = (e) => {
-    e.preventDefault();
-  };
-
   updateValue(valuesToAdd) {
     const nextScalingValue = Object.assign({}, this.props.scaling, valuesToAdd);
 
@@ -57,7 +53,7 @@ class NodeCountSelector extends React.Component {
 
     if (this.props.autoscalingEnabled === true) {
       return (
-        <form onSubmit={this.handleFormSubmit}>
+        <div>
           <TwoInputArea>
             <InnerTwoInputArea>
               <NumberPicker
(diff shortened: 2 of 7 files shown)


src/components/shared/NodeCountSelector.js
```diff
@@ -57,7 +59,7 @@ class NodeCountSelector extends React.Component {
 
     if (this.props.autoscalingEnabled === true) {
       return (
-        <form onSubmit={this.handleFormSubmit}>
```
axbarsan, line 60: This shouldn't be a `<form />` element. This is an input, it should be part of a `<form />` element together with other inputs. Using a `<form />` element inside another one is invalid HTML markup

src/components/shared/NodeCountSelector.js
```diff
@@ -16,15 +16,17 @@ const DEFAULT_VALUE_CONSTRAINTS = {
  * max values for scaling limits.
  */
 class NodeCountSelector extends React.Component {
-  handleFormSubmit = (e) => {
-    e.preventDefault();
-  };
-
   updateValue(valuesToAdd) {
-    const nextScalingValue = Object.assign({}, this.props.scaling, valuesToAdd);
+    setTimeout(() => {
```
axbarsan, line 20: Forcing the callback to be executed after `react` makes all the updates.

This is because the 2 underlying `<NumberPicker />` components execute the callback when their `min`/`max` props change (because they also need to validate again). This also includes the first time that the `min`/`max` props are set, which is when the components get mounted.

Because the 2 callbacks (for 2 `<NumberPicker />` components) get executed one right after another, the value of the scaling in the props does not have time update in between calls. Therefore the callback returns that the input is invalid when it's actually valid.

Long story short, this makes sure that the scaling is valid by default and it doesn't prevent node pool creation.

src/components/shared/NodeCountSelector.js
```diff
@@ -29,13 +31,22 @@ class NodeCountSelector extends React.Component {
...
+      });
+    });
   };
 
   updateScalingMax = (numberPicker) => {
-    this.updateValue({
-      max: numberPicker.value,
-      maxValid: numberPicker.valid,
+    clearTimeout(this.updateTimeout);
+    this.updateTimeout = setTimeout(() => {
```
axbarsan, line 45: Forcing the callback to be executed after `react` makes all the updates.

This is because the 2 underlying `<NumberPicker />` components execute the callback when their `min`/`max` props change (because they also need to validate again). This also includes the first time that the `min`/`max` props are set, which is when the components get mounted.

Because the 2 callbacks (for 2 `<NumberPicker />` components) get executed one right after another, the value of the scaling in the props does not have time update in between calls. Therefore the callback returns that the input is invalid when it's actually valid.

Long story short, this makes sure that the scaling is valid by default and it doesn't prevent node pool creation."

src/components/UI/Inputs/NumberPicker/index.tsx
```diff
@@ -187,7 +187,7 @@ const NumberPicker = React.forwardRef<HTMLInputElement, INumberPickerProps>(
     };
 
     useEffect(() => {
-      if (prevMax !== max || prevMin !== min) {
+      if ((prevMax && prevMax !== max) || (prevMin && prevMin !== min)) {
```
axbarsan, line 190: This is to prevent `calling the callback` when the component is rendered for the first time

marians (approved):
LGTM if it works for you

PR #2540, opened 2021-07-05
clusters: add support for configuring spot instances on NP creation via the MAPI page

axbarsan:

This PR adds support for configuring Azure spot instances when creating node pools from the MAPI cluster detail page. It uses the same UI components as the current GS API implementation, so nothing too crazy on that part.

The PR also includes a small change in the way we display the maximum spot instance pricing in node pools that have already been created. Because the max price is a `Quantity` type, and not a regular string or number, it represents a serialized value of a number, with certain suffixes that represent the number that the prefix should be multiplied by to get the actual value.

> For example, `"10u"` represents `0.00001`.

The utility for parsing these types of data structures has been copied from the official Kubernetes JS client, with a few added suffixes.

## Preview

<details>
<summary>Spot instances disabled</summary>

image

</details>

<details>
<summary>Using current on-demand pricing as the max price</summary>

image

</details>

<details>
<summary>With some max value</summary>

image

</details>

### DIFF
--- a/src/components/MAPI/workernodes/WorkerNodesCreateNodePool.tsx
+++ b/src/components/MAPI/workernodes/WorkerNodesCreateNodePool.tsx
@@ -14,6 +14,7 @@ import {
 } from 'MAPI/types';
 import {
   generateUID,
+  getClusterReleaseVersion,
   getProviderClusterLocation,
   getProviderNodePoolLocation,
 } from 'MAPI/utils';
@@ -24,6 +25,7 @@ import { useSelector } from 'react-redux';
 import { Providers } from 'shared/constants';
 import { PropertiesOf } from 'shared/types';
 import { getProvider } from 'stores/main/selectors';
+import { supportsNodePoolSpotInstances } from 'stores/nodepool/utils';
 import Button from 'UI/Controls/Button';
 
 import { INodePoolPropertyValue, NodePoolPatch } from './patches';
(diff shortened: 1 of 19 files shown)


src/components/MAPI/workernodes/WorkerNodesCreateNodePool.tsx
```diff
@@ -299,6 +309,16 @@ const WorkerNodesCreateNodePool: React.FC<IWorkerNodesCreateNodePoolProps> = ({
                 cluster={cluster}
                 margin={{ top: 'small' }}
               />
+
+              {supportsSpotInstances && (
```
axbarsan, line 313: We only display the spot instances configuration components for clusters that support it (release version >= `14.1.0`)

marians:
Looks great in general.

Detail nitpick:

- When I select "Use current on-demand pricing as max", the value shown in the amount input is `-1`. It used to be zero before. I would prefer no value to be visible in the field, as both -1 and 0 are misleading.

I want to use the occasion to improve our wording here:

image

Adding my suggestions to the code. (Might not find all places...)

marians:
Turns out I can't find all the strings to add sugegstions in code, so doing it here.

- `Spot instances` (Azure only) -> `Spot virtual machines`
- `Maximum price per hour` -> `Price limit`
- `The maximum price per hour that a single node pool VM instance can reach before it is deallocated.` -> `The highest amount per hour (in USD) you are willing to pay, per virtual machine.`
- `Use current on-demand pricing as max` -> `Use the on-demand price as limit`

marians (approved):
Also unblocking this

marians:
The renaming of `Spot instances` to `Spot virtual machines` (or `Spot VMs` where space is limited would have ramifications beyond this PR (e. g. the node pools table).

I'm also fine with addressing that separately of course.

axbarsan:
> * When I select "Use current on-demand pricing as max", the value shown in the amount input is `-1`. It used to be zero before. I would prefer no value to be visible in the field, as both -1 and 0 are misleading.

That's unsupported at the moment because of how the `CurrencyInput` works. I changed it to display `0`, as it does for the GS API implementation.

> I want to use the occasion to improve our wording here:
> 
> image

✅ Done

PR #2541, opened 2021-07-06
Refactor changeEmail and changePassword forms

oponder:
towards: https://github.com/giantswarm/giantswarm/issues/12347


oponder:
sorry about the broken tests @giantswarm/chapter-front-end this should be good to go now

src/components/AccountSettings/ChangePasswordForm.tsx
```diff
@@ -1,31 +1,68 @@
...
 import Section from 'UI/Layout/Section';
 
-class ChangePassword extends React.Component {
+interface IChangePasswordPropsUser {
+  email: string;
+}
+
+interface IChangePasswordProps {
+  user: IChangePasswordPropsUser;
+  giantswarmLogin: (email: string, password: string) => {};
```
axbarsan, line 22: Can we import this directly? Why does it have to come from props?

src/components/AccountSettings/ChangePasswordForm.tsx
```diff
@@ -110,85 +148,82 @@ class ChangePassword extends React.Component {
...
-          <form className='change_password_form' onSubmit={this.submit}>
+          <form onSubmit={this.submit}>
             <TextInput
               id='current_password'
               label='Current Password'
               onChange={this.validate}
-              ref={(p) => {
-                this.current_password = p;
-              }}
+              ref={this.currentPassword}
```
axbarsan, line 218: Why do we need `refs` for handling the input values? Why not make the inputs controlled, and store their values in state?

oponder:
@axbarsan I didn't want to refactor the component too much, so just taking the way it was made and putting it into the typescript world.

It's a bit odd, not very much in the style we do it nowadays.. I think it was one of the first components I ever wrote back in the early days of Happa.

But yeah I figured not worth it to bring it up to date fully, since this is about the email and password login and will be removed eventually from Happa altogether.

axbarsan (approved):
ok, don't want to block here

PR #2544, opened 2021-07-06
Consolidate an extracted label style that is only being used in one place

oponder:
@axbarsan I noticed this style wasn't really being used in any other places except for `ClusterCreationLabelSpan.tsx ` so I moved it in to that file, to avoid extra layers.

If you prefer still having it in it's own file in `src/styles` feel free to let me know and we can close the PR.


oponder:
I actually think this whole component `ClusterCreationLabelSpan` isn't being used anywhere by the way.

axbarsan:
> I actually think this whole component `ClusterCreationLabelSpan` isn't being used anywhere by the way.

It doesn't seem to be used, can we get rid of it?

axbarsan (approved):

PR #2545, opened 2021-07-06
Stop displaying inactive versions as upgradable for admins

axbarsan:
As seen on https://gigantic.slack.com/archives/C7865GLSY/p1625574670362400

This is a left-over from the days when we had the `Change version` button in the `Upgrade cluster` modal. It was then possible for admins (GS staff) to see the `Upgrade available` warning for inactive/pre-release versions, in order to test upgrades on testing installations, or execute upgrades to pre-release versions on the customers' behalf.

Now, as we have upgrading to beta versions (or inactive versions) available through the `Release details` modal, it is no longer necessary to display the warning (as it might confuse users).

To be clear, this PR removes the ability of admins to see the `Upgrade available` warning when newer versions are inactive or pre-releases.


anvddriesch (approved):
Not familiar with happa but LGTM!

PR #2546, opened 2021-07-06
clusters: add getting started guide specific for MAPI clusters

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/17531

This PR adds the MAPI-specific getting started guide for clusters.
The information displayed is *almost* the same as the GS API one. The difference is that the MAPI one doesn't have the alternative keypair creation method (since it is unsupported via MAPI at the moment).

Implementation wise, the new MAPI-specific components are brand new, and they use `grommet` for styling (the old mostly used SASS and were dependent on the clusters coming from the global state).

> ⚠️  In the screenshots the cluster base URL includes `localhost`. This is not the case when deployed in production.

## Preview

<details>
<summary>Overview</summary>

image

</details>

<details>
<summary>Configure kubectl</summary>

image

</details>

<details>
<summary>Install ingress</summary>

image

</details>

<details>
<summary>Create example application</summary>

image

</details>

<details>
<summary>Final step</summary>

image

</details>

### DIFF
--- a/src/components/MAPI/utils.ts
+++ b/src/components/MAPI/utils.ts
@@ -1,13 +1,16 @@
 import { HttpClientFactory } from 'lib/hooks/useHttpClientFactory';
 import { IOAuth2Provider } from 'lib/OAuth2/OAuth2';
+import { IHttpClient } from 'model/clients/HttpClient';
 import * as capiv1alpha3 from 'model/services/mapi/capiv1alpha3';
 import * as capiexpv1alpha3 from 'model/services/mapi/capiv1alpha3/exp';
 import * as capzv1alpha3 from 'model/services/mapi/capzv1alpha3';
 import * as capzexpv1alpha3 from 'model/services/mapi/capzv1alpha3/exp';
-import { Constants } from 'shared/constants';
+import { Constants, Providers } from 'shared/constants';
+import { PropertiesOf } from 'shared/types';
 import { IState } from 'stores/state';
 
 import {
+  Cluster,
   ControlPlaneNodeList,
   NodePool,
   NodePoolList,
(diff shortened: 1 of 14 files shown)


src/components/MAPI/utils.ts
```diff
@@ -230,6 +233,24 @@ export function fetchProviderNodePoolsForNodePoolsKey(
   return keys.join();
 }
 
+export async function fetchCluster(
```
axbarsan, line 236: Will refactor other places to use this function for fetching CAPI cluster resources, so we can have different versions per provider.

src/components/UI/Display/MAPI/clusters/GettingStarted/GettingStartedNextSteps.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  return (
+    <Breadcrumb
+      data={{
+        title: 'NEXT STEPS',
+        pathname: match.url,
+      }}
+    >
+      <Box>
+        <Heading level={1}>🎉 Congratulations</Heading>
```
axbarsan, line 21: Fun fact: Converting this emoji to ASCII character codes crashes the compiler 😁

src/components/shared/Tabs/index.tsx
```diff
@@ -38,6 +40,7 @@ const Tabs: React.FC<ITabsProps> = ({
       id='tabs'
       mountOnEnter={true}
       unmountOnExit={true}
+      {...props}
```
axbarsan, line 43: Let's make this customizable

marians:
The renaming of `Spot instances` to `Spot virtual machines` (or `Spot VMs` where space is limited would have ramifications beyond this PR (e. g. the node pools table).

I'm also fine with addressing that separately of course.

marians:
Spacing around a code block:

image

The space above and below should be the same optically. It appears too be too much in the top. Changing the codeblock's margin-top from 25px to 12px fixed it for me.

marians (approved):
Rest LGTM

PR #2547, opened 2021-07-07
clusters: add support for really basic cluster creation via the MAPI cluster creation page

axbarsan:

This PR adds support for basic cluster creation via the MAPI cluster creation page. It includes defaulting, viewing the cluster name in a disabled form and modifying the cluster description.

As for the implementation, it is pretty much the same as with node pool creation. I re-used the same `patch` pattern as I used there. The only twist is that we have to fetch all releases and determine the latest (active and non-prerelease) release that we can use. Until that data is fetched, the `Create cluster` button is disabled.

## Preview

<details>
<summary>The form</summary>

image

</details>

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -61,7 +61,7 @@ module.exports = {
     },
     featureFlags: {
       FEATURE_MAPI_AUTH: true,
-      FEATURE_MAPI_CLUSTERS: true,
+      FEATURE_MAPI_CLUSTERS: false,
       FEATURE_MONITORING: true,
     },
   },
--- a/src/components/MAPI/clusters/CreateCluster/index.tsx
+++ b/src/components/MAPI/clusters/CreateCluster/index.tsx
@@ -90,17 +90,12 @@ function makeInitialState(
   const name = generateUID(5);
 
   const resourceConfig = { ...config, name, releaseVersion: '' };
-  const controlPlaneNode = createDefaultControlPlaneNode(
-    provider,
-    resourceConfig
-  );
   const providerCluster = createDefaultProviderCluster(
     provider,
     resourceConfig
   );
-  const cluster = createDefaultCluster({
-    providerCluster,
-  });
+  const controlPlaneNode = createDefaultControlPlaneNode({ providerCluster });
+  const cluster = createDefaultCluster({ providerCluster });
 
   return {
     provider,
--- a/src/components/MAPI/clusters/CreateCluster/index.tsx
+++ b/src/components/MAPI/clusters/CreateCluster/index.tsx
@@ -286,7 +286,6 @@ const CreateCluster: React.FC<ICreateClusterProps> = (props) => {
               providerCluster={state.providerCluster}
               controlPlaneNode={state.controlPlaneNode}
               onChange={handleChange(ClusterPropertyField.Name)}
-              disabled={true}
             />
             <WorkerNodesCreateClusterDescription
               id={`cluster-${ClusterPropertyField.Description}`}
--- a/src/components/MAPI/clusters/CreateCluster/index.tsx
+++ b/src/components/MAPI/clusters/CreateCluster/index.tsx
@@ -1,13 +1,16 @@
+import { useAuthProvider } from 'Auth/MAPI/MapiAuthProvider';
 import { push } from 'connected-react-router';
 import { Box, Heading, Text } from 'grommet';
 import produce from 'immer';
 import ErrorReporter from 'lib/errors/ErrorReporter';
 import { FlashMessage, messageTTL, messageType } from 'lib/flashMessage';
+import { useHttpClientFactory } from 'lib/hooks/useHttpClientFactory';
 import RoutePath from 'lib/routePath';
 import { extractErrorMessage } from 'MAPI/organizations/utils';
 import { Cluster, ControlPlaneNode, ProviderCluster } from 'MAPI/types';
 import { generateUID } from 'MAPI/utils';
-import React, { useReducer } from 'react';
+import * as releasev1alpha1 from 'model/services/mapi/releasev1alpha1';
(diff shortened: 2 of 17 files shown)


src/components/MAPI/clusters/CreateCluster/index.tsx
```diff
@@ -0,0 +1,330 @@
...
+      case 'startCreation':
+        draft.isCreating = true;
+        break;
+      case 'endCreation':
+        draft.isCreating = false;
+        break;
+      case 'setLatestRelease':
+        // Apply the version to the CRs if no version was set before.
+        if (!draft.latestRelease) {
+          withClusterReleaseVersion(action.value)(
```
axbarsan, line 143: Since we need to fetch all the releases first in order to determine the latest release version, once we have all the data necessary we update the resources.

src/model/services/mapi/capzv1alpha3/types.ts
```diff
@@ -74,7 +74,7 @@ export interface IAzureClusterSpec {
   networkSpec?: INetworkSpec;
   resourceGroup?: string;
   subscriptionID?: string;
-  contronPlaneEndpoint?: capiv1alpha3.IApiEndpoint;
+  controlPlaneEndpoint?: capiv1alpha3.IApiEndpoint;
```
axbarsan, line 77: Typo 😬

src/model/services/mapi/capzv1alpha3/types.ts
```diff
@@ -165,6 +168,10 @@ export interface IAzureMachineSpec {
   vmSize: string;
   osDisk: IOSDisk;
   sshPublicKey: string;
+  /**
+   * @deprecated
+   *  */
+  location: string;
```
axbarsan, line 174: Our admission controller requires this to be an empty string 🤓

jest.config.js
```diff
@@ -61,7 +61,7 @@ module.exports = {
     },
     featureFlags: {
       FEATURE_MAPI_AUTH: true,
-      FEATURE_MAPI_CLUSTERS: true,
+      FEATURE_MAPI_CLUSTERS: false,
```
axbarsan, line 64: No idea why this was on

src/components/MAPI/clusters/utils.ts
```diff
@@ -155,3 +168,267 @@ export function isClusterCreating(cluster: capiv1alpha3.ICluster): boolean {
...
+  }
+) {
+  if (provider === Providers.AZURE) {
+    return createDefaultAzureCluster(config);
+  }
+
+  throw new Error('Unsupported provider.');
+}
+
+function createDefaultAzureCluster(config: {
```
axbarsan, line 188: Various resource defaulting

marians:
image

Since our cluster ID visualization is quite recognizable, I'd like to use it here in the form context already. So instead of an input field, the `ClusterIDLabel` component should be used here to display the new name. This makes it super easy for customers to understand that this is going to be the identifier they will see in that style in happa.

Since the element is quite narrow, I'd prefer placing the `Name` label and the ClusterIDLabel component in the same line.

Can we do that?

marians:
When opening the form, the cursor is automatically placed behind "Please add description". This requires users to delete the input value, then add their own.

image

I'd prefer to use a [placeholder](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-placeholder) text instead, which is displayed as long as the user has not written any input. Focusing the input field automatically is good and should be done still.

axbarsan:
> Since our cluster ID visualization is quite recognizable, I'd like to use it here in the form context already. So instead of an input field, the `ClusterIDLabel` component should be used here to display the new name. This makes it super easy for customers to understand that this is going to be the identifier they will see in that style in happa.

I don't think that's a very good idea.
Since we want to let users provide custom cluster resource names (as we already do with `kubectl-gs`) at some point, we would have to switch to a text input anyway.

Switching to a `<ClusterIDLabel />` now means that the user has to get used to having that in, and then later they have to get used to having an input instead. Why not get the users familiar with having an input right away?

axbarsan:
> When opening the form, the cursor is automatically placed behind "Please add description". This requires users to delete the input value, then add their own.
>
> I'd prefer to use a [placeholder](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-placeholder) text instead, which is displayed as long as the user has not written any input. Focusing the input field automatically is good and should be done still.

Having a placeholder means that the input will be empty (will have no value) at first. I'd rather have a default value, so the user can just press on `Create cluster` right away.

I've changed the implementation so when the page is loaded, the `Description` input is focused, and all its contents are fully selected.

image

marians:
> Since we want to let users provide custom cluster resource names (as we already do with `kubectl-gs`) at some point, we would have to switch to a text input anyway.

And even then I'd like to be able to create a visual link between the input field and the visual token they will look for later. An editable ClusterIDLabel component, so to say.

> Having a placeholder means that the input will be empty (will have no value) at first. I'd rather have a default value, so the user can just press on Create cluster right away.

I agree that the user should not have to fill something in. But that does not require having an actual value in the input. The default should be applied if the user submits an empty description field. (This could even happen on the admission controller level consistently)

Interaction wise I would prefer using the placeholder attribute.

axbarsan:
> And even then I'd like to be able to create a visual link between the input field and the visual token they will look for later. An editable ClusterIDLabel component, so to say.

OK, I'll commit to that. It looks a bit strange, though

image

> I agree that the user should not have to fill something in. But that does not require having an actual value in the input. The default should be applied if the user submits an empty description field. (This could even happen on the admission controller level consistently)

We are trying to be closer to the Kubernetes API, and we can let the user know that, by not doing any magic in the background. What they see in the UI should be what is set in the created resources. I see no problem with the current implementation

marians:
Thank you!

marians (approved):

PR #2549, opened 2021-07-12
Refactor the Tab components to use grommet instead of bootstrap

oponder:
towards: giantswarm/giantswarm#12347


oponder:
This PR tries to keep the tabs looking the same, but it does change the tabs in the getting started guides to have the same visual style as tabs elsewhere throughout the app. I figured consolidating our tab implementations towards grommet would be more desirable than having two different tab looks and implementations.

Screenshot 2021-07-12 at 7 49 39 PM

axbarsan:
Taking this over

axbarsan:
> As I understand it, this should affect all tabs in happa, right?

Yeah, this is for every one of them

> For my taste, the font color of the two tabs not selected is too dark here, making the `macOS` and `Windows` tab look like they are disabled. I'd make id a good deal brighter, but not as bright as the active tab's font.

Updated ✅ 

image

image

PR #2565, opened 2021-07-20
clusters: don't execute cluster-specific GS API requests if MAPI cluster FF is turned on

axbarsan:
This PR disables all the GS API cluster (and everything else related - node pools, releases) if the MAPI cluster management feature flag is turned on.

With this change and the feature flag enabled, we only execute GS API requests for:
- `/v4/info/` - when the app loads initially
- `/v5/users/` - on the `Users` page
- `/v5/analytics/` - when reporting analytics

Bonus points: This makes the app load super quickly (compared to before)


marians (approved):
No idea, just unblocking this. Also tested on gremlin.

PR #2570, opened 2021-07-23
Set URL prefix for sentry sourcemaps

axbarsan:


webpack.common.ts
```diff
@@ -123,6 +123,7 @@ const makeFeatureFlags = () => {
 };
 
 export const compilerConfig: Config = {
+  sourceMaps: true,
```
axbarsan, line 126: lol

PR #2573, opened 2021-07-26
Fix cluster name in MAPI UI to display `Cluster name` label on hover.

kuosandys:
Closes [giantswarm/giantswarm#18138](https://github.com/giantswarm/giantswarm/issues/18138)

### DIFF
--- a/src/components/UI/Display/Cluster/ClusterIDLabel/index.tsx
+++ b/src/components/UI/Display/Cluster/ClusterIDLabel/index.tsx
@@ -56,7 +56,7 @@ interface IClusterIDLabelProps extends React.ComponentPropsWithoutRef<'span'> {
 const ClusterIDLabel: React.FC<IClusterIDLabelProps> = ({
   clusterID,
   copyEnabled,
-  variant = ClusterIDLabelType.ID,
+  variant,
   ...props
 }) => {
   const [hasContentInClipboard, setClipboardContent] = useCopyToClipboard();
@@ -129,6 +129,10 @@ ClusterIDLabel.propTypes = {
   clusterID: PropTypes.string.isRequired,
   copyEnabled: PropTypes.bool,
   variant: PropTypes.oneOf(Object.values(ClusterIDLabelType)),
+};
+
+ClusterIDLabel.defaultProps = {
+  variant: ClusterIDLabelType.ID,
 };
 
 export default ClusterIDLabel;
--- a/src/components/UI/Display/Cluster/ClusterIDLabel/index.tsx
+++ b/src/components/UI/Display/Cluster/ClusterIDLabel/index.tsx
@@ -42,14 +42,21 @@ const Label = styled.span<{ clusterID: string }>`
   border-radius: 0.2em;
 `;
 
+export enum ClusterIDLabelType {
+  Name = 'name',
+  ID = 'ID',
+}
+
 interface IClusterIDLabelProps extends React.ComponentPropsWithoutRef<'span'> {
   clusterID: string;
   copyEnabled?: boolean;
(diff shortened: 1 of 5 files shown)


src/components/UI/Display/Cluster/ClusterIDLabel/index.tsx
```diff
@@ -71,7 +71,7 @@ const ClusterIDLabel: React.FC<IClusterIDLabelProps> = ({
       <OverlayTrigger
         overlay={
           <Tooltip id='idtooltip'>
-            <>Cluster ID: {clusterID}</>
+            <>Cluster name: {clusterID}</>
```
axbarsan, line 74: I'm sorry for not giving more context in the issue 😅 

In the legacy GS API features we use the `Cluster ID` naming, and in the new K8s Management API features  we call them `Cluster name`. It's technically the same thing, but for the MAPI implementations we aim to be closer to the terminology of K8s resources.

> More on K8s resource naming:
> https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names

That being said, the `ClusterIDLabel` component is used by both GS API and MAPI worlds, so just changing `ID` to `Name` doesn't work, because it wouldn't make sense any more in the legacy world (which is what our customers currently use).

Ideally there should be a way to configure which label is shown, either by specifying the label directly as a prop (`label='Name'`), or which variant of the component to use - with an enum (`variant={ClusterIDLabelType.Name}`).

axbarsan (changes requested):

marians (approved):

src/components/UI/Display/Cluster/ClusterIDLabel/index.tsx
kuosandys: Got it - thanks for clarifying!

src/components/UI/Display/Cluster/ClusterIDLabel/index.tsx
```diff
@@ -42,14 +42,21 @@ const Label = styled.span<{ clusterID: string }>`
...
 interface IClusterIDLabelProps extends React.ComponentPropsWithoutRef<'span'> {
   clusterID: string;
   copyEnabled?: boolean;
+  variant?: ClusterIDLabelType;
 }
 
 const ClusterIDLabel: React.FC<IClusterIDLabelProps> = ({
   clusterID,
   copyEnabled,
+  variant = ClusterIDLabelType.ID,
```
axbarsan, line 59: Throughout the codebase we use a component's `defaultProps` static property to set default values for props. [They are then handled internally by `react`](https://reactjs.org/docs/typechecking-with-proptypes.html#default-prop-values), and behave in the same way.

https://github.com/giantswarm/happa/blob/d8f609f8bd5a8d96797f46413811309bb49065d5/src/components/UI/Display/MAPI/clusters/ClusterDetail/ClusterDetailWidget/index.tsx#L64

The way you implemented it is cleaner and a bit closer to 2021, but for consistency, it would be great if we could use `defaultProps` here, as well.

If you think it's valuable, we can switch our codebase to using function argument defaults. It would definitely be more enjoyable to use with TypeScript.

axbarsan (commented):
Really cool 👍 

Thank you for going the extra mile and updating `storybook`, as well

src/components/UI/Display/Cluster/ClusterIDLabel/index.tsx
kuosandys: Let's stick with `defaultProps` for now for consistency's sake. I'll make the appropriate changes 👍

axbarsan (approved):
LGTM

PR #2577, opened 2021-07-28
Update storybook to v6.3.6

marians:


axbarsan:
Why do all these dependency updates manually? We have dependabot

marians:
> Why do all these dependency updates manually? We have dependabot

Can dependabot update several related dependencies together?

There was one PR open which tried updating one sentry dependency, but not the others.

axbarsan:
> > Why do all these dependency updates manually? We have dependabot
> 
> Can dependabot update several related dependencies together?
> 
> There was one PR open which tried updating one sentry dependency, but not the others.

No, it can't update multiple at once. But also, different versions are compatible with each other, so they don't have to be updated all at once

axbarsan (approved):

PR #2576, opened 2021-07-27
apps: write tests for cluster app management

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/18136

Nothing too crazy, just to create a base of happy path tests, so we can refactor easier.

PR #2578, opened 2021-07-28
change right single quatation mark to single quatation mark

wind0r:
When installing an app the description text under Namespace showed an \u2019 instead of a right single quotation.

While changing the String to a Template-String I also noticed that happa codebase uses single quotation for doesn't and this was the only place with a right single quotation.

image


marians:
Thanks @wind0r 

I think @axbarsan has just modified this in a recent PR. https://github.com/giantswarm/happa/pull/2576/files#diff-6079fb7296066161815fddf63c753fe01d0b4310c4f80a5db8dc5024f5fd13ceR127

PR #2582, opened 2021-08-02
Get Organization namespace from CRD

kuosandys:
Closes [giantswarm/giantswarm#18131](https://github.com/giantswarm/giantswarm/issues/18131).

The `Organization` CRD has a `namespace` field that should be used to get an organization's namespace. This will replace the current method of calculating the namespace from the organization name with a function.


testUtils/mockHttpCalls/securityv1alpha1/organization.ts
```diff
@@ -5,4 +5,7 @@ export const getOrganizationByName = {
     name: 'org1',
   },
   spec: {},
+  status: {
+    namespace: 'org-org1',
+  },
```
marians, line 10: I think the test case would be even more meaningful if the actual namespace value would be different from the pattern `org-` + org name.

src/components/MAPI/clusters/ClusterDetail/ClusterDetailOverview.tsx
```diff
@@ -36,14 +36,31 @@ const ClusterDetailOverview: React.FC<{}> = () => {
   const clientFactory = useHttpClientFactory();
   const auth = useAuthProvider();
 
-  const namespace = getOrgNamespaceFromOrgName(orgId);
-
   const clusterClient = useRef(clientFactory());
```
axbarsan, line 39: We shouldn't re-use the same client for both the cluster and organization requests. There is a limitation on the `HttpClientImpl`'s side, due to a poor decision I made when I first designed it: It is stateful, so when you configure a request, before executing it, you mutate `HttpClientImpl`'s internals.

The problem here can be that the organization request configures the client in a way, executes the request, and then when the cluster request configures it in a different way, but doesn't overwrite all the previously set configuration options, so the request might return unexpected results.

For now we can use a separate client for organizations, and a separate one for clusters.

src/components/MAPI/clusters/ClusterDetail/index.tsx
```diff
@@ -88,14 +88,44 @@ const ClusterDetail: React.FC<{}> = () => {
   const clientFactory = useHttpClientFactory();
   const auth = useAuthProvider();
 
-  const namespace = getOrgNamespaceFromOrgName(orgId);
-
   const clusterClient = useRef(clientFactory());
```
axbarsan, line 91: Same thing here

src/components/MAPI/clusters/GettingStarted/index.tsx
```diff
@@ -100,14 +100,44 @@ const GettingStarted: React.FC<IGettingStartedProps> = () => {
   const clientFactory = useHttpClientFactory();
   const auth = useAuthProvider();
 
-  const namespace = getOrgNamespaceFromOrgName(orgId);
-
   const clusterClient = useRef(clientFactory());
```
axbarsan, line 103: Same here

axbarsan (commented):
Really cool in general, just one concern

src/components/MAPI/clusters/ClusterDetail/ClusterDetailOverview.tsx
kuosandys: Thanks for the explanation!

testUtils/mockHttpCalls/securityv1alpha1/organization.ts
kuosandys: I think the test for `ClusterDetail` should be testing whether the namespace returned by the first mocked response can be used for getting the second response. Since they're both being mocked for the test, it would make more sense to test that, rather than the specific output.

Thanks for pointing this out - I'll make the changes for the test!

src/components/MAPI/clusters/ClusterDetail/ClusterDetailOverview.tsx
```diff
@@ -34,15 +34,18 @@ const ClusterDetailOverview: React.FC<{}> = () => {
   const { orgId, clusterId } = match.params;
 
   const clientFactory = useHttpClientFactory();
+  const orgClientFactory = useHttpClientFactory();
+
   const auth = useAuthProvider();
 
   const clusterClient = useRef(clientFactory());
+  const orgClient = useRef(orgClientFactory());
```
axbarsan, line 42: You can re-use the same factory. `clientFactory()` will create a new instance of `HttpClientImpl`, so the mutation thingy is not a problem anymore. It's just a matter of not using the same instance of `HttpClientImpl`

```suggestion
  const orgClient = useRef(clientFactory());
```

axbarsan (approved):
Really nice 👍

PR #2589, opened 2021-08-03
Add guide on deleting the current org

marians:
Follow-up to https://github.com/giantswarm/giantswarm/issues/16564

Adds a guide to the org details page for deleting the org via the Management API

## Preview

image


kuosandys (commented):
Looks good :+1:

src/components/MAPI/organizations/guides/DeleteOrganizationGuide.tsx
```diff
@@ -0,0 +1,66 @@
...
+      {...props}
+    >
+      <CLIGuideStepList>
+        <LoginGuideStep />
+        <CLIGuideStep
+          title='2. Delete the organization'
+          command={`kubectl delete organization ${organizationName}`}
+        >
+          <Text>
+            As a result, the organization and the associated namespace will be
```
axbarsan, line 53: Should we also mention that if there are BYOC credentials set up, then the organization will not be deleted until credentials are also deleted?

marians: That's not the only restriction, is it? Also clusters owned by the org would prevent deletion, right?

marians:
Added a note regarding deletion restrictions and updated the preview screenshot

src/components/MAPI/organizations/guides/DeleteOrganizationGuide.tsx
```diff
@@ -0,0 +1,72 @@
...
+        <CLIGuideStep
+          title='2. Delete the organization'
+          command={`kubectl delete organization ${organizationName}`}
+        >
+          <Text>
+            As a result, the organization and the associated namespace including
+            all resources in that namespace will be deleted.
+          </Text>
+          <Text>
+            <strong>Note:</strong> if there are any clusters associated with the
```
axbarsan, line 57: What about provider credentials? If those are set, we prevent a `Organization` CR from being deleted.

src/components/MAPI/organizations/guides/DeleteOrganizationGuide.tsx
```diff
@@ -0,0 +1,72 @@
...
+            organization via the label{' '}
+            <code>giantswarm.io/organization={`${organizationName}`}</code>, the
```
axbarsan, line 59:
```suggestion
            <code>giantswarm.io/organization={organizationName}</code>, the
```

src/components/MAPI/organizations/guides/DeleteOrganizationGuide.tsx
marians: I'll add another phrase

src/components/MAPI/organizations/guides/DeleteOrganizationGuide.tsx
marians: Done

axbarsan (approved):

PR #2607, opened 2021-08-04
Improve cluster deletion by requiring multiple confirmation steps

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/18297

This PR implements the GS API part of the change, with generic components that will be used for the MAPI implementation, as well.

The old deletion button was removed, and the Redux actions were modified to not require modals anymore.

⚠️  **Note:**
<details>
<summary>For SSO users, on AWS installations, users see `n/a` apps installed. That is because App management is currently a hybrid between MAPI and GS API there, and adding extra support for that adds too much complexity to the code. As we will bring MAPI cluster management to AWS soon, this would be a `won't-fix` from me</summary>

image

</details>

## Preview

<details>
<summary>Just chilling</summary>

image

</details>

<details>
<summary>Are you sure?</summary>

image

</details>

<details>
<summary>Are you REALLY sure?</summary>

image

</details>

<details>
<summary>Are you REALLY sure? Yes!</summary>

image

</details>


src/components/Cluster/ClusterDetail/ClusterDetailView.js
```diff
@@ -362,22 +352,6 @@ class ClusterDetailView extends React.Component {
                     workerNodesDesired={this.getDesiredNumberOfNodes()}
                   />
                 )}
-
-                <Section title='Delete This Cluster' flat>
```
axbarsan, line 366: Bye-bye 👋

src/components/UI/Display/MAPI/clusters/ClusterDetail/ClusterDetailDeleteAction.tsx
```diff
@@ -0,0 +1,205 @@
...
+  extends React.ComponentPropsWithoutRef<typeof Box> {
+  name: string;
+  description: string;
+  creationDate: string;
+  workerNodesCount: number;
+  nodePoolsCount: number;
+  userInstalledAppsCount?: number;
+  onDelete: () => void;
+  isLoading?: boolean;
+  variant?: ClusterDetailDeleteActionNameVariant;
```
axbarsan, line 44: You can configure which term is used to identify a cluster: `Name` or `ID`

src/stores/cluster/selectors.ts
```diff
@@ -146,3 +147,29 @@ export const selectIsClusterUpgrading = (clusterID: string) => (
 
   return isClusterUpdating(cluster);
 };
+
+export function selectClusterUserInstalledApps(clusterID: string) {
```
axbarsan, line 151: Moved this into a separate file so I can re-use it

src/stores/cluster/types.ts
```diff
@@ -174,20 +168,20 @@ export interface IBatchedClusterCreationErrorAction {
...
   type: typeof CLUSTER_DELETE_SUCCESS;
-  clusterId: string;
+  id: string;
   timestamp: string;
 }
 
 export interface IClusterDeleteErrorAction {
   type: typeof CLUSTER_DELETE_ERROR;
-  clusterId: string;
+  id: string;
```
axbarsan, line 184: Switched up the interfaces a bit: this gives us automatic loading flags using the `entityloading` reducer.

src/components/UI/Display/MAPI/clusters/ClusterDetail/ClusterDetailDeleteAction.tsx
marians: Very cool! Thanks for making it adapt to the legacy UI.

marians (approved):

JosephSalisbury:
Incredibly small nitpick - do we want a ':' at the end of the 'please enter the cluster ID' message?

axbarsan:
> Incredibly small nitpick - do we want a ':' at the end of the 'please enter the cluster ID' message?

It was in the spec but I decided not to include it because it looked weird. Do you think it helps?

JosephSalisbury:
if you think it looks weird, then totally fine with dropping it - i saw it in the spec / would _maybe_ kind of expect it, but i don't have a strong opinion on a single colon :D

PR #2608, opened 2021-08-05
Improve MAPI cluster deletion by requiring multiple confirmation steps

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/18297

This PR replaces the existing MAPI cluster deletion logic with the same one that we just added to the GS API. It looks the same as before 👍


src/components/MAPI/clusters/ClusterDetail/ClusterDetailActions.tsx
```diff
@@ -0,0 +1,222 @@
...
+        messageType.ERROR,
+        messageTTL.LONG,
+        errorMessage
+      );
+
+      setIsDeleting(false);
+    }
+  };
+
+  const isLoading = isLoadingResources || isDeleting;
```
axbarsan, line 171: If you come from the `Overview` tab, then you won't see any loading animations at first, because all the data is already cached. However, if you visit the `/actions` page directly, the deletion button will be disabled and you will see a loading animation.

src/components/MAPI/clusters/ClusterDetail/ClusterDetailOverview.tsx
```diff
@@ -104,15 +103,6 @@ const ClusterDetailOverview: React.FC<{}> = () => {
         basis='100%'
       />
       <ClusterDetailWidgetCreated cluster={cluster} basis='100%' />
-
-      {cluster && (
```
axbarsan, line 108: Bye 👋

src/components/MAPI/clusters/ClusterDetail/utils.ts
```diff
@@ -34,21 +36,55 @@ export async function updateClusterDescription(
 export async function deleteCluster(
```
axbarsan, line 36: Refactored this a bit, so it would be consistent with `deleteNodePool()`

src/components/MAPI/clusters/Clusters.tsx
```diff
@@ -95,7 +95,7 @@ const Clusters: React.FC<{}> = () => {
     clusterListIsValidating;
 
   const sortedClusters = useMemo(
-    () => clusterList?.items.sort(compareClusters),
+    () => clusterList?.items.slice().sort(compareClusters),
```
axbarsan, line 98: After deleting a cluster and mutating the cache, the newly created array is a `ReadonlyArray`, which means that we need to copy it before sorting, so it won't throw any errors.

src/components/UI/Display/MAPI/clusters/ClusterDetail/ClusterDetailDeleteActionClusterName.tsx
```diff
@@ -32,7 +32,7 @@ const ClusterDetailDeleteActionClusterName: React.FC<IClusterDetailDeleteActionC
 
   return (
     <Box direction='row' gap='small' align='baseline' {...props}>
-      <Text>If yes, please enter the cluster {variant}</Text>
+      <Text>If yes, please enter the cluster {variant}:</Text>
```
axbarsan, line 35: Bonus for Joe

marians (approved):

axbarsan:
> Looks good! Btw something minor, I just noticed that hovering over the relative time displays the creation timestamp in the tooltip as well - is this intentional?
> 
> Screen Shot 2021-08-05 at 15 13 41

That's a side effect of using `relativeDate()` for printing the date relative to now. We use the same function everywhere in the app, regardless if the context needs a tooltip or not.

It is indeed a bit weird, and we also have it on the `Created` widget, on the `Overview` tab of the MAPI `Cluster detail` view. It is a bit out of the scope of this PR, do you want to create an issue for it?

kuosandys:
> > Looks good! Btw something minor, I just noticed that hovering over the relative time displays the creation timestamp in the tooltip as well - is this intentional?
> > Screen Shot 2021-08-05 at 15 13 41
> 
> That's a side effect of using `relativeDate()` for printing the date relative to now. We use the same function everywhere in the app, regardless if the context needs a tooltip or not.
> 
> It is indeed a bit weird, and we also have it on the `Created` widget, on the `Overview` tab of the MAPI `Cluster detail` view. It is a bit out of the scope of this PR, do you want to create an issue for it?

That's alright, I think aligning everything makes sense. Thanks for clarifying!

PR #2609, opened 2021-08-05
Fix cluster details page - Subscription ID sometimes not appearing

kuosandys:
Closes [giantswarm/giantswarm#18326](https://github.com/giantswarm/giantswarm/issues/18326).

After creating a cluster, the Subscription ID field's placeholder remains visible until the user navigates away and back to the page. This is because the `subscriptionID` of the resource is not immediately available after cluster creation.

This fix makes a distinction between whether the `subscriptionID` is `undefined` (field not available yet, but the rest of the information from the `providerCluster` is) or the `providerCluster` is undefined.

Only `subscriptionID` is undefined:
Screen Shot 2021-08-06 at 09 47 41

`providerCluster` is undefined:
Screen Shot 2021-08-06 at 09 47 15


src/components/MAPI/clusters/ClusterDetail/ClusterDetailWidgetProvider.tsx
```diff
@@ -118,22 +125,30 @@ const ClusterDetailWidgetProvider: React.FC<IClusterDetailWidgetProviderProps> =
...
-          </StyledLink>
-        )}
+      <ClusterDetailWidgetOptionalValue
+        value={accountID}
+        loaderWidth={300}
+        replaceEmptyValue={false}
+      >
+        {(value) =>
+          value === '' ? (
+            <NotAvailable />
```
kuosandys, line 135: Passing in the `NotAvailable` component directly to keep it visually aligned with the "Subscription ID" label.

marians (approved):

src/components/MAPI/clusters/ClusterDetail/ClusterDetailWidgetProvider.tsx
```diff
@@ -84,7 +85,13 @@ const ClusterDetailWidgetProvider: React.FC<IClusterDetailWidgetProviderProps> =
 }) => {
   const region = providerCluster?.spec.location;
 
-  const accountID = providerCluster?.spec.subscriptionID;
+  const accountID = useMemo(() => {
+    if (!providerCluster) return undefined;
+    if (!providerCluster?.spec.subscriptionID) return '';
+
+    return providerCluster?.spec.subscriptionID;
```
axbarsan, line 92: This is no longer optional due to the falsy value check above

```suggestion
    if (!providerCluster.spec.subscriptionID) return '';

    return providerCluster.spec.subscriptionID;
```

axbarsan (approved):

PR #2613, opened 2021-08-09
Mapi skeleton loading animations

kuosandys:
Closes [giantswarm/giantswarm#18145](https://github.com/giantswarm/giantswarm/issues/18145).

In the MAPI UI, various loading placeholder components are used. These implementations can be aligned by using the same common components.

When possible, the `<OptionalValue>` component is used. From the original issue: "It displays falsy values as `n/a`, flashes when values update and has configurable loading skeleton sizes." The child component `LoadingPlaceholder` is used when the use of the `<OptionalValue>` component is not possible.

Value is loading:
Screen Shot 2021-08-09 at 10 10 03

Value loaded:
Screen Shot 2021-08-09 at 10 10 14

Value is falsy:
Screen Shot 2021-08-09 at 10 10 22


src/components/UI/Display/OptionalValue/stories/OptionalValue.stories.tsx
```diff
@@ -0,0 +1,46 @@
+import { Meta, Story } from '@storybook/react/types-6-0';
```
axbarsan, line 1:
```suggestion
import { Meta, Story } from '@storybook/react';
```

This is no longer necessary 👍

axbarsan (approved):
Damn 🚀

PR #2614, opened 2021-08-09
Don't display tooltips with the full date on relative dates when they are not needed

axbarsan:
As noticed by Sandy in https://github.com/giantswarm/happa/pull/2608#issuecomment-893451618, we always display a tooltip with the full date while hovering on relative dates, even when the full date is displayed near the relative one.

Turns out we actually have an utility that only formats the relative date, and does not return a React component.
So with this PR, we no longer display full date tooltips over relative dates, if the full date is already displayed in the UI.


kuosandys (approved):
😄

marians (approved):

marians:
Details, details, details! ❤️

PR #2615, opened 2021-08-09
Fix version picker closing itself when ticking 'Use test versions'

axbarsan:
Quick and easy fix. Previously when ticking the `Use test versions` checkbox (which is only available to GS staff, hence the `kind/refactor` label), the dropdown would close itself before you could interact with any of the list items. This fixes that.

### DIFF
--- a/src/components/UI/Controls/VersionPicker/VersionPicker.tsx
+++ b/src/components/UI/Controls/VersionPicker/VersionPicker.tsx
@@ -160,13 +160,6 @@ const VersionPicker: React.FC<IVersionPickerProps> = ({
     setIncludeTestVersions(event.target.checked);
   };
 
-  const handleBlur = (onBlurHandler: () => void) => () => {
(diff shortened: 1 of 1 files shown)


src/components/UI/Controls/VersionPicker/VersionPicker.tsx
```diff
@@ -160,13 +160,6 @@ const VersionPicker: React.FC<IVersionPickerProps> = ({
     setIncludeTestVersions(event.target.checked);
   };
 
-  const handleBlur = (onBlurHandler: () => void) => () => {
-    setTimeout(() => {
-      onBlurHandler();
-      // eslint-disable-next-line no-magic-numbers
-    }, 300);
-  };
```
axbarsan, line 168: No idea what this was solving, but whatever the problem was, it doesn't happen anymore

kuosandys (approved):

PR #2616, opened 2021-08-09
Disable username and password login when SSO feature flag is enabled

axbarsan:
## Before

image

## After

image


marians (approved):
My understanding was that we would add a separate feature flag.

But this way here is fine with me, too. If we run into a customer that wants a transition period, we can add an extra feature flag for turning on/off the legacy login.

axbarsan:
> But this way here is fine with me, too. If we run into a customer that wants a transition period, we can add an extra feature flag for turning on/off the legacy login.

Keeping it simple 😬

PR #2617, opened 2021-08-09
Add CLI guides for listing and creating clusters

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/18343

Nothing too crazy, following the spec. The cluster creation guide gets data such as the provider, organization and cluster name from the state (so when you refresh the page, you see that you get the same UID in the guide as you do in the form)

## Preview

<details>
<summary>List clusters</summary>

image

</details>

<details>
<summary>Create cluster</summary>

image

</details>

### DIFF
--- a/src/components/MAPI/clusters/guides/CreateClusterGuide.tsx
+++ b/src/components/MAPI/clusters/guides/CreateClusterGuide.tsx
@@ -1,8 +1,16 @@
 import { Text } from 'grommet';
 import * as docs from 'lib/docs';
 import LoginGuideStep from 'MAPI/guides/LoginGuideStep';
+import {
+  getCurrentInstallationContextName,
+  makeKubectlGSCommand,
+  withContext,
+  withFormatting,
+  withTemplateCluster,
+} from 'MAPI/guides/utils';
 import PropTypes from 'prop-types';
 import React from 'react';
+import { useSelector } from 'react-redux';
 import { Providers } from 'shared/constants';
 import { PropertiesOf } from 'shared/types';
 import CLIGuide from 'UI/Display/MAPI/CLIGuide';
@@ -15,14 +23,24 @@ interface ICreateClusterGuideProps
   provider: PropertiesOf<typeof Providers>;
   clusterName: string;
   organizationName: string;
+  releaseVersion?: string;
+  description?: string;
+  labels?: Record<string, string>;
+  controlPlaneAZs?: string[];
 }
 
 const CreateClusterGuide: React.FC<ICreateClusterGuideProps> = ({
   provider,
   clusterName,
   organizationName,
+  releaseVersion,
+  description,
+  labels,
+  controlPlaneAZs,
   ...props
 }) => {
+  const context = useSelector(getCurrentInstallationContextName);
+
   return (
     <CLIGuide
       title='Create a cluster via the Management API'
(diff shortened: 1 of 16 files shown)


src/components/MAPI/clusters/guides/CreateClusterGuide.tsx
```diff
@@ -0,0 +1,90 @@
...
+              href: docs.kubectlGSInstallationURL,
+              external: true,
+            },
+            {
+              label: 'kubectl gs template cluster command',
+              href: docs.kubectlGSTemplateClusterURL,
+              external: true,
+            },
+            {
+              label: 'Cluster CRD schema',
```
axbarsan, line 43: Also added a link to the `Cluster` CRD, to keep it consistent with the organizations guides

marians:
Checking now

marians:
As discussed in SIG UX, let's add `--context CONTEXT` here already. I think the flag should appear as far to the left as possible. E. g.

    kubectl gs --context gs-gremlin get clusters --all-namespaces

marians:
In the cluster creation form context, the idea was that the `kubectl gs template cluster` command in the guide adapts to the user's input in the form. I can't see that happening.

axbarsan:
> In the cluster creation form context, the idea was that the `kubectl gs template cluster` command in the guide adapts to the user's input in the form. I can't see that happening.

The organization and name do adapt. Did you also want to include all the other flags that `kubectl gs` supports?

marians:
> Did you also want to include all the other flags that kubectl gs supports?

Yes, all the flags that are relevant to the form displayed. (I'm not sure if that covers all flags or leaves some unused.)

src/components/MAPI/guides/utils.ts
```diff
@@ -0,0 +1,212 @@
...
+export type KubectlGSCommandModifier = (
+  parts: readonly string[]
+) => readonly string[];
+
+/**
+ * Construct a `kubectl gs` command example.
+ *
+ * You can use modifiers to add arguments or flags to the command.
+ * */
+export function makeKubectlGSCommand(
```
axbarsan, line 18: I created a new builder function for creating `kubectl gs` examples.

It uses modifiers to add arguments/flags to the commands, and each command has a custom modifier. There's also a modifier for automatically formatting the command output, so it spans multiple lines.

axbarsan:
I updated the code and now we have a (over-engineered) way to create `kubectl gs` command templates for guides. Please take a look ✌️

src/components/MAPI/guides/utils.ts
```diff
@@ -0,0 +1,212 @@
...
+      hasArgs = true;
+      newParts.push(part);
+      // If the next value is not a flag, add an argument delimiter.
+      if (peekPart && !isFlag(peekPart)) {
+        newParts.push(' ');
+      }
+    }
+
+    if (hasFlags && !hasArgs) {
+      newParts.unshift(formattingLineBreak);
```
axbarsan, line 207: I hate this, but I guess it works for now

marians:
All right, the good news first: It seems to be working. Nice!

Here for some improvements regarding cluster creation.

1. The idea was to keep the `template cluster` command as short as possible. So if a default is used, the according flag should not be applied. Specifically there is
  - `--control-plane-az 2` despite the selection being the default `Automatic selection` (which should be done in the backend as far as I understand). Here I'm actually not sure whether I should see this as a logical error or simply as a cosmetic detail. I don't know if this is even implemented properly in `kubectl gs template cluster`.
  - `--release 14.1.4` despite it being the default. This is mostly cosmetic, however there is a difference between using the command without a `--release` (which means: apply the default at the time of execution) and having a release set.

2. The `--context` flag.
  - Is not supported in the `template cluster` command, as I just learned. Probably as there is no API communication.
  - Should be present in the `kubectl apply` command in step 3.

3. Indentation of the commands is missing

None of these is blocking.

marians (approved):

axbarsan:
> 1. The `--context` flag.
> 
> * Is not supported in the `template cluster` command, as I just learned. Probably as there is no API communication.
> * Should be present in the `kubectl apply` command in step 3.

Totally missed the `template cluster` one 😆 Updated the other one.

> 2. Indentation of the commands is missing

Fixed ✅

axbarsan:
> * `--release 14.1.4` despite it being the default. This is mostly cosmetic, however there is a difference between using the command without a `--release` (which means: apply the default at the time of execution) and having a release set.

But the user sees the latest version selected in the cluster creation form (e.g. `15.0.0`). I would be confused if when I selected the latest release version, the command example would remove the `--release` flag. The user doesn't know that we use defaulting to set the latest version when `--release` is not used.

PR #2618, opened 2021-08-09
Access control: Improve deletion confirmation prompt for deleting service account subjects

kuosandys:
Closes [giantswarm/giantswarm#16790](https://github.com/giantswarm/giantswarm/issues/16790).

The confirmation prompt displayed when removing a binding between a service account subject and a role should be more specific, especially to clarify that removing the binding does not remove the service account that may have been created when the binding was created.

### Confirmation prompt
Replacing the text "Are you sure?" with a more specific message for removing bindings (service account subjects only):
Screen Shot 2021-08-09 at 21 45 04

### Confirmation buttons
The label for the delete button has also been changed from `Yes, delete it` to `Remove`.
The `Remove` and `Cancel` buttons should have the same height now after Adrian's button refactoring 👍 

### Z-index
Changing the z-index of the tooltip shown when hovering over the delete icon, so it appears behind the prompt:
Screen Shot 2021-08-09 at 21 45 35

### DIFF
--- a/src/components/UI/Display/MAPI/AccessControl/AccessControlSubjectSetItem.tsx
+++ b/src/components/UI/Display/MAPI/AccessControl/AccessControlSubjectSetItem.tsx
@@ -34,6 +34,10 @@ const StyledAnchor = styled(Anchor)`
   }
 `;
 
+const StyledDrop = styled(Drop)`
+  z-index: 1071; /* To appear above the Bootstrap Tooltip, which has a z-index of 1070 */
+`;
+
 interface IAccessControlSubjectSetItemProps
   extends Omit<IAccessControlSubjectSetRenderer, 'name'>,
     React.ComponentPropsWithoutRef<typeof Box> {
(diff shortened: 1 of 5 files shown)


src/components/UI/Display/MAPI/AccessControl/AccessControlSubjectSetItem.tsx
```diff
@@ -34,11 +34,16 @@ const StyledAnchor = styled(Anchor)`
   }
 `;
 
+const StyledDrop = styled(Drop)`
+  z-index: 1071; /* To appear above the Bootstrap Tooltip, which has a z-index of 1070 */
```
axbarsan, line 38: That's very specific 😆

axbarsan (approved):
LGTM ✅ 🥇

src/components/UI/Display/MAPI/AccessControl/AccessControlSubjectSetItem.tsx
kuosandys: I know 😅 it's probably not the smartest solution, so definitely let me know if there's a better way!

PR #2627, opened 2021-08-12
Access control: Improve service account flash messages

kuosandys:
Closes [giantswarm/giantswarm#16792](https://github.com/giantswarm/giantswarm/issues/16792).

Flash messages for actions related to binding roles with service accounts subjects can be more specific.

For service accounts, the messages `Subject added successfully` (and its plural variation) for creation, and `Subject <name> deleted successfully` for deletion have been replaced by the following:

#### 1. Binding an existing service account to a role:
Screen Shot 2021-08-12 at 14 50 28

#### 2. Creating a new service account and binding it to a role:
Screen Shot 2021-08-12 at 14 51 15

It is also possible to enter multiple service account names into the input field to bind them to a role. In this case, the service account names are aggregated into one flash message (a similar aggregation is performed for creating and binding multiple service accounts):
Screen Shot 2021-08-12 at 14 51 52
 
In the case where there is a combination of creating role bindings for existing and newly created service accounts, 2 flash messages will be displayed:
Screen Shot 2021-08-12 at 14 52 25

#### 3. Removing a service account binding:
Screen Shot 2021-08-12 at 14 50 38

### Service account creation errors:
The phrase `Could not add subject` has been replaced with a more specific message:
Screen Shot 2021-08-12 at 14 49 04


src/components/MAPI/organizations/AccessControl/AccessControlRoleSubjects.tsx
```diff
@@ -191,23 +223,64 @@ const AccessControlRoleSubjects: React.FC<IAccessControlRoleSubjectsProps> = ({
...
+        const createdAccountsMessage = getBindServiceAccountSuccessMessage(
+          values,
+          statuses,
+          ui.AccessControlRoleSubjectStatus.Created
+        );
+        if (createdAccountsMessage.length > 0) {
+          new FlashMessage(
+            createdAccountsMessage,
+            messageType.SUCCESS,
+            messageTTL.MEDIUM
```
kuosandys, line 239: The messages are a bit longer, so I increased the length of time the flash message is displayed

src/components/MAPI/organizations/AccessControl/utils.ts
```diff
@@ -550,16 +544,11 @@ export async function ensureServiceAccount(
   auth: IOAuth2Provider,
   name: string,
   namespace: string
-): Promise<corev1.IServiceAccount> {
+): Promise<ui.AccessControlRoleSubjectStatus> {
```
axbarsan, line 547: Really cool ✅

src/components/MAPI/organizations/AccessControl/AccessControlRoleSubjects.tsx
```diff
@@ -155,7 +184,10 @@ interface IAccessControlRoleSubjectsProps
   roleName: string;
   namespace: string;
   permissions: ui.IAccessControlPermissions;
-  onAdd: (type: ui.AccessControlSubjectTypes, names: string[]) => Promise<void>;
+  onAdd: (
+    type: ui.AccessControlSubjectTypes,
+    names: string[]
+  ) => Promise<void | ui.AccessControlRoleSubjectStatus[]>;
```
axbarsan, line 190: Is it required to still allow `Promise<void>`? When do we resolve without a value?

src/components/MAPI/organizations/AccessControl/AccessControlRoleSubjects.tsx
```diff
@@ -146,6 +146,35 @@ const mapValueToSetItem = (stateValue: IStateValue) => (
   };
 };
 
+const getBindServiceAccountSuccessMessage = (
+  values: string[],
+  statuses: ui.AccessControlRoleSubjectStatus[],
+  statusFilter: ui.AccessControlRoleSubjectStatus
+): string => {
+  const accounts = values.filter((_, i) => statuses[i] === statusFilter);
```
axbarsan, line 154: We do not enforce in any way that for a value in `values` we also have a status in `statuses`, and that it is present at the same index. This isn't very safe, as the implementation of this depends on `handleAdd()` in `src/components/MAPI/organizations/AccessControl/index.tsx`.

Do you think it would be worth combining the value and status into a single data structure?

src/components/MAPI/organizations/AccessControl/utils.ts
```diff
@@ -550,16 +543,11 @@ export async function ensureServiceAccount(
   auth: IOAuth2Provider,
   name: string,
   namespace: string
-): Promise<corev1.IServiceAccount> {
+): Promise<ui.IAccessControlServiceAccounts> {
```
kuosandys, line 546: Changed this to return an object that associates the service account name with its status.

src/components/MAPI/organizations/AccessControl/index.tsx
```diff
@@ -70,7 +71,12 @@ const AccessControl: React.FC<IAccessControlProps> = ({
     names: string[]
   ) => {
     try {
-      if (!activeRole || !data) return Promise.resolve();
+      if (!activeRole || !data) return Promise.resolve([]);
```
kuosandys, line 74: This was returning `Promise<void>` - changing this to resolve to the same type so the case for a `void` value would not be necessary

src/components/MAPI/organizations/AccessControl/AccessControlRoleSubjects.tsx
```diff
@@ -146,6 +146,53 @@ const mapValueToSetItem = (stateValue: IStateValue) => (
   };
 };
 
+const formatAccountNames = (accountNames: string[]): string => {
+  return accountNames
+    .map((accountName) => `<code>${accountName}</code>`)
+    .join(', ');
+};
+
+const getBindServiceAccountSuccessMessages = (
```
kuosandys, line 155: Changed this to return an array of strings - only the non-empty strings will result in a flash message

axbarsan (approved):
LGTM 🎂

marians (approved):

src/components/MAPI/organizations/AccessControl/AccessControlRoleSubjects.tsx
```diff
@@ -153,42 +153,59 @@ const formatAccountNames = (accountNames: string[]): string => {
...
+  accounts: ui.IAccessControlServiceAccount[],
+  statuses = Object.values(ui.AccessControlRoleSubjectStatus)
 ): string[] => {
-  const messages = statusFilters.map((statusFilter) => {
-    const accounts = statuses.filter(
-      (account) => account.status === statusFilter
-    );
+  const accountsByStatus = {} as Record<
+    ui.AccessControlRoleSubjectStatus,
+    string[] | null
```
axbarsan, line 161: When can a value be null?

src/components/MAPI/organizations/AccessControl/AccessControlRoleSubjects.tsx
```diff
@@ -153,42 +153,59 @@ const formatAccountNames = (accountNames: string[]): string => {
...
+  >;
+
+  for (const account of accounts) {
+    accountsByStatus[account.status] = accountsByStatus[account.status]?.concat(
+      account.name
+    ) || [account.name];
```
axbarsan, line 167: Why use `.concat()` instead of `.push()`?

kuosandys: I don't think `.push()` returns the array after adding a new element, so we wouldn't be able to assign it back to the key

src/components/MAPI/organizations/AccessControl/AccessControlRoleSubjects.tsx
kuosandys: I was thinking since I'm looping through accounts to populate `accountsByStatus`, there is the possibility that it does not contain every value from the enum `AccessControlRoleSubjectStatus`, but I think `null` is the wrong choice here.

Would a better change be to remove that and loop through the entries of `accountsByStatus` rather than all the possible `statuses`?

src/components/MAPI/organizations/AccessControl/AccessControlRoleSubjects.tsx
```diff
@@ -146,6 +146,63 @@ const mapValueToSetItem = (stateValue: IStateValue) => (
...
+): string[] => {
+  const accountsByStatus = {} as Record<
+    ui.AccessControlRoleSubjectStatus,
+    string[]
+  >;
+  for (const account of accounts) {
+    if (!accountsByStatus.hasOwnProperty(account.status)) {
+      accountsByStatus[account.status] = [];
+    }
+    accountsByStatus[account.status].push(account.name);
```
kuosandys, line 166: Rewrite this with `push()` instead of `concat()`

src/components/MAPI/organizations/AccessControl/AccessControlRoleSubjects.tsx
```diff
@@ -146,6 +146,63 @@ const mapValueToSetItem = (stateValue: IStateValue) => (
...
+  }
+
+  const messages = [];
+
+  for (const [status, filteredAccounts] of Object.entries(accountsByStatus)) {
```
kuosandys, line 171: Loop through `accountsByStatus` instead of `statuses`

PR #2630, opened 2021-08-12
Use grommet's DataTable component for complex tables, instead of the bootstrap one

axbarsan:
This PR customizes `grommet`'s `<DataTable />` component, so we can use it for complex tables. As opposed to our regular `<Table />` component, it offers stuff like sortable/searchable/resizable columns. Check [this](https://v2.grommet.io/datatable) out for more info about it.

It is an easy replacement for `react-bootstrap-table`, as the only differences between the two are some changed configuration property names. This is currently only used in the legacy parts of our app, but now that we have it, we can use it in other places, as well.

As for styling, it tries to look the same as the regular `<Table />` (with extra fun like sorting arrows).

This is an example of how it looks.

image


package.json
```diff
@@ -89,7 +89,6 @@
     "platform": "1.3.6",
     "react": "17.0.2",
     "react-bootstrap": "0.33.1",
-    "react-bootstrap-table-next": "4.0.3",
```
axbarsan, line 92: 🥳

marians (approved):
LGTM, didn't test

PR #2636, opened 2021-08-16
Use `GenericResponseError` for typing MAPI error responses

kuosandys:
Closes [giantswarm/giantswarm#18144](https://github.com/giantswarm/giantswarm/issues/18144).

This PR updates the error response typing for MAPI features that were previously typed as `GenericResponse`. These are replaced with the `GenericResponseError` type, which extends the built-in `Error` object.


marians (approved):
Fine if it works :)

axbarsan (approved):
Noice

PR #2642, opened 2021-08-17
Use function to produce CRD schema docs URL

marians:
This replaces CRD docs schema constants with a function.

Rationale: All CRD schema pages are created programmatically. It feels wrong to maintain lots of these URLs as constants.

The CRD names are maintained in a hierarchical const object, grouped by publisher domain, to allow for DRY use and IDE auto-completion.

### DIFF
--- a/src/components/MAPI/clusters/guides/InspectClusterGuide.tsx
+++ b/src/components/MAPI/clusters/guides/InspectClusterGuide.tsx
@@ -50,16 +50,12 @@ const InspectClusterGuide: React.FC<IInspectClusterGuideProps> = ({
             },
             {
               label: 'AzureCluster CRD schema',
-              href: docs.crdSchemaURL(
-                'azureclusters.infrastructure.cluster.x-k8s.io'
-              ),
+              href: docs.crdSchemaURL(docs.crds.xk8sio.azureCluster),
               external: true,
             },
             {
               label: 'AzureMachine CRD schema',
-              href: docs.crdSchemaURL(
-                'azuremachines.infrastructure.cluster.x-k8s.io'
-              ),
+              href: docs.crdSchemaURL(docs.crds.xk8sio.azureMachine),
               external: true,
             },
             {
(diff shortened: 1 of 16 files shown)


marians:
@kuosandys What do you think here?

kuosandys:
> @kuosandys What do you think here?

Since they all have the same base URL, this makes sense. But I think a small drawback would be the lack of auto-completion - the path for each CRD schema doc would have to be passed into the function every time as a string.

If the path to the CRD doc is generated automatically and matches the CR's full name, would it make sense in the long run to keep track of the names as constants? Then that could be passed into the function as an argument instead, e.g. `getCRDSchemaURL(CR.SomeResource)`

marians:
@kuosandys You pretty much confirm what I suspected...

What do you think about the way proposed in https://github.com/giantswarm/happa/pull/2642/commits/99607e6d23da767e58eac1296ba6f517fb40d0c0 ?

kuosandys:
> @kuosandys You pretty much confirm what I suspected...
> 
> What do you think about the way proposed in [99607e6](https://github.com/giantswarm/happa/commit/99607e6d23da767e58eac1296ba6f517fb40d0c0) ?

That's definitely more preferable than strings 👍

src/components/MAPI/clusters/guides/InspectClusterGuide.tsx
```diff
@@ -45,17 +45,21 @@ const InspectClusterGuide: React.FC<IInspectClusterGuideProps> = ({
...
               label: 'Cluster CRD schema',
-              href: docs.clusterCRDSchemaURL,
+              href: docs.crdSchemaURL(docs.crds.xk8sio.cluster),
               external: true,
             },
             {
               label: 'AzureCluster CRD schema',
-              href: docs.azureClusterCRDSchemaURL,
+              href: docs.crdSchemaURL(
+                'azureclusters.infrastructure.cluster.x-k8s.io'
```
kuosandys, line 54: Could this one be grouped under `xk8sio` as well?

marians: Yes

kuosandys (approved):

PR #2645, opened 2021-08-19
Add CLI guide for creating a node pool

kuosandys:
Towards [giantswarm/giantswarm#18343](giantswarm/giantswarm#18343).

This PR adds a CLI guide to the Cluster Details > Worker Nodes tab > Add Node Pool form, and provides information on how to template and create a node pool using the Management API.

The contents of the node pool manifest aligns with user input from the Add Node Pool form.

Collapsed:
Screen Shot 2021-08-19 at 10 23 18

Expanded:
Screen Shot 2021-08-19 at 16 38 26
Screen Shot 2021-08-19 at 16 38 46

### DIFF
--- a/src/components/MAPI/workernodes/guides/CreateNodePoolGuide.tsx
+++ b/src/components/MAPI/workernodes/guides/CreateNodePoolGuide.tsx
@@ -95,7 +95,7 @@ const CreateNodePoolGuide: React.FC<ICreateNodePoolGuide> = ({
         />
         <CLIGuideStep
           title='3. Apply the manifest'
-          command={`kubectl apply --context ${context} -f cluster-${clusterName}-nodepool.yaml`}
+          command={`kubectl --context ${context} apply -f cluster-${clusterName}-nodepool.yaml`}
         >
           <Text>As a result, a node pool will get created.</Text>
         </CLIGuideStep>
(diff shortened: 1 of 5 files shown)


marians:
Looking good already!

Regarding the behaviour: The general idea is that we only show required flags. If a flag sets a value just like the default would, then we should omit it.

Example: When nodes count min=3 is the default, we should not have the flag. Currently the flag only disappears when min is 0. Same for max. Our default is 10 so the flag should not be present when max=10, but it should be present if max is set to 0.

kuosandys:
> Looking good already!
> 
> Regarding the behaviour: The general idea is that we only show required flags. If a flag sets a value just like the default would, then we should omit it.
> 
> Example: When nodes count min=3 is the default, we should not have the flag. Currently the flag only disappears when min is 0. Same for max. Our default is 10 so the flag should not be present when max=10, but it should be present if max is set to 0.

@marians I've made the changes and updated the screenshots - the flags should only be displayed if they differ from the default values

marians:
Availability zones in form and CLI guide don't match:

image

image

marians:
Wait, my bad. That's actually working as expected.

marians:
image

The `--context gs-ghost` flag should ideally be in the 2nd position, right after `kubectl`.

marians (approved):
Awesome!

src/components/MAPI/workernodes/guides/CreateNodePoolGuide.tsx
```diff
@@ -0,0 +1,120 @@
...
+              nodesMin,
+              nodesMax,
+              output: `cluster-${clusterName}-nodepool.yaml`,
+            }),
+            withFormatting()
+          )}
+        />
+        <CLIGuideStep
+          title='3. Apply the manifest'
+          command={`kubectl --context ${context} apply -f cluster-${clusterName}-nodepool.yaml`}
```
kuosandys, line 98: Updated `context` flag's position 👍

PR #2647, opened 2021-08-19
Create feature flag value for mapiAuth (SSO)

marians:
Towards https://github.com/giantswarm/giantswarm/issues/17978

This PR

- removes the logic that all testing installations automatically have SSO auth enabled, while production ones don't.
- adds a feature flag configuration value `mapiAuth`

This requires a config change before releasing.

### DIFF
--- a/helm/happa/values.yaml
+++ b/helm/happa/values.yaml
@@ -39,6 +39,7 @@ happa:
   host: ""
   letsencrypt: ""
   featureFlags:
+    mapiAuth: false
     mapiClusters: false
 
 happaapi:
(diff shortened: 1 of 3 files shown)


helm/happa/values.yaml
```diff
@@ -39,6 +39,7 @@ happa:
   host: ""
   letsencrypt: ""
   featureFlags:
+    mapiAuth: false
     mapiClusters: false
```
kopiczko, line 43: Does one feature make sense without the other?

marians: Yes

marians: mapiClusters also cannot be used on AWS, while mapiAuth can be used on all providers

kopiczko (approved):

PR #2651, opened 2021-08-20
Modify "List clusters ..." MAPI guide to list clusters for the current org namespace only

marians:
The title says it.

The happa cluster list shows all clusters in the selected organization. With this change, the CLI guide matches that behaviour.

In addition, this method should also be more robust, as it would also work for a users that does not have cluster-admin permissions.

Towards https://github.com/giantswarm/giantswarm/issues/18343


src/components/MAPI/clusters/Clusters.tsx
```diff
@@ -216,7 +216,7 @@ const Clusters: React.FC<{}> = () => {
           </Keyboard>
         </Box>
         <Box margin={{ top: 'medium' }} direction='column' gap='small'>
-          <ListClustersGuide />
+          <ListClustersGuide namespace={`org-${selectedOrgName}`} />
```
axbarsan, line 219: It is not guaranteed that the namespace is correct, as `selectedOrgName` is not normalized (it can be an old organization name with special characters).

In other places we fetch the `Organization` CR and we get the namespace from its `.status` field. It is a little bit much, but this way we can always be 100% sure that the namespace is correct.
https://github.com/giantswarm/happa/blob/846601537cc37c5921d0256aec53c74e768062fe/src/components/MAPI/clusters/ClusterDetail/index.tsx#L103

We could get away with getting the org namespace from one of the clusters in the list, but that means that when there are no clusters, we can't display the CLI guide.

marians: Very good point.

Yes, we should definitely get the correct namespace name from the CR. I think that goes beyond the scope of what I can do efficiently. I'd be happy of you or @kuosandys could take this one over then.

axbarsan (approved):

PR #2656, opened 2021-08-24
Hide Users page if the customer SSO feature flag is enabled

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/18597

Simply put, right now we have a `Users` page visible to GS staff. With the changes in this PR, if the customer SSO feature flag is on, the `Users` page is not displayed, and the `/users/` route is not accessible.


marians (approved):
LGTM if it works for you

PR #2685, opened 2021-08-27
Add support for using configuration values from a production environment

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/18636

This PR adds support for using the production configuration while working on happa locally. This works by fetching the production `ConfigMap` from an installation, and using its values to template the configuration object (similar to what happens in production, but we're fetching the `ConfigMap` using `kubectl` instead of from the file system).

#### ⚠️ Note: This is a breaking change to how local development currently works.

Before this change, we were using `HAPPA_PROXY_INSTALLATION=<installation-codename> yarn start` to start local development, which depended on `opsctl`.

With the change, we no longer depend on `opsctl`, but we depend on `kubectl`, being logged into an installation's management cluster and having the required permissions to read happa's production `ConfigMap` (all of us at GS have this permissions).

The old way of starting local development was removed, and even though the new way is a bit more limited, it gives us an environment closer to production. For example, we now have all the instance types / vm sizes locally. This is also necessary for getting rid of the `/v4/info/` GS API endpoint, because in this way we can have those values in development, as well.


webpack.dev.ts
```diff
@@ -59,26 +61,41 @@ const config: webpack.Configuration = merge(common, {
...
+        mapiEndpoint: 'http://localhost:8888',
+        mapiAudience: 'http://localhost:9999',
+        passageEndpoint: 'http://localhost:5001',
+        mapiAuthRedirectURL: 'http://localhost:7000',
+        environment: 'development',
+        happaVersion: 'development',
+        enableRealUserMonitoring: false,
       },
-    ]),
+      proxies: [
```
axbarsan, line 74: Here we configure all the proxies necessary for bypassing CORS requirements on backend endpoints.

scripts/webpack/ConfigurationPlugin.ts
```diff
@@ -42,32 +68,85 @@ export class ConfigurationPlugin implements webpack.WebpackPluginInstance {
...
-          transformedContents,
-          false
-        );
-        compilation.emitAsset(this.options.outputFilename, modifiedSource);
-      }
+    const providedKubectlContext = process.env.hasOwnProperty(
+      'HAPPA_KUBECTL_CONTEXT'
     );
+
+    if (providedKubectlContext) {
```
axbarsan, line 113: If we provided a context, then we fetch the `ConfigMap` from the installation's management cluster. Otherwise, we don't 😬

scripts/webpack/ConfigurationPlugin.ts
```diff
@@ -42,32 +68,85 @@ export class ConfigurationPlugin implements webpack.WebpackPluginInstance {
...
+      logger.info(
+        `Fetching production configuration using kubectl context '${process.env.HAPPA_KUBECTL_CONTEXT}'`
+      );
+      config = await getProdConfiguration(process.env.HAPPA_KUBECTL_CONTEXT!);
+    }
+
+    logger.info('Computing configuration values');
+    this.configurationValues = await getConfigurationValues(config);
+
+    if (providedKubectlContext) {
```
axbarsan, line 123: We only instantiate proxies if a `kubectl` context is provided

scripts/getProdConfiguration.ts
```diff
@@ -0,0 +1,17 @@
+import { execSync } from 'child_process';
+import yaml from 'js-yaml';
+
+/**
+ * This function fetches the production `ConfigMap` using the
+ * given `kubectl` context.
+ * */
+export async function getProdConfiguration(context: string): Promise<string> {
```
axbarsan, line 8: This is what fetches the `ConfigMap` from the management cluster

kuosandys (approved):

PR #2697, opened 2021-08-30
Access control: order subjects by name and briefly highlight last added subject(s)

kuosandys:
Closes [giantswarm/giantswarm#16793](https://github.com/giantswarm/giantswarm/issues/16793).

This PR introduces ordering for access control subjects - they are sorted alphabetically by name rather than by order of entry. When new subject item(s) are added, they are also added to this sorted list, and a fading highlight effect is used to show users the last added subject item(s).

Subject items:
Screen Shot 2021-08-30 at 09 19 54

Last added item(s) (uses a fading highlight effect):
Screen Shot 2021-08-30 at 09 21 05


src/components/MAPI/organizations/AccessControl/AccessControlRoleSubjects.tsx
```diff
@@ -145,6 +145,11 @@ const mapValueToSetItem = (stateValue: IStateValue) => (
   };
 };
 
+const sortByName = (
```
axbarsan, line 148: Nitpicky: This doesn't really sort anything, right? It just compares 2 subjects

axbarsan (approved):
Works for me

src/components/MAPI/organizations/AccessControl/AccessControlRoleSubjects.tsx
kuosandys: Good point! I'll rename it to `compareSubjects` so it's more appropriate

PR #2699, opened 2021-08-30
Get rid of the need to use the info endpoint

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/18636

Depends on https://github.com/giantswarm/config/pull/412 (if you deploy this branch without using the `config` branch, you won't have any availability zones)

This PR:
- Adds (almost) all the info data that was previously in the `/v4/info/` GS API response -> except for the Kubernetes versions EOL info (tracked in https://github.com/giantswarm/roadmap/issues/336)
- Replaces all the usage of the global state for installation info -> the global config object is now used (with an exception for the Kubernetes versions EOL info)

Note: The `/v4/info/` request is still executed for non-SSO users, but only the K8s versions EOL info is used from its response. This is so we won't break the non-MAPI UI features that have to keep displaying EOL dates.


scripts/webpack/ConfigurationPlugin.ts
```diff
@@ -90,7 +90,7 @@ export class ConfigurationPlugin implements webpack.WebpackPluginInstance {
     // Template the file with the configuration values.
     const transformedContents = await templateIndex(
       contents,
-      this.configurationValues
+      this.configurationValues as IConfigurationValues
```
axbarsan, line 93: Didn't catch this last time

helm/happa/templates/configmap.yaml
```diff
@@ -43,3 +43,27 @@ data:
     {{- else }}
     feature-monitoring: true
     {{- end }}
+
+    info:
```
axbarsan, line 47: One of the important parts

scripts/getConfigurationValues.ts
```diff
@@ -27,6 +27,33 @@ export interface IConfigurationValues {
...
+        default: number;
+      };
+      instanceType: {
+        options: string;
+        default: string;
+      };
+      vmSize: {
+        options: string;
+        default: string;
+      };
```
axbarsan, line 54: These are the serialized versions of the values

src/components/MAPI/utils.ts
```diff
@@ -518,22 +517,17 @@ export function getProviderNodePoolLocation(
   }
 }
 
-// TODO(axbarsan): Get this info from the environment, rather than the info response.
```
axbarsan, line 521: I knew something back then

src/index.ejs
```diff
@@ -52,6 +52,45 @@
...
+          general: {
+            provider: '{!= info.general.provider !}',
+            installationName: '{!= info.general.installationName !}',
+            availabilityZones: {
+              default: {!= info.general.availabilityZones.default !},
+              max: {!= info.general.availabilityZones.zones.split(' ').filter(Boolean).length !},
+              zones: [
+                {! info.general.availabilityZones.zones.split(' ').filter(Boolean).map((zone) => { !}
+                  '{!= zone !}',
+                {! }) !}
```
axbarsan, line 65: Here's where the weird stuff starts. Deserializing values.

Not worried about performance since this code is not executed at run time.

kuosandys (approved):

PR #2704, opened 2021-08-31
Fix cluster and node pool description tooltips and error messages in MAPI UI

kuosandys:
The use of the 'name' label to refer to the descriptor for clusters and node pools is no longer accurate in the MAPI UI. This PR updates the tooltips and error messages for clusters and node pools to use "description" instead.

Cluster description:
Screen Shot 2021-08-31 at 16 44 42
Screen Shot 2021-08-31 at 16 45 03

Node pool description:
Screen Shot 2021-08-31 at 16 45 49
Screen Shot 2021-08-31 at 16 48 03


src/components/UI/Inputs/ViewEditName.tsx
```diff
@@ -67,11 +73,17 @@ class ViewAndEditName extends Component<
     return null;
   };
 
-  static validate(value: string): { valid: boolean; error: string } {
+  static validate(
+    value: string,
+    variant: string
```
axbarsan, line 78: Is this a `string` on purpose? Or should it be `ViewAndEditNameVariant`?

kuosandys: Nope that should be `ViewAndEditNameVariant`!

PR #2718, opened 2021-09-02
Fix organization usage when SSO is used but MAPI cluster management is disabled

axbarsan:
As seen on https://gigantic.slack.com/archives/C13MG9F0W/p1630578781011000

Turns out that while working on https://github.com/giantswarm/happa/pull/2683 I didn't take into account happa hybrid mode (SSO enabled, but MAPI cluster management disabled -> currently only used by GS staff, on AWS and KVM).

For migrated organizations, that used to have names that didn't adhere to the DNS standard, there were no clusters showing up. Also, cluster creation was not happening in the right organization. This PR fixes those issues.


calvix (approved):
thx

kuosandys (approved):

PR #2723, opened 2021-09-03
Apps: create widget for uninstalling an app

kuosandys:
Towards [giantswarm/giantswarm#18522](https://github.com/giantswarm/giantswarm/issues/18522).

This PR includes the widget for uninstalling an app from the Cluster Details > Apps tab

Step 1: Uninstall..
Screen Shot 2021-09-03 at 10 02 49

Step 2: Confirm Uninstall
Screen Shot 2021-09-03 at 10 59 54


src/components/MAPI/apps/ClusterDetailAppList.tsx
```diff
@@ -22,6 +22,8 @@ const ClusterDetailAppList: React.FC<IClusterDetailAppListProps> = ({
 }) => {
   const [activeIndex, setActiveIndex] = useState(-1);
 
+  const resetActiveIndex = () => setActiveIndex(-1);
```
kuosandys, line 25: This resets the index so the next app that takes the index of the uninstalled app isn't unexpectedly set to be the active (expanded) app list item

src/components/UI/Controls/ConfirmationPrompt/index.tsx
```diff
@@ -10,6 +10,7 @@ interface IConfirmationPromptProps
   cancelButton?: React.ReactNode;
   onConfirm?: () => void;
   onCancel?: () => void;
+  contentProps?: React.ComponentPropsWithoutRef<typeof Box>;
```
kuosandys, line 13: Added a new prop to pass in contentProps to style the inner Box (i.e. the padding and background colour)

src/components/MAPI/apps/ClusterDetailAppListWidgetUninstall.tsx
```diff
@@ -0,0 +1,144 @@
...
+      await deleteAppWithName(
+        appClient,
+        auth,
+        app.metadata.namespace!,
+        app.metadata.name
+      );
+
+      mutate(
+        applicationv1alpha1.getAppListKey({ namespace: app.metadata.namespace })
+      );
```
axbarsan, line 50: One problem with this: after deleting an app, the deleted app won't be updated until fetching the app list again. Which means that on slower internet connections, the user will still see the unchanged app in the list after deleting it.

We can update the cache manually, to get instant feedback in the UI.

Here's an example of how we do this in other places:
https://github.com/giantswarm/happa/blob/5912045903bbf35f01e33633cf6c79b10136fccb/src/components/MAPI/clusters/ClusterDetail/utils.ts#L51

In this example we mutate the cluster resource manually, adding its `deletionTimestamp` property value (which makes it be displayed as deleted), and then we update the local cache with the updated value.

src/components/MAPI/apps/ClusterDetailAppListWidgetUninstall.tsx
```diff
@@ -0,0 +1,144 @@
...
+        <Box direction='row' justify='between' align='center'>
+          <LoadingPlaceholder width={500} height={20} />
+          <LoadingPlaceholder width={100} height={36} />
+        </Box>
+      ) : (
+        <>
+          <ConfirmationPrompt
+            title={
+              <Text weight='bold' margin={{ bottom: 'small' }}>
+                Are you sure you want to uninstall {app.metadata.name} from
```
axbarsan, line 97: What do you think about changing it to this?

```suggestion
                Are you sure you want to uninstall <code>{app.metadata.name}</code> from
```

src/components/MAPI/apps/ClusterDetailAppListWidgetUninstall.tsx
```diff
@@ -0,0 +1,144 @@
...
+              justify='between'
+              align='center'
+              animation={{ type: 'fadeIn', duration: 300 }}
+            >
+              <Text>
+                Uninstalling the app from this cluster will not remove the
+                configuration resources.
+              </Text>
+              <Button secondary onClick={handleUninstall}>
+                Uninstall...
```
axbarsan, line 134: This is the first thing that Marian will see 😆 We use the UTF-8 ellipsis character in other places.
```suggestion
                Uninstall…
```

src/components/MAPI/apps/__tests__/ClusterDetailAppListWidgetUninstall.tsx
```diff
@@ -0,0 +1,183 @@
...
+  });
+
+  it('displays the confirmation for uninstalling an app when the uninstall button is clicked', () => {
+    const app = generateApp();
+
+    render(getComponent({ app }));
+
+    const uninstallButton = screen.getByRole('button', {
+      name: 'Uninstall...',
+    }) as HTMLButtonElement;
```
axbarsan, line 129: Do we need this assertion?

```suggestion
    });
```

src/components/UI/Controls/ConfirmationPrompt/index.tsx
```diff
@@ -126,6 +122,7 @@ const ConfirmationPrompt = React.forwardRef<
 ConfirmationPrompt.defaultProps = {
   confirmButton: 'Confirm',
   cancelButton: 'Cancel',
+  contentProps: { pad: 'medium', background: 'background-front' },
```
axbarsan, line 125: Very minor thing: When this component is used, and we want to override the background color, we have to also set `pad` again. Also, if we wanted to add a new property, we would have to set `pad` and `background` again.

In other places we merge default props with the overrides at the component level
https://github.com/giantswarm/happa/blob/5912045903bbf35f01e33633cf6c79b10136fccb/src/components/UI/Inputs/RadioInput/index.tsx#L96

This makes sure that if we ever want to update the defaults, or add another default, we don't have to edit all files that use the component.

axbarsan (changes requested):

axbarsan (commented):
Works great and looks good overall, just have a few comments

axbarsan (approved):
Amazing 👍

PR #2728, opened 2021-09-06
apps: display deleted cluster apps in a different way

axbarsan:
Towards giantswarm/giantswarm#18522

This PR adds a different display style for apps that are currently deleted (have the `deletionTimestamp` set). It also introduces sorting to the list of cluster apps. The list is now sorted by the app name in an ascending order, leaving deleted apps at the end of the list (same as clusters on the home page).

## Preview

<details>
<summary>Details</summary>

image

</details>

### DIFF
--- a/src/components/MAPI/apps/ClusterDetailAppListItem.tsx
+++ b/src/components/MAPI/apps/ClusterDetailAppListItem.tsx
@@ -1,6 +1,7 @@
 import { AccordionPanel, Box, Text } from 'grommet';
+import { relativeDate } from 'lib/helpers';
 import * as applicationv1alpha1 from 'model/services/mapi/applicationv1alpha1';
-import React from 'react';
+import React, { useLayoutEffect, useRef } from 'react';
 import styled from 'styled-components';
 import OptionalValue from 'UI/Display/OptionalValue/OptionalValue';
 
@@ -22,6 +23,12 @@ const StyledBox = styled(Box)`
   gap: ${({ theme }) => theme.global.edgeSize.small};
 `;
 
+const Header = styled(Box)`
+  &[aria-disabled] {
+    cursor: default;
+  }
+`;
+
 interface IClusterDetailAppListItemProps
   extends React.ComponentPropsWithoutRef<typeof Box> {
   app?: applicationv1alpha1.IApp;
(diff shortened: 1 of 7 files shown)


src/components/MAPI/apps/ClusterDetailAppListItem.tsx
```diff
@@ -37,15 +44,40 @@ const ClusterDetailAppListItem: React.FC<IClusterDetailAppListItemProps> = ({
...
+  const accordionRef = useRef<HTMLDivElement>(null);
+
+  useLayoutEffect(() => {
+    if (!accordionRef) return;
+
+    const accordionButton = accordionRef.current?.querySelector('button');
+    if (!accordionButton) return;
+
+    accordionButton.disabled = isDisabled;
+  }, [isDisabled]);
```
axbarsan, line 59: A bit weird, but `grommet` doesn't provide a way to disable the accordion button

src/components/MAPI/apps/ClusterDetailAppListItem.tsx
```diff
@@ -37,15 +44,40 @@ const ClusterDetailAppListItem: React.FC<IClusterDetailAppListItemProps> = ({
...
+
+  const handleHeaderClick = (e: React.MouseEvent<HTMLElement>) => {
+    if (isDisabled) {
```
axbarsan, line 62: Blocking click events if the app is deleting, so the user can not open the accordion

kuosandys (commented):
This looks great! Though the deleted app only shows up at the bottom of the list for a brief second for me, before disappearing altogether. Would it be possible to prolong that?

axbarsan:
> This looks great! Though the deleted app only shows up at the bottom of the list for a brief second for me, before disappearing altogether. Would it be possible to prolong that?

What do you mean? Making the the app hang around in the UI for a longer time than it does in the K8s management cluster?

Right now `deleted xxx time ago` state will only really be visible if deleting an app takes longer than expected (i.e. `app-operator` has some trouble deleting an app's configmap). In other cases, the app would just be removed from the UI right away.

kuosandys:
> > This looks great! Though the deleted app only shows up at the bottom of the list for a brief second for me, before disappearing altogether. Would it be possible to prolong that?
> 
> What do you mean? Making the the app hang around in the UI for a longer time than it does in the K8s management cluster?
> 
> Right now `deleted xxx time ago` state will only really be visible if deleting an app takes longer than expected (i.e. `app-operator` has some trouble deleting an app's configmap). In other cases, the app would just be removed from the UI right away.

Ah ok I misunderstood the purpose for the change here - thanks for clarifying!

kuosandys (approved):

PR #2729, opened 2021-09-06
Compute namespace if the Organization CR does not have the required status field

axbarsan:
As seen on https://gigantic.slack.com/archives/C02MS174K/p1630671483018400

The problem was that we weren't taking into account that `Organization` CRs might not have the `status.namespace` field populated. Which meant that we were creating `SelfSubjectRulesReviews` using empty namespaces (which resulted in errors).

This PR addresses that, by computing the organization name manually if the CR does not provide it. Also, now if one of the `SelfSubjectRulesReview` request fails, we still compute permissions for other namespaces. One failure does no longer mean that we block access to the whole app.

### DIFF
--- a/src/stores/main/actions.ts
+++ b/src/stores/main/actions.ts
@@ -53,7 +53,7 @@ import {
 } from 'utils/localStorageUtils';
 
 import { LoggedInUserTypes } from './types';
-import { computePermissions } from './utils';
+import { computePermissions, getNamespaceFromOrgName } from './utils';
 
 export function selectCluster(clusterID: string): MainActions {
   return {
@@ -417,7 +417,9 @@ export function fetchPermissions(
   return async (dispatch, getState) => {
     const orgs = Object.values(selectOrganizations()(getState()));
 
-    const namespaces = orgs.map((o) => o.namespace ?? '');
+    const namespaces = orgs.map(
+      (o) => o.namespace ?? getNamespaceFromOrgName(o.id)
+    );
     // Also get permissions for the default namespace.
     namespaces.push('default');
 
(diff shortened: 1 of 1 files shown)


src/stores/main/actions.ts
```diff
@@ -441,7 +443,17 @@ export function fetchPermissions(
...
     });
 
-    const reviews = await Promise.all(requests);
+    const reviewRequests = await Promise.allSettled(requests);
+    const reviews: [
+      namespace: string,
+      review: authorizationv1.ISelfSubjectRulesReview
+    ][] = [];
+    for (const reviewRequest of reviewRequests) {
+      if (reviewRequest.status === 'fulfilled') {
```
axbarsan, line 452: We compute permissions for the requests that didn't fail

src/stores/main/actions.ts
```diff
@@ -417,7 +417,9 @@ export function fetchPermissions(
   return async (dispatch, getState) => {
     const orgs = Object.values(selectOrganizations()(getState()));
 
-    const namespaces = orgs.map((o) => o.namespace ?? '');
+    const namespaces = orgs.map(
+      (o) => o.namespace ?? getNamespaceFromOrgName(o.id)
```
kopiczko, line 421: If this is empty, doesn't it mean it isn't reconciled and the ns may be not even created?

axbarsan: Sure, that might be the case, and the request would fail. But with the new logic, it doesn't break anything

tfussell (approved):

kopiczko (approved):

PR #2734, opened 2021-09-07
Get k8s version info from giantswarm/config

axbarsan:
Towards https://github.com/giantswarm/roadmap/issues/336

This PR makes use of the kubernetes version information coming in from `giantswarm/config`. It also replaces the lookup used by GS API features to display k8s version info to using the global config. The rest of the leftover info endpoint stuff will be removed in a separate PR.


src/stores/main/selectors.ts
```diff
@@ -54,26 +54,6 @@ export function getAllowedInstanceTypeNames(): string[] {
   }
 }
 
-export function getK8sVersionEOLDate(version: string) {
```
axbarsan, line 57: Moved this out of here as it's no longer a `selector`

src/index.ejs
```diff
@@ -66,7 +66,14 @@
...
             },
             dataCenter: '{!= info.general.dataCenter !}',
-            kubernetesVersions: [],
+            kubernetesVersions: {
+              {! Object.entries(JSON.parse(info.general.kubernetesVersions)).map(([version, versionInfo]) => { !}
+                  '{!= version !}': {
+                    minorVersion: '{!= versionInfo.minorVersion !}',
+                    eolDate: '{!= versionInfo.eolDate !}',
+                  },
+              {! }) !}
```
kopiczko, line 75: `{! }) !}` is my favourite line of code

src/lib/config.ts
```diff
@@ -0,0 +1,15 @@
...
+  if (!version) return null;
+
+  const k8sVersions = window.config.info.general.kubernetesVersions;
+  if (!k8sVersions) return null;
+
+  const versionParts = version.split('.');
+  if (versionParts.length < 2) return null;
+  const minor = `${versionParts[0]}.${versionParts[1]}`;
+
+  const versionInfo = k8sVersions[minor];
```
kopiczko, line 11: Will it work when it's constructed like below?

```
                  '{!= version !}': {
                    minorVersion: '{!= versionInfo.minorVersion !}',
                    eolDate: '{!= versionInfo.eolDate !}',
                  },             
```

Shouldn't it be `k8sVersions[version]`?

axbarsan: `version` is actually the minor version here
https://github.com/giantswarm/config/blob/main/include/k8s-versions.yaml.template

But you're right, there's nothing enforcing that, I should change the template to always have the minor version as the key

src/index.ejs
```diff
@@ -66,7 +66,14 @@
               ]
             },
             dataCenter: '{!= info.general.dataCenter !}',
-            kubernetesVersions: [],
+            kubernetesVersions: {
+              {! Object.values(JSON.parse(info.general.kubernetesVersions)).map((versionInfo) => { !}
+                  '{!= versionInfo.minorVersion !}': {
+                    minorVersion: '{!= versionInfo.minorVersion !}',
```
kopiczko, line 72: Isn't this field write-only now?

axbarsan: What do you mean? `versionInfo` is a value of the map created by `JSON.parse(info.general.kubernetesVersions)`

kopiczko: I mean `minorVersion` field.

axbarsan: This file is a template, and it uses `ejs` as a templating language. Here is where the `kubernetesVersions` map is generated, and for every value of the map parsed from `JSON.parse(info.general.kubernetesVersions)`, we would render a new entry into our `kubernetesVersions` map.

In the output code, we would have a similar object:
```typescript
...

kubernetesVersions: {
	"1.15": {
		minorVersion: "1.15",
		eolDate: "2020-03-12,
	},
	"1.16": {
		minorVersion: "1.16",
		eolDate: "2020-07-12,
	},
	"1.17": {
		minorVersion: "1.17",
		eolDate: "2020-11-12,
	},
},

...
```

So setting `minorVersion` here is just the same as setting the value of any random key in a JS object. Nothing is read/write-only

kuosandys (approved):

PR #2735, opened 2021-09-07
Block possibility to edit the `cluster.x-k8s.io/watch-filter` label.

whites11:
We don't want the user to edit the `cluster.x-k8s.io/watch-filter` as it will break functionality of CAPI clusters.


axbarsan (approved):
Nice

PR #2736, opened 2021-09-07
apps: use normalized app versions for comparisons

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/18827

The problem: Some versions have a `v` prefix, while others don't.
Solution: Always remove the `v` prefix before doing semver comparisons ✅

### DIFF
--- a/src/components/MAPI/apps/utils.ts
+++ b/src/components/MAPI/apps/utils.ts
@@ -734,6 +734,14 @@ export function isAppChangingVersion(app: applicationv1alpha1.IApp): boolean {
   );
 }
 
+export function normalizeAppVersion(version: string): string {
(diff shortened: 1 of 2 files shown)


src/components/MAPI/apps/utils.ts
```diff
@@ -734,6 +734,14 @@ export function isAppChangingVersion(app: applicationv1alpha1.IApp): boolean {
   );
 }
 
+export function normalizeAppVersion(version: string): string {
+  if (version.toLowerCase().startsWith('v')) {
+    return version.substring(1);
```
axbarsan, line 739: Slightly faster than doing `.slice(1)`

kopiczko (approved):

PR #2752, opened 2021-09-10
apps: display error details when YAML validation fails

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/17553

This PR exposes the actual reasons for why YAML validation failed when uploading app values/secrets. This is handled on app installation and app updating. It is also implemented for the GS API UI, not because I wanted to, but because the `YAMLFileUpload` component is shared between the GS API and MAPI implementations, so it was actually easier to have everything consistent.

#### Before

image

#### After

image

### DIFF
--- a/src/components/MAPI/apps/utils.ts
+++ b/src/components/MAPI/apps/utils.ts
@@ -1,4 +1,5 @@
 import produce from 'immer';
+import { YAMLException } from 'js-yaml';
 import ErrorReporter from 'lib/errors/ErrorReporter';
 import { HttpClientFactory } from 'lib/hooks/useHttpClientFactory';
 import { IOAuth2Provider } from 'lib/OAuth2/OAuth2';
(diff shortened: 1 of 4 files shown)


src/components/MAPI/apps/utils.ts
```diff
@@ -803,3 +804,28 @@ export function compareApps(
 
   return a.metadata.name.localeCompare(b.metadata.name);
 }
+
+export function formatYAMLError(err: unknown): string {
+  if (err instanceof YAMLException) {
+    interface IYAMLExceptionInternals extends YAMLException {
```
axbarsan, line 810: The `YAMLException` type from the library doesn't actually expose these fields, but we know they're there 🤓

kuosandys (approved):

marians (approved):
Nice!

PR #2755, opened 2021-09-13
Improve Apps browsing performance: virtualize the Apps list

kuosandys:
Towards [giantswarm/giantswarm#17429](https://github.com/giantswarm/giantswarm/issues/17429).

This PR introduces a component for displaying Apps in the Apps page with a virtualized list. This reduces the number of elements on the page at a given time, thus reducing the number of elements that need to be re-rendered with every interaction.

The appearance and behaviour of the App list should remain the same as before. If you scroll very very fast, you might notice some dropped frames. This can be improved by increasing the number of rows that are rendered outside of the current viewport, but this comes at the cost of increasing the number of elements on the page.


src/components/UI/Display/Apps/AppList/AppGrid.tsx
```diff
@@ -0,0 +1,138 @@
...
+const StyledContainer = styled.div<{ gridGap: number }>`
+  margin: 0 -${(props) => props.gridGap / 2}px;
+`;
+
+const StyledGrid = styled(VariableSizeGrid)<{ minHeight: number }>`
+  height: 100% !important;
+
+  > div {
+    height: ${(props) => props.minHeight}px !important;
+  }
```
kuosandys, line 39: The height for the outer `div` element has to be overwritten to allow it to be scrolled by scrolling the page (otherwise it would be a scrollable window within the page).

The height for the inner `div` element of the virtualized list increases as the user scrolls. This presents a problem when navigating back to this page because the scroll position has a larger number than the initial height of the element, and it will not restore the correct scroll position. Setting the height to be the total height of the elements ensures that the page will have a greater height than the scroll position.

src/components/UI/Display/Apps/AppList/AppGrid.tsx
```diff
@@ -0,0 +1,138 @@
...
+import { WindowScroller } from 'react-virtualized';
+import {
+  areEqual,
+  GridChildComponentProps,
+  VariableSizeGrid,
+} from 'react-window';
+import styled from 'styled-components';
+import { IAppProps } from 'UI/Display/Apps/AppList/App';
+
+interface IAppGrid {
```
axbarsan, line 11:
```suggestion
interface IAppGridProps {
```

src/components/UI/Display/Apps/AppList/AppGrid.tsx
```diff
@@ -0,0 +1,138 @@
...
+    const patchedStyle = Object.assign({}, style, {
+      padding: `0 ${gridGap / 2}px`,
+    });
+
+    return item ? <div style={patchedStyle}>{render(item, index)}</div> : null;
+  },
+  areEqual
+);
+
+const AppGrid: React.FC<IAppGrid> = ({
```
axbarsan, line 65:
```suggestion
const AppGrid: React.FC<IAppGridProps> = ({
```

src/components/UI/Display/Apps/AppList/AppGrid.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+  const gridRef = useRef<VariableSizeGrid>(null);
+  const containerRef = useRef<HTMLDivElement>(null);
+
+  const handleScroll = ({ scrollTop }: IScrollPosition) => {
+    if (gridRef.current) gridRef.current.scrollTo({ scrollTop });
+  };
+
+  useLayoutEffect(() => {
+    const handleResize = () => {
```
axbarsan, line 82: This is not entirely necessary right now, since performance is great, but for future improvement:
This function could be debounced, so we won't be calling `getBoundingClientRect()` so often, since that triggers reflows

src/components/UI/Display/Apps/AppList/AppsListPage.tsx
```diff
@@ -105,18 +110,24 @@ const AppsList: React.FC<IAppsListPageProps> = (props) => {
...
+              itemMinWidth={270}
+              gridGap={20}
+              render={(app: IAppProps, i: number) => (
+                <App
+                  name={app.name}
+                  appIconURL={app.appIconURL}
+                  catalogTitle={app.catalogTitle}
+                  catalogIconUrl={app.catalogIconUrl}
+                  to={app.to}
+                  key={app.name + i.toString()}
```
axbarsan, line 126: This could actually be the `.to` property, since it is always unique, so we won't need to use the index
```suggestion
                  key={app.to}
```

axbarsan (approved):
This is a huge improvement! I only have a few small comments, but otherwise this is great 👍

PR #2756, opened 2021-09-13
Use the new 'template organization' command in kubectl-gs in the CLI guide for creating an organization

marians:
Depends on https://github.com/giantswarm/kubectl-gs/pull/470

Before

image

After

image


axbarsan (approved):
:chefs_kiss:

PR #2759, opened 2021-09-13
Refactor all modals to use the <GenericModal /> component

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/18913

Making all modals use the same main component, so later, we only have to change the `<GenericModal />` internals to migrate away from `bootstrap`


kuosandys (approved):

marians (approved):
Sounds good!

PR #2762, opened 2021-09-14
Improve Apps browsing performance: preload App icons

kuosandys:
Towards [giantswarm/giantswarm#17429](https://github.com/giantswarm/giantswarm/issues/17429).

This PR changes how App icons are rendered by preloading them before appending them to the page. This prevents the App icons from re-rendering on the page on load/error.

This PR also fixes a bug for the `AppGrid` component when displaying non-full rows.


src/components/UI/Display/Apps/AppList/AppGrid.tsx
```diff
@@ -98,7 +98,7 @@ const AppGrid: React.FC<IAppGridProps> = ({
   }, []);
 
   const columnCount = Math.floor(width / (itemMinWidth + gridGap)) || 1;
-  const rowCount = items.length / columnCount;
+  const rowCount = Math.ceil(items.length / columnCount);
```
kuosandys, line 101: This should be rounded up so the last row is displayed even if it is not filled. I tried to fix this yesterday, but made the wrong decision

axbarsan (approved):
Really cool 👍

PR #2773, opened 2021-09-17
Align flash messages for updating installed apps

kuosandys:
Closes [giantswarm/giantswarm#18967](https://github.com/giantswarm/giantswarm/issues/18967).

The messages for successfully updating an installed app are aligned with other information shown in the UI. The length of time for which the flash messages are displayed is also increased, and they are visible for the same amount of time as an unsuccessful update message.

The messages can still be pinned by pointing the cursor on the message (no changes needed here).

Successful upgrade:
Screen Shot 2021-09-17 at 12 08 28

Successful downgrade:
Screen Shot 2021-09-17 at 12 08 44

### DIFF
--- a/src/components/MAPI/apps/ClusterDetailAppListWidgetVersionInspector.tsx
+++ b/src/components/MAPI/apps/ClusterDetailAppListWidgetVersionInspector.tsx
@@ -218,7 +218,7 @@ const ClusterDetailAppListWidgetVersionInspector: React.FC<IClusterDetailAppList
         new FlashMessage(
           `<code>${app.metadata.name}</code> on cluster <code>${app.metadata.namespace}</code> will be ${updateAction} to version <code>${truncatedVersion}</code>.`,
           messageType.SUCCESS,
-          messageTTL.SHORT
+          messageTTL.LONG
         );
       } catch (err) {
         setAppUpdateIsLoading(false);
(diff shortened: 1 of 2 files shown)


src/components/MAPI/apps/ClusterDetailAppListWidgetVersionInspector.tsx
```diff
@@ -202,10 +206,19 @@ const ClusterDetailAppListWidgetVersionInspector: React.FC<IClusterDetailAppList
         setAppUpdateIsLoading(false);
         handleCancelSwitchingVersion();
 
+        const truncatedVersion = truncate(
```
axbarsan, line 209: Why not use the `<Truncated />` component? You can specify a custom wrapper element for it `<Truncated as='code' />`

There's an example lower down in the file

kuosandys: `FlashMessage` is expecting a string for the text input, so I thought it was more straight-forward to use a string. I'll change it to use the `Truncate` component 👍

axbarsan: Actually, you're right... we don't allow React components in there, only `<code />` elements. My bad, sorry 👍

axbarsan (approved):

PR #2783, opened 2021-09-27
Use `athena` for submitting RUM events

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/18033
Needs https://github.com/giantswarm/config/pull/479 and https://github.com/giantswarm/athena/pull/43 for testing.

This PR changes the RUM event reporting logic, so it would use `athena` instead of `api`. It contains a GraphQL client, logic for interacting with `athena` and a bit of refactoring of the way we triggered RUM events.

Functionality should be the same as before.


src/@types/graphql.d.ts
```diff
@@ -0,0 +1,3 @@
+declare module '*.graphql' {
```
axbarsan, line 1: So TS knows that we get strings from `.graphql` files

src/@types/realUserMonitoring.d.ts
```diff
@@ -1,11 +0,0 @@
-interface IRum {
```
axbarsan, line 1: These are leftovers from when we used DataDog

src/model/clients/GraphQLClient.ts
```diff
@@ -0,0 +1,76 @@
...
+}
+
+export interface IGraphQLClient {
+  executeQuery<T = unknown>(
+    query: string,
+    variables?: GraphQLVariables
+  ): Promise<T>;
+}
+
+export class GraphQLClientImpl implements IGraphQLClient {
```
axbarsan, line 41: Didn't want to add another big-ass dependency for this

webpack.common.ts
```diff
@@ -96,6 +96,11 @@ const config: webpack.Configuration = {
           mimetype: 'image/svg+xml',
         },
       },
+      {
+        test: /\.(graphql|gql)$/,
```
axbarsan, line 100: This is so when we import from `.graphql` files, they actually get embedded in the app as strings

src/model/services/athena/mutations/CreateAnalyticsEvent.graphql
```diff
@@ -0,0 +1,10 @@
+mutation CreateAnalyticsEvent($input: AnalyticsEventInput!) {
```
axbarsan, line 1: The GraphQL mutation for submitting an event

marians (approved):

kuosandys (approved):

PR #2786, opened 2021-09-28
Fix semver comparison

kuosandys:
When comparing a version without labels (e.g. `12.0.0`) to a version of the same number with labels (e.g. `12.0.0-something`), the `compare` function is returning "0", indicating that the versions are equal. This creates a bug where it is not possible to downgrade an app's version:

Screen Shot 2021-09-28 at 09 30 22

This is fixed by changing the order of operands in the logical operation so that the last value returned is a boolean.

### DIFF
--- a/src/lib/__tests__/semver.ts
+++ b/src/lib/__tests__/semver.ts
@@ -54,5 +54,21 @@ describe('semver', () => {
         '47.48.11.2',
       ]);
     });
+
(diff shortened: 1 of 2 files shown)


axbarsan (approved):
Really nice find 👍

Can we also have a test case for it?

axbarsan (approved):

PR #2787, opened 2021-09-28
Add alpha support for node pools on CAPZ clusters

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/18770

Starting with the Giant Swarm Azure alpha release `v20.0.0`, we support full blown CAPZ clusters and node pools, and since these use `v1alpha4` `MachinePool` and `AzureMachinePool` CRs, which are now part of a different API group, we need to make some changes to support them.

This PR adds the new `v1alpha4` types and utility functions, and adapts the logic so the new types would also work.

> ⚠️ Note: CAPZ node pools are in read only mode for the alpha release: you can't create/update/delete node pools from the UI, and the CLI guides for those operations are also hidden.

### DIFF
--- a/src/components/MAPI/workernodes/ClusterDetailWorkerNodes.tsx
+++ b/src/components/MAPI/workernodes/ClusterDetailWorkerNodes.tsx
@@ -12,6 +12,7 @@ import {
   fetchProviderClusterForClusterKey,
   fetchProviderNodePoolsForNodePools,
   fetchProviderNodePoolsForNodePoolsKey,
+  isNodePoolMngmtReadOnly,
 } from 'MAPI/utils';
 import { GenericResponseError } from 'model/clients/GenericResponseError';
 import * as capiv1alpha3 from 'model/services/mapi/capiv1alpha3';
@@ -305,6 +306,8 @@ const ClusterDetailWorkerNodes: React.FC<IClusterDetailWorkerNodesProps> =
       setIsCreateFormOpen(false);
     };
 
+    const isReadOnly = cluster && isNodePoolMngmtReadOnly(cluster);
+
     return (
       <DocumentTitle title={`Worker Nodes | ${clusterId}`}>
         <Breadcrumb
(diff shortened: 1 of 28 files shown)


src/shared/constants/index.ts
```diff
@@ -38,6 +38,8 @@ export const Constants = {
   AWS_HA_MASTERS_VERSION: '11.4.0',
   AWS_HA_MASTERS_MAX_NODES: 3,
 
+  AZURE_CAPZ_VERSION: '20.0.0',
```
axbarsan, line 41: We only fetch the new `v1alpha4` CRs if the cluster release version is `>=20.0.0`

src/model/services/mapi/capzv1alpha4/key.ts
```diff
@@ -0,0 +1,21 @@
+import { IAzureMachinePool } from './';
+
+export const tagAutoscalerMinSize = 'min';
+export const tagAutoscalerMaxSize = 'max';
+
+export function getAzureMachinePoolScaling(
```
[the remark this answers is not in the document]
whites11: That's how cluster autoscaler for azure works

kuosandys (commented):
Really cool 👍 Just a question - I noticed that the `Add node pool` button is only disabled for the display when there are node pools. Should it also be disabled in the `WorkerNodesNodePoolListPlaceholder` component?

src/components/MAPI/workernodes/ClusterDetailWorkerNodes.tsx
```diff
@@ -471,12 +476,17 @@ const ClusterDetailWorkerNodes: React.FC<IClusterDetailWorkerNodesProps> =
...
                 />
-                <ModifyNodePoolGuide
-                  clusterNamespace={cluster.metadata.namespace!}
-                />
-                <DeleteNodePoolGuide
-                  clusterNamespace={cluster.metadata.namespace!}
-                />
+
+                {!isReadOnly && (
+                  <>
```
kuosandys, line 481: I think the spacing between the CLI guides changed a bit here
Screen Shot 2021-09-28 at 20 28 15

axbarsan: Lol 😬 thank you

axbarsan:
> I noticed that the `Add node pool` button is only disabled for the display when there are node pools. Should it also be disabled in the `WorkerNodesNodePoolListPlaceholder` component?

Oh shit, nice find 🤦

kuosandys (approved):

PR #2790, opened 2021-09-28
Show the number of apps to be upgraded in cluster overview

kuosandys:
Closes [giantswarm/giantswarm#18960](https://github.com/giantswarm/giantswarm/issues/18960).

This PR adds an additional summary statistic to the cluster details `Overview` tab > `Apps` widget, and provides information on the number of apps that can be upgraded (i.e. has a newer version available).

Screen Shot 2021-09-28 at 23 45 42

The `UPGRADABLE` label is used instead of `AWAITING UPGRADE` because the latter would span two lines and would be visually out of alignment with the other summary statistics.


marians:
Nice word, `UPGRADABLE` :) Thanks for picking up the issue!

If there are no upgradable apps, what do we show?

src/components/MAPI/apps/ClusterDetailWidgetApps.tsx
```diff
@@ -74,6 +76,30 @@ const ClusterDetailWidgetApps: React.FC<IClusterDetailWidgetAppsProps> = (
...
 
+  const upgradableAppsKey = appList
+    ? getUpgradableAppsKey(appList.items)
+    : null;
+
+  const { data: upgradableApps, error: upgradableAppsError } = useSWR<
+    string[],
+    GenericResponseError
+  >(upgradableAppsKey, () =>
+    getUpgradableApps(appList!.items, clientFactory, auth)
```
axbarsan, line 87: This works well, but I have a concern, because we're fetching so many `AppCatalogEntry` CRs, that we're also fetching in other parts of the app, and we're not making use of the caching in `swr`. Is there any way that we can leverage the cache?

kuosandys:
> Nice word, `UPGRADABLE` :) Thanks for picking up the issue!
> 
> If there are no upgradable apps, what do we show?

Currently it would show `0 UPGRADABLE`

marians (approved):

axbarsan:
Also, I believe we should display how many user-installed apps are upgradable, so we would have consistency with the other app counters. Currently we display all apps that can be upgraded (including the release pre-installed ones).

For example here I installed 3 apps, and only one of them is upgradable, but the UI says that there are 4, because it also takes into consideration the pre-installed ones.

image

axbarsan (approved):
Really nice 😶‍🌫️

PR #2810, opened 2021-10-06
Fix webpack development config

axbarsan:
Development mode no longer works. This got broken with https://github.com/giantswarm/happa/pull/2807.


marians:
Thanks for cleaning up! Obviously I didn't test enough.

PR #2812, opened 2021-10-06
Refactor tooltips

kuosandys:
Closes [giantswarm/giantswarm#19147](https://github.com/giantswarm/giantswarm/issues/19147).

This PR replaces the `bootstrap` components used for tooltips with `grommet` components.

### Why not `<Tip />`
`grommet` has a `<Tip />` component that could be used for this purpose, but it does not provide a responsive tooltip caret/arrow out-of-the-box, and sometimes does not render the tooltip contents if the mouse event begins from a child element of the target element (it listens to the `mouse enter` event rather than the `mouse-over` event as the bootstrap implementation does).

### `<Tooltip />` and `<TooltipContainer />`
The tooltip functionality is comprised of 2 components:
- `<Tooltip />` is the tooltip element itself, and is styled to look similar to the existing tooltips, the main difference being the font (it now uses Roboto - the same font as the rest of the UI).
- `<TooltipContainer />` provides the tooltip functionality, namely with displaying the tooltip when the target element is hovered over.


src/components/UI/Display/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,179 @@
...
+      >
+        {children}
+      </Text>
+    ) : (
+      children
+    );
+
+  return (
+    <>
+      <StyledDrop
```
kuosandys, line 126: Using 2 separate `<Drop>` components to allow the caret to align itself to the centre of the target element, and for the tooltip contents to also do so if possible, but to not overflow off the page if it is wider than the target element

src/components/UI/Display/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,179 @@
...
+      >
+        <TooltipCaret
+          role='presentation'
+          aria-hidden='true'
+          width='10px'
+          height='10px'
+          margin={{ [marginPosition]: 'small' }}
+          placement={placement as Position}
+          background='transparent'
+          tooltipColor={background as string}
```
kuosandys, line 141: These values are provided by `defaultProps`

src/components/UI/Display/Tooltip/TooltipContainer.tsx
```diff
@@ -0,0 +1,40 @@
...
+  target?: React.RefObject<HTMLElement>;
+}
+
+const TooltipContainer: React.FC<
+  React.PropsWithChildren<ITooltipContainerProps>
+> = ({ content, target, children }) => {
+  const [tooltipVisible, setTooltipVisible] = useState(false);
+  const createdTargetRef = useRef<HTMLElement>(null);
+
+  const tooltipTargetRef = target ?? createdTargetRef;
```
kuosandys, line 14: If a target element is specified, the tooltip will align to that. Otherwise, it will align to the `children` elements

kuosandys:
> In the storybook named "Tooltip" I can't see any tooltip showing up

I realized splitting the element and the functionality into 2 separates stories was probably not the best idea. It is just one now, under Tooltip.

PR #2837, opened 2021-10-14
Switch back to the babel compiler

axbarsan:
Although `swc` is a great project, it is not mature enough yet. Although it is much faster than `babel` and it produces much smaller bundles (due to better dead code elimination), new releases often introduce breaking changes (we haven't been able to update for a couple months, already), and styled components support is pretty bad (remember how we have to update all snapshots in almost every PR?).

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -3,7 +3,8 @@ import { MemoryRouter } from 'react-router';
 import theme from 'styles/theme';
 import ThemeProvider from 'styles/ThemeProvider';
 
-import 'babel-polyfill';
+import 'core-js/stable';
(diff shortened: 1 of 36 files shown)


axbarsan:
> why were the imported mocks renamed?

Funny enough, jest doesn't allow referencing out-of-scope variables in `mock()` factories. `swc` was ignoring this, but `babel` follows the spec.

The explanation and solution, quoted from `jest`:
>  Note: This is a precaution to guard against uninitialized mock variables. If it is ensured that the mock is required lazily, variable names prefixed with `mock` (case insensitive) are permitted.

So out-of-scope variables need to have the `mock` prefix in order to work 😅

PR #2844, opened 2021-10-18
Use LoadBalancer ports from cluster object

ubergesundheit:
This PR changes the LoadBalancer ports displayed on KVM clusters.

I noticed the ports displayed on the cluster overview differed from the ports displayed in the Ingress tab.


glitchcrab:
lgtm

PR #2849, opened 2021-10-20
clusters: add support for displaying details about AWS control plane nodes via MAPI

axbarsan:
Towards giantswarm/roadmap#388

This PR enables all the stats about control plane nodes (including if they are high-availability). It also renames `master node` to `control plane node` everywhere.


axbarsan:
⏩

PR #2885, opened 2021-11-03
client certificates: only display it on AWS when the command is supported

axbarsan:
Towards https://github.com/giantswarm/roadmap/issues/388

This PR makes it so we only display the client certificate creation guide if the cluster supports it. On AWS the cluster must have a release newer or equal to `v16.0.1`, and the cluster must be in an organization namespace.


src/components/MAPI/keypairs/ClusterDetailKeyPairs.tsx
```diff
@@ -137,101 +168,105 @@ const ClusterDetailKeyPairs: React.FC<IClusterDetailKeyPairsProps> = () => {
           pathname,
         }}
       >
-        <Box>
+        <>
```
axbarsan, line 171: Unrelated, but `<Breadcrumb />` requires a single child, so this was displaying an error in the console

kuosandys (approved):

PR #2901, opened 2021-11-08
Remove reviewers from happa dependency PRs

marians:
See title


axbarsan (commented):
We should remove all reviewers, since the team is automatically added as a reviewer (because of CODEOWNERS)

marians:
Good point

axbarsan (approved):

PR #2902, opened 2021-11-08
Fix flash messages duplication and content overflow

kuosandys:
This PR fixes a bug with flash messages, where duplicate messages were being displayed after the user closes the first one. It also adds the `word-break` property to flash message content to prevent text content from overflowing.

### DIFF
--- a/src/lib/Queue.ts
+++ b/src/lib/Queue.ts
@@ -66,7 +66,9 @@ class QueueImpl<T> implements IQueue<T>, Iterable<T> {
    * @param entry
    */
   public includes(entry: T) {
-    return this.entries.includes(entry);
+    return this.entries.some(
+      (currentEntry) => JSON.stringify(currentEntry) === JSON.stringify(entry)
+    );
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/lib/Queue.ts
```diff
@@ -66,7 +66,9 @@ class QueueImpl<T> implements IQueue<T>, Iterable<T> {
    * @param entry
    */
   public includes(entry: T) {
-    return this.entries.includes(entry);
+    return this.entries.some(
+      (currentEntry) => JSON.stringify(currentEntry) === JSON.stringify(entry)
+    );
```
axbarsan, line 71: In the context of `FlashMessagesController` this is fine, since we control how we create queue entries. But this is a generic class that should be usable in other places, as well. See this example:

image

The objects should be equal, but they're not, since the keys are in a different order.

We could use a deep comparison algorithm here. We already use `underscore`'s `isEqual()` in another place, I think it would be better to use it here, as well.

axbarsan (approved):

PR #2905, opened 2021-11-08
Improve app installation modal: keyboard accessibility and scrollable content

kuosandys:
This PR makes the cluster picker in app installation modal accessible with keyboard navigation. This is done by replacing some of the `div` elements with `grommet`'s `<Box>` component.

The content/body of the app installation modal is also set to scroll on vertical overflow. Previously, if the modal height exceeds the user's screen height, the elements within the content/body of the modal will shrink or overlap. The scrollbar is styled to align with the modal's style, and is similar to the scrollbar used in the `<CodeBlock>` component.

<details>
<summary>before</summary>
Screen Shot 2021-11-08 at 20 26 20
</details>

<details>
<summary>with scrollable content</summary>
Screen Shot 2021-11-08 at 20 33 20
</details>


src/components/UI/Layout/Modal/index.tsx
```diff
@@ -15,6 +15,22 @@ const CloseButtonText = styled(Text)`
   line-height: normal;
 `;
 
+const Content = styled(Box)`
+  ::-webkit-scrollbar {
```
axbarsan, line 19: *cries in firefox*

axbarsan (commented):
When you focus on a cluster (i.e. by TABbing to it), it would be nice to be able to select it with the `Space` key, as well. Right now it is only possible to do so with the `Enter` key.

image

src/components/UI/Layout/Modal/index.tsx
kuosandys: I added some properties to support Firefox as well, but I don't think there's an equivalent property to style scrollbar radius. It looked a bit obtrusive with the sharp corners, so I made the scrollbar track transparent instead:
Screen Shot 2021-11-09 at 11 59 29

axbarsan (approved):

PR #2910, opened 2021-11-09
Make permissions auto-refreshable

axbarsan:
Towards https://github.com/giantswarm/roadmap/issues/551

This PR refactors the way we deal with permissions.

We previously made all the necessary MAPI requests, computed and stored all the permissions in the Redux store once, as part of the app's initial load.

We now want to make these permissions automatically update in the UI, as they are updated in MAPI. We also want to have an easy way of dealing with permissions in multiple components, while also taking into account differences between providers.

This PR is an attempt towards that, moving all the permission storage/fetching logic into a React hook, which uses `swr` internally. The permissions are updated every minute, but also when the window is focused or when reconnecting to the internet (if you were previously disconnected).

This PR also includes 2 examples of resource-specific permission hooks, as well as how we can compose one using another. These hooks are usable in any component, and we can use the results to hide or show different parts of the UI.
I believe these should be based on business-level resources (i.e. clusters, node pools or apps), and not on the individual MAPI resources.


src/components/MAPI/utils.ts
```diff
@@ -1089,3 +1089,38 @@ export function supportsClientCertificates(cluster: Cluster): boolean {
       return false;
   }
 }
+
+/**
+ * Compute an organization namespace from the given organization name.
+ * This also makes the org name follow the [DNS label standard](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names).
+ * @param name
+ */
+export function getNamespaceFromOrgName(name: string): string {
```
axbarsan, line 1098: Moved from another file

src/components/MAPI/permissions/utils.ts
```diff
@@ -0,0 +1,306 @@
+import { HttpClientFactory } from 'lib/hooks/useHttpClientFactory';
```
axbarsan, line 1: Everything here was moved from different files

src/components/Layout.tsx
```diff
@@ -0,0 +1,171 @@
+import { useAuthProvider } from 'Auth/MAPI/MapiAuthProvider';
```
axbarsan, line 1: Refactored this into a functional component (and TypeScript), to be able to use hooks

src/components/MAPI/clusters/permissions/usePermissionsForClusters.ts
```diff
@@ -0,0 +1,200 @@
+import { usePermissions } from 'MAPI/permissions/usePermissions';
+import { hasPermission } from 'MAPI/permissions/utils';
+import { Providers } from 'shared/constants';
+import { PropertiesOf } from 'shared/types';
+
+import { usePermissionsForCPNodes } from './usePermissionsForCPNodes';
+
+export function usePermissionsForClusters(
```
axbarsan, line 8: A more complex example of how a permission hook might look like

src/components/MAPI/permissions/usePermissions.ts
```diff
@@ -0,0 +1,62 @@
...
+import { IState } from 'stores/state';
+import useSWR from 'swr';
+
+import { IPermissionMap } from './types';
+import { fetchPermissions } from './utils';
+
+// eslint-disable-next-line no-magic-numbers
+const REFRESH_INTERVAL = 60 * 1000; // In ms.
+
+export function usePermissions() {
```
axbarsan, line 20: This hook is the base for all permission hooks, and it handles fetching, storing and updating permissions

kuosandys (approved):
This is really cool! 👍

PR #2919, opened 2021-11-10
Display scheduled cluster upgrades

kuosandys:
Closes https://github.com/giantswarm/roadmap/issues/465.

This PR makes use of the annotations that can be added to `Cluster` CRD's to schedule upgrades. If a cluster has an upgrade scheduled, the cluster status "Upgrade scheduled" will be displayed instead of the status "Upgrade available".

`Upgrade cluster` button:
If a cluster has these annotations, it is still possible to initiate an upgrade, but only to a version that is older than the scheduled version. In order to allow the upgrade to a newer version (including the version that the scheduled upgrade specifies), we remove the annotations at the same time we set the release label, initiating the upgrade request.
Screen Shot 2021-11-11 at 13 10 22
Screen Shot 2021-11-11 at 13 10 08


PrashantR31:
Nice one :)

PrashantR31 (approved):

marians:
> If a cluster has these annotations, it is still possible to initiate an upgrade, but only to a version that is older than the scheduled version.

Why this restriction?

axbarsan (approved):
Tested and works great 👍

kuosandys:
> > If a cluster has these annotations, it is still possible to initiate an upgrade, but only to a version that is older than the scheduled version.
> 
> Why this restriction?

The targeted version for the scheduled upgrade (in the annotation `alpha.giantswarm.io/update-schedule-target-release`) is validated against the current version (in the label `release.giantswarm.io/version`). This prevents us from scheduling an "upgrade" to a version older than or equal to the current version - i.e. the targeted version for the upgrade must always be newer than the current version.

When we initiate an upgrade for a cluster, we do so by changing the release label. However, if there are also annotations in place for a scheduled upgrade, the request could be denied if the resulting cluster version is now newer than or equal to the targeted version for the scheduled upgrade.

anvddriesch:
I really like this!

marians:
Thanks for the explanation! Anyway, I think the outcome is a bit counter-intuitive.

I'm thinking of this scenario: We have scheduled an upgrade for a cluster, but now I find that I want to upgrade earlier than initially planned, so I wanter to trigger the upgrade manually.

In order to manually upgrade to the scheduled version, we may remove the annotation (for the scheduled upgrade) and set the release in one step. Is that possible?

marians:
Another thing I'm wondering: we are showing text like `The cluster will be upgraded to vX.X.X in about 1 hour`. Is that exact enough? Or should we show the exact time in that case? As we are in an overlay already, we can't offer another overlay to show the exact date for the duration phrase (which is what we should normally do).

kuosandys:
> Thanks for the explanation! Anyway, I think the outcome is a bit counter-intuitive.
> 
> I'm thinking of this scenario: We have scheduled an upgrade for a cluster, but now I find that I want to upgrade earlier than initially planned, so I wanter to trigger the upgrade manually.
> 
> In order to manually upgrade to the scheduled version, we may remove the annotation (for the scheduled upgrade) and set the release in one step. Is that possible?

Sorry, I should have explained the second part better. The action you mentioned above (upgrading to the target version earlier, or an even newer version) is possible - we do a check in that case and remove the annotations + set the release version.

marians:
Now I am confused. My understanding of "If a cluster has these annotations, it is still possible to initiate an upgrade, but only to a version that is older than the scheduled version." so far, in an example:

- Cluster is at v1.2.3
- Annotations say the cluster will be upgraded to v1.2.5 soon
- User cannot upgrade to v1.2.5 in happa, but can upgrade to v1.2.4

kuosandys:
> Now I am confused. My understanding of "If a cluster has these annotations, it is still possible to initiate an upgrade, but only to a version that is older than the scheduled version." so far, in an example:
> 
> * Cluster is at v1.2.3
> * Annotations say the cluster will be upgraded to v1.2.5 soon
> * User cannot upgrade to v1.2.5 in happa, but can upgrade to v1.2.4

I was describing the limitations, and in the following sentence I explained what we changed to overcome it:
In order to allow the upgrade to a newer version (including the version that the scheduled upgrade specifies), we remove the annotations at the same time we set the release label, initiating the upgrade request.

Edited for more clarity, let me know if I can explain better in a call

marians (approved):
All right, I got it now. Thank you!

My nitpick regarding the non-exact time remains, but it shouldn't block this.

kuosandys:
> Another thing I'm wondering: we are showing text like `The cluster will be upgraded to vX.X.X in about 1 hour`. Is that exact enough? Or should we show the exact time in that case? As we are in an overlay already, we can't offer another overlay to show the exact date for the duration phrase (which is what we should normally do).
> 
> Edit: the scenario I'm thinking of is "As a cluster admin, I want to communicate to others when a cluster will. be upgraded.". With the info given, I'll have a hard time finding that our exactly.

I think in this case it's better show the date scheduled 👍

marians (approved):

PR #2924, opened 2021-11-11
Add staff support for impersonating users

axbarsan:
Closes https://github.com/giantswarm/roadmap/issues/564

This PR adds a new `Impersonation` section to the `Experiments` page.
With it we can act as other users, test permissions or debug problems.

This is only visible to Giant Swarm staff, for now. We can create better UI support for customers later on.

Impersonation is also persisted between sessions. If you get stuck, you can delete the `localStorage` entry from the DevTools Inspector.

[Read more about Kubernetes API impersonation](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#user-impersonation)

## Preview

image

### DIFF
--- a/src/components/MAPI/Experiments.tsx
+++ b/src/components/MAPI/Experiments.tsx
@@ -6,6 +6,7 @@ import React, { useEffect, useState } from 'react';
 import { Breadcrumb } from 'react-breadcrumbs';
 import { AccountSettingsRoutes } from 'shared/constants/routes';
 import * as featureFlags from 'shared/featureFlags';
+import { mutate } from 'swr';
 import Button from 'UI/Controls/Button';
 import {
   Table,
@@ -16,6 +17,8 @@ import {
 } from 'UI/Display/Table';
 import CheckBoxInput from 'UI/Inputs/CheckBoxInput';
 import TextInput from 'UI/Inputs/TextInput';
+
+import { usePermissionsKey } from './permissions/usePermissions';
 
 interface IExperimentsProps {}
 
@@ -45,13 +48,17 @@ const Experiments: React.FC<IExperimentsProps> = () => {
     })();
   }, [auth]);
 
-  const handleSetImpersonation = (e: React.FormEvent<HTMLFormElement>) => {
+  const handleSetImpersonation = async (
+    e: React.FormEvent<HTMLFormElement>
+  ) => {
     e.preventDefault();
 
-    auth.setImpersonationMetadata({
+    await auth.setImpersonationMetadata({
       user: impersonationUser,
       groups: impersonationGroup ? [impersonationGroup] : undefined,
     });
+
+    mutate(usePermissionsKey);
 
     new FlashMessage(
       'Impersonation configured successfully.',
@@ -60,11 +67,13 @@ const Experiments: React.FC<IExperimentsProps> = () => {
     );
   };
 
-  const handleClearImpersonation = () => {
+  const handleClearImpersonation = async () => {
     setImpersonationUser('');
     setImpersonationGroup('');
 
-    auth.setImpersonationMetadata(null);
+    await auth.setImpersonationMetadata(null);
+
+    mutate(usePermissionsKey);
 
     new FlashMessage(
       'Impersonation removed successfully.',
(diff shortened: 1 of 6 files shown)


src/lib/OAuth2/OAuth2.ts
```diff
@@ -40,12 +40,21 @@ export interface IOAuth2SigningKey {
   e: string;
 }
 
+export interface IOAuth2ImpersonationMetadata {
```
axbarsan, line 43: Luckily impersonation is also a OAuth2 thing, so we can add it here 🥳

src/model/services/mapi/generic/ensureClientAuth.ts
```diff
@@ -24,6 +24,15 @@ export async function ensureClientAuth(
 
   client.setAuthorizationToken(user.authorizationType, user.idToken);
 
+  const impersonationMetadata = await auth.getImpersonationMetadata();
```
axbarsan, line 27: Decorating all MAPI requests with the impersonation configuration (if it's set)

marians:
Giving it a test drive now

marians:
A conceptual thing: I think it should it be possible to set more than one group name. SSO users can be member of multiple groups, too. So for testing it would help to be able to apply that. I see that the storage object stores a groups array. So all we have to do I guess is allow input of multiple names. We could simply use white space as separator. Plus add the according info next to the input field.

---

My first test was to create a rolebinding for the `read-all` group in an organization (`giantswarm` on `gollum`). Like this:

```nohighlight
$ kubectl -n org-giantswarm create rolebinding marian-read-only-test --role read-all --group read-only-users
rolebinding.rbac.authorization.k8s.io/marian-read-only-test created
```

Then I used the impersonation form with these settings:

- Username: `username-does-not-matter`
- Group: `read-only-users`

I hit the `Set` button and navigated to the home page (Cluster). There I got a lot of error messages saying something like "user 'username-does-not-matter' is not allowed to get resource XYZ". Then the page disappeared and the Insufficient permissions screen appeared instead.

image

When going through "Try logging in again" I'm running into the same "Insufficient permissions" screen again after logging in. So I was locked out.

I guess this is more or less all expected, because the user is in fact lacking permissions required for happa.

Can we improve the "Insufficient permissions" screen for this case? I think we could:

- Show the impersonated user name and groups instead of the one from the ID token
- If impersonation is in place, tell the user how to  quit impersonation, or offer a button to quit impersonation simply.

---

Question: what are the exact permissions again happa needs for every user so they don't see the "Insufficient permissions" screen?

src/components/MAPI/Experiments.tsx
```diff
@@ -80,6 +122,61 @@ const Experiments: React.FC<IExperimentsProps> = () => {
...
+                label='Username'
+                id='username'
+                name='username'
+                help='The user defined in the RoleBinding.'
+                required={true}
+                value={impersonationUser}
+                onChange={(e) => setImpersonationUser(e.target.value)}
+              />
+              <TextInput
+                label='Group'
```
marians, line 163:
```suggestion
                label='Groups'
```

src/components/MAPI/Experiments.tsx
```diff
@@ -28,6 +32,44 @@ const Experiments: React.FC<IExperimentsProps> = () => {
...
+    new FlashMessage(
+      'Impersonation configured successfully.',
+      messageType.SUCCESS,
+      messageTTL.MEDIUM
+    );
+  };
+
+  const handleClearImpersonation = () => {
+    auth.setImpersonationMetadata(null);
+
```
kuosandys, line 65: Could we clear the form in the UI as well? 🙂

axbarsan:
> A conceptual thing: I think it should it be possible to set more than one group name. SSO users can be member of multiple groups, too. So for testing it would help to be able to apply that. I see that the storage object stores a groups array. So all we have to do I guess is allow input of multiple names. We could simply use white space as separator. Plus add the according info next to the input field.

Yes, the K8s API supports multiple groups, but it requires multiple headers with the same name (`Impersonate-Group`). This is not possible at the moment in our HTTP Client, because internally it uses a map for handling headers. Supporting multiple groups is an improvement and can be added later. A single group is fine for now.

I'll take a look at improving the `Insufficient permissions` screen

src/components/MAPI/Experiments.tsx
axbarsan: Nah that would be too smart 😆

axbarsan:
If impersonating an user, the `Insufficient permissions` page now looks like this:

image

kuosandys (approved):

marians:
> the K8s API supports multiple groups, but it requires multiple headers with the same name

Oh wow!

PR #2929, opened 2021-11-12
Fix deleted cluster description stuck in loading state

kuosandys:
Closes https://github.com/giantswarm/giantswarm/issues/19780.

This PR fixes a bug with deleted clusters' cluster descriptions displaying as loading indefinitely. This was caused by not making a distinction between a provider cluster that is loading and a provider cluster that has been deleted.

On Azure, the cluster description is found in the `Cluster` resource, so we are still able to display the description even after the cluster resources have been deleted. On AWS, the cluster description is found on the provider cluster resource, so after deletion, we display the `N/A` label appropriately.


src/components/MAPI/clusters/ClusterList/ClusterListItem.tsx
```diff
@@ -81,7 +81,7 @@ const ClusterListItem: React.FC<IClusterListItemProps> = ({
 }) => {
   const name = cluster?.metadata.name;
   const description = useMemo(() => {
-    if (!cluster || !providerCluster) return undefined;
+    if (!cluster || typeof providerCluster === 'undefined') return undefined;
```
axbarsan, line 84: Do we need to do this in the cluster detail view, as well?

axbarsan (commented):
So now, what does an `undefined` provider-specific cluster mean? How about a `null` one?

src/components/MAPI/clusters/ClusterList/ClusterListItem.tsx
kuosandys: Yes we do 😅

axbarsan (approved):

PR #2933, opened 2021-11-15
Fix node pool creation: spot instances should be disabled by default

kuosandys:
In the node pool creation form, there is an option to enable spot instances, and this option should be disabled by default in both the UI and the resulting CRD created by the form.

The percentage of spot instances is equal to the percentage of instances that are not on-demand. So in order to disable spot instances by default, we set `onDemandPercentageAboveBaseCapacity` to 100. When the option is enabled, we set `onDemandPercentageAboveBaseCapacity` to 0.


marians (approved):
From a shallow look, looks good to me

ubergesundheit (approved):
Looks very good :+1:

PR #2938, opened 2021-11-16
Compute AWS node pool stats taking into account similar instance types

axbarsan:


axbarsan:
Similar types are actually guaranteed to have the same specs, so this is not necessary

src/components/MAPI/clusters/utils.ts
```diff
@@ -135,16 +151,21 @@ export function getWorkerNodesMemory(
...
+          ) as INodePoolMachineTypesAWS;
 
-          count += machineTypeProperties.memory * readyReplicas;
+          for (const instanceType of providerNPMachineTypes.all) {
+            const machineTypeProperties = machineTypes[instanceType];
+            if (!machineTypeProperties) {
+              return -1;
+            }
+
+            count += machineTypeProperties.memory * readyReplicas;
```
axbarsan, line 167: This was wrong anyway

PR #2966, opened 2021-11-24
Make the password reset form work properly

axbarsan:
As seen on https://gigantic.slack.com/archives/C02EVLE9W/p1637756165387300


axbarsan:
YOLO

PR #2972, opened 2021-11-25
Make access control UI aware of permissions. Re-enable binding subjects.

axbarsan:
This PR re-enables being able to bind and remove bindings for all subjects. It also now includes more granular control over what users can and can't do, all relative to their RBAC permissions.

What we control:
- Fetching `Role`s/`ClusterRole`s or both, depending on what the user has access to
- Not allowing users to remove subjects, if they don't have access to do it
- Not allowing users to bind subjects, if they don't have access to do it
- Hiding the lists of subjects if there are no entries and the user doesn't have access to binding others
- Only allowing users to bind existing service accounts, if they don't have access to creating new ones
- Not fetching service account name suggestions if the user doesn't have access to do it

For the actual RBAC permissions required for all these, scroll down.

Please note that this is not meant to be the best example UX/DX in dealing with permissions in our codebase. It is just a step towards more granular permissions in the UI.

### DIFF
--- a/src/components/MAPI/organizations/AccessControl/AccessControlRoleSubjects.tsx
+++ b/src/components/MAPI/organizations/AccessControl/AccessControlRoleSubjects.tsx
@@ -175,7 +175,6 @@ const getStatusMessages = (
     {
       [ui.AccessControlRoleSubjectStatus.Created]: [],
       [ui.AccessControlRoleSubjectStatus.Bound]: [],
-      [ui.AccessControlRoleSubjectStatus.Updated]: [],
     }
   );
 
--- a/src/components/MAPI/organizations/AccessControl/AccessControlRoleSubjects.tsx
+++ b/src/components/MAPI/organizations/AccessControl/AccessControlRoleSubjects.tsx
@@ -434,6 +434,21 @@ const AccessControlRoleSubjects: React.FC<IAccessControlRoleSubjectsProps> = ({
     }
   }, [serviceAccountSuggestionsError]);
 
+  const validateServiceAccountSubjects = (values: string[]): string => {
+    if (serviceAccountPermissions.canCreate) return '';
+
+    if (!serviceAccountSuggestions) {
+      return `You don't have the required permissions to create new service accounts. You can only bind existing ones.`;
+    }
+
+    const uniqueSuggestions = new Set(serviceAccountSuggestions);
+    if (values.some((v) => !uniqueSuggestions.has(v))) {
+      return `You don't have the required permissions to create new service accounts. You can only bind existing ones.`;
+    }
+
+    return '';
+  };
+
   return (
     <Box direction='column' gap='medium' pad={{ top: 'small' }} {...props}>
       {canListSubjects(groupCollection, groupPermissions) && (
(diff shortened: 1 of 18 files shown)


src/components/UI/Display/MAPI/AccessControl/types.ts
```diff
@@ -44,21 +44,27 @@ export interface IAccessControlRoleItem {
...
+  canBind: boolean;
   canDelete: boolean;
   canList: boolean;
 }
 
+export interface IAccessControlRolePermissions {
+  canList: boolean;
+}
+
 export interface IAccessControlPermissions {
```
axbarsan, line 57: This wouldn't be a good example for handling permissions. But since the access control part of the UI is built differently compared to other parts of the app, we need to get dirty with it

src/model/services/mapi/rbacv1/getClusterRoleBindingList.ts
```diff
@@ -15,7 +15,17 @@ export function getClusterRoleBindingList(
...
-  return getResource<IClusterRoleBindingList>(client, auth, url.toString());
+  const list = await getResource<IClusterRoleBindingList>(
+    client,
+    auth,
+    url.toString()
+  );
+  for (const item of list.items) {
+    item.kind = ClusterRoleBinding;
+    item.apiVersion = 'rbac.authorization.k8s.io/v1';
+  }
```
axbarsan, line 26: Turns out these weren't set, and we need them in the new logic 😅

src/components/UI/Display/MAPI/AccessControl/AccessControlSubjectAddForm.tsx
```diff
@@ -116,7 +116,12 @@ const AccessControlSubjectAddForm: React.FC<IAccessControlSubjectAddFormProps> =
     return (
       <Box {...props}>
         {isAdding ? (
-          <Box direction='row' gap='small' align='center'>
+          <Box
+            direction='row'
+            gap='small'
+            align='center'
+            width={{ max: 'medium' }}
```
axbarsan, line 123: To make custom error messages look ok

src/components/MAPI/permissions/utils.ts
```diff
@@ -202,7 +202,7 @@ export function hasNamespacePermission(
 export function hasPermission(
   permissions: IPermissionMap,
   namespace: string,
-  verb: string,
+  verb: PermissionVerb,
```
axbarsan, line 205: I missed this before

src/components/MAPI/organizations/permissions/usePermissionsForAccessControl.ts
```diff
@@ -0,0 +1,123 @@
+import { usePermissions } from 'MAPI/permissions/usePermissions';
+import { hasPermission } from 'MAPI/permissions/utils';
+import { Providers } from 'model/constants';
+import * as ui from 'UI/Display/MAPI/AccessControl/types';
+
+export function usePermissionsForAccessControl(
```
axbarsan, line 6: The actual permissions required are in this file

src/components/MAPI/organizations/AccessControl/AccessControlRoleSubjects.tsx
```diff
@@ -150,75 +151,80 @@ const compareSubjects = (
...
+    <React.Fragment key={name}>
+      <code>{name}</code>
       {idx !== arr.length - 1 && ', '}
     </React.Fragment>
   ));
 };
 
-const getBindServiceAccountSuccessMessages = (
-  accounts: ui.IAccessControlServiceAccount[]
+const getStatusMessages = (
```
axbarsan, line 163: Re-using the message logic we had for service accounts for all subject types

kuosandys (approved):

marians (approved):

PR #2973, opened 2021-11-25
Improve cluster labels widget

kuosandys:
Closes https://github.com/giantswarm/giantswarm/issues/19766.

This PR brings some improvements to the cluster labels widget in the cluster details overview tab. This is purely a UI change and the functionality remains the same. There is no change to the cluster label editing and deletion overlays besides some formatting (i.e. making sure that the validation error text is fully displayed and that there is space between the "Delete" and "Cancel" buttons in the deletion overlay).

Screen Shot 2021-11-25 at 21 14 57

Hovering over the label:
Screen Shot 2021-11-25 at 21 15 29


kuosandys:
@marians The colours used for the various states are definitely up for further discussion 👍

src/components/UI/Display/Cluster/ClusterLabels/DeleteLabelButton.tsx
```diff
@@ -1,10 +1,40 @@
...
-import Button from 'UI/Controls/Button';
+import { Anchor } from 'grommet';
+import React, { FC } from 'react';
+import styled from 'styled-components';
+import { Tooltip, TooltipContainer } from 'UI/Display/Tooltip';
 
-interface IButtonProps extends ComponentPropsWithRef<typeof Button> {}
+interface IButtonProps {}
+
+const StyledAnchor = styled(Anchor)`
```
kuosandys, line 8: This is mostly taken from the styled Anchor element we use for deleting access control subjects

marians:
Will take a closer look tomorrow

marians:
Running it now

axbarsan (commented):
Thank you for doing this, it looks great. However, I'm personally not very sure about 3 things:
- The contrast between the the label value and label key is too large. When you visit the cluster detail page, it's the first thing that you notice, and I don't think it should be that way.
- I believe the `Add` button is now the only pill-shaped button in the app, right? I believe it looks a bit out of place (which is also because of its styling)
- The `+` icon in the `Add` button is much larger than the icon in other buttons
image
vs
image

src/components/UI/Display/Cluster/ClusterLabels/ClusterLabel.tsx
```diff
@@ -0,0 +1,143 @@
...
+    color: ${({ theme }) => theme.colors.darkBlueDarker2};
+  }
+`;
+
+const StyledTooltip = styled(Tooltip)`
+  z-index: 1069 !important;
+`;
+
+interface IClusterLabelProps {
+  label: React.ReactNode | string;
```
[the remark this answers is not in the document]
kuosandys: Good point 😅

marians:
I just noticed that the `Add` button pill is still 32 pixels high (it should be identical with the label pill, 30px) and the font-size is also too big (15px instead of 14px).

kuosandys:
Another option for the dark blue would be to align the darker colour with the background of the Kubernetes API endpoint, which would decrease the contrast a bit:
Screen Shot 2021-11-26 at 19 06 10
(For reference, this is `shade2` #224a61)

Going even further, and making the light background a low opacity #eee with white text (taking this from the hover state of secondary buttons):
Screen Shot 2021-11-26 at 20 31 29

I think that the "Add" button could be better aligned with the rest of the UI - especially with the "Add node pool" button above using the same `+` icon, the size discrepancy is quite noticeable. In terms of the shape and styling, a square button inlined with the pill elements looks out of place, which was the decision behind using a round button. I think we should give this more thought.

src/components/UI/Display/Cluster/ClusterLabels/AddClusterLabelButton.tsx
```diff
@@ -34,14 +34,14 @@ const AddClusterLabelButton: React.FC<IAddClusterLabelButtonProps> = ({
...
-      <Text margin={{ right: 'xsmall' }} size='24px' color='text-weak'>
+      <Text margin={{ right: 'xxsmall' }} size='21px' color='text-weak'>
         <i
           className='fa fa-add-circle'
           role='presentation'
           aria-hidden='true'
         />
       </Text>
-      <Text alignSelf='center' color='text-weak'>
+      <Text size='14px' alignSelf='center' color='text-weak'>
```
axbarsan: Why should this be 14px? All of our text elements using `grommet` use 15px?

axbarsan (commented):
> In terms of the shape and styling, a square button inlined with the pill elements looks out of place, which was the decision behind using a round button.

But should labels be pills? A possible alternative solution would be to follow a similar pattern to subjects in the access control UI.

image

kuosandys:
Thank you for the insightful questions - I think there's still a lot to be done here if we want to bring a proper redesign to this widget. We also haven't touched on the adding/editing overlays yet, and those should also be taken into consideration.

I would like to close this PR for now in favour of https://github.com/giantswarm/happa/pull/2981 to implement more modest changes to improve readability, and revisit this again when we can give it a full rework :heart:

marians:
@kuosandys Your first suggestion for adjusting the color value appears fine to me. (Feel free to change it that way.) The second option actually appears less readable to me. Wouldn't use that one.

image

@axbarsan Regarding the pill shape and the font size: It's fine for labels to look different than subjects assigned to roles in Access Control, because they are a different thing. Also the font size is adjusted to have the resulting shape size balanced overall. I consider it quite readable.

We'll also continue to work towards consistency, both here and in the Access Control UI and elsewhere.

PR #2982, opened 2021-11-29
Only set impersonation group header if groups is defined

kuosandys:
I noticed that if `groups` is not set, the condition for setting "Impersonate-Group" in headers still evaluates to true, and impersonation doesn't behave as expected.

### DIFF
--- a/src/model/services/mapi/generic/ensureClientAuth.ts
+++ b/src/model/services/mapi/generic/ensureClientAuth.ts
@@ -30,7 +30,7 @@ export async function ensureClientAuth(
 
     if (
       impersonationMetadata.groups &&
-      impersonationMetadata.groups?.length > 0
+      impersonationMetadata.groups.length > 0
     ) {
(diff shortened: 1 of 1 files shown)


src/model/services/mapi/generic/ensureClientAuth.ts
```diff
@@ -28,8 +28,11 @@ export async function ensureClientAuth(
   if (impersonationMetadata) {
     client.setHeader('Impersonate-User', impersonationMetadata.user);
 
-    if (impersonationMetadata.groups?.length !== 0) {
-      client.setHeader('Impersonate-Group', impersonationMetadata.groups![0]);
+    if (
+      impersonationMetadata.groups &&
+      impersonationMetadata.groups?.length > 0
```
axbarsan, line 33:
```suggestion
      impersonationMetadata.groups.length > 0
```

axbarsan (approved):

PR #2991, opened 2021-12-01
Fix G8sControlPlane CR's infrastructure reference namespace when creating non-org-namespaced AWS clusters

kuosandys:
Closes https://github.com/giantswarm/roadmap/issues/607.

Cluster resources for non-org-namespaced clusters (i.e. AWS `v15.x.x` and older) should be placed in the `default` namespace. This PR fixes a bug in cluster creation when such a release is selected, so we also update the `G8sControlPlane` CR's infrastructure reference to point to the `default` namespace for these clusters.


src/components/MAPI/clusters/CreateCluster/patches.ts
```diff
@@ -65,6 +65,13 @@ export function withClusterReleaseVersion(
       controlPlaneNode.metadata.namespace = hasNonNamespacedResources
         ? defaultNamespace
         : orgNamespace;
+      if (
+        providerCluster &&
+        controlPlaneNode.kind === infrav1alpha3.G8sControlPlane
+      ) {
+        controlPlaneNode.spec.infrastructureRef.namespace =
```
kuosandys, line 72: We set the namespace for the control plane node's infrastructure reference to be the same one as `AWSControlPlane` CR's namespace.

marians (approved):

PR #2992, opened 2021-12-01
Update ingress resources

ubergesundheit:
This PR does not touch the actual code of the web UI. 

The changes are in the helm chart:

- Change certificate secret type to kubernetes/tls. This allows cert-exporter to monitor a possible expiry
- Update the Ingress resources to networking.k8s.io/v1. This prevents installing on kubernetes below 1.19
- Update ClusteRole and ClusterRoleBinding apiVersion
- Use `ingress.tls` values for certificate secrets. This is the same as used by athena and dex-app
- Put happaapi Ingress and secret into `giantswarm` namespace instead of `default`


kuosandys (approved):
👍

helm/happa/Chart.yaml
```diff
@@ -7,3 +7,4 @@ version: [[ .Version ]]
 appVersion: [[ .AppVersion ]]
 annotations:
   config.giantswarm.io/version: 1.x.x
+kubeVersion: ">=1.19.0-0"
```
marians, line 10: ❤️

helm/happa/templates/happaapi-ingress.yaml
```diff
@@ -1,12 +1,8 @@
-{{ if .Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress" }}
 apiVersion: networking.k8s.io/v1
-{{ else }}
-apiVersion: networking.k8s.io/v1beta1
-{{ end }}
 kind: Ingress
 metadata:
   name: happaapi
-  namespace: default
+  namespace: giantswarm
```
marians, line 5: Have you tested the namespace change thoroughly?

This ingress is a proxy to the Kubernetes API server (see `...backend.service.name: kubernetes`). I have tried moving the ingress resource to the giantswarm namespace in the past, and as a result, happaapi.* didn't work any more.

ubergesundheit: No I didn't test this, but good point. I'll see what I can find!

ubergesundheit: You're right! Reverted my changes.

PR #2997, opened 2021-12-02
Support preview releases as read-only

kuosandys:
Towards https://github.com/giantswarm/roadmap/issues/538.

This PR supports the display of preview releases. Interactions with clusters using preview release versions are not currently supported, but these clusters will still be visible from the UI if they were created via other means. The following changes have been made to support preview releases and these clusters:

### 1. Cluster list item:
We do not display node pool/worker nodes/CPU/RAM information for these clusters. We also do not display the "Get started" button. Instead, we add a label indicating that the cluster is using a preview release.
Screen Shot 2021-12-08 at 12 21 43

### 2. Navigation to the cluster details pages for these clusters are disabled:
Clicking on the cluster list item is disabled. If a user tries to enter the URL to navigate to the cluster details pages, we redirect them pack to the cluster list page.

### 3. Release selector in cluster creation form:
We display a release's state in the release selector within the cluster creation form. The last active release is automatically selected, as it was before. Selection of the preview release is disabled. The following is the release list as viewed by non-Giant Swarm staff.
Screen Shot 2021-12-08 at 12 18 51

Preview releases are visible to everyone, while deprecate, WIP, and non-preview pre-releases are only visible to Giant Swarm staff. The following is the release list as viewed by Giant Swarm staff.
Screen Shot 2021-12-08 at 12 12 13

### 4. App installation modal does not display clusters using preview release versions:
We only display clusters in the app install modal if we can currently install apps on them (e.g. a cluster being deleted will not be displayed). Following this logic, we also do not show clusters using preview release versions in this list.

An additional test case was added to ensure that we do not display an "Upgrade available" warning if the next version available for upgrade is a preview release.


marians:
@kuosandys Can we pair at some point to review this one? I figure that might be easiest - at leat for me :)

marians (approved):
We talked things through. It'll be great!

PR #3035, opened 2021-12-14
Make cluster details > actions tab aware of permissions for a read-only user

kuosandys:
Towards [giantswarm/roadmap#551](https://github.com/giantswarm/roadmap/issues/551).

This PR makes the cluster details > actions tab aware of permissions. The UI is adapted in the following ways for a read-only user:

**Disable "Delete..." button and adapt warning message:**
Screen Shot 2021-12-13 at 20 38 30
The cursor is also displayed in the `not-allowed` style on hover over the button.

**Display warning message in CLI guide:**
Screen Shot 2021-12-13 at 20 40 39

**Checking for permissions for node pools and apps:**
Although the deletion confirmation step (the step showing number of worker nodes and apps installed in the cluster) is not shown for a read-only user, the data is fetched in the same parent component. Thus, we check whether the user has permissions to read this data before fetching.

### DIFF
--- a/src/components/MAPI/permissions/utils.ts
+++ b/src/components/MAPI/permissions/utils.ts
@@ -1,6 +1,7 @@
 import { getNamespaceFromOrgName } from 'MAPI/utils';
 import * as authorizationv1 from 'model/services/mapi/authorizationv1';
 import { LoggedInUserTypes } from 'model/stores/main/types';
+import { AccessControlRoleItemVerb } from 'UI/Display/MAPI/AccessControl/types';
 import { HttpClientFactory } from 'utils/hooks/useHttpClientFactory';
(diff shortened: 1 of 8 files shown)


marians (approved):

src/components/MAPI/permissions/utils.ts
```diff
@@ -301,3 +302,66 @@ export function hasAppAccess(
...
+          group,
+          resource,
+        },
+      };
+
+      const accessReviewResponse =
+        await authorizationv1.createSelfSubjectAccessReview(
+          httpClientFactory(),
+          auth,
+          request
```
kuosandys, line 338: The permissions map we use to calculate user permissions does not include the WC namespace, so we have to calculate permissions for `apps` a little differently: we create a `SelfSubjectAccessReview` in the WC namespace for a specific resource

PR #3045, opened 2021-12-17
Use normalized app version when comparing app versions

kuosandys:
Closes https://github.com/giantswarm/roadmap/issues/647.

This PR normalizes app versions by trimming the `v` prefix from an installed app's current version (`app.spec.version`) before using it in comparisons. This fixes the bug for installed apps being displayed as switching to a newer version when it was the same version with a `v` prefix.


ubergesundheit:
Unsure if the version could be normalized once when the app cr information is loaded instead of doing it on every rerender? I haven't looked too deep into the actual code, if you feel this is the right way to do this, I'll happily give you my approval :)

kuosandys:
> Unsure if the version could be normalized once when the app cr information is loaded instead of doing it on every rerender? I haven't looked too deep into the actual code, if you feel this is the right way to do this, I'll happily give you my approval :)

Thanks for checking this out! I think this could be a good improvement, and help reduce repetition as well - I'll work on the changes 🙂

ubergesundheit (approved):
Sorry for being not specific enough. Is there maybe a place when the app CR data is loaded (redux store?) so that this normalization function needs to be executed just once the data is loaded?

Again, I don't want to block you, so here is my approval :)

kuosandys:
> Sorry for being not specific enough. Is there maybe a place when the app CR data is loaded (redux store?) so that this normalization function needs to be executed just once the data is loaded?
> 
> Again, I don't want to block you, so here is my approval :)

Ah ok, thanks for clarifying! So the `app` CR data isn't loaded from anywhere "central". As with other MAPI UI components, we use `SWR` here, which loads the data from cache while a fetch request revalidates and refreshes the data as needed. We can then access the data from any component using a hook.

Since the `app` CR data is constantly being refreshed, it felt cleaner to derive the normalized app version from `app.spec.version` rather than storing that value somewhere (e.g. redux store), and manually checking and updating the value each time `app` CR data changes. It is also not a very computationally heavy function (it just checks and trims strings), but I think it was a good suggestion to not re-evaluate it too much 👍 

Hope this makes sense, let me know if I can clarify anything!

ubergesundheit:
Makes total sense! :)

PR #3060, opened 2022-01-03
Reload organizations list when impersonation is set or cleared

kuosandys:
Closes https://github.com/giantswarm/giantswarm/issues/20244.


PrashantR31 (approved):
Cool, thanks for that @kuosandys

PR #3082, opened 2022-01-11
Fix upgradable apps count in cluster details overview > apps widget

kuosandys:
Closes https://github.com/giantswarm/roadmap/issues/680.

This PR fixes a bug in the cluster details overview page > apps widget, where the number of upgradable apps displayed was incorrect. This is similar to the issue described in https://github.com/giantswarm/roadmap/issues/647. Some error handling was also added to prevent `n/a` values from being incorrectly displayed when some of the requests for determining the number of upgradable apps fail due to insufficient permissions.


PrashantR31 (approved):
:+1:

PR #3083, opened 2022-01-11
Move headers set in ingress configuration snippets to nginx.conf

kuosandys:
Towards https://github.com/giantswarm/giantswarm/issues/20283.

This PR moves the headers added in `nginx.ingress.kubernetes.io/configuration-snippet` to the `nginx.conf` file. This is to support installing `happa` on OpenStack MC's, where ingress snippets are disabled.


marians:
@kuosandys Are you sure it's necessary to apply the `add_headers` directives 3 times? My understanding is that if they are added to the `server` context, it should apply in all subordinate `location` contexts.

kuosandys:
> @kuosandys Are you sure it's necessary to apply the `add_headers` directives 3 times? My understanding is that if they are added to the `server` context, it should apply in all subordinate `location` contexts.

@marians I was following some information from the [nginx docs about `add_header`](https://nginx.org/en/docs/http/ngx_http_headers_module.html#add_header), but I'm not an expert here, so I might be misunderstanding it 😅 From what's written, it seems to imply that `add_header` directives are only inherited if the current level doesn't set them?

marians:
Thanks for pointing that out!

> These directives are inherited from the previous configuration level if and only if there are no add_header directives defined on the current level.

Then it's definitely right this way.

One wish: can we more the NGINX related files into a subfolder? I think we should try to keep stuff out of the repo root as good as possible.

nginx/headers.conf
```diff
@@ -0,0 +1,4 @@
+add_header X-Frame-Options sameorigin always;
+add_header X-XSS-Protection "1; mode=block" always;
+add_header X-Content-Type-Options "nosniff" always;
+add_header Referrer-Policy "no-referrer-when-downgrade" always;
```
tfussell, line 4:
```suggestion
add_header Referrer-Policy "no-referrer-when-downgrade" always;

```

tfussell (approved):
Tested on `gamma` (OpenStack). LGTM.

nginx/headers.conf
```diff
@@ -0,0 +1,4 @@
+add_header X-Frame-Options sameorigin always;
+add_header X-XSS-Protection "1; mode=block" always;
```
marians, line 2:
```suggestion
# For Storybook, as it's using iFrames
add_header X-Frame-Options sameorigin always;

add_header X-XSS-Protection "1; mode=block" always;
```

marians (approved):

PR #3088, opened 2022-01-12
Make organization overview aware of permissions for a read-only user

kuosandys:
Towards giantswarm/roadmap#551.

This PR refactors the organizations overview page to be aware of permissions and adapts the UI for a read-only user in the following ways:

**Disable 'Add organization' button and display warning message**
Screen Shot 2022-01-12 at 17 44 23
The cursor is also displayed in the not-allowed style on hover.

**Display warning message in CLI guide for organization creation**
Screen Shot 2022-01-12 at 17 44 11

As with the other permission-awareness-related PR's so far, this PR includes refactoring of the logic for how we fetch resources. For detailed explanations on changes, please see the comments below for the sections of the code that were refactored/adapted, and feel free to ask any questions for clarification!


src/components/MAPI/organizations/permissions/usePermissionsForOrganizations.ts
```diff
@@ -0,0 +1,84 @@
...
+    ),
+    () =>
+      fetchAccessForResource(
+        httpClientFactory,
+        auth,
+        '',
+        verbs,
+        'security.giantswarm.io',
+        'organizations'
+      ),
```
kuosandys, line 49: Since `organizations` is not a namespaced resource, we cannot accurately derive the permissions from the permissions map as we do for namespaced resources. Instead, we perform a `SelfSubjectAccessReview` for resources and verbs we're interested in.

src/components/MAPI/organizations/utils.ts
```diff
@@ -64,3 +72,101 @@ export function computeClusterCountersForOrganizations(clusters?: Cluster[]) {
...
   }, {});
 }
+
+export async function fetchClusterListForOrganizations(
+  httpClientFactory: HttpClientFactory,
+  auth: IOAuth2Provider,
+  cache: Cache,
+  provider: PropertiesOf<typeof Providers>,
+  organizations: Record<string, IOrganization>
+): Promise<ClusterList> {
```
kuosandys, line 82: This is the function for fetching cluster list for organizations. First we perform a `SelfSubjectAccessReview` to determine if the user has permissions to list clusters at the cluster scope. If they do, we fetch the cluster list as we do before. If they do not, then we iterate through the namespaces they do have access to, and the cluster list from each namespace before aggregating.

The reason for performing the access check first and performing a request at the cluster scope if the user has permissions, is for performance reasons. For admin users with access to everything, the number of organizations they belong to would likely be extensive. During testing on `gauss`, it was taking a significantly longer time to iterate through all the organizations.

src/components/MAPI/organizations/utils.ts
```diff
@@ -64,3 +72,101 @@ export function computeClusterCountersForOrganizations(clusters?: Cluster[]) {
...
+  const requests = Object.entries(organizations).map(
+    async ([organizationName, organizationEntry]) => {
+      const clusterListKey = fetchClusterListKey(
+        provider,
+        organizationEntry.namespace,
+        organizationName
+      );
+      const cachedClusterList: capiv1alpha3.IClusterList | undefined =
+        cache.get(clusterListKey);
+
```
kuosandys, line 121: Trying to improve performance here as well by checking if there is already cached data in `SWR` for clusters for a specific namespace, before launching another fetch request. If data is fetched, then we write it to the cache. This is similar to how we check for `appcatalogentries` in the cache before fetching.

src/model/stores/organization/actions.tsx
```diff
@@ -175,7 +175,7 @@ export function organizationsLoadMAPI(
         // to check if the orgs have a 'ui.giantswarm.io/original-organization-name'
         // annotation.
         const orgGetRequests = organizationNames.map((orgName) =>
-          securityv1alpha1.getOrganization(client, auth, orgName)
+          securityv1alpha1.getOrganization(new HttpClientImpl(), auth, orgName)
         );
```
kuosandys, line 179: This was returning unexpected results - we have to pass in a new instance of `HttpClientImpl` for each request here.

src/components/MAPI/organizations/OrganizationListCreateOrg.tsx
```diff
@@ -80,7 +82,7 @@ const OrganizationListCreateOrg: React.FC<IOrganizationListCreateOrgProps> = ({
   const handleSubmit = async (e: React.FormEvent<HTMLElement>) => {
     e.preventDefault();
 
-    if (!isValid) return;
+    if (!isValid || !canCreateOrganizations) return;
```
kuosandys, line 85: Extra check here for the function handling organization creation. Since we are fetching permissions every minute or so, there could be a possibility that from the time the page is rendered to the submission request is made that the user's permissions has changed.

src/components/MAPI/organizations/utils.ts
PrashantR31: :+1:

PrashantR31 (approved):

PR #3089, opened 2022-01-12
Push to openstack-app-collection

tfussell:
We would like for new releases to be deployed automatically to the openstack-app-collection so we don't have to update the app collection manually.


kuosandys (approved):
Thanks @tfussell ! LGTM if it works for you 👍

PR #3108, opened 2022-01-19
Add a 'pipeline' tag to Sentry events, to distinguish stable/testing

marians:
This PR adds a [custom tag](https://docs.sentry.io/platforms/javascript/enriching-events/tags/) named `pipeline` to any Sentry event. We can use that to filter out issues created in test installations from notifications.


kuosandys (approved):
Thanks for doing this!

PR #3117, opened 2022-01-21
Set happaapi request size limit to 4m

marians:
Attempts to fix https://github.com/giantswarm/giantswarm/issues/20464


marians:
FYI: this is actually not ready for review.

marians:
This is not the solution.

PR #3118, opened 2022-01-21
Make organizations > access control permission-aware (part 2)

kuosandys:
Towards https://github.com/giantswarm/roadmap/issues/551.

This PR builds on https://github.com/giantswarm/happa/pull/2972 in making the access control UI aware of permissions for a read-only user. The UI adapts as follows for the following scenarios:

**User can read (but not edit) `Roles` and `RoleBindings` in the organization namespace, and can read `ClusterRoles`:**
A warning message is displayed, 'Add' buttons for each subject type and 'x' icons on subject items are removed.
Screen Shot 2022-01-21 at 19 29 26

**User can read and edit `Roles` and `RoleBindings` in the organization namespace, but cannot access `ClusterRoles`:**
If a `ClusterRole` is referenced in a `RoleBinding` when a user does not have permissions to read the `ClusterRole` itself, we display a "fake" `ClusterRole` in the roles list for completeness. We only have information for the role kind and name, so the "managed by" and role description information cannot be displayed and are omitted.

The 'Permissions' tab is also disabled (hovering displays the cursor in the `not-allowed` style).
Screen Shot 2022-01-21 at 19 35 20

**User can read `Roles` and `RoleBindings` in the organization namespace, but not `ClusterRoles`:**
This is a combination of the above scenarios: the user cannot bind subjects and also cannot access the 'Permissions' tab. The warning message indicates the combination of permission restrictions.
Screen Shot 2022-01-21 at 19 32 00

**Add warning message to CLI guide for binding roles to subjects:**
Screen Shot 2022-01-21 at 20 08 55


PrashantR31 (approved):
:+1:

PR #3124, opened 2022-01-24
Improve handling of fetch errors during login

kuosandys:
Towards https://github.com/giantswarm/giantswarm/issues/20464.

This PR improves the handling of fetch errors thrown while a user is logging into the application. While we can't reliably check for specific network errors (e.g. `net::ERR_CONNECTION_CLOSED`), these errors fall under the `TypeError` exception thrown by `fetch()`. We improve handling by including the original error message in the error reported to Sentry, and displaying a more appropriate error message to the user in this case during login:

Screen Shot 2022-01-25 at 09 51 27

One thing to bear in mind is that these errors can occur for a number of different reasons (more information can be found [here](https://developer.mozilla.org/en-US/docs/Web/API/fetch#exceptions)), so it wouldn't be appropriate to display a message that is too specific, and should rather prompt the user to reach out to support if this occurs during the login process.

### DIFF
--- a/src/model/clients/HttpClient.ts
+++ b/src/model/clients/HttpClient.ts
@@ -305,6 +305,15 @@ export class HttpClientImpl implements IHttpClient {
         return Promise.reject(res);
       }
 
+      if ((err as Error).name === 'TypeError') {
+        res.message = `Your request appears to be invalid: ${
+          (err as Error).message
+        }`;
+        res.headers = Object.assign({}, headers);
+
+        return Promise.reject(res);
+      }
+
       // We got a non-2xx status code.
       if (err instanceof Response) {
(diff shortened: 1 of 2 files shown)


marians:
Nice! I'd like to have something like "if this problem persists" in the message, to not trigger support queries for one-off issues.

kuosandys:
> Nice! I'd like to have something like "if this problem persists" in the message, to not trigger support queries for one-off issues.

@marians I've updated the error message to: "We couldn't execute a request. Please try again later, and contact support at support@giantswarm.io if the problem persists" (the screenshot above is also updated). What do you think?

marians (approved):
Great! Thank you

PR #3134, opened 2022-01-25
Add documentation for permissions

kuosandys:
Closes https://github.com/giantswarm/roadmap/issues/551.

This PR adds some documentation to our FE docs, and seeks to explain how permissions are fetched and used in happa. It also includes a table mapping user actions to the permissions needed to perform those actions.


PrashantR31 (approved):
LGTM, nice page @kuosandys

PR #3141, opened 2022-01-27
Improve app uninstallation confirmation info

kuosandys:
Closes https://github.com/giantswarm/roadmap/issues/691.

Currently, the only additional information we show for confirming an app uninstallation is the ambiguous phrase: "There is no undo." This PR seeks to improve the message shown, informing users about what happens to any configuration resources after uninstallation, and that the app can be easily re-installed .

Both ConfigMap and Secret are used to configure the app:
Screen Shot 2022-01-27 at 17 03 09

Just ConfigMap exists (the case where just Secret exists is similar):
Screen Shot 2022-01-27 at 17 05 34

No configurations exist:
Screen Shot 2022-01-27 at 13 43 15


marians:
How about a positive wording? Instead of "will not remove" we could use something like "will leave ... in place". Usually that's easier to digest.

marians (approved):

PR #3156, opened 2022-02-01
Fix default g8sControlPlane replicas for creating HA masters and remove setting AWSControlPlane AZ's

kuosandys:
Towards https://github.com/giantswarm/giantswarm/issues/20594.

This PR fixes the bug where clusters could not be created when the number of availability zones for an AWS installation was less than 3 (i.e. installations in the `cn-north-1` region).

What changed for cluster creation:
- we set the default number of replicas specified in `g8sControlPlane` to be 3 when the HA masters option is selected
- we remove the `availabilityZones` field from `AWSControlPlane` - this should be empty, and defaulted by aws-admission-controller.
- we send the creation request for `g8sControlPlane` before `AWSControlPlane`  

Regarding the first change: toggling between the two options (HA and single) already sets the number of replicas correctly - the bug was isolated to just the initial templating.

The changes of this PR were tested on `giraffe` - the built image is tagged as `1.35.8-f352d929768e72a37d281d636684de8693db1981` in the happa repository on aliyun.


src/components/MAPI/clusters/utils.ts
```diff
@@ -545,7 +538,7 @@ function createDefaultG8sControlPlane(config: {
       },
     },
     spec: {
-      replicas: config.awsControlPlane.spec.availabilityZones!.length,
+      replicas: Constants.AWS_HA_MASTERS_MAX_NODES,
```
kuosandys, line 541: This is now hard-coded to 3

src/components/MAPI/clusters/utils.ts
```diff
@@ -512,7 +505,7 @@ function createDefaultAWSControlPlane(config: {
       },
     },
     spec: {
-      availabilityZones: azs,
+      // availabilityZones is defaulted by aws-admission-controller
```
kuosandys, line 508: Removed `availabilityZones` - just adding a comment for our future selves

src/components/MAPI/clusters/CreateCluster/patches.ts
```diff
@@ -115,17 +112,8 @@ export function withClusterControlPlaneNodesCount(count: number): ClusterPatch {
...
-          controlPlaneNode.spec.replicas = count;
-          break;
-        case infrav1alpha3.AWSControlPlane:
-          controlPlaneNode.spec.availabilityZones = determineRandomAZs(
-            count,
-            supportedAZs
-          );
-          break;
+      if (controlPlaneNode.kind === infrav1alpha3.G8sControlPlane) {
+        controlPlaneNode.spec.replicas = count;
```
kuosandys, line 116: Nothing added, just removing the logic related to setting `availabilityZones` when HA/single masters is toggled

PrashantR31 (approved):
Thanks for solving this @kuosandys

marians (approved):

njuettner (approved):
lgtm 👍🏻

PR #3166, opened 2022-02-03
Prevent formatting invalid dates

kuosandys:
Closes https://github.com/giantswarm/giantswarm/issues/20635.

This PR adds a check before attempting to call the `utcToZonedTime()` function from the `date-fns-tz` library. This function calls the built-in `Intl.DateTimeFormat.prototype.formatToParts()` function, which will throw a RangeError if we pass it an Invalid Date object.

As an Invalid Date object is still an instance of Date, we also check whether the date's representation in ms is a finite number.


marians (approved):
TIL

PR #3205, opened 2022-02-18
(chore) bump architect-orb to v4.13.0

rossf7:
This is needed so we can remove the manifests dir from the app collections.

https://github.com/giantswarm/aws-app-collection/tree/master/manifests

This holds the Argo application CRs and is now obsolete.


rossf7:
Going with a ping

PR #3209, opened 2022-02-21
Access control: support adding subjects with names containing spaces

kuosandys:
Closes https://github.com/giantswarm/roadmap/issues/818.

This PR supports adding subject names containing spaces by removing whitespace characters from the allowed delimiters for adding multiple subjects in the access control UI.
Screen Shot 2022-02-21 at 16 17 21

Note that the support for adding names with spaces only applies to adding groups and users, where the allowed characters are determined by the identity providers. Allowed characters for `ServiceAccounts` don't include whitespace characters, so technically we could continue using them as delimiters. But to reduce the mental load for users having to remember different delimiters for different fields, we also remove whitespace characters as allowed delimiters here.


marians:
I agree that the handling of separators should be consistent between all subject types.

While supporting a semicolon as separator is nice in theory, I wouldn't mention it in the field instructions. The instructions for something so simple shouldn't become so complicated.

marians (approved):
Tested, works as advertised :)

PR #3214, opened 2022-02-23
cluster details > ingress tab: fix information displayed during cluster creation

kuosandys:
Closes https://github.com/giantswarm/roadmap/issues/819.

This PR fixes the bug in cluster details > ingress tab, where no useful information was being displayed after nginx IC app was installed.

The information we display in this tab is derived from the cluster's `spec.controlPlaneEndpoint.host` field, and this can be empty while a cluster is still being created, or if we haven't refreshed this data yet. To mitigate this, we refetch the cluster data after the Nginx IC app is installed to make sure we have updated information. We also display loading placeholders, as this is more appropriate than displaying empty fields:
Screen Shot 2022-02-23 at 14 43 38

While testing out the changes of this PR, I also encountered the scenario when the cluster namespace has not been created yet, and thus app installation is not yet possible. In this case, we return a more specific error message informing users of the reason why the nginx IC app cannot be installed yet:
Screen Shot 2022-02-28 at 11 07 32


marians:
How load would the placeholders stay visible in practice? I think if it's rather minutes than seconds, then it's not the right pattern for the case.

5 to 10 minutes sounds a bit much though. The cluster namespace hopefully would only take a few seconds to be created.

kuosandys:
> How load would the placeholders stay visible in practice? I think if it's rather minutes than seconds, then it's not the right pattern for the case.

When I was testing, refetching the cluster data after nginx IC installation was enough to ensure we have the updated controlplaneEndpoint field. I thought about using placeholders if fetching the information is slower than re-rendering the page, but we could also display them as N/A if it’s more appropriate.

> 5 to 10 minutes sounds a bit much (for the missing namespace). The cluster namespace hopefully would only take a few seconds to be created.

It was around a minute at most when I was testing on gauss. I reused the same message we display when installing an app from the app catalog, but I’d be happy to align the message in both places to something more realistic 👍

marians (approved):

PR #3228, opened 2022-03-01
Don't rely on `Rolebinding.subjects` being defined

kuosandys:
Closes https://github.com/giantswarm/giantswarm/issues/21113, related to the changes in `rbac-operator` from https://github.com/giantswarm/rbac-operator/pull/292.

This PR fixes the error we saw on `gauss` while testing the new changes in `rbac-operator`. It is possible for a `RoleBinding` to exist, and not be bound to any subjects, and the `subjects` field will be omitted if it's empty. We update the interfaces for `IClusterRoleBinding` and `IRoleBinding` to take this into account, and add some checks to prevent us from trying to use the `undefined` value.

### DIFF
--- a/src/model/services/mapi/rbacv1/types.ts
+++ b/src/model/services/mapi/rbacv1/types.ts
@@ -50,7 +50,7 @@ export interface IClusterRoleBinding {
   kind: typeof ClusterRoleBinding;
   metadata: metav1.IObjectMeta;
   roleRef: IRoleRef;
-  subjects: ISubject[];
+  subjects?: ISubject[];
 }
 
 export const ClusterRoleBindingList = 'ClusterRoleBindingList';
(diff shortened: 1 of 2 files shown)


src/model/services/mapi/rbacv1/types.ts
```diff
@@ -84,7 +84,7 @@ export interface IRoleBinding {
   kind: typeof RoleBinding;
   metadata: metav1.IObjectMeta;
   roleRef: IRoleRef;
-  subjects: ISubject[];
+  subjects?: ISubject[];
```
kuosandys, line 87: From what I understand [here](https://github.com/kubernetes/api/blob/b8c40e080bc5e830097df540d4ef804034cb5bdb/rbac/v1/types.go#L131), `subjects` can be omitted if empty

marians (approved):

PR #3251, opened 2022-03-10
access control: improve roles list

kuosandys:
Closes https://github.com/giantswarm/roadmap/issues/656.

This PR brings some improvements to the roles list in the access control UI to make it easier to see which roles are bound in the namespace. The following changes have been made:
- display number of service accounts bound to a role
- subject labels (i.e. "Groups", "Users", "Service Accounts") have been replaced by subject icons, with a tooltip displaying the label on hover
- display subject icon and count in a "lighter" colour if there are no subjects of that type bound to a role
- briefly highlight the subject icon when the count changes

Screen Shot 2022-03-10 at 16 26 36 PM

Hovering over a role list subject icon:
Screen Shot 2022-03-10 at 16 27 10 PM

In addition, a small bug was fixed to prevent a newly added subject from being repeatedly highlighted when toggling between items in the roles list.

### DIFF
--- a/src/components/UI/Display/MAPI/AccessControl/AccessControlRoleListItem.tsx
+++ b/src/components/UI/Display/MAPI/AccessControl/AccessControlRoleListItem.tsx
@@ -1,6 +1,7 @@
 import { Box, Card, CardBody, CardHeader, Heading, Text } from 'grommet';
 import * as React from 'react';
 import styled from 'styled-components';
+import RefreshableLabel from 'UI/Display/RefreshableLabel';
 import { Tooltip, TooltipContainer } from 'UI/Display/Tooltip';
 
 import { IAccessControlRoleItem } from './types';
@@ -126,8 +127,10 @@ const AccessControlRoleListItem = React.forwardRef<
                 size='small'
                 aria-label={`Groups: ${groupCountFormatted}`}
               >
-                <i className='fa fa-group' role='presentation' />{' '}
-                {groupCountFormatted}
+                <RefreshableLabel value={groupCount}>
+                  <i className='fa fa-group' role='presentation' />{' '}
+                  {groupCountFormatted}
+                </RefreshableLabel>
               </Text>
             </TooltipContainer>
             <TooltipContainer
@@ -142,8 +145,10 @@ const AccessControlRoleListItem = React.forwardRef<
                 size='small'
                 aria-label={`Users: ${userCountFormatted}`}
               >
-                <i className='fa fa-user' role='presentation' />{' '}
-                {userCountFormatted}
+                <RefreshableLabel value={userCount}>
+                  <i className='fa fa-user' role='presentation' />{' '}
+                  {userCountFormatted}
+                </RefreshableLabel>
               </Text>
             </TooltipContainer>
             <TooltipContainer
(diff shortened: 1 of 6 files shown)


src/components/UI/Display/MAPI/AccessControl/AccessControlRoleListItem.tsx
```diff
@@ -49,53 +48,117 @@ interface IAccessControlRoleListItemProps
...
+                </Text>
+              )}
+            </Box>
+          </StyledHeading>
+        </CardHeader>
+        {hasBoundSubjects && (
+          <CardBody margin={{ top: 'xsmall' }}>
+            <Box direction='row'>
+              {groupCount > 0 && (
+                <Box width='30%'>
```
gusevda, line 110: What do you think about getting rid of these wrapper elements `<Box width='30%'>`? We can separate children with some fixed gap, for example. I think it could make the layout more independent of the childs count.

gusevda:
I like the idea of hiding a `<CardBody />` when there are no bound subjects.

marians:
Apart from what we spoke about, one other thing came to my mind:

For the tooltip, how about a slightly longer text (which would explain a bit more): Instead of just `Groups` and `Users` (for example), show e. g.

- `This cluster role is bound to 1 group`
- `This role is bound to 2 users`

src/components/UI/Display/RefreshableLabel.js
```diff
@@ -51,7 +51,7 @@ function RefreshableLabel({ children, value, className }) {
   useEffect(() => {
     const sToMs = 1000;
     const compareData = () => {
-      if (prevValue && value !== prevValue) {
+      if (typeof prevValue !== 'undefined' && value !== prevValue) {
         setHasDataChanged(true);
       }
```
kuosandys, line 56: `0` is being coerced to `false`, and subject counts changing from `0` to `1` was not appropriately being highlighted. I think by checking for `undefined` is more appropriate to check whether there is a previous value.

kuosandys:
> I like the idea of hiding a `<CardBody />` when there are no bound subjects.

Unfortunately it was looking a little messy to have items of varying height in the roles list 😅 In the end it looked better to keep everything at a consistent height so as to not increase mental load for users.

gusevda (approved):
Looks great!

I think we can create a separate ticket to refactor places in the code where we pluralize labels. There are several places in the project and it's done slightly different everywhere.

marians:
Everything we talked about is working as expected. However, I tested again and found more things I'd like to discuss.

- When I add a subject to a role, we show a flash success message. This is actually distracting and probably not necessary I think.
- The dark blue scroll bar I see (when using a touch pad) is "touching" the list items. I think it would look better with a few pixels of margin between the item border and the scroll bar.

image

marians (approved):
BTW, my previous comment should not block this PR in any way

kuosandys:
> When I add a subject to a role, we show a flash success message. This is actually distracting and probably not necessary I think.
> The dark blue scroll bar I see (when using a touch pad) is "touching" the list items. I think it would look better with a few pixels of margin between the item border and the scroll bar.

I think these are good things to consider, and I'd like to do so in more depth (e.g. we also show flash messages on subject deletion, so it would be nice to think about these in conjunction, and we use custom scrollbars on some other components that could be aligned better).

I'll create separate issues for these two 👍

PR #3252, opened 2022-03-10
Align managed app catalog label

gusevda:
Towards [giantswarm/roadmap/issues/825](https://github.com/giantswarm/roadmap/issues/825).

The "managed" label in the "Installed Apps" section has different style from the similar labels in other places. In this PR this label was changed so now it's being rendered by the same component that we use for other labels.

Before:
Screenshot 2022-03-10 at 15 40 12

After:
Screenshot 2022-03-10 at 15 39 53

### DIFF
--- a/src/components/UI/Display/Apps/AppList/CatalogLabel.tsx
+++ b/src/components/UI/Display/Apps/AppList/CatalogLabel.tsx
@@ -60,18 +60,19 @@ const CatalogLabel: React.FC<ICatalogLabelProps> = (props) => {
   const text = (
     <span>
       {props.catalogName}
-      {props.isManaged && <CatalogType>MANAGED</CatalogType>}
-      {props.catalogName === 'Giant Swarm Catalog' && (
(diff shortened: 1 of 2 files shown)


src/components/UI/Display/Apps/AppList/CatalogLabel.tsx
```diff
@@ -60,18 +60,19 @@ const CatalogLabel: React.FC<ICatalogLabelProps> = (props) => {
   const text = (
     <span>
       {props.catalogName}
-      {props.isManaged && <CatalogType>MANAGED</CatalogType>}
-      {props.catalogName === 'Giant Swarm Catalog' && (
+      {props.isManaged || props.catalogName === 'Giant Swarm Catalog' ? (
         <CatalogType>MANAGED</CatalogType>
```
kuosandys, line 64: Nice 👍

kuosandys (approved):
Looks good!

PR #3268, opened 2022-03-21
Fix extra LOGIN_SUCCESS redux actions

gusevda:
Towards [giantswarm/roadmap/issues/870](https://github.com/giantswarm/roadmap/issues/870).

In our `OAuth2 Provider` we have `getLoggedInUser` method that returns `User` object for the currently authenticated user. This method is being called a lot, at least once for every HTTP request to ensure that that the user is authenticated. Right now, probably unintentionally, on each `getLoggedInUser` call we dispatch `LOGIN_SUCCESS` redux action which causes a lot of unnecessary rerenders.  It happens here - 
1) `getLoggedInUser` call [fires](https://github.com/giantswarm/happa/blob/main/src/utils/OAuth2/OAuth2.ts#L158) `UserLoaded` event;
2) `LOGIN_SUCCESS` action is being [dispatched](https://github.com/giantswarm/happa/blob/main/src/components/Auth/MAPI/MapiAuthProvider.tsx#L42) in the `MapiAuthProvider`.

In this pull request I changed `getLoggedInUser` method so it fires `UserLoaded` event only if the current session is expired.

Redux actions on page load before the change:
Screenshot 2022-03-22 at 10 46 23

After the change:
Screenshot 2022-03-22 at 10 47 45


kuosandys (approved):
Nice 👍

PR #3277, opened 2022-03-23
Cluster list loading behaviour

gusevda:
Towards [giantswarm/giantswarm/issues/17708](https://github.com/giantswarm/giantswarm/issues/17708).

Changes in this PR:

**1. Cluster list preloader was changed from "item placeholders" to a spinner**
Reasons for this change:
- we can not know in advance how many placeholders to display;
- after loading, not only list of clusters can be displayed, but also "No clusters" or "Error" view.

Before:
Screenshot 2022-03-23 at 14 15 46

After:
Screenshot 2022-03-23 at 14 18 01

**2. Cluster list transitions between different states were improved**

Initial loading of the clusters page
1. preloader appears without transitions
2. list of clusters appears with the slide transition

Comming back to the clusters page from another page
1. list of clusters displayed immediately without any transitions

Switching to another organization
1. list of clusters disappears with the slide transition
2. preloader appears with the fade-in transition if the loading takes more than 300ms
3. new list of clusters appears with the slide transition

Switching to another organization where list of clusters was loaded before
1. list of clusters disappears with the slide transition
2. new list of clusters appears with the slide transition

New cluster added/deleted
1. only one cluster item appears/disappears with the slide transition

No clusters view
1. appears/disappears with the fade-in/fade-out transition


src/components/MAPI/clusters/Clusters.tsx
```diff
@@ -89,62 +60,48 @@ const Clusters: React.FC<{}> = () => {
...
-    data: providerClusterList,
-    error: providerClusterListError,
-    isValidating: providerClusterListIsValidating,
-  } = useSWR<IProviderClusterForClusterName[], GenericResponseError>(
-    providerClusterKey,
+  } = useSWR<IClusterListWithProviderClusters, GenericResponseError>(
+    clusterListKey,
     () =>
-      fetchProviderClustersForClusters(clientFactory, auth, clusterList!.items)
+      fetchClusterListWithProviderClusters(
```
kuosandys, line 69: We're storing the response of `fetchClusterListWithProviderClusters()` with the key returned by `fetchClusterListKey()`. I think this could create problems when we try to access the cached data in other parts of the app, because the responses of `fetchClusterListWithProviderClusters()` and `fetchClusterList()` have different interfaces. Could we create a different key for this new fetch function?

gusevda: good find, will fix

src/components/MAPI/utils.ts
```diff
@@ -348,6 +348,37 @@ export function fetchClusterKey(
...
+  namespace?: string,
+  organization?: string
+): Promise<IClusterListWithProviderClusters> {
+  const clusterList = await fetchClusterList(
+    httpClientFactory,
+    auth,
+    provider,
+    namespace,
+    organization
+  );
```
kuosandys, line 369: Small suggestion here - since we call this function in other places in the app, maybe we could leverage `swr`'s cache here as well?

src/components/MAPI/clusters/Clusters.tsx
gusevda: I decided to go back to two separate `useSWR` calls to keep consistency with other places. I had to change the way we map results of two calls - I added `clustersRef` ref where I store the previous value and return it when the first fetch has finished and the second one is still in progress.

src/components/MAPI/utils.ts
gusevda: good point, but not actual anymore since I returned to the previous logic to fetch cluster list and provider clusters separately.

kuosandys (approved):
Really nice improvement! LGTM 👍

PR #3282, opened 2022-03-24
Improve experimental impersonation experience for staff

kuosandys:
Closes https://github.com/giantswarm/roadmap/issues/823.

This PR makes a small improvement towards making the impersonation experience more accurate. Note that this feature is currently experimental, and available to GS staff only.

In parts of the UI, certain elements are displayed (or displayed differently) to users who are GS staff. An example is the list of app catalogs: GS staff can see all catalogs, while non-GS staff can only see the Giant Swarm and Playground catalogs.

**How do we know if a user is GS staff**
We determine whether the current logged in user is GS staff by checking if they belong to the `kubernetes.auth.controlPlaneAdminTargetGroup` group defined in the `config` repo.

**Current state of the impersonation feature**
A caveat with the impersonation feature is that we currently don't take this into account to show/hide elements: in other words, GS staff impersonating someone who is not GS staff is still able to see these "GS staff-only" elements. This means the UI under impersonation is not displayed accurately, and IMO this affects the overall impersonation experience.

**What changes are made by this PR**
This PR makes some changes to improve this. First, we store the impersonation state in the Redux store. We update it when the app loads, and when impersonation is set/cleared. This is to prevent reading from `localStorage` too often, and to prevent having to call and await the asynchronous method `auth.getImpersonationMetadata` repeatedly. We follow a similar logic with getting the current logged in user from `auth`, and loading it into Redux store.

Then, we can use this impersonation state throughout the app to determine if we are currently impersonating someone who is not GS staff, and to appropriate hide UI elements that should not be visible to them. These elements are:
- the app catalog list (only the Giant Swarm and Playground catalogs are visible to non-GS staff)
- in the cluster creation form, the releases list (non-GS staff do not see WIP and deprecated releases, and they do not see the warning message about this)
- cluster upgradable status (this is affected by the available releases for upgrade for staff v.s. non-staff)


marians:
Sorry, can't review this

kuosandys:
> Sorry, can't review this

@marians I'm sorry to hear that. Is there an improvement you have in mind for making this PR more reviewable, or should I be asking for reviews elsewhere? Team rainbow was automatically pinged as codeowners 🙂

src/model/stores/main/actions.ts
```diff
@@ -303,6 +328,11 @@ export function resumeLogin(
       user = mapOAuth2UserToUser(mapiUser);
       dispatch(loginSuccess(user));
 
+      const metadata = await auth.getImpersonationMetadata();
+      if (metadata) {
+        dispatch(setImpersonation(metadata));
```
gusevda, line 333: I'm not sure, but maybe we also need to handle impersonation in this block above?
```
// Try to resume GS user first.
user = fetchUserFromStorage();
if (user) {
      dispatch(loginSuccess(user));
      return Promise.resolve(user);
}
```

src/components/Auth/MAPI/MapiUnauthorized.tsx
```diff
@@ -80,6 +81,8 @@ const MapiUnauthorized: React.FC<IMapiUnauthorizedProps> = ({
       messageTTL.MEDIUM
     );
 
+    dispatch(clearImpersonationAction());
+
     reloadOrganizations();
   };
```
gusevda, line 88: In this component we use local state for the impersonation metadata
```
const [impersonationMetadata, setImpersonationMetadata] =
useState<IOAuth2ImpersonationMetadata | null>(null);
```

Maybe we can take it from the redux store now?

src/components/MAPI/Experiments.tsx
```diff
@@ -110,6 +117,8 @@ const Experiments: React.FC<IExperimentsProps> = () => {
       messageTTL.MEDIUM
     );
 
+    dispatch(clearImpersonation());
+
     reloadOrganizations();
   };
```
gusevda, line 124: Local state values in this component - `impersonationUser`, `impersonationGroup` - maybe it's better to take them from redux store instead of from auth provider?
```
const auth = useAuthProvider();
const [impersonationUser, setImpersonationUser] = useState<string>('');
const [impersonationGroup, setImpersonationGroup] = useState<string>('');

useEffect(() => {
    (async () => {
      const metadata = await auth.getImpersonationMetadata();

      setImpersonationUser(metadata?.user ?? '');
      setImpersonationGroup(metadata?.groups?.[0] ?? '');
    })();
}, [auth]);
```

PrashantR31:
> 



> > Sorry, can't review this
> 
> @marians I'm sorry to hear that. Is there an improvement you have in mind for making this PR more reviewable, or should I be asking for reviews elsewhere? Team rainbow was automatically pinged as codeowners slightly_smiling_face

Let us do it this way:

gusevda (approved):
Looks great! 👍

marians:
@kuosandys 

> I'm sorry to hear that. Is there an improvement you have in mind for making this PR more reviewable ...

Don't worry. Just wanted to tell you that I feel I can't contribute here, as it's a "behind the scenes" thing.

PR #3298, opened 2022-03-30
Fix permissions list scrolling

gusevda:
Towards [giantswarm/roadmap/issues/820](https://github.com/giantswarm/roadmap/issues/820).

In the permissions tab we use the `<InfiniteScroll>` component from `Grommet` to lazy-render the items. It's currently configured to remove "off-screen" items. Since the permissions list is not inside a scrollable container (window scroll used instead) `<InfiniteScroll>` component removes items while they are still visible. As a result we can see empty gaps.
154312465-86183a09-6e05-46a8-837a-686c856d05e2

In this PR `replace` feature was disabled to improve scrolling experience. This should not bring any performance issues since the items we render in the permissions list are simple and not very interactive. From the `Grommet` [docs](https://v2.grommet.io/infinitescroll#replace):
> `replace` should be set to true within Drop containers and false otherwise.

### DIFF
--- a/src/components/UI/Display/MAPI/AccessControl/AccessControlRolePermissions.tsx
+++ b/src/components/UI/Display/MAPI/AccessControl/AccessControlRolePermissions.tsx
@@ -84,7 +84,6 @@ const AccessControlRolePermissions: React.FC<
       </TableHeader>
       <TableBody>
         <InfiniteScroll
-          replace={true}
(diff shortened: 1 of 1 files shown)


marians (approved):
LGTM

kuosandys (approved):
Nice 👍

PR #3299, opened 2022-03-30
Don't rely on `apiVersion` to distinguish between provider-specific resources

kuosandys:
Towards https://github.com/giantswarm/roadmap/issues/953.

This PR changes how we determine which provider-specific action to take (e.g. fetching additional resources). Previously, we checked the resource's `apiVersion`, which contains the resource group name and the API version (e.g. `cluster.x-k8s.io/v1alpha3`). However, this made us reliant on the specific API version specified. In addition, the CAPI-related resources have the same resource group name, so relying on `apiVersion` alone to determine provider-specific resources is not viable in the long run.

Similar to an earlier [PR](https://github.com/giantswarm/happa/pull/3173), our utility functions are refactored so that we check resources by their `kind`:
- for Azure: `AzureCluster`, `MachinePool`, `AzureMachinePool`
- for AWS: `AWSCluster`, `MachineDeployment`, `AWSMachineDeployment`

Note that instead of checking against lowercased string literals as we did in the previous PR, we now check against exported constant variables from our MAPI services. This is to guard against typos, and to leverage TS type-narrowing to reduce errors.


gusevda (approved):
LGTM, great improvement!

PR #3300, opened 2022-03-30
Keep happa available during upgrades

marians:
Closes https://github.com/giantswarm/roadmap/issues/972

This PR

- Fixes the anti-affinity selector that should avoid all pods running on the same node
- Adds a `PodDisruptionBudget` template with `minAvailable: 1`


whites11 (approved):
Thanks Marian <3

PR #3303, opened 2022-03-31
Add upgrade cluster cli guide

gusevda:
Towards [giantswarm/roadmap/issues/574](https://github.com/giantswarm/roadmap/issues/574).

Upgrade cluster CLI guide was added to the cluster details page. The guide is visible only when a new version is available - the logic is the same as for the "Upgrade cluster…" button in the RELEASE section.

Screenshot 2022-04-01 at 14 16 23


kuosandys (commented):
In our CLI guides, we display a warning to the user if they don't have sufficient permissions to perform the actions listed. For example:
Screen Shot 2022-03-31 at 17 35 21

Would it be possible to display a warning in this guide if the user does not have the permissions to upgrade the cluster? Let me know if I could clarify anything about this.

gusevda:
> Also, I was wondering if we could improve the consistency of the command's formatting. We haven't defined a standard, but I noticed we use this format in many other places... 
> 
> ...where `--context` is on the same line as `kubectl ...` and the resource name is on the same line as the resource.
> 

@kuosandys @marians what do you think?
Screenshot 2022-04-01 at 10 38 24

marians:
That looks cleaner!

marians (approved):

gusevda:
> We also have the [kubectl gs update cluster](https://docs.giantswarm.io/ui-api/kubectl-gs/update-cluster/) command. Do we want to provide a link/reference to that?

@marians what do you think about it?

marians:
🙈 Oh jeez, I totally forgot. Yes, I'd say that's the canonical way of upgrading a cluster via the CLI.

kuosandys (approved):

kuosandys (approved):
Thanks for updating the command! LGTM 👍

PR #3314, opened 2022-04-04
Set cluster labels via the Management API: use `kubectl label` command

marians:
This simplifies the MAPI guide by using the `label` command instead of `patch`. Also the placeholders are turned into uppercase letters.

Also I'm removing the conditional for the resource name, as `clusters.cluster.x-k8s.io` works on all providers and is more robust than relying on `clusters`.


PrashantR31 (approved):
:+1:

src/components/MAPI/clusters/guides/SetClusterLabelsGuide.tsx
```diff
@@ -65,19 +65,14 @@ const SetClusterLabelsGuide: React.FC<ISetClusterLabelsGuideProps> = ({
...
-              provider === Providers.AWS
-                ? 'clusters.cluster.x-k8s.io'
-                : 'cluster'
-            } ${clusterName} \\
+            label clusters.cluster.x-k8s.io ${clusterName} \\
             --namespace ${clusterNamespace} \\
-            --type merge \\
-            --patch '{"metadata": {"labels": {"foo": "bar"}}}'
+            LABEL=VALUE'
           `}
```
kuosandys, line 71:
```suggestion
            LABEL=VALUE
          `}
```

marians: Good catch, thanks!

kuosandys (approved):

marians:
@kuosandys Could you please check whether I have to update a test here? It doesn't appear related to me.

PR #3322, opened 2022-04-06
Fix token refreshing request issued twice with the same refresh token

kuosandys:
Closes https://github.com/giantswarm/giantswarm/issues/21489.

This PR fixes the bug where token refreshing requests were being issued for a second time after 1 second with the same refresh token, causing users to be redirected back to the login page unnecessarily. This was only happening when specific conditions were met, and can be fixed by handling token renewal manually to ensure that only one request is made with one refresh token.

**Why did this happen - the short version:**
The extra token refreshing request was caused by getting the current logged in user immediately after token refreshing has been initiated, but not completed. This results in the `AccessTokenEvents` class from the `oidc-client-js` library to start a countdown timer of 1 second to trigger renewal at the next second.

<details><summary><strong>Why did this happen: the long version:</strong></summary>

- The first call was initiated by `SilentRenewalService` from the `oidc-client` library, which helps us handle token renewal in the background. It works by initiating a timer to when the token is expiring (60 seconds before token is expired), and checks on an interval (5 seconds) for when the timer runs out. When it does, `UserManager.signInSilent()` is called to renew the token. This was enabled by setting `automaticSilentRenew` to true when we instantiated the `UserManager` class.
- The second call was initiated by `ensureClientAuth()`, which is a check we run on every fetch request to ensure "that the user is authenticated, their token is not expired, and that the given HTTP client uses the authentication token." We get the current logged in user via `OAuth2.getLoggedInUser()`, which calls `UserManager.getUser()`, loading the user found in storage. The user loaded event triggers `AccessTokenEvents.load()`, which will initiate a timer counting down to token expiring if there is still time to renew the user's token (i.e. the stored user's `expires_in` field is greater than 0). If at the time of evaluation, there is less than the default 60 seconds to renew the token, a timer is initiated for 1 second, **triggering renewal at the next second**.
- If the second call is initiated before the completion of the first, the user loaded from storage will still have the old `expires_in` data, which is now less than 60 seconds away, triggering a token renewal at the next second with the same refresh token. This call will then fail, because `dex` only allows a refresh token to be claimed once.
</details>

**Why doesn't this happen all the time?**
This involves a bit of convenient timing: if we make fetch requests while the token refreshing request has been initiated but not completed yet (and the refresh token in storage has not been updated), then the subsequent token refreshing requests will fail until the token is updated, and we make requests using the updated token. For example, after being directed to the `/login` page, navigating to `/` triggers fetch requests using the new token. Fetch requests being made to auto-refresh data on the page would also trigger a renewal with the new token, if they were made after the token has been updated. This is dependent on network speed, as well as the timing of the fetch requests.

**How can I recreate this error?**
Check out the `main` branch, set the `PERMISSIONS_REFRESH_INTERVAL` to a large number (e.g. 50 mins.). Call `OAuth2.getLoggedInUser()` at the end of `OAuth2.onAccessTokenExpiring()`, and throttle network speed. Wait for the access token to expire.


marians (approved):
I'm truly amazed!

PR #3329, opened 2022-04-08
Auto refresh relative dates

gusevda:
Towards [giantswarm/roadmap/issues/971](https://github.com/giantswarm/roadmap/issues/971).

We have a `<Date />` component that displays a date in a relative form, e.g. 5 minutes ago, or 1 year ago. After being displayed, relative date stays the same until next rerender - it's unpredictable when a rerender can happen or it can not happen at all. As a result relative dates can be outdated. In this PR I changed the `<Date />` component so it updates the displayed value every minute. "Auto refresh" happens only for dates that are in 45 minutes from current time, because after 45 minutes, formatted date looks like `about 1 hour`, `about [2..24] hours`, `1 day` and so forth and it's not necessary to auto refresh it anymore.

Some other small changes:
- `<Date />` component was reused in some other places where we display relative dates;
- cluster creation date on the cluster details page was capitalized

### DIFF
--- a/src/components/MAPI/clusters/ClusterDetail/ClusterDetailWidgetCreated.tsx
+++ b/src/components/MAPI/clusters/ClusterDetail/ClusterDetailWidgetCreated.tsx
@@ -10,6 +10,12 @@ const StyledDot = styled(Dot)`
   padding: 0;
 `;
 
+const CapitalizedText = styled.div`
+  &:first-letter {
+    text-transform: uppercase;
+  }
+`;
+
 interface IClusterDetailWidgetCreatedProps
   extends Omit<
     React.ComponentPropsWithoutRef<typeof ClusterDetailWidget>,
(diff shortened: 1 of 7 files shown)


src/components/MAPI/clusters/ClusterDetail/ClusterDetailWidgetCreated.tsx
```diff
@@ -10,6 +10,12 @@ const StyledDot = styled(Dot)`
   padding: 0;
 `;
 
+const CapitalizedText = styled.div`
+  &:first-letter {
+    text-transform: uppercase;
+  }
+`;
```
kuosandys, line 17: Could we just style the `<Date>` component here? Just a small nitpick, but I was wondering if we could avoid another wrapper 😅

kuosandys (approved):
Really awesome, and aligns really well with other "auto-refreshing" info we display on the page 👍 LGTM if you're happy - my suggestion above should not be a blocker.

marians (approved):

PR #3338, opened 2022-04-12
App catalog: display app installation status for current selected cluster

kuosandys:
Closes https://github.com/giantswarm/roadmap/issues/785.

This PR brings some changes to the app catalog page and app details page in order to communicate whether an app is already installed in the selected cluster, if any has been selected.

**How is a cluster selected?**
A cluster is selected when a user clicks on the "Install app" button in a cluster's details page to navigate to the app catalogs page. A cluster is also selected when a user clicks on the "Install in (this) cluster" button in an app's details page and selects a cluster from the list. This logic was already in place, and remains unchanged.

**App details page:**
If there is a cluster selected, we display the selected cluster and the install button text changes to  "Install in **this** cluster". Clicking on the "x" deselects the cluster.
Screen Shot 2022-04-13 at 12 03 23

If the app is already installed in the selected cluster, we display the appropriate "Installed in cluster" message instead of the "Install in this cluster" button. Deselecting the cluster would restore the "Install in cluster" button.
Screen Shot 2022-04-13 at 12 02 54

If there is no cluster selected, or if the user clicks on "Pick another cluster" from the installation modal, we display a list of clusters. If the app is already installed in a cluster, the cluster picker item is displayed as disabled.
Screen Shot 2022-04-12 at 11 51 56

**App catalogs page**
If there is a cluster selected, we display the selected cluster, as well as a checkmark on each app if it is already installed on the cluster.
Screen Shot 2022-04-13 at 11 56 13


marians (approved):
Awesome!

src/components/UI/Display/Apps/AppList/AppsListPage.tsx
```diff
@@ -11,7 +11,9 @@ import Toolbar from './Toolbar';
 
 const LOADING_COMPONENTS = new Array(9).fill(0);
 
-const Wrapper = styled.div``;
+const Wrapper = styled.div`
+  margin-top: -10px;
```
gusevda, line 15: Is it possible to not have negative margin here? As I understand it is here to compensate `<AppInstallationSelectedCluster/>` component's `margin-top`, right? But maybe we can just remove `top` property [here](https://github.com/giantswarm/happa/pull/3338/files#diff-4b91d69bf401498c7d748bed07fabed1e6dbc800b819e880d255bd65811aa580R314)?

gusevda (commented):
Two small comments, otherwise looks good to me. Great job!

src/components/MAPI/apps/AppList/__tests__/AppDetail.tsx
```diff
@@ -132,4 +148,108 @@ describe('AppDetail', () => {
...
+      screen.getAllByLabelText('Loading...')
+    );
+
+    expect(
+      withMarkup(screen.getByText)(
+        ' Your selected cluster is test1 and install status will be shown for this cluster.'
+      )
+    ).toBeInTheDocument();
+  });
 });
```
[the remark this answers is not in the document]
kuosandys: Oh no they should definitely be different 😅 I'll finish the test case for deselecting the cluster

src/components/UI/Display/Apps/AppList/AppsListPage.tsx
kuosandys: I used it to try and even out the vertical margins when there's no selected cluster banner - this is how it would look without the negative margin:
Screen Shot 2022-04-13 at 15 43 39
The position of `<Wrapper>` is influenced by the padding on the `<main>` element (120px).

Another approach I was thinking of was to increase the bottom margin of the text instead (to 30px), but only when there's no selected cluster banner:
Screen Shot 2022-04-13 at 15 55 35
Removing the `margin-top` here:
Screen Shot 2022-04-13 at 15 57 41

What do you think?

PR #3343, opened 2022-04-13
Improve build time in CI

kuosandys:
Towards https://github.com/giantswarm/roadmap/issues/1042.

This PR makes some changes in an effort to reduce build time/credit usage in CircleCI:
- only build and push fe-docs when creating a release
- combine pushing to docker and quay with `build-container` step to prevent building image and attaching workspace multiple times (attaching the workspace takes up most of the time in these steps)
- combine building and pushing fe-docs to quay to prevent building image and attaching workspace twice

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -57,6 +57,7 @@ workflows:
             - gather-facts
           <<: *job_filters
       - build-container:
+          context: architect
           requires:
             - test
             - lint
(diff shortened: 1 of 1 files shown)


kuosandys:
It hasn't run enough times to produce a reliable average for the duration, but it looks a bit faster than the 11m35 average from the last 30 days.
Screen Shot 2022-04-13 at 23 08 12

marians:
> It hasn't run enough times to produce a reliable average for the duration, but it looks a bit faster than the 11m35 average from the last 30 days.

Regarding the cost/credits, the total duration from start to finish is not that important. More important is the sum of the individual steps, with some happening in parallel. And with regard to that, it looks like this PR is already a great improvement!

I would also suggest to merge the three short steps prettier, typecheck, and lint into one. Maybe also test.

Look at prettier:

image

More than half of the duration is spent on overhead (getting ready).

So if we merge these, we can save several times the overhead.

As a result, of course, these would fail fast. That means: If your branch has a problem with style, it will fail in the prettier step. And you won't find out until the next run whether tests succeed or not. This could potentially become very annoying if each step has something to complain about.

marians (approved):

gusevda (approved):
Looks good

kuosandys:
> > It hasn't run enough times to produce a reliable average for the duration, but it looks a bit faster than the 11m35 average from the last 30 days.
> 
> Regarding the cost/credits, the total duration from start to finish is not that important. More important is the sum of the individual steps, with some happening in parallel. And with regard to that, it looks like this PR is already a great improvement!
> 
> I would also suggest to merge the three short steps prettier, typecheck, and lint into one. Maybe also test.
> 
> Look at prettier:
> 
> image
> 
> More than half of the duration is spent on overhead (getting ready).
> 
> So if we merge these, we can save several times the overhead.
> 
> As a result, of course, these would fail fast. That means: If your branch has a problem with style, it will fail in the prettier step. And you won't find out until the next run whether tests succeed or not. This could potentially become very annoying if each step has something to complain about.

Thanks for the suggestions! I think with the pre-commit hook already running prettier, linting, and typecheck, we can combine these steps together. If any of these fail in CI, it would most likely be due to not having these dependencies up to date locally.

I'd keep test as a separate step for now, as these are run separately and can fail due to other reasons.

kuosandys:
Screen Shot 2022-04-14 at 16 23 07

PR #3349, opened 2022-04-14
Limit renovate bot number of concurrent PRs to 1

gusevda:
Renovate bot is configured to create 3 parallel PRs. After merging a PR, we need to update other branches with the base branch, which triggers another builds on CI. So we have several unnecessary CI builds every day.

In this PR I propose to limit amount of PRs created to only one. Once first PR is merged we can go to the Dependency Dashboard and trigger other PR's creation manually, one after one.


marians (approved):
This means that we have to take care that there is no stale renovate bot PR open, as it would block other dependency updates.

gusevda:
> This means that we have to take care that there is no stale renovate bot PR open, as it would block other dependency updates.

What do you mean by other dependency updates?

marians:
I mean: If we have one renovate PR open and don't get it merged, the list of dependencies to get updated will grow and no new PRs will be created.

gusevda:
Sure, it also applies to the limit of 3. We deal with the dependencies every day, so shouldn't be a problem.

PR #3362, opened 2022-04-19
Improve upgrade cluster CLI guide

gusevda:
In this PR CLI guide for upgrading a cluster was changed:
- note about `--scheduled-time` modifier was added;
- link to the [docs](https://docs.giantswarm.io/ui-api/kubectl-gs/update-cluster/) was added.

Screenshot 2022-04-19 at 17 25 02


marians:
Nice! In the text I would replace "should be" with "must be".

marians (approved):

PR #3363, opened 2022-04-19
Check installed app's `.spec.name` against app catalog entry name

kuosandys:
Closes https://github.com/giantswarm/roadmap/issues/1052.

This PR fixes the bug described in the linked issue, where an installed app was not being appropriately displayed in the app catalogs. This was caused by matching the App CR's `.metadata.name` instead of its `.spec.name` against the AppCatalogEntry's name.

The mock AppCatalogEntry list items data we use in our tests was also updated to match the data we would expect to receive in a real call, and the affected test cases updated.


marians (approved):
Awesome!

gusevda (approved):
LGTM

PR #3364, opened 2022-04-19
Allow retrying cluster and nodepool creation

kuosandys:
Closes https://github.com/giantswarm/roadmap/issues/1014.

This PR changes cluster and node pool creation to allow for resubmission with the same inputs in case of an error. We also update the error flash message prompting users to try again or contact support:
Screen Shot 2022-04-19 at 19 53 47

The `createCluster` and `createNodePool` functions are changed to allow toggling behaviour with the `isRetrying` boolean flag. This is to handle errors on retries that an object already exists. In this case we don't return the error as a rejected promise, and fetch the corresponding existing resource.


marians (approved):
As I'm not sure how to test, I'll just approve.

gusevda (approved):
Looks good!

PR #3380, opened 2022-04-25
Update React packages to v18 major

gusevda:
Towards [giantswarm/roadmap/issues/982](https://github.com/giantswarm/roadmap/issues/982).

In this PR:
- React was updated to v18;
- React related libraries were updated to recent versions which support React v18;
- Type declarations for `connected-react-router` were patched to work with React v18;


kuosandys (approved):
Tested on `gauss` and everything looks great! Thank you for making the update 🎉

PR #3385, opened 2022-04-26
Parse permissions use cases file

kuosandys:
Closes https://github.com/giantswarm/roadmap/issues/1067.

This PR parses the permissions use case `.yaml` file into the application and makes the data available on the window's `config` object in JSON format. The data can then be accessed as a JS object with the `getPermissionsUseCases` utility function, and the `IPermissionsUseCase` interface is introduced to help work with the object.

Since we may want to make permissions use cases configurable for different installations in the future (e.g. with the `happa-configmap` as we do now for other configurable values), the use cases file is parsed during the Docker image build step rather than application build. In local development, the file parsing is done in the `ConfigurationPlugin`.


scripts/prepare.ts
```diff
@@ -103,12 +110,39 @@ async function tryReadConfigFile(): Promise<string> {
...
+  const useCasesObject = {
+    'permissions-use-cases-json': JSON.stringify(
+      useCases['useCases'],
+      replacer
+    ),
+  };
+
+  log('Read permissions use cases file successfully.');
+
+  return yaml.dump(useCasesObject);
```
gusevda, line 133: This code duplicates the one from `getPermissionsUseCasesConfig` function. Is there a reason for that?

kuosandys: Not a good one 😬 they were different at the beginning and I forgot to refactor. Will consolidate them into one

gusevda (approved):
Looks great!

PR #3388, opened 2022-04-27
Create stories file for use case definition YAML

marians:
Towards https://github.com/giantswarm/roadmap/issues/1049

This adds some developer documentation for the use case definition file we introduced recently.


src/stories/UseCaseDefinitionYAML.stories.mdx
```diff
@@ -0,0 +1,105 @@
...
+Under the `scope` field we define whether the permissions for this use case are defined in the cluster scope, or in a paerticular namespace, or in any namespace.
+
+Using `cluster: true`, we specify that the permissions are to be defined in the cluster scope. Example:
+
+```yaml
+scope:
+  cluster: true
+```
+
+In contrast, if a use case's permissions are to be defined in namespaces, `scope.cluster` can be omitted ot set to `false`. In addition, the `namespaces` have to be enumerated, or an `*` string must be used to indicate that the permissions can be looked up in any namespace.
```
kuosandys, line 87:
```suggestion
In contrast, if a use case's permissions are to be defined in namespaces, `scope.cluster` can be omitted or set to `false`. In addition, the `namespaces` have to be enumerated, or an `*` string must be used to indicate that the permissions can be looked up in any namespace.
```

kuosandys (approved):
Looks great! Thank you ❤️

PR #3392, opened 2022-04-27
Display CGroups version for node pools

kuosandys:
Towards https://github.com/giantswarm/roadmap/issues/935.

This PR adds an extra column to the node pool list for a cluster, displaying the node pool's CGroups version:
Screen Shot 2022-04-28 at 09 48 59

If the node pool is using CGroups `v1`, we also display a warning icon, as well as a deprecation message on hover:
Screen Shot 2022-04-28 at 09 49 14


marians:
Nice!

I think we don't need to use a `<code>` wrapper here, and so the background should also be the normal row background. WDYT?

marians (approved):

kuosandys:
> Nice!
> 
> I think we don't need to use a `<code>` wrapper here, and so the background should also be the normal row background. WDYT?

I think it's not entirely necessary, and has the added effect of looking more aligned with the information to the right. I've removed the wrapper and updated the screenshots

PR #3400, opened 2022-04-29
Handle app catalog entry versions with `v-`  prefixes

kuosandys:
Background: https://gigantic.slack.com/archives/C02BBLEE2KW/p1651154989199529

We had previously made a change to handle discrepancies in an installed app's `.spec.version` and `.status.version` (caused by Flux) by stripping the `v-` prefix when making comparisons between versions. However, this created an issue when app catalog entries have a `v-` prefix. GS app catalogs don't have entries with a `v-` prefix, but this poses an issue for customers who create their app catalog entries with the prefix.

This PR fixes the issue by adding an extra check and only normalizes (strips the `v-` prefix) of an app's `.spec.version` if it is managed by Flux. Otherwise, the app versions are unmodified so they can be properly matched with app catalog entry versions.


QuentinBisson (approved):
LGTM if tested  but it's been a while since I've done react :D

PR #3429, opened 2022-05-12
Get permissions for a user/group

kuosandys:
Towards https://github.com/giantswarm/roadmap/issues/1086.

This PR implements a way to get permissions for a subject (user/group), so we can use the result to determine and display permissions for use cases in our "Inspecting Permissions" UI.

**How can I use this data?**
The information is exposed via the `useUseCasesPermissions` hook, which combines namespaced permissions and permissions at the cluster scope in order to determine overall permissions for a given subject. If no subject is given, the resulting permissions object is the user's own permissions.

**How do we get a subject's namespaced permissions?**
First we perform [`LocalSubjectAccessReviews`](https://pkg.go.dev/k8s.io/api/authorization/v1#LocalSubjectAccessReview) to determine which organizations the subject has access to. Then, in each organization's namespace, we get all `RoleBindings` that reference the subject, and aggregate the permissions referenced in the bindings. This is similar to how we get permissions at the cluster scope (in which case we get all `ClusterRoleBindings` that reference the subject).

While this involves a bit of computation on our side, IMO it's a better approach than querying the API hundreds of times with `LocalSubjectAccessReviews` for every unique `resource`/`apiGroup`/`verb` combination of each use case's permissions in each namespace.


gusevda (approved):
LGTM. Great improvement by adding `useUseCasesPermissions` hook. 

Small refactoring suggestion, shouldn't block merging of this PR. PermissionsOverviewGlobal and PermissionsOverviewOrganizations components look redundant now, since we don't have permissions logic inside of them anymore. What do you think about removing them?

kuosandys:
> LGTM. Great improvement by adding `useUseCasesPermissions` hook.
> 
> Small refactoring suggestion, shouldn't block merging of this PR. PermissionsOverviewGlobal and PermissionsOverviewOrganizations components look redundant now, since we don't have permissions logic inside of them anymore. What do you think about removing them?

That's a good point! I'll refactor 👍

kuosandys:
@gusevda I've refactored so that the `<PermissionsOverview />` component calls the `<PermissionsUseCases />` UI component directly, and I've moved the tests we had for each tab as well.

PR #3439, opened 2022-05-17
Determine whether a user can inspect a user's/group's permissions

kuosandys:
Towards https://github.com/giantswarm/roadmap/issues/1086.

This PR adds a way to determine whether a user can inspect a user's/group's permissions based on the access they have to the following resources:
- `list`: ClusterRoles, ClusterRoleBindings, Roles, RoleBindings at the cluster scope
- `create`: LocalSubjectAccessReviews in the `default` namespace (this is used to determine the Organizations accessible to the subject for whom they're inspecting permissions).


src/components/MAPI/permissions/utils.ts
```diff
@@ -521,6 +521,51 @@ export function hasAppAccess(
...
+      'rolebindings'
+    ):
+    case hasPermission(
+      permissions,
+      'default',
+      'create',
+      'authorization.k8s.io',
+      'localsubjectaccessreviews'
+    ):
+      return true;
```
gusevda, line 563: Just to ensure, it should return `true` if any one of these `hasPermission` calls returns `true`, not all of them, right?

kuosandys: That's a really good point - it should return `true` only if all the conditions are met 😬 Thanks for keeping a sharp eye!

gusevda (approved):
Looks great!

PR #3473, opened 2022-06-01
Service priority label

gusevda:
**Cluster details changes:**

- Service priority label is displayed in the "Interpreted" style
- "Display raw labels" switcher was added to see all the labels
Screenshot 2022-06-03 at 12 40 46

- In the raw view, all the labels are displayed. It is not possible to edit or delete originally hidden labels.
Screenshot 2022-06-03 at 12 48 42

- It's only possible to edit a value for the priority label. The key input in the edit popup is disabled.
Screenshot 2022-06-03 at 12 53 31

**Cluster list changes:**
- Cluster creation time was replaced by the information about node pools to make a view more compact.
- Labels information was added to the cluster list item. Only important labels are displayed here, other labels are hidden. Service priority label is displayed in the "interpreted" style. It's not possible to edit/remove a label in this view.

Before:
Screenshot 2022-06-03 at 11 59 39

After:
Screenshot 2022-06-03 at 12 16 31


marians:
I have one UX nitpick wish: When hovering a non-editable label, currently there is the same dotted/dashed underlining as with editable ones. I would like there to be no underlining in the case of non-editable labels.

marians:
When I switched from the details page to the cluster list (home) page, the labels container flashed briefly.

image

What's a bit strange here is the fact that the highlighted container has the full width of the parent item, as shown in the screenshot above. With only two labels in the div, this looked quite arbitrary.

This is a minor issue, nothing blocking.

marians (approved):
I commented on two possible improvements. Overall, works great, just as expected :)

src/model/constants/index.ts
```diff
@@ -71,7 +71,15 @@ export const Constants = {
   FLATCAR_CONTAINERLINUX_CGROUP_V2_VERSION: '3033.2.0',
 
   // Used to filter cluster labels in api responses
-  RESTRICTED_CLUSTER_LABEL_KEY_SUBSTRING: 'giantswarm.io',
+  RESTRICTED_CLUSTER_LABEL_KEY_SUBSTRINGS: [
+    'giantswarm.io',
+    'kustomize.toolkit.fluxcd.io',
+  ],
```
kuosandys, line 77: Could we mark these array constants as `readonly string[]`?

src/utils/labelUtils.ts
```diff
@@ -131,6 +131,22 @@ const isQualifiedName: IValidationFunction = (key) => {
   };
 };
 
+export const labelKeyIsRestricted = (key: string) => {
+  return Constants.RESTRICTED_CLUSTER_LABEL_KEYS.some(
+    (restrictedKeyItem) => key.toLowerCase() === restrictedKeyItem
+  );
+};
```
kuosandys, line 138: Since these utility functions return booleans, what do you think about using the naming scheme `is...` or `has...`?

src/components/MAPI/clusters/utils.ts
```diff
@@ -1002,3 +1004,92 @@ export enum ClusterStatus {
   UpgradeScheduled = 'UPGRADE_SCHEDULED',
   UpgradeAvailable = 'UPGRADE_AVAILABLE',
 }
+
+export function getClusterLabels(
+  cluster: capiv1beta1.ICluster
+): IClusterLabelMap {
+  return capiv1beta1.getClusterLabels(cluster);
+}
```
kuosandys, line 1012: Would it be possible to change the return type of `capiv1beta1.getClusterLabels()` to `IClusterLabelMap`? So we can prevent redeclaring this function 😅

kuosandys:
Really awesome changes 🚀 The test cases really helped to summarize the changes made, and thank you for adding Storybook docs for the new component as well!

Along with a few small suggestions above, I noticed that the release components seem to be missing their outlines. I think it might be an unintended side effect from the changes to `<ValueLabel />`:
Before:
Screen Shot 2022-06-03 at 20 44 54
Currently:
Screen Shot 2022-06-03 at 20 44 48

src/components/MAPI/clusters/utils.ts
gusevda: fixed

src/model/constants/index.ts
gusevda: Fixed. I marked the whole `Constants` object as readonly. It required some type adjustments in a few places. Here is [the commit](https://github.com/giantswarm/happa/pull/3473/commits/61f1e339df8408d8470f8e980c27d0a510a0eec8).

src/utils/labelUtils.ts
kuosandys: Hmm that’s a harder one.. What about`hasRestrictedLabelKeySubstring` or `hasRestrictedSubstringForLabelKey`? If that’s too long I think `hasRestrictedSubstring` would work too in the context

kuosandys (approved):
Looks great 👍 Thanks for making the changes!

kuosandys:
I noticed while testing https://github.com/giantswarm/happa/pull/3473 that the "Create cluster" CLI guide now shows all labels, and doesn't currently work 😬 
Screen Shot 2022-06-10 at 10 28 16

We could set the `--service-priority` flag in the command to just set the service priority, and avoid showing all the other labels.

gusevda:
> I noticed while testing #3473 that the "Create cluster" CLI guide now shows all labels, and doesn't currently work 😬 Screen Shot 2022-06-10 at 10 28 16
> 
> We could use the `--service-priority` flag in the command to just set the service priority, and avoid setting all the other labels.

Fixed

PR #3493, opened 2022-06-08
Future-proof code accessing Catalog.spec.storage

kubasobon:
Towards https://github.com/giantswarm/giantswarm/issues/21782

Catalog CRs will be extended with a new field `.spec.repositories` containing an array of catalog sources in contrast to a single source defined in `.spec.storage`. Since catalog sources can have various types now (`helm`, `oci`, etc.) we need to find `helm` source if it exists in the array.

### DIFF
--- a/src/model/stores/appcatalog/actions.tsx
+++ b/src/model/stores/appcatalog/actions.tsx
@@ -144,7 +144,15 @@ export function catalogLoadIndex(
 }
 
 async function loadIndexForCatalog(catalog: IAppCatalog): Promise<IAppCatalog> {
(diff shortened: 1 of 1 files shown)


kubasobon:
@giantswarm/team-rainbow I did my best, but since I know close to nothing about TypeScript, please let me know if this is stupid.

kubasobon:
I figured out that Happa is still using `AppCatalog` CRs on the backend, not Catalog CRs. I will have to retract my PR - AppCatalogs are untouched.

kubasobon:
Oh, great! I'll take another swing at it then and ping you in the PR.

PR #3494, opened 2022-06-08
Future-proof Catalog CR by adding .spec.repositories

kubasobon:
Towards https://github.com/giantswarm/giantswarm/issues/21782 (as a follow-up to the closed https://github.com/giantswarm/happa/pull/3493)

Catalog CRs will be extended with a new field `.spec.repositories` containing an array of catalog sources in contrast to a single source defined in `.spec.storage`. Since catalog sources can have various types now (`helm`, `oci`, etc.) we need to find the `helm` source if it exists in the array.


kubasobon:
@kuosandys I've extended the `ICatalogSpec`, but I can't find any place in code where `ICatalogSpec.storage` is used explicitly. My guess is you're relying on `AppCatalogEntry` resources, but I may be wrong.

Is there anything else I should update? AppCatalog definitions will stay the same.

kuosandys:
> @kuosandys I've extended the `ICatalogSpec`, but I can't find any place in code where `ICatalogSpec.storage` is used explicitly. My guess is you're relying on `AppCatalogEntry` resources, but I may be wrong.
> 
> Is there anything else I should update? AppCatalog definitions will stay the same.

Thank you! You're right, we don't currently make use of `ICatalogSpec.storage` since we mostly fetch AppCatalogEntries, and it's also not possible to modify/create Catalogs from happa. Speaking of - are there any changes to AppCatalogEntries resources?

kubasobon:
> Speaking of - are there any changes to AppCatalogEntries resources?

We are discussing internally whether AppCatalogEntries need to be updated in any way. Since they only reference the Catalog CR and are agnostic to the storage solution I think we are fine leaving them as they are.

kuosandys (approved):
thank you ❤️

PR #3501, opened 2022-06-09
Add service priority select box into create cluster form

gusevda:
The ability to set service priority has been introduced to the create cluster form in this PR. The service priority is set to `highest` by default.

Screenshot 2022-06-09 at 17 36 04

### DIFF
--- a/src/components/MAPI/clusters/utils.ts
+++ b/src/components/MAPI/clusters/utils.ts
@@ -246,7 +246,6 @@ export function createDefaultProviderCluster(
     name: string;
     organization: string;
     releaseVersion: string;
-    servicePriority: string;
   }
 ) {
   switch (provider) {
@@ -264,7 +263,6 @@ function createDefaultAzureCluster(config: {
   name: string;
   organization: string;
   releaseVersion: string;
-  servicePriority: string;
 }): capzv1beta1.IAzureCluster {
   return {
     apiVersion: 'infrastructure.cluster.x-k8s.io/v1beta1',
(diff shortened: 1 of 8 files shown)


src/components/MAPI/clusters/utils.ts
```diff
@@ -317,6 +321,7 @@ function createDefaultAWSCluster(config: {
         [capiv1beta1.labelClusterName]: config.name,
         [infrav1alpha3.labelOrganization]: config.organization,
         [infrav1alpha3.labelReleaseVersion]: config.releaseVersion,
+        [capiv1beta1.labelServicePriority]: config.servicePriority,
```
kuosandys, line 324: We only want to set the service priority label on the main cluster (`cluster.clusterx-k8s.io`) resource. Can we avoid setting this label on the provider cluster (`azurecluster.infrastructure.cluster.x-k8s.io` and `awscluster.infrastructure.giantswarm.io`) resources?

src/model/constants/index.ts
```diff
@@ -26,6 +28,7 @@ export const Constants = {
   DEFAULT_NODEPOOL_NAME: 'Unnamed node pool',
   DEFAULT_NODEPOOL_DESCRIPTION: 'Please add description',
   DEFAULT_CLUSTER_DESCRIPTION: 'Please add description',
+  DEFAULT_CLUSTER_SERVICE_PRIORITY: 'highest',
```
kuosandys, line 31: This constant isn't currently being used anywhere 😅

src/components/MAPI/utils.ts
```diff
@@ -909,6 +909,12 @@ export function getClusterDescription(
   }
 }
 
+export function getClusterServicePriority(cluster: Cluster): string {
+  const labels = capiv1beta1.getClusterLabels(cluster);
+
+  return labels[capiv1beta1.labelServicePriority];
+}
+
```
kuosandys, line 917: We have some functions for getting cluster labels in `src/model/services/mapi/capiv1beta1/key.ts` - do you think it makes sense to place this with those?

PR #3505, opened 2022-06-13
Support node pools for Azure release v17.1.0

kuosandys:
Closes https://github.com/giantswarm/giantswarm/issues/22349.

This PR adds support for node pools for Azure release v17.1.0.

For clusters using this release version or newer, we use the `MachinePool` and `AzureMachinePool` CRs from the API groups `cluster.x-k8s.io` and `cluster.infrastructure.x-k8s.io`, rather than the experimental API groups (prefixed with `exp.`) for all operations regarding node pools. This includes creation, updating description and scaling limits, and deletion.

Most of the changes are behind the scenes. The only noticeable user-facing change is that CLI guides for updating and deleting node pools will display the appropriate CRs depending on the release version.


src/model/constants/index.ts
```diff
@@ -43,6 +43,7 @@ export const Constants = {
   AWS_HA_MASTERS_MAX_NODES: 3,
 
   AZURE_CAPZ_VERSION: '20.0.0',
+  AZURE_NON_EXP_MACHINE_POOLS_VERSION: '17.1.0',
```
kuosandys, line 46: This is the minimum version required for supporting `MachinePools` and `AzureMachinePools` from the non-experimental API groups. For testing purposes, this may need to be temporarily changed

whites11 (commented):
I don't feel like reviewing TS, but thanks for your effort Sandy <3

gusevda (approved):
LGTM

PR #3517, opened 2022-06-16
Fix node pool creation for Azure: Spark not created

kuosandys:
We've previously introduced retry logic to the node pool creation form in order to allow users to keep the information in their form and to retry the creation in case it fails.

A bug was reported where after a node pool has been created, the Spark resource was not being created for subsequent node pools unless the user navigated away from the page and back.

This PR fixes this bug by only setting the retrying state if an error was encountered during creation.


src/model/constants/index.ts
```diff
@@ -43,7 +43,7 @@ export const Constants = {
   AWS_HA_MASTERS_MAX_NODES: 3,
 
   AZURE_CAPZ_VERSION: '20.0.0',
-  AZURE_NON_EXP_MACHINE_POOLS_VERSION: '17.1.0',
+  AZURE_NON_EXP_MACHINE_POOLS_VERSION: '17.0.999',
```
kuosandys, line 46: For testing purposes - to be reverted before merge

gusevda (approved):
LGTM

PR #3518, opened 2022-06-16
Fix configuring an installed app: object modified error

kuosandys:
Closes https://github.com/giantswarm/roadmap/issues/1010.

This PR fixes the bug where configuring an installed app (with ConfigMap or Secret) sometimes fails with the error: "Operation cannot be fulfilled on apps.application.giantswarm.io: the object has been modified".

**Why does this happen**
This error occurs when the App resource is modified by something else in the time between us fetching and updating it. This time period is very short (the `UPDATE` request happens immediately after the `GET`), but because the ConfigMap/Secret is ensured right before we update the App resource, this would lead to an update of the App's annotations (e.g. `app-operator.giantswarm.io/latest-configmap-version`) and `resourceVersion`. Whether the App resource we `GET` is up to date depends on how fast our network speed is v.s. how fast the operator(s) managing the App resource is.

**How do we fix it**
To mitigate this racing behaviour, we patch the App resource with just the fields we want to update. So instead of using `UPDATE` to submit the whole resource, we use `PATCH` to just update `.spec.userConfig.configMap`/`spec.userConfig.secret` with JSON merge patch as the patching strategy.

I also removed some unused code for an app details modal that we don't display anymore and fixed a broken link in the app configuration CLI guide.


src/components/MAPI/apps/AppValueConfigurator.tsx
```diff
@@ -139,10 +139,10 @@ const AppValueConfigurator: React.FC<
   const handleFileUpload = async (values: string, cb: () => void) => {
     if (hasConfig) {
       await onReplaceConfig?.(values);
+    } else {
+      await onUploadConfig?.(values);
```
kuosandys, line 143: These are bound to the same function and we were submitting the request twice

gusevda (approved):
The code looks good to me!

PR #3534, opened 2022-06-22
Synchronize selected chart version in app install form and app details page

kuosandys:
Closes https://github.com/giantswarm/giantswarm/issues/22487 and https://github.com/giantswarm/giantswarm/issues/21746.

We display a dropdown menu for selecting an app's chart version in both the app details page and the app install form (displayed after clicking on the "Install app in cluster" button). This PR synchronizes the chart version selected in these two places in order to reduce the likelihood of an incorrect version being selected for app installation (e.g. when a user clicks on the "Install app in cluster" button after inspecting an older version on the app details page, the expectation would be to also see the same version selected by default in the app install form).

Screen Shot 2022-06-22 at 18 41 38

**How to test this PR**
The changes of this PR can be tested by deploying this branch of `happa` using `opsctl deploy happa@app-install-modal-chart-version -i INSTALLATION`.


hervenicol (approved):
I can't say anything about the code, but it looks like the application now behaves the way I was expecting it to :+1:

marians (approved):

PR #3546, opened 2022-06-28
Add "Use web UI" use case to permissions overview

kuosandys:
Closes https://github.com/giantswarm/roadmap/issues/1185.

This PR adds the "Use web UI" use case to the list of use cases displayed in the permissions overview page. This use case is only evaluated and displayed for User and Group subject types (it doesn't make sense to display it for Service Accounts, and is redundant when inspecting permissions for yourself).

Screen Shot 2022-06-28 at 16 09 13


marians (approved):
LGTM from what I see in the PR description. Not tested.

PR #3557, opened 2022-06-30
Color palette refactoring

gusevda:
Towards [roadmap/issues/1142](https://github.com/giantswarm/roadmap/issues/1142).

The purpose of this PR is to improve the way we organize and use colors in the project. The idea is to keep color palette definition in one place, and provide semantic ways to use it across the project. 

Changes in this PR:
1. Colors that were recently added to the deprecated `theme.colors` list were moved to the Grommet's theme configuration - `theme.global.colors`.
3. Colors values from the Grommet's theme configuration were moved to a separate `COLORS` object, so they can be easier reused in the future.
4. Places where theme colors were used in styled components were refactored to use `normalizeColor` utility function from Grommet. This way it's more safe to use semantic color names and it's alligned with how Grommet uses colors behind the scene. 
```
// Before:
const StyledDiv = styled.div`
    background: ${({ theme }) => theme.global.colors['background-front'].dark};
`;

// After:
const StyledDiv = styled.div`
    background: ${({ theme }) => normalizeColor('background-front', theme)};
`;
```
5. Storybook colors information was updated.
6. `<ColorSample />` component was improved to display color titles better. Now where contrast between title and background is not big enough, dark color for title is used. Before we had to manually set color value for title in such cases.


kuosandys (approved):
Really cool 🎉 Thank you for updating FE docs as well!

PR #3571, opened 2022-07-05
Empty list items type error

gusevda:
Sometimes when we fetch resources that supposed to have items field inside we get a response with items field being undefined. It happens rarely and for different resources in different use cases. So the root cause of a problem is not clear. From the previous investigation we realised that the another attempt to fetch a list resource usually ends up with a valid response. Back then we implemented a fix for a ReleaseList resource. Current idea is to implement such logic for all list resources we have to collect more statistics on this issue.

In this PR another utility function was added to fetch list resources. `getListResource` function is calling our current `getResource` function  and if the response doesn't contain items field, another attempt is being made. The fact of a failed attempt is reported to Sentry behind the scenes and no errors are shown to a user. `getListResource` function makes 5 attempts to fetch a valid response until it fails with a timeout error. Every next attempt happens with a small delay - first retry after one second, second after two seconds and so on.


marians (approved):
Approach appears legit to me

src/model/services/mapi/generic/getListResource.ts
```diff
@@ -0,0 +1,55 @@
...
+  auth: IOAuth2Provider,
+  url: string,
+  attempt = 1
+): Promise<T> {
+  return getResource<T>(client, auth, url).then((data) => {
+    if (data.items === undefined) {
+      ErrorReporter.getInstance().notify(
+        `Invalid Response. Items are missing in the getListResource response.`,
+        { url, attempt }
+      );
```
kuosandys, line 25: Good idea to report the number of attempts! What do you think about including the entire data object returned? Would it help us get more insight or would it be redundant?

src/model/services/mapi/generic/__tests__/getListResource.ts
```diff
@@ -0,0 +1,67 @@
...
+import { IOAuth2Provider } from 'utils/OAuth2/OAuth2';
+import TestOAuth2 from 'utils/OAuth2/TestOAuth2';
+
+import { getListResource } from '../getListResource';
+import * as utils from '../getResource';
+
+const getResource = jest.spyOn(utils, 'getResource');
+
+describe.only('getListResource', () => {
+  jest.setTimeout(20000);
```
kuosandys, line 11: Is the `.only` required here?

gusevda: thanks for noticing, will fix 😬

src/model/services/mapi/generic/getListResource.ts
gusevda: good idea, let's do it

kuosandys (approved):
LGTM 🎉

PR #3574, opened 2022-07-06
Fix uploading a file with the same name for configuring an installed app

kuosandys:
Closes https://github.com/giantswarm/roadmap/issues/1212.

This PR fixes a bug with uploading files to configure an installed app. When users re-upload a file with the same name, the `onChange` event is not fired. This is because the `<input />` element's value is set to the file's path, and the value has not changed (even if the file contents have).

To circumvent this, we reset the value of `<input />` to an empty string whenever the "Upload/Replace values/secret values" button is clicked.

This PR also fixes a related bug where we try to index an empty array of files - in this case it makes more sense to exit early if there are no files to upload.


LolloneS (approved):
_Grazie mille!_

gusevda (approved):
LGTM

PR #3576, opened 2022-07-07
Add support for determining permissions for CAPG resources

kuosandys:
Closes https://github.com/giantswarm/roadmap/issues/1214 (towards https://github.com/giantswarm/roadmap/issues/1192).

This PR adds support for determining permissions to CAPG resources, specifically clusters, control plane nodes, and node pools.

For permissions to `get` and `list` resources, this is accomplished in a similar way as we do for existing providers.

For mutating actions (`create`, `update`, `delete`), a new permissions hook is added for determining permissions to cluster apps (Apps and ConfigMaps). Cluster management should be done via cluster apps, so it makes sense to check for permissions to these resources rather than the resulting resources. The [`values.yaml`](https://github.com/giantswarm/cluster-gcp/blob/main/helm/cluster-gcp/values.yaml) file for the `cluster-gcp` app provides more information about what is configurable.


gusevda (approved):
Looks great!

PR #3581, opened 2022-07-11
CAPG: Display node pools stats

kuosandys:
Closes https://github.com/giantswarm/roadmap/issues/1221 and closes https://github.com/giantswarm/roadmap/issues/1187.

This PR adds support for displaying node pool statistics (i.e. node pool count, worker nodes count, worker nodes CPU and worker nodes RAM). This is reflected in the cluster list (homepage), the cluster detail > worker nodes widget, and the organizations > general overview.

It also includes an improvement for displaying large memory amounts (i.e. > 1000GB) in TB. Memory amounts between 1 and 10TB will be displayed to 1 decimal place, and memory amounts above that will be displayed to the nearest integer.
Screen Shot 2022-07-11 at 17 09 16


marians (approved):
LGTM

Is there a unit test for getHumanReadableMemory?

kuosandys:
> LGTM
> 
> Is there a unit test for getHumanReadableMemory?

Most of its logic relies on `humanFileSize()` which has some tests, but I'll also add some for `getHumanReadableMemory()` to cover the specific cases we use it for 👍

gusevda (approved):
Looks great!

PR #3611, opened 2022-07-22
Display original app name

gusevda:
Closes [giantswarm/roadmap/issues/1057](https://github.com/giantswarm/roadmap/issues/1057).

In this PR application's original name was added to the information displayed about an installed application.  For the "original name" - `.spec.name` field of the App resource is used. `.metadata.name` field, that was previously displayed as an app name, is now displayed as "Installed as". In the collapsed view, "Installed as" name is displayed only if it's different from the original name.

App item is collapsed:
Screenshot 2022-07-22 at 12 51 07

App item is expanded:
Screenshot 2022-07-22 at 12 51 25


kuosandys (approved):
Looks great! LGTM 🎉

PR #3612, opened 2022-07-22
Display cluster app and Kubernetes versions information for CAPG

kuosandys:
Towards https://github.com/giantswarm/roadmap/issues/1246

This PR adds support for displaying the cluster app and control plane Kubernetes versions for CAPG clusters in the cluster list (home page) and cluster details overview page. These replace the display of release information, as Releases will not be used for CAPG.

Cluster list (home page):
Screen Shot 2022-07-22 at 16 16 31
Cluster details overview > versions widget:
Screen Shot 2022-07-22 at 16 16 31
Loading placeholder widget component:
Screen Shot 2022-07-22 at 16 16 31

To support the possibility of having to support an installation that includes both Releases and cluster apps (e.g. CAPA with AWS or CAPZ with Azure), we determine on a per-cluster basis which UI variant to display. If the Cluster has the `app: cluster-PROVIDER` label, we display the cluster app variant.

In cases where we fetch Releases to determine whether a cluster is using a preview release version, we first check if Releases are even supported by the provider - this is to prevent us from making unnecessary requests if the information will not be useful at all.


src/components/MAPI/clusters/ClusterDetail/ClusterDetailOverview.tsx
```diff
@@ -142,6 +157,13 @@ const ClusterDetailOverview: React.FC<React.PropsWithChildren<{}>> = () => {
             clusterName={cluster.metadata.name}
             clusterNamespace={cluster.metadata.namespace!}
           />
+          {!isClusterApp && clusterVersion && (
```
gusevda, line 160: Should we also hide this guide when `isClusterApp` is undefined?

gusevda: `clusterVersion` is undefined in this case, so it's not a problem

gusevda (approved):
Looks great!

PR #3628, opened 2022-08-02
Fix cluster upgrade modal disclaimer

gusevda:
In the cluster upgrade modal we display a disclaimer for API downtime when a cluster has only one master node. There is an issue with this message - it's being shown regardless of how many master nodes a cluster has.

Screenshot 2022-08-02 at 17 55 44

This PR fixes this issue. If a cluster has more then one master nodes, the modal will look like this:
Screenshot 2022-08-02 at 18 03 22


anvddriesch (approved):
From my limited understanding it looks fine. :)

PR #3640, opened 2022-08-09
Display cluster app and Kubernetes version information in Organizations > General tab

kuosandys:
Closes https://github.com/giantswarm/roadmap/issues/1246.

This PR displays information about cluster app and Kubernetes version in the Organizations > General tab. This section is only displayed if at least one cluster in the organization is created from a cluster app. We also don't display the 'Releases' section if the provider does not support releases.

For example, when there is one cluster app and Kubernetes version in use:
Screen Shot 2022-08-09 at 17 10 33

When there are 2 different versions in use, we enumerate the versions:
Screen Shot 2022-08-09 at 17 19 17

When there are 3 or more versions in use, we display the number of versions in use, as well as the oldest and newest versions:
Screen Shot 2022-08-09 at 17 17 18

Kubernetes versions are also displayed using a similar logic, and the EOL date is displayed on mouseover in a tooltip.

### DIFF
--- a/src/components/MAPI/organizations/OrganizationDetailGeneral/__tests__/index.tsx
+++ b/src/components/MAPI/organizations/OrganizationDetailGeneral/__tests__/index.tsx
@@ -830,6 +830,12 @@ describe('OrganizationDetailGeneral on GCP', () => {
 
     nock(window.config.mapiEndpoint)
       .get(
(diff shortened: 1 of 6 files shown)


gusevda (approved):
Looks great!

PR #3650, opened 2022-08-15
Display node pool details for CAPG

kuosandys:
Closes https://github.com/giantswarm/roadmap/issues/1299.

This PR adds support for displaying most node pool details for CAPG, specifically:
- description
- instance type
- zones (availability zones)
- current and desired number of nodes

Compared with vintage, the cgroups and min/max columns are hidden.
Screen Shot 2022-08-19 at 17 57 27

The main focus of this PR is to display the cluster details > node pool tab without the page crashing. This was done by reusing existing functionality (from previously implemented CAPZ node pool support) to display the node pools as read-only, rather than trying to refactor the node pool creation form component.

The command displayed in the CLI guide for listing node pools has also updated:
Screen Shot 2022-08-16 at 17 04 45


src/components/UI/Display/MAPI/CLIGuide/CLIGuidesList.tsx
```diff
@@ -31,56 +31,61 @@ const CLIGuidesList: React.FC<React.PropsWithChildren<ICLIGuidesListProps>> = ({
...
-      <Accordion
-        onActive={(indices: number[]) => setIsExpanded(indices.length > 0)}
-        basis='100%'
-        round='xsmall'
-        background={isExpanded ? 'accent-strong' : 'background-strong'}
-        {...props}
-      >
-        <AccordionPanel
-          label={
+      {filteredChildren.length > 1 ? (
```
kuosandys, line 38: Trying to make `<CLIGuidesList/>` "smarter" (potentially over-engineered 😬 ), so it can determine whether to render the wrapper component based on the number of `<CLIGuides/>` visible. If there is only one CLI guide in the list, then we don't show the wrapper component.

src/components/MAPI/workernodes/ClusterDetailWorkerNodes.tsx
```diff
@@ -382,9 +375,11 @@ const ClusterDetailWorkerNodes: React.FC<
       );
     }, [nodePoolList?.items, providerNodePools]);
 
-    const clusterReleaseVersion = cluster
-      ? capiv1beta1.getReleaseVersion(cluster)
-      : undefined;
+    const isReleasesSupportedByProvider = supportsReleases(provider);
```
kuosandys, line 378: We fetch releases to determine node pool cgroups versions, but if Release CRs are not supported by the provider, we should refrain from making the request to fetch them altogether.

gusevda (approved):
Looks great to me!

src/components/UI/Display/MAPI/CLIGuide/CLIGuidesList.tsx
marians: I like it!

marians:
Is the idea that the MIN and MAX column will stay there?

kuosandys:
> Is the idea that the MIN and MAX column will stay there?

I opened this PR before our discussion and the updates from Alex, but the plan is to remove the MIN/MAX and cgroups columns. I'll make the necessary changes and update the PR description + screenshots above.

marians (approved):
Works as expected

gusevda (approved):
LGTM

PR #3657, opened 2022-08-18
Change cluster preinstalled apps UI

gusevda:
Closes https://github.com/giantswarm/roadmap/issues/1277.

This PR changes how the preinstalled apps are displayed on the cluster details page. Previously we displayed some static information based on the Release CR, information like - release component name, version and icon. With the changes from this PR, preinstalled apps are displayed the same way as other (user installed) apps. Some additional changes were implemented to the `App` component to work well with preinstalled apps:
- `App` component UI was limited so it's not possible to modify preinstalled apps (upgrade, change configuration or uninstall);
- information about the upstream version of an app was added to the app details.

Before:
Screenshot 2022-08-18 at 15 19 05

After:
Screenshot 2022-08-18 at 15 18 24

Extra refactoring:
- cluster's `hasOptionalIngress` setting was removed since it's always `True` for all the releases that are currently in use for all the providers.


src/components/MAPI/apps/ClusterDetailReleaseApps.tsx
```diff
@@ -0,0 +1,104 @@
...
+  const clientFactory = useHttpClientFactory();
+  const auth = useAuthProvider();
+
+  const releaseClient = useRef(clientFactory());
+  const { canGet: canGetReleases } = usePermissionsForReleases(
+    provider,
+    'default'
+  );
+
+  const isReleasesSupportedByProvider = supportsReleases(provider);
```
kuosandys, line 47: Thanks for checking if releases are supported before fetching them! In the case that they aren't though, would it still make sense to render this component?

src/components/MAPI/apps/ClusterDetailAppListWidgetVersion.tsx
```diff
@@ -22,11 +26,20 @@ interface IClusterDetailAppListWidgetVersionProps
...
 
 const ClusterDetailAppListWidgetVersion: React.FC<
   React.PropsWithChildren<IClusterDetailAppListWidgetVersionProps>
-> = ({ app, catalogNamespace, canListAppCatalogEntries, ...props }) => {
+> = ({
+  app,
+  catalogNamespace,
+  canListAppCatalogEntries,
+  displayUpgradableStatus = true,
+  displayUpstreamVersion = false,
```
kuosandys, line 40: Just for my understanding, is the main function of the `displayUpstreamVersion` prop for rendering a variant of this version widget?

src/components/MAPI/apps/ClusterDetailReleaseApps.tsx
gusevda: Actually, this component will be used only on vintage. For CAPI providers there will be another component where we will fetch `default-apps-xxx` bundle. So this check is redundant probably.

kuosandys (approved):
Looks good to me 👍 and thanks for removing the unused optional ingress code! I just have the 2 questions above, but they shouldn't block merging this PR

src/components/MAPI/apps/ClusterDetailAppListWidgetVersion.tsx
gusevda: This is correct

marians:
Will give it a short test drive

marians (approved):
Tested, works great!

I noticed that the text `These apps and services are preinstalled on your cluster and managed by Giant Swarm.` is not exactly appropriate. I think it fits for an iteration where we not only listed apps. By now I think we should just simplify it to

`These apps are preinstalled on your cluster and managed by Giant Swarm.`

PR #3658, opened 2022-08-18
Allow installing app multiple times if `clusterSingleton` restriction is not set on AppCatalogEntry

kuosandys:
Towards https://github.com/giantswarm/roadmap/issues/1130.

Previously, we implemented the [feature](https://github.com/giantswarm/happa/pull/3338) to display the installation status of an app for a selected cluster. However, we didn't take into account the `clusterSingleton` restriction on the app catalog entry when determining installation status. This blocked users from installing an app multiple times for the same workload cluster, even when it does not have the `clusterSingleton` restriction.

This PR fixes this bug by only disabling the app installation button if this restriction is set on the app catalog entry. Otherwise, the installation button is still actionable.

When the app is installed in a selected cluster, and its app catalog entry has the `clusterSingleton` restriction, we display an appropriate message and disable selecting the cluster from the cluster picker form:
Screen Shot 2022-08-18 at 16 21 52

When the app is installed in a selected cluster, and its app catalog entry does not have the `clusterSingleton` restriction, we display an appropriate message, but the button is clickable, and users can select the cluster from the cluster picker form:
Screen Shot 2022-08-18 at 16 21 25


gusevda (approved):
Looks great!

PR #3659, opened 2022-08-18
Refactor how cluster organization is determined

kuosandys:
Closes https://github.com/giantswarm/roadmap/issues/1322.

This PR refactors how we determine a cluster's organization, so that it doesn't rely on the `giantswarm.io/organization` label on the Cluster CR. In happa, we fetch clusters using the namespaces of organizations that a user has access to, so we can determine a cluster's organization by checking within that list of organizations which organization's namespace (Organization CR's `status.namespace`) matches the cluster's namespace (Cluster's `spec.namespace`).

Note that for older AWS clusters in the `default` namespace, we still rely on the label, as we cannot determine their organization from the namespace.


src/components/MAPI/apps/AppInstallModal.tsx
```diff
@@ -158,6 +148,7 @@ const AppInstallModal: React.FC<
 
   const onClose = () => {
     setVisible(false);
+    setQuery('');
```
kuosandys, line 151: Small bug in the app installation modal - this resets the input field for filtering the cluster list when the user closes the modal

src/components/MAPI/utils.ts
```diff
@@ -404,40 +404,60 @@ export async function fetchClusterList(
...
+        })
+      );
+    }
+    if (organization?.id) {
+      requests.push(
+        capiv1beta1.getClusterList(httpClientFactory(), auth, {
+          namespace: 'default',
+          labelSelector: {
+            matchingLabels: {
+              [capiv1beta1.labelOrganization]: organization.id,
```
kuosandys, line 432: This fixes a bug with fetching older AWS clusters that are still in the `default` namespace - we check for matches with both the case-sensitive organization display name (`organization.name`, which has the value of the `ui.giantswarm.io/original-organization-name` annotation) and the organization CR name (`organization.id`)

gusevda (approved):
Looks great!

PR #3676, opened 2022-08-24
Update permissions use cases for GCP

kuosandys:
### What does this PR do?

Update the permissions use cases so that it works for GCP. This includes:
- adding `gcpcluster` to use case permissions where provider-specific cluster resources are used
- filtering out use cases containing permissions for `releases` when the provider does not support Release CRs

Users viewing happa's Permissions page on GCP installations will not see the "Releases" section and "Inspect releases" use case displayed. No change on vintage (AWS and Azure) installations.

### How does it look like?

Permissions use cases on GCP:
Screen Shot 2022-08-24 at 16 53 47

For reference, permissions use cases on vintage.
Screen Shot 2022-08-24 at 16 54 04

Closes https://github.com/giantswarm/roadmap/issues/1339 which is part of our efforts towards [supporting GCP in happa](https://github.com/giantswarm/roadmap/issues/1192)

Nothing extra beyond the general code review and verifying UI changes.

No

Not necessarily


marians (approved):
LGTM

gusevda (approved):
Looks great!

PR #3681, opened 2022-08-25
App bundles UI and GCP support for cluster apps

gusevda:
### What does this PR do?
1. App bundle UI was implemented;
2. GCP support for displaying apps on the cluster details page was added;
3. Apps statistics on the Organization overview page were removed;
4. Apps widget on the cluster overview page was changed:
    - it displays the amount of all App CRs available instead of the user installed only
    - it displays the amount of Not Deployed apps instead of the amount of Deployed apps
    - number colors we changed to bring user's attention to important items

(Please set a descriptive PR title. Use this space for additional explanations.)

App bundles implementation will not effect current users since the only ones who use app bundles - don't use web-ui at the moment.
Information we display in the Apps widget was changed which can confuse a user. Now we calculate statistics based on all App CRs installed. We decided to do this change, because recently, we started to display preinstalled App CRs on the cluster Apps page. Before, we used to display only user installed App CRs, preinstalled apps were displayed differently. Now, the information we display on the cluster Apps page and in the Apps widget on the cluster Overview page will match.

### How does it look like?

Example of an app bundle:
Screenshot 2022-08-25 at 12 24 39

Example of the Apps widget changes:
Screenshot 2022-08-25 at 12 26 12


marians (approved):
Fantastic! Tested on gohan.

marians:
Regarding "Apps widget changes can introduce some confusion, since we changed the information we display there." - This itself can cause some confusion, too :) I suggets that you link to the issue where we summarized the info (if we did) or explain what is changed and why.

gusevda:
> Regarding "Apps widget changes can introduce some confusion, since we changed the information we display there." - This itself can cause some confusion, too :) I suggets that you link to the issue where we summarized the info (if we did) or explain what is changed and why.

Thanks for the review, I updated the description a little bit.

src/components/MAPI/clusters/ClusterDetail/ClusterDetailActions.tsx
```diff
@@ -137,7 +138,27 @@ const ClusterDetailActions: React.FC<
 
   const { canList: canListApps } = usePermissionsForApps(provider, clusterId);
 
-  const appListGetOptions = { namespace: clusterId };
+  const appsNamespace =
+    typeof isClusterApp === 'undefined'
+      ? undefined
+      : isClusterApp
+      ? org?.status?.namespace
```
kuosandys, line 145:
```suggestion
      ? namespace
```
We already have the value of `org?.status?.namespace` assigned to the variable `namespace` on line 65 - do we want to reuse it?

src/components/MAPI/apps/ClusterDetailApps.tsx
```diff
@@ -142,24 +179,45 @@ const ClusterDetailApps: React.FC<
...
-            errorMessage={extractErrorMessage(appListError)}
-            appsPermissions={appsPermissions}
-            releaseVersion={releaseVersion}
-          />
-
-          <Box margin={{ top: 'medium' }} direction='column' gap='small'>
-            <ListAppsGuide namespace={clusterId} />
-          </Box>
+          {typeof isClusterApp !== 'undefined' &&
+            typeof clusterVersion !== 'undefined' && (
```
kuosandys, line 183: Is it possible to use `isClusterApp && clusterVersion` here, or would it affect the UI loading behaviour?

src/components/MAPI/clusters/ClusterDetail/ClusterDetailActions.tsx
gusevda: good catch

src/components/MAPI/apps/ClusterDetailApps.tsx
gusevda: I can change it to
```
typeof isClusterApp !== 'undefined' && clusterVersion
```

`isClusterApp` can be `true | false | undefined`, in this case only `undefined` is not allowed.

src/components/MAPI/apps/ClusterDetailWidgetApps.tsx
```diff
@@ -143,7 +209,7 @@ const ClusterDetailWidgetApps: React.FC<
     )
       return -1;
 
-    if (!upgradableApps) return undefined;
+    if (!upgradableApps) return 0;
```
kuosandys, line 212: I think this means the upgrades available number in the apps widget will display `0` while we calculate this. Is this intentional?

src/components/MAPI/apps/ClusterDetailApps.tsx
kuosandys: Thanks for explaining! It's not a blocker from me, just wondered if we could simplify it a little bit

kuosandys (approved):
Looks great, and now we have support for displaying app bundles and installed apps on GCP 🎉  Also really awesome that we can support multiple nesting levels for app bundles!

src/components/MAPI/apps/ClusterDetailWidgetApps.tsx
gusevda: You are right, fixed

PR #3697, opened 2022-09-06
Make cluster management read-only for GCP

kuosandys:
### What does this PR do?

This PR makes cluster management "read-only" in the following ways:
- If a cluster is managed via a "cluster app" (i.e. it has the `app: cluster-PROVIDER` label), cluster description editing, cluster label editing, and cluster deletion is disabled.
- If an installation has a CAPI provider (i.e. not `AWS`, `Azure`, or `KVM`), cluster creation is disabled.

Users on GCP installations will not be able to edit/delete cluster resources from the web UI. Users on vintage installations should notice no change.

### How does it look like?

Cluster list page does not display the "Launch new cluster" button, and the placeholder message for an empty cluster list does not prompt users to click the button:
Screen Shot 2022-09-06 at 16 30 57

Cluster description is displayed as uneditable text:
Screen Shot 2022-09-06 at 15 58 33

Cluster labels widget does not display the "Add" button, and displays labels as text:
Screen Shot 2022-09-06 at 15 57 53

Cluster details "Actions" tab is not displayed (usually displayed after "Ingress" tab), and we don't support routing to the `/actions` path:
Screen Shot 2022-09-06 at 15 58 56

Closes https://github.com/giantswarm/roadmap/issues/1350, which is a task towards our first milestone of providing a read-only web UI for GCP.

The changes here are meant to be temporary in the sense that they will be in place until cluster management via apps is implemented for the web UI.

The UI changes can be verified by checking out this branch and starting local development against a GCP installation (e.g. `gohan`), or alternatively by using `opsctl deploy`. The logic for determining whether cluster management is read-only is described at the start of this PR.

I've also added a new Storybook entry documenting the variants of the `ViewAndEditName` component, which is used for displaying editable text (e.g. cluster description).


marians (approved):
LGTM

PR #3701, opened 2022-09-07
Update list of cluster labels hidden by default

kuosandys:
### What does this PR do?

This PR updates the list of restricted cluster labels to include label key `app` and label keys starting with `app.kubernetes.io` and `helm.sh`.

As with other restricted cluster labels, users will no longer see labels matching the above-mentioned criteria displayed by default and will also not be able to edit them from the web UI. These labels can still be viewed by toggling "Display raw labels".

### How does it look like?
Screen Shot 2022-09-07 at 14 13 57

In the case where all the cluster labels are hidden by default (when "Display raw labels" is toggled off), but the cluster still has labels, we adapt the message to more accurately read "This cluster has no _editable_ labels" rather than the current "This cluster has no labels".
Screen Shot 2022-09-07 at 14 14 12

Closes https://github.com/giantswarm/roadmap/issues/1369, originally raised by [this comment](https://github.com/giantswarm/roadmap/issues/1192#issuecomment-1227062992) noting that on GCP installations we've noticed some more cluster labels that should be hidden by default.

The usual UI change verification (`gohan` is a good installation on which to test the changes of this PR) and a second pair of eyes on the code is always appreciated :)


marians (approved):
Great little enhancement included! LGTM

PR #3706, opened 2022-09-08
Make node pool management read-only for GCP

kuosandys:
### What does this PR do?

This PR makes node pool management "read-only" for node pools belonging to clusters that are managed by an app (i.e. it has the `app: cluster-PROVIDER` label). Node pool creation, deletion, and updating an existing node pool's description and scaling are disabled.

Users on GCP installations will not be able to add/edit/delete node pool resources from the web UI. Users on vintage installations should notice no change.

### How does it look like?
- the "Add node pool" button and any placeholder messages prompting users to create node pools when the cluster doesn't have any node pools is hidden
- the node pool list item menu button is hidden
- node pool description is displayed as uneditable text
Screen Shot 2022-09-08 at 16 17 17

Closes https://github.com/giantswarm/roadmap/issues/1382, which is a task towards our first milestone of providing a read-only web UI for GCP.

The changes here are meant to be temporary in the sense that they will be in place until cluster management via apps is implemented for the web UI.

The UI changes can be verified by checking out this branch and starting local development against a GCP installation (e.g. `gohan`), or alternatively by using `opsctl deploy`.

### DIFF
--- a/src/components/MAPI/utils.ts
+++ b/src/components/MAPI/utils.ts
@@ -1249,13 +1249,17 @@ export function isCAPZCluster(cluster: Cluster): boolean {
   return compare(releaseVersion, Constants.AZURE_CAPZ_VERSION) >= 0;
 }
 
-export function isCAPGCluster(cluster: Cluster): boolean {
+function isCAPGCluster(cluster: Cluster): boolean {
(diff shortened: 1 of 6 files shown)


marians (approved):
LGTM

PR #3707, opened 2022-09-08
Fix layout bug in non-MAPI UI control plane widget

kuosandys:
### What does this PR do?

It fixes a layout bug in the non-MAPI UI's cluster details control plane widget.

Users will see a more polished UI :)

### How does it look like?
Currently:
Screen Shot 2022-09-08 at 17 37 01

With the changes of this PR:
Screen Shot 2022-09-08 at 17 55 17
Screen Shot 2022-09-08 at 17 55 10

There's no tracking issue, I found it while browsing the non-MAPI UI.

Verify that the layout bug is fixed.


marians (approved):
❤️

PR #3713, opened 2022-09-09
Display app logo for app items in catalogs if provided

kuosandys:
### What does this PR do?

It checks whether an AppCatalogEntry has the `ui.giantswarm.io/logo` annotation and uses the URL provided to display an image for the app in the catalog. If the annotation is not provided, it falls back to the app's icon (i.e. what is currently used).

As this annotation has not been added to AppCatalogEntries yet, users should notice no change in the current UI. Once app developers use the `ui.giantswarm.io/logo`, the image at the URL provided will be displayed in the app catalog list view and app details page.

Towards https://github.com/giantswarm/roadmap/issues/1248.


src/model/services/mapi/applicationv1alpha1/key.ts
```diff
@@ -44,6 +45,12 @@ export function getAppCatalogEntryReadmeURL(
   return appCatalogEntry.metadata.annotations?.[annotationReadme];
 }
 
+export function getAppCatalogEntryLogoURL(
+  appCatalogEntry: IAppCatalogEntry
+): string | undefined {
+  return appCatalogEntry.metadata.annotations?.[annotationLogo];
```
gusevda, line 51: Do you think it's possible to move icon fallback inside this function?
like this:
```
   return appCatalogEntry.metadata.annotations?.[annotationLogo] ?? appCatalogEntry.spec.chart.icon;
```

kuosandys: Good idea 👍 I'll make the changes

src/components/MAPI/apps/AppList/utils.tsx
```diff
@@ -249,3 +250,13 @@ export function mapAppCatalogEntriesToAppPageApps(
     };
   });
 }
+
+export function getAppCatalogEntryLogoURL(
+  appCatalogEntry: applicationv1alpha1.IAppCatalogEntry
+): string {
```
kuosandys, line 256: As the fallback logic was more UI-related, I thought it would be more appropriate to move the function to the app list utilities file

gusevda (approved):
Looks great!

PR #3731, opened 2022-09-15
Support permissions for apps: app resources in organization namespace

kuosandys:
### What does this PR do?

This PR modifies how we determine permissions for apps to support the case when app-related resources (`App` + `ConfigMaps` + `Secrets`) are in the organization namespace.

For vintage clusters, app resources were placed in the namespace of the workload cluster they belong to. To determine permissions for apps, we created `SelfSubjectAccessReviews` for the corresponding resources in the workload cluster namespaces.

For CAPI, app resources are placed in the organization's namespace. We can now determine permissions the same way we do for other resources located in the organization's namespace - by using the permission object exposed by the `usePermissions()` hook.

With the changes of this PR, we should now be checking in the appropriate namespaces for the user's permissions to apps, and we should no longer be creating unnecessary `SelfSubjectAccessReviews`.

Users using the web UI on vintage installations should notice no change. Users with limited permissions on GCP installations should now see a more accurate UI representative of their permissions to app-related resources.

Closes https://github.com/giantswarm/roadmap/issues/1400, towards https://github.com/giantswarm/roadmap/issues/1192.


src/components/MAPI/apps/ClusterDetailIngress.tsx
```diff
@@ -51,17 +54,27 @@ const ClusterDetailIngress: React.FC<
...
-  const { clusterId } = useParams<{ clusterId: string }>();
+  const { clusterId, orgId } = useParams<{
+    clusterId: string;
+    orgId: string;
+  }>();
+
+  const organizations = useSelector(selectOrganizations());
+  const appsNamespace = isClusterApp
+    ? organizations[orgId]?.namespace
+    : clusterId;
```
gusevda, line 69: `isClusterApp` can be undefined here, so I suggest we do an additional check:
```
const appsNamespace =
    typeof isClusterApp === 'undefined'
      ? undefined
      : isClusterApp
      ? organizations[orgId]?.namespace
      : clusterId;
```

gusevda (commented):
One small comment to look into before merging, other then that looks great!

src/components/MAPI/apps/InstallIngressButton.tsx
```diff
@@ -65,7 +67,11 @@ const InstallIngressButton: React.FC<
   const appListClient = useRef(clientFactory());
   const appListGetOptions = { namespace: clusterID };
 
-  const appsPermissions = usePermissionsForApps(provider, clusterID);
+  const appsPermissions = usePermissionsForApps(
+    provider,
+    appsNamespace ?? '',
+    isClusterApp
+  );
```
kuosandys, line 74: Missed updating the permissions hook in `<InstallIngressButton />` earlier - I referenced the app installation PR for the extra props we need to pass in, hopefully it wouldn't create major conflicts

gusevda (approved):
LGTM

PR #3741, opened 2022-09-19
App installation support on GCP

gusevda:
### What does this PR do?

This PR adds support for the apps installation on GCP:
- installation of an app from the catalog;
- installation of the `nginx-ingress-controller-app` app from the cluster details page (ingress tab) and from the "getting started" guide.

Changes in the implementation:
- on GCP an app is being installed in the organization namespace rather than cluster namespace;
- on GCP configMap used for the `nginx-ingress-controller-app` app is the same as for other apps - `CLUSTER_ID-cluster-values`. On vintage - `ingress-controller-values` is being used.
- `giantswarm.io/cluster` label is being added to the App CR during app installation (it's required on GCP).

Users on GCP should now be able to install applications via standard interface.

Closes https://github.com/giantswarm/roadmap/issues/1398.


src/components/MAPI/apps/ClusterDetailIngress.tsx
```diff
@@ -45,41 +51,109 @@ interface IClusterDetailIngressProps
...
+  const clusterKey =
+    canGetClusters && namespace
+      ? fetchClusterKey(provider, namespace, clusterId)
+      : null;
+
+  const { data: cluster, error: clusterError } = useSWR<
+    Cluster,
+    GenericResponseError
+  >(clusterKey, () =>
+    fetchCluster(clientFactory, auth, provider, namespace!, clusterId)
```
kuosandys, line 108: In `<ClusterDetailApps />` we get the `isClusterApp` prop from the parent component and get the organization's namespace from the current selected organization. Is it also possible to do the same here, or is there another reason for the different implementation?

src/components/GettingStarted/Steps/InstallIngress.js
```diff
@@ -18,6 +18,10 @@ const InstallIngressButtonWrapper = styled.div`
 `;
 
 const InstallIngress = (props) => {
+  console.log('AAAAAAAA');
+  console.log('AAAAAAAA');
+  console.log('AAAAAAAA');
+  console.log('AAAAAAAA');
```
kuosandys, line 24:
```suggestion
```
AAA!

gusevda: nice 🙈

src/components/MAPI/apps/ClusterDetailIngress.tsx
gusevda: Good idea, should work

src/components/UI/Display/MAPI/clusters/GettingStarted/GettingStartedInstallIngress.tsx
```diff
@@ -1,16 +1,102 @@
...
+import ErrorReporter from 'utils/errors/ErrorReporter';
+import { FlashMessage, messageTTL, messageType } from 'utils/flashMessage';
+import { useHttpClient } from 'utils/hooks/useHttpClient';
+import { useHttpClientFactory } from 'utils/hooks/useHttpClientFactory';
 
 interface IGettingStartedInstallIngressProps {}
 
 const GettingStartedInstallIngress: React.FC<
   React.PropsWithChildren<IGettingStartedInstallIngressProps>
 > = () => {
```
kuosandys, line 23: Since this component now includes more than just display functionality, what do you think about moving it to `src/components/MAPI/clusters/GettingStarted/` ?

gusevda: Sounds good, makes sense

kuosandys (approved):
LGTM!

PR #3748, opened 2022-09-23
Fix displaying app's catalog name and versions info for users with limited permissions

kuosandys:
### What does this PR do?

This PR fixes the display of an installed app's catalog name and version info for users without permissions to get the app's Catalog or AppCatalogEntries, by only fetching these resources from namespaces the user has access to (i.e. `default`, organization namespaces, and `giantswarm` if they are GS staff).

If the user does not have permissions to view these resources, then we display the catalog name using the App CR's `.spec.catalog`, and the upstream version using `.status.appVersion`.

Users with limited permissions should now be able to see the catalog name of an app, even if they do not have permissions to the catalog. In the browser's console, the user should also not see 403 errors from fetching Catalog and AppCatalogEntry resources they do not have access to.

We also don't display the empty selector menu with other versions of the app, and disable the non-functioning link ('Details' button) to the app's details if the user does not have permissions for the app's catalog entries. A tooltip message is displayed when hovering over the 'Details' button
Screen Shot 2022-09-22 at 16 30 43
Screen Shot 2022-09-23 at 15 40 04

Closes https://github.com/giantswarm/giantswarm/issues/22793.


gusevda (approved):
Looks great!

PR #3749, opened 2022-09-23
Support displaying app installation status for app resources in organization namespaces

kuosandys:
### What does this PR do?

This PR adds support for displaying app installation status for app resources in organization namespaces. This was previously implemented for app resources in workload cluster namespaces.

Users viewing the UI on GCP installations should now see app installation status displayed correctly.

Closes https://github.com/giantswarm/roadmap/issues/1434.


gusevda (approved):
LGTM

PR #3765, opened 2022-09-30
Change web UI release notes link

kuosandys:
### What does this PR do?

This PR changes the release notes link to point to the happa releases page for the version, replacing the link to the docs releases page.

When clicking on the "release notes" link in the footer, users should be directed to happa's releases page for the specified version.

No tracking issue, reported internally on Slack. Due to our documentation page's workflow, sometimes the link does not point to an existing page. This ensures that the the link always points to an existing page with the release notes.


gusevda (approved):
LGTM

PR #3776, opened 2022-10-04
Add support for fetching CAPA cluster resources

kuosandys:
### What does this PR do?

This PR adds support for fetching CAPA cluster resources, specifically by refactoring the `fetchProviderClusterforCluster` function to add a case for CAPA clusters. It also updates the permission hook for determining permissions to cluster resources, and adds some test mocks for CAPA clusters.

Hopefully this is sufficient enough to allow us to work concurrently on displaying cluster, node pools, and control planes details.

On a CAPA installation, users should now be able to see clusters listed in the cluster list page instead.

Closes https://github.com/giantswarm/roadmap/issues/1450.


gusevda (approved):
Looks good!

PR #3781, opened 2022-10-05
Display details for CAPA clusters

gusevda:
### What does this PR do?

This PR adds support for displaying a description and a status for CAPA clusters. Only `Cluster creating…` status is supported for now.

### How does it look like?

Screenshot 2022-10-05 at 19 05 02

Closes https://github.com/giantswarm/roadmap/issues/1474.

### DIFF
--- a/src/components/MAPI/utils.ts
+++ b/src/components/MAPI/utils.ts
@@ -1093,6 +1093,8 @@ export function getClusterDescription(
 
   const { kind, apiVersion } = infrastructureRef;
   switch (true) {
+    case kind === capav1beta1.AWSCluster &&
+      apiVersion === capav1beta1.ApiVersion:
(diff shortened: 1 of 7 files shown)


src/components/MAPI/utils.ts
```diff
@@ -1093,6 +1093,8 @@ export function getClusterDescription(
 
   const { kind, apiVersion } = infrastructureRef;
   switch (true) {
+    case kind === capav1beta1.AWSCluster &&
+      apiVersion === capav1beta1.ApiVersion:
```
kuosandys, line 1097: Since the cluster description is an annotation on the Cluster resource, and we use the same Cluster CR for all providers, what do you think about making this case the default behaviour? That way we could avoid having a long list of cases here as we support more CAPI providers

kuosandys (approved):

PR #3798, opened 2022-10-11
Fix YAMLException error parsing for app installation

kuosandys:
### What does this PR do?
This PR fixes the `TypeError` encountered when we try to parse `YAMLException`s, by replacing manual formatting with the [`toString` method](https://github.com/nodeca/js-yaml/blob/0d3ca7a27b03a6c974790a30a89e456007d62976/lib/exception.js#L50) available on the `YAMLException class`.

It also fixes not being able to upload a file with the same name (but with different contents), and resets the errors when the user navigates away from the app installation form.

Closes https://github.com/giantswarm/roadmap/issues/1499.


gusevda (approved):
LGTM, very nice improvement 👍

PR #3815, opened 2022-10-19
Allow specifying providers for permissions use cases and permissions rules

kuosandys:
### What does this PR do?

This PR adds an additional `providers` field to our permissions use cases `.yaml` file, so that we can specify whether a use case, or a permission rule for a use case, is specific to certain providers.

As we worked on adding support for more CAPI providers in happa, we've encountered cases where it was necessary to filter permission use cases or the permissions rules of a use case by a provider.

For example, the permission use case to "Inspect releases" is not applicable to CAPI providers, as the releases concept does not exist in CAPI. Another instance is the "Inspect clusters" use case where we had to filter out certain permission rules so that we can correctly determine permissions for the current provider (i.e. ignoring permission rules for `awsclusters` when the provider is Azure).

We previously defined this logic in a utility function for filtering permissions, but it was not obvious to see, and thus difficult to extend. This PR proposes to move this to be declared in the permission use cases `.yaml` file so that it can be easily changed in the future.

### How does it look like?
An example of a use case specifying that it is only applicable for certain providers:
```yaml
- name: "Inspect releases"
  description: "Read workload cluster releases"
  category: "releases"
  scope:
    cluster: true
  providers: # this use case is only applicable for aws and azure
    - aws
    - azure
  permissions: # ...
```

An example of a use case specifying that certain permission rules is only applicable for certain providers:
```yaml
- name: "Inspect clusters"
  description: "Read resources that form workload clusters"
  category: "workload clusters"
  scope:
    namespaces:
      - "*"
  permissions:
    - apiGroups: # this permission rule does not specify providers - it will be applied to all
        - cluster.x-k8s.io
      resources:
        - clusters
      verbs:
        - get
        - list
    - apiGroups:
        - infrastructure.cluster.x-k8s.io
      resources:
        - azureclusters
        - azuremachines
      verbs:
        - get
        - list
      providers: # this permission rule specifies providers, and will only be used to determine permissions on azure
        - azure
    - apiGroups:
        - infrastructure.cluster.x-k8s.io
      resources:
        - awsclusters
      verbs:
        - get
        - list
      providers: # this permission rule specifies providers, and will only be used to determine permissions on capa
        - capa
```

There's no change to the web UI, and as the permissions use cases file currently only resides within happa's codebase, there is no change to "users" besides developers of happa.

The documentation page in Storybook for permissions use cases has been updated to provide a short explanation of the providers field and some examples.


gusevda (approved):
This is a great improvement! Thanks for updating documentation as well! Fantastic job 👏

PR #3818, opened 2022-10-20
Tests for apps resources in organization namespaces

gusevda:
### What does this PR do?

In this PR:
- existing tests for `components/MAPI/apps` were refactored to share app generation logic;
- tests for `ClusterDetailApps` were added, for both cases - when resources are in cluster namespaces (Azure cluster as an example) and organization namespaces (CAPA cluster as an example);
- additional tests for `ClusterDetailWidgetApps` component were added to cover cases when resources are in organization namespaces (CAPA cluster is used as an example).

No changes should be introduced to a user by this PR.

Closes https://github.com/giantswarm/roadmap/issues/1508.


kuosandys (approved):
Looks good to me 👍  Thanks for refactoring the shared app generating logic as well, it really cleaned up the test files quite a bit!

PR #3822, opened 2022-10-21
Generate TS types for MAPI resources from CRDs

kuosandys:
### What does this PR do?

This PR seeks to improve the process of how we generate TypeScript types for custom resources used with happa's MAPI client. TS types can now be generated by running `yarn generate-types`.

### How does it work?
The file `scripts/generateTypes/mapi-resources.yaml` keeps track of a list of resources we want to generate TS types for. These are organized by API group alias (e.g. `capav1beta1`) to work with happa's existing directory structure. For each resource, a URL to the CRD in `yaml` format should be provided.

The `generate-types` script fetches the CRD from the provided URL, and parses the Open API v3 schema for the appropriate version into TS types using the [json-schema-to-typescript](https://github.com/bcherny/json-schema-to-typescript) library. The types are then written to the `types.ts` file in the appropriate API group directory in `src/model/services/mapi/`. Additional types for the resource's `List` interface is also generated, and the appropriate import/export statements are also written.

While the write operations are performed sequentially, the fetching + parsing for each API group is done in parallel to save time. If memory usage becomes an issue in the future we can also further optimize.

The `generate-types` script has been run for CAPG and CAPA resources as an example. While I don't think it's necessary to migrate all the existing MAPI resources TS types to be generated by this script, I hope the passing typecheck on this PR demonstrates its compatibility with happa's code base.

Towards https://github.com/giantswarm/roadmap/issues/1528.

There's an additional task item in the issue linked above to add documentation for the entire process, and will be addressed in a separate PR.


scripts/generateTypes/getTypesForResource.ts
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * Provide custom TS types for specified fields
+ * @param config
+ */
+function getTsTypesConfig(config: { kind: string }): JSONSchema4 {
+  return {
+    properties: {
+      apiVersion: {
+        tsType: 'typeof ApiVersion',
```
kuosandys, line 28: We can use the `tsType` field to provide custom TS types for a property

gusevda (approved):
Fantastic, works like a charm!

One thing I noticed while testing. After I had run the `yarn generate-types`, newly regenerated `types.ts` files had a little bit different formatting. Looks like they were auto formatted by prettier before being committed. What do you think about running `yarn prettier` at the end of the script?

kuosandys:
> Fantastic, works like a charm!
> 
> One thing I noticed while testing. After I had run the `yarn generate-types`, newly regenerated `types.ts` files had a little bit different formatting. Looks like they were auto formatted by prettier before being committed. What do you think about running `yarn prettier` at the end of the script?

That's a good idea, I relied on the precommit hook running the formatting step. I'll update the script 👍

PR #3834, opened 2022-10-26
Unify default apps fetching for vintage and CAPI

gusevda:
### What does this PR do?

In this PR WC cluster apps fetching was simplified and aligned between CAPI and vintage. With [recent changes](https://github.com/giantswarm/kubectl-gs/releases/tag/v2.24.0) in `kubectl gs` default app bundle also has `giantswarm.io/cluster` label similar to other cluster apps. It allows us to fetch default app bundle together with other apps. Previously we had to do it separately and add it to the list of all apps. Now apps fetching is the same for CAPI and vintage.

The logic of splitting the apps into user installed apps and preinstalled apps was also changed.

On vintage we used to select preinstalled apps based on a release CR. In this PR this logic was removed as redundant, since the same result can be achieved without it. Now an app is considered to be preinstalled if it has `giantswarm.io/managed-by` label that equals to `cluster-operator`. All other apps are considered as user installed apps.

For CAPI the logic is the same, but the criteria for preinstalled apps is different. An app is considered to be preinstalled if it has `giantswarm.io/managed-by` label that equals to `cluster-apps-operator` OR `app.kubernetes.io/name` label that equals to the default app bundle name for the provider (e.g. default-apps-aws). The name check covers the default app bundle and its child apps.

Small additional change - provider name was removed from the default apps section note.
Screenshot 2022-10-26 at 16 48 34

After [this release](https://github.com/giantswarm/kubectl-gs/releases/tag/v2.24.0) of `kubectl gs`, apps view for newly created clusters was broken. This PR should fix this issue.

Ticket about refactoring of apps fetching - https://github.com/giantswarm/roadmap/issues/1527
Ticket about the bug introduced by `kubectl gs` release - https://github.com/giantswarm/roadmap/issues/1501


kuosandys (approved):

gusevda:
Based on the recent discussion during KaaS sync, it's not guaranteed that `giantswarm.io/cluster` label is present in a default-apps bundel. But there will be actions from our side to make this label required. If we proceed with changes from this PR, for all the clusters created with `kubectl gs`, apps view works as expected (tree view).
Screenshot 2022-10-31 at 13 34 47

If there is a cluster without a label for some reason, then the apps view will look like this (flat view without a default-apps bundle CR).
Screenshot 2022-10-31 at 13 36 47

@marians what do you think?

Regarding the logic behind splitting app CRs into two lists - installed by a user and default apps, most probably it will change in the future, but it works right now on vintage, GCP and AWS CAPA. So I suggest we keep it for now.

marians (approved):

PR #3840, opened 2022-10-28
Fix dropdown elements being underlined

mogottsch:
### What does this PR do?

Links in dropdowns were previously underlined when the mouse hovered over them. This PR removes the underline.
The responsible style `text-decoration: none` was already present, but overwritten by a global styles for anchor elements. This PR adds the `!important` directive, so that it is no longer overwritten.

Users will no longer see underlined menu links, which is visually more appealing.

### How does it look like?

**Before**:
image
image

**After**:
image
image

closes https://github.com/giantswarm/giantswarm/issues/22598

In addition to the problem described in the issue, this PR fixes the same problem for the sorting dropdown in the app catalog.

Confirm that the usage of `!important` is acceptable in this situation and not bad practice.

No.

Yes, added `kind/ux-enhancment`.


kuosandys (commented):
hey Mortiz, thanks for looking into this, and thanks for putting all the details in the PR descirption! Regarding your question about using `!important`, we try to avoid it when possible, though sometimes it can be unavoidable.

From the screenshots provided, it seems that the issue with the underlined text only appears on hover. What do you think about placing the `text-decoration: none` line to the block for hover styles starting at line 39? Would that work as an alternative instead of using `!important` for this case?

mogottsch:
That works and is way better, thank you!
I just pushed the changes.

kuosandys (approved):

PR #3842, opened 2022-10-28
Align style of keywords and cluster labels

mogottsch:
### What does this PR do?

App keywords and cluster labels are logically similar. This PR changes the styles of the keyword component to the styles of cluster labels.

Users will see logically similar components also visually similar.

### How does it look like?
**Before**:
Keywords (left), cluster labels (right)
image
**After**:
Keywords (left), cluster labels (right)
image

closes https://github.com/giantswarm/roadmap/issues/1191

As you can see it is not possible to use the background color of the cluster label component for the keyword component, because then the background color of the keyword component would be indistinguishable from the background.
It would be possible to change the background color of the cluster label component to the background color of the keyword component. 
I did not do that, because I think changing the color of the cluster labels is less visually appealing. Do you agree?

No.

Added `kind/ux-enhancement`

### DIFF
--- a/src/components/UI/Display/Apps/AppDetailNew/AppDetailPage.tsx
+++ b/src/components/UI/Display/Apps/AppDetailNew/AppDetailPage.tsx
@@ -125,11 +125,14 @@ const Detail = styled.div`
 const Keyword = styled.span`
   margin-right: 10px;
   margin-bottom: 10px;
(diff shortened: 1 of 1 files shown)


kuosandys (approved):
Since they're displayed against different background colours, I agree with your decision to leave the colours as-is 👍

PR #3843, opened 2022-10-28
Generate client functions for MAPI resources

kuosandys:
### What does this PR do?

This PR seeks to improve how we generate client functions for custom resources used with happa's MAPI client. The script `generate-mapi` now generates TS types and client functions for the resources defined in `scripts/generate/mapi-resources.yaml`.

In addition to the changes from https://github.com/giantswarm/happa/pull/3822, we extend `scripts/generate/mapi-resources.yaml` to allow specifying verbs for which we'd like to generate corresponding client functions for (e.g. specifying `get` for `SomeResource` would generate the client function `getSomeResource`).

The script has been run on the CAPA and CAPG resources defined in the above-mentioned file as an example.

As the client functions are dependent on the MAPI client implementation details, how we generate them is also dependent on happa's current codebase. If there is a need in the future, we could extract the MAPI client entirely and make the client function generation less happa-specific, but I consider the work for that to be outside the scope of this PR.

Towards https://github.com/giantswarm/roadmap/issues/1528, as a follow up to https://github.com/giantswarm/happa/pull/3822.


scripts/generate/templates.ts
```diff
@@ -0,0 +1,267 @@
...
+  resourceName: string,
+  resourceNamePlural: string,
+  namespaced: boolean,
+  verb: ClientFunctionVerbs
+): string {
+  const clientFunctionName = getClientFunctionMethodName(resourceName, verb);
+
+  return `export function ${clientFunctionName}(
+  client: IHttpClient,
+  auth: IOAuth2Provider,${namespaced ? '\nnamespace: string,' : ''}
```
gusevda, line 102: It would be nice to move `namespace` parameter to a separate line and remove `\n` formatting, to improve readability, e.g.:
```
  auth: IOAuth2Provider,
  ${namespaced ? 'namespace: string,' : ''}
  name: string
```

gusevda (approved):
Looks awesome!

I left only one comment, that's applicable in other places in the `templates.ts` file. I think if it's possible to get rid of `\n` formatting it would improve readability, and formatting can be handled by `prettier` which is already happening. 

But this shouldn't block merging. Great job!

PR #3856, opened 2022-11-02
Update cluster creation message for GCP

kuosandys:
### What does this PR do?

This PR updates the cluster creation message displayed for GCP clusters to inform users that the creation process usually takes 5 minutes instead of 15 minutes.

Users should now see a more accurate representation of the cluster creation duration.

### How does it look like?
Screenshot 2022-11-02 at 15 30 39
Screenshot 2022-11-02 at 15 30 13

Closes https://github.com/giantswarm/roadmap/issues/1438.


gusevda (approved):
Looks good to me!

PR #3861, opened 2022-11-04
Migrate happa to app-build-suite

mogottsch:
### What does this PR do?

This PR uses [app-build-suite](https://github.com/giantswarm/app-build-suite) as the executor for the [push-to-catalog](https://github.com/giantswarm/architect-orb/blob/master/docs/job/push-to-app-catalog.md#executor-optional-either-architect-or-app-build-suite-defaultarchitect) job.
app-build-suite offers additional linting and validation of helm charts. It will also enable us to implement chart tests more easily.

None.

Towards https://github.com/giantswarm/giantswarm/issues/20068
Closes https://github.com/giantswarm/roadmap/issues/1585

~~The only change that I think could potentially have an impact is the new `securityContext` setting in both happa and fe-docs deployments.~~
```
          securityContext:
            readOnlyRootFilesystem: true
```
~~As I see it both containers run nginx and should only write to the mounted volumes.~~
~~Am I correct?~~
This change crashed the container and was reverted.

No.

I don't think so.


kuosandys (approved):
Looks good from what I can tell 👍 I've added the `kind/dev-change` label to the PR and the changes won't be mentioned in the release notes

anvddriesch (approved):

kuosandys (changes requested):
@mogottsch I think I may have approved this prematurely, but I deployed this branch for testing (with `opsctl deploy`) and was seeing deployment failed errors:

In the `happa` app's status:
```
status:
  appVersion: 1.52.8
  release:
    lastDeployed: "2022-11-09T10:13:14Z"
    reason: 'Deployment failed for `1.52.7-601369efb5edb6628813154079a41ca8fc0cb0cc`
      with `unknown-error`: "values don''t meet the specifications of the schema(s)
      in the following chart(s):\nhappa:\n- security.restrictAccess.api: Invalid type.
      Expected: string, given: boolean\n- happa.letsencrypt: Invalid type. Expected:
      string, given: boolean\n- happaapi.letsencrypt: Invalid type. Expected: string,
      given: boolean\n- kubernetes.versions: Invalid type. Expected: string, given:
      object\n"'
    status: deployed
  version: 1.52.8
```

I figured it's better to raise the issue now rather than later 😅 Have you noticed something similar during your testing?

mogottsch:
After the changes, I was able to deploy this branch on `ginger` successfully.

mogottsch:
Migration `fe-docs` is still a WIP

mogottsch:
Waiting for [this](https://github.com/giantswarm/architect-orb/pull/399) PR

.abs/main.yaml
```diff
@@ -0,0 +1,4 @@
+replace-chart-version-with-git: true
+replace-app-version-with-git: true
+chart-dir: ./helm/athena
```
uvegla, line 3: I am thinking this should be `./helm/happa`, no?

mogottsch: Yes, missed that when copy-pasting. Thank you.
Weird that I didn't notice it in any other way.

mogottsch: I guess in the `push-to-app-catalog` job it is overwritten.

If we continue to maintain two charts in this repository, it might be best to remove the `chart-dir` from the `.abs.` file.

mogottsch:
I successfully deployed and tested `fe-docs` on a ginger workload cluster.

gusevda (approved):
Looks good

PR #3863, opened 2022-11-04
Provide link to catalog and catalog item for installed apps

kuosandys:
### What does this PR do?

This PR adds links to an installed app:
- on the app name: link to the app's details page for the corresponding `AppCatalogEntry`
- on the catalog name: link to the catalog page with the catalog pre-selected

Users are now able to navigate to the installed app's details page by clicking on the app's name:
Screenshot 2022-11-04

Clicking on the installed app's catalog name...
Screenshot 2022-11-04
...should allow them to navigate to the catalog page with the appropriate catalog pre-selected:
Screenshot 2022-11-04

Users who do not have permission to view the catalog item (`AppCatalogEntry` CR) or the app's catalog will still see these names displayed as plain text. This is to avoid presenting users with links to view resources they do not have sufficient permissions to see.

Closes https://github.com/giantswarm/roadmap/issues/1389.


gusevda (approved):
Looks great!

PR #3867, opened 2022-11-07
Add basic chart tests

mogottsch:
### What does this PR do?

This PR adds basic chart tests. The tests ensure that the Kubernetes API  is reachable and that at least one pod of happas deployment is ready.

None

Towards https://github.com/giantswarm/giantswarm/issues/20068

I had to edit `values.yaml`, because some validation did not accept `.happa.host` and `.happaapi.host` to be an empty string. Both have now the value `"example.com"`. I'm not sure whether this is ok.
Same goes for `registry.pullSecret.dockerConfigJson`, which now is set to `"e30="`. `e30=` is `{}` base64 encoded.
In general, do we provide defaults or examples in `values.yaml`?

No.


mogottsch:
Also, it is not so nice for happa's folder structure, that the `app-test-suite` tests have to be in the folder `tests/ats`, because happa's directory for tests is called `test`. 
As much as I know that is currently hard coded. 

What do you think would be an optimal solution from happa's point of view?

kuosandys:
Thanks for adding some chart tests for happa!

> In general, do we provide defaults or examples in values.yaml?

Would it be possible to use a separate `values.yaml` file for the test? For example in `dex-operator` I see that we can [reference a different values file for the test](https://github.com/giantswarm/dex-app/blob/15db61e7905a84e69820c20716d56d09692433a4/.ats/main.yaml#L11). If we could also do that here, I think this would keep the original `values.yaml` cleaner.

> Also, it is not so nice for happa's folder structure, that the app-test-suite tests have to be in the folder tests/ats

I'm not too familiar with app test suite, but if the folder value is hard-coded, I'm ok with the test files being where they are now. The fact that they're in their own folder should be sufficient to keep them organized 🙂

mogottsch:
I separated the `valus.yaml` file. Thanks for pointing that out. I looked for that option myself but didn't find it in the documentation.

I looked in the code of ats and realized that `/tests/` is not hard coded. Do you have a preference where the ats tests should go?
We could for example just put them in `/ats`, instead of `/tests/ats`.

kuosandys:
> Do you have a preference where the ats tests should go? We could for example just put them in `/ats`, instead of `/tests/ats`. I just think having a `/test` and `/tests` directory is confusing.

`/ats` works for me 👍 It's similar to `/.ats` but since they contain related files I'm cool with that. Definitely better than `/test` and `/tests`

mogottsch:
Alright, I've moved the folder :+1:

kuosandys (approved):

PR #3868, opened 2022-11-07
Fix cluster details Ingress tab

gusevda:
### What does this PR do?

This PR adds support for displaying Ingress information on CAPA.

Ingress tab on a cluster details page should display correct information instead of the loading indicators.

### How does it look like?

Before:
Screenshot 2022-11-07 at 18 21 10

After:
Screenshot 2022-11-07 at 18 24 36

Closes https://github.com/giantswarm/roadmap/issues/1575.


src/components/MAPI/apps/ClusterDetailIngress.tsx
```diff
@@ -123,6 +129,18 @@ const ClusterDetailIngress: React.FC<
...
+  const k8sEndpoint = useMemo(() => {
+    if (rest.k8sEndpoint) {
+      return rest.k8sEndpoint;
+    }
+
+    return cluster
+      ? isCAPGCluster(cluster) || isCAPACluster(cluster)
+        ? getClusterK8sAPIUrl(cluster, provider)
+        : capiv1beta1.getClusterK8sAPIUrl(cluster)
+      : undefined;
```
kuosandys, line 141: Instead of having 2 similar functions both named `getClusterK8sAPIUrl`, would it be possible to combine them and move the provider checks there?

gusevda: After refactoring:
`getClusterK8sAPIUrl ` was moved to the `MAPI/utils.ts` and is used to display KUBERNETES API on a cluster Overview page.
`getClusterK8sEndpoint` is now used for displaying instructions on a cluster Ingress tab and was moved to the `MAPI/apps/utils.ts`.

kuosandys (approved):

PR #3875, opened 2022-11-09
Remove infrastructure.giantswarm.io/v1alpha2 types and checks

kuosandys:
### What does this PR do?

This PR removes the TS types we have for `infrastructure.giantswarm.io/v1alpha2` resources, and removes places in the code where we check for `awscluster` resources from this API group (this was the only resource from this API group that we were checking for).

Closes https://github.com/giantswarm/roadmap/issues/1472.


gusevda (approved):
Perfect 👍

PR #3883, opened 2022-11-11
Improve how we determine when clusters should display "Upgrade available" status

kuosandys:
### What does this PR do?

This PR improves how we determine when clusters are upgradable, and thus when we show an "Upgrade available" prompt and the "Upgrade cluster" button, by omitting inactive versions when determining upgradable versions.

Users should no longer be prompted to upgrade to inactive versions. Note that only admin users (GS staff) can see non-active releases, so this change only impacts admin users.

### How does it look like?
For example, for a cluster that has an upgrade available:
Screenshot 2022-11-11 at 16 32 10

Clicking on the "Upgrade cluster" button will now prompt to upgrade to the next active version. Previously, it would prompt to upgrade to the next version, active or not:
Screenshot 2022-11-11 at 16 24 03

To maintain ease of testing non-active releases from the web UI, it is still possible to upgrade to non-active versions, by clicking on the current release version details, and selecting a release from the upgrade options list:
Screenshot 2022-11-11 at 16 24 31

If the next available upgrade version is not an active release, the "Upgrade available" status, and the "Upgrade cluster" button will not be shown.

Closes https://github.com/giantswarm/roadmap/issues/1352, from internal feedback that we shouldn't be prompting GS staff to upgrade to non-active releases.


src/components/MAPI/releases/ClusterDetailWidgetRelease.tsx
```diff
@@ -159,10 +159,19 @@ const ClusterDetailWidgetRelease: React.FC<
   }, [isAdmin, isImpersonatingNonAdmin, provider, releaseList, releaseVersion]);
 
   const nextVersion = useMemo(() => {
-    return supportedUpgradeVersions.find(
+    if (!releaseList || !releaseVersion) return undefined;
```
kuosandys, line 162: `supportedUpgradeVersions` contain all upgradable versions (active and inactive), while the `nextVersion` should be determined from only the active versions

src/components/MAPI/releases/ClusterDetailWidgetRelease.tsx
```diff
@@ -159,10 +159,19 @@ const ClusterDetailWidgetRelease: React.FC<
...
+
+    const activeVersions = getSupportedUpgradeVersions(
+      releaseVersion,
+      provider,
+      false,
```
mogottsch, line 167: Maybe we could add a comment here explaining that we set admin to `false` because we don't want to prompt anyone, including admins "Upgrade available" for inactive versions.
What do you think?

gusevda (approved):
Looks good to me

PR #3905, opened 2022-11-24
Add support for app bundles installation

gusevda:
### What does this PR do?

With the changes from this PR now it is now possible to install app bundles. Differences for an app bundle installation:
1. App CR instance of an app bundle must be given a unique name within the Management Cluster.
2. App CR should have `app-operator.giantswarm.io/version: 0.0.0` label and the `.spec.kubeConfig.inCluster` field set to `true`.
3. App CR’s `.metadata.namespace` and `.spec.namespace` fields must be equal and reference the cluster namespace on vintage and organization namespace on CAPI.
4. A user configuration to the App CR must provide `clusterName` and `organization` values.

App catalog entries that represent app bundles were visible in the UI but their installation were not possible. Now it should work correctly and not cause any errors.

### How does it look like?

Unique app name in a context of a workload cluster is suggested and there is a note that emphasis that the name should be unique . There is no namespace field for an app bundle installation form.
Screenshot 2022-11-24 at 11 10 49

We don't additionally validate on the client side that the app name is unique across other workload clusters, but if a user tries to install an app bundle with the name that is already taken, it will fail and an error will be displayed.
Screenshot 2022-11-24 at 11 16 12

https://github.com/giantswarm/giantswarm/issues/21478


src/components/MAPI/apps/AppInstallModal.tsx
```diff
@@ -552,7 +572,15 @@ const AppInstallModal: React.FC<
           break;
 
         default:
-          errorMessage = `Something went wrong while trying to install your app. Please try again later or contact support: support@giantswarm.io`;
+          errorMessage = (
+            <>
+              Something went wrong while trying to install your app. Please try
+              again later or contact support: support@giantswarm.io
+              <Text size='small' as='div' margin={{ top: 'medium' }}>
+                {metav1.getStatusMessage((error as GenericResponseError)?.data)}
```
kuosandys, line 580: In other places, we extract the error message from a k8s API response using the `extractErrorMessage` function from `src/components/MAPI/utils.ts`. Would it be possible to also do that here?

gusevda: oh, good point, thank you!

gusevda:
> Also, I know we're just adding support for installation with this PR, but I was wondering if you knew whether we'd also need to adapt app uninstallation/deletion?

From what I experienced, app bundle deletion works well. But let me double check it.

kuosandys (approved):

PR #3922, opened 2022-12-01
Only build fe docs when creating tags

mogottsch:
### What does this PR do?

Edit circle ci config to only build fe-docs when creating tags.

When testing the abs migration, I had to build fe-docs on every commit, to test it. I forgot to revert that change.


kuosandys (approved):
👍

PR #3927, opened 2022-12-02
Add form to create clusters via app bundles

kuosandys:

This PR adds a form for creating clusters via app bundles, which is generated from the cluster app's schema.

Screenshot 2023-01-25 at 16 17 17

To facilitate testing, we've included a test schema, as well as the functionality to select the schema's provider and a branch name (from the cluster app schema's GitHub repository). Also, note that clicking on the "Create cluster" button on the form currently does not create any resources.

The form can be found at `/organizations/ORGANIZATION/clusters/new-via-app-bundles`, or from a link in the "Experiments" section (visible to admin users only).


gusevda:
Questions to address on next refinement:
- Do we want to use UpDownWidget instead of `<input type="number" />` for `integer` or `number` fields? If yes, how can we configure it to not depend on the schema.
- Let's think on the UI for array fields.
- Order of elements? Do we want to display fields as they appear in a schema or do we want to prioritize some of them? How can we do it for all the providers?
- It is possible to use Accordion for object fields with the ability to collapse/expand groups of inputs. Is it something we want to experiment with?

marians:
FYI Since this is WIP, I removed myself from the reviewers list to reduce notifications.

src/components/UI/JSONSchemaForm/test.schema.json
```diff
@@ -0,0 +1,190 @@
...
+                    "enum": [
+                        "firstValue",
+                        "secondValue",
+                        "thirdValue"
+                    ],
+                    "title": "String enum",
+                    "type": "string"
+                },
+                "stringFormat": {
+                    "description": "String field with specific format (email, uri, date, ipv4, etc.).",
```
marians, line 166: To make it clear which format is valid here:

```suggestion
                    "description": "String field with format 'ipv4'.",
```

marians:
I think the text "Submit the form to see validation errors." is repetitive and could be placed in the top description once, but doesn't have to appear in several property descriptions.

src/components/UI/JSONSchemaForm/test.schema.json
```diff
@@ -0,0 +1,190 @@
...
+                },
+                "stringLength": {
+                    "description": "String field with minLength and maxLength validation. Submit the form to see validation errors.",
+                    "maxLength": 5,
+                    "minLength": 2,
+                    "title": "String with min/max length",
+                    "type": "string"
+                },
+                "stringPattern": {
+                    "description": "String field with pattern validation. Submit the form to see validation errors.",
```
marians, line 179:
```suggestion
                    "description": "String field with pattern validation for regular expression '^[a-z0-9]{10,20}$'.",
```

src/components/UI/JSONSchemaForm/test.schema.json
```diff
@@ -0,0 +1,190 @@
...
+                    "title": "Integer enum",
+                    "type": "integer"
+                },
+                "number": {
+                    "description": "Number field with title and description.",
+                    "title": "Number",
+                    "type": "number"
+                },
+                "numberLimit": {
+                    "description": "Number field with minimum and maximum values validation. Submit the form to see validation errors.",
```
marians, line 85:
```suggestion
                    "description": "Number field with minimum value 2 and maximum value 5.",
```

src/components/UI/JSONSchemaForm/test.schema.json
```diff
@@ -0,0 +1,190 @@
...
+                        },
+                        "name": {
+                            "type": "string"
+                        }
+                    },
+                    "title": "Object",
+                    "type": "object"
+                },
+                "objectRequired": {
+                    "description": "Object field with required properties. Submit the form to see validation errors.",
```
marians, line 110:
```suggestion
                    "description": "Object where one property is required.",
```

marians:
Just observed the following behaviour with Test schema: I filled in all properties to match the constraints and then clicked the "Create cluster" submit button. As a result, the form got reset and the form data preview was empty. I would have expected the form and data preview to stay in the previous state, just with all error messages removed.

kuosandys:
> Just observed the following behaviour with Test schema: I filled in all properties to match the constraints and then clicked the "Create cluster" submit button. As a result, the form got reset and the form data preview was empty. I would have expected the form and data preview to stay in the previous state, just with all error messages removed.

@marins Normally when the cluster creation form is submitted, the form data is reset and the user is redirected to the cluster details page. Since we're not currently submitting anything, we can persist the form data, and only reset the form on page refresh. We could also think about adding a flash message as a response to clicking the "Create cluster" button, as otherwise the user would not receive any indication that the button has been clicked, or that the submission had no effect in reality.

marians:
> We could also think about adding a flash message as a response to clicking the "Create cluster" button, as otherwise the user would not receive any indication that the button has been clicked, or that no resources have been created.

That would make sense. This is indeed behaviour specific to the prototype. In a real world scenario, on successful submission, you don't want to have the form in front of you any more. But here we need the submission button only to trigger validation, and after validation the property values should all stay as input, even if validation succeeded.

marians (approved):
I guess the 🚧 WIP - do not merge! 🚧  part can be removed now :)

gusevda (approved):

PR #3943, opened 2022-12-08
WIP Cluster app creation form fixes

gusevda:
### What does this PR do?

In this PR:
- `clusterName` and `clusterDescription` ordering was added based on the `uiSchema`.
- `ClusterNameWidget` was added to automatically generate `clusterName`. This is mostly to demonstrate how it could look like. Do we want it to be autogenerated? Do we also want to edit it manually?
- `generateUID` utility function was fixed. Before it used to cause an infinite loop when length parameter was an even number.

### How does it look like?

Screenshot 2022-12-08 at 16 49 01

### DIFF
--- a/src/components/MAPI/clusters/CreateCluster/CreateClusterAppForm.tsx
+++ b/src/components/MAPI/clusters/CreateCluster/CreateClusterAppForm.tsx
@@ -9,6 +9,7 @@ import {
   fetchAppCatalogEntrySchema,
   fetchAppCatalogEntrySchemaKey,
 } from 'MAPI/apps/AppList/utils';
+import { generateUID } from 'MAPI/utils';
 import { GenericResponseError } from 'model/clients/GenericResponseError';
 import { IHttpClient } from 'model/clients/HttpClient';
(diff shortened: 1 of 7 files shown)


src/components/MAPI/clusters/CreateCluster/CreateClusterAppForm.tsx
```diff
@@ -225,6 +232,13 @@ const CreateClusterAppForm: React.FC<ICreateClusterAppFormProps> = ({
   const [formDataPreviewFormat, setFormDataPreviewFormat] =
     useState<FormDataPreviewFormat>(FormDataPreviewFormat.Json);
 
+  const uiSchema = {
+    'ui:order': ['clusterName', 'clusterDescription', '*'],
+    clusterName: {
+      'ui:widget': ClusterNameWidget,
+    },
+  };
+
```
kuosandys, line 241: Does this have to be declared inside the component?

kuosandys (approved):
Looks good in general 👍 Just a concern about the `uiSchema` being declared inside the component, but we could also separate it as we work on tooling for `uiSchema`.

PR #3952, opened 2022-12-12
Customise array and object fields

gusevda:
### What does this PR do?
Changes to the `CreateClusterAppForm` component were added:
- `NumberPicker` component is now being used for number fields by default;
- `array` and `object` fields are customised to use an `Accordion` component;

### How does it look like?

Number fields:
Screenshot 2022-12-13 at 12 31 25

Object field collapsed:
Screenshot 2022-12-13 at 12 23 56

Object field expanded:
Screenshot 2022-12-13 at 12 23 37

Array field expanded and collapsed:
Screenshot 2022-12-13 at 12 25 18


src/components/UI/JSONSchemaForm/ArrayFieldTemplate/index.tsx
```diff
@@ -1,39 +1,95 @@
...
+  schema,
   onAddClick,
 }) => {
+  const [activeIndexes, setActiveIndexes] = useState<number[]>([]);
+
+  const itemsAreObjects =
+    schema.items &&
+    typeof schema.items !== 'boolean' &&
+    !Array.isArray(schema.items) &&
+    schema.items.type === 'object';
```
gusevda, line 30: In the future I would like to find a better way to determine if the array items are objects or primitives (string, boolean, etc.)

gusevda:
There are a couple of things that I will try to find better solutions for. They are mentioned in the comments. And there is also `Accordion` markup that is duplicated in two components, if we decide that we can continue with this implementation, I will try to refactor this part as well.

kuosandys (approved):
Looks great!

PR #3961, opened 2022-12-16
add-custom-cluster-issuer

calvix:
### What does this PR do?

allow configure custom clusterIssuer for happa certs

### DIFF
--- a/helm/happa/values.yaml
+++ b/helm/happa/values.yaml
@@ -43,6 +43,8 @@ gcp:
 
 ingress:
   tls:
+    letsencrypt: false
+    clusterIssuer: ""
     crtPemB64: ""
     keyPemB64: ""
 
@@ -57,7 +59,6 @@ kubernetes:
 happa:
   address: ""
   host: ""
-  letsencrypt: false
   featureFlags:
     mapiAuth: false
     mapiClusters: false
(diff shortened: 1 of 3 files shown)


helm/happa/values.yaml
```diff
@@ -43,6 +43,8 @@ gcp:
 
 ingress:
   tls:
+    letsencrypt: false
+    clusterIssuer: ""
```
calvix, line 47: migrate values under ingress to be aligned with `dex-app` and `athena`

helm/happa/values.yaml
```diff
@@ -57,7 +59,6 @@ kubernetes:
 happa:
   address: ""
   host: ""
-  letsencrypt: false
```
calvix, line 60: can't find a reason why we would want a different settings on happa and happaapi so moving all under ingress block

helm/happa/values.yaml
```diff
@@ -43,6 +43,8 @@ gcp:
 
 ingress:
   tls:
+    letsencrypt: false
```
calvix, line 46: its a little bit weird that the default is `false` since `dex-app` and `athena` has default `true`, but keeping same value not to mess with the state

kuosandys (approved):

PR #3972, opened 2022-12-21
Resolve external schema references

kuosandys:
### What does this PR do?

This PR adds basic support for resolving external schema references. This feature is currently and unfortunately not supported by `react-json-schema-form`, so we implement our own external schema resolution, which functions as follows:
- we collect all `$ref` from a cluster app schema that references a URL of http/https protocol (e.g. `$ref: https://giantswarm.schema.io/someSchema`)
- we fetch the schema from each URL and patch the original (parent) schema, so that the `$ref` now references a relative schema defined in the parent schema's `$def`

The tests accompanying this PR might shed some more light on the resolution behaviour.

Caveats:
~~- we don't currently handle external schema that reference other external schema - we only fetch from the URLs found in the parent schema~~
- currently we throw an error if any of the external schema cannot be fetched. We can choose handle this with more nuance in the future (e.g. proceeding with an incomplete schema, but notifying the user).

This PR can be tested using provider `GCP`'s `main` branch.  Expanding and adding items to `machineDeployments.customNodeLabels` used to break the form due to external schema referenced. Now that should no longer be the case, and the schema from https://schema.giantswarm.io/labelvalue/v0.0.1 should be used for a `customerNodeLabels` item.

Screenshot 2022-12-21 at 14 15 20

### DIFF
--- a/src/components/MAPI/apps/AppList/utils.tsx
+++ b/src/components/MAPI/apps/AppList/utils.tsx
@@ -1,5 +1,3 @@
-import { RJSFSchema } from '@rjsf/utils';
-import { merge } from 'lodash';
 import { IHttpClient } from 'model/clients/HttpClient';
 import { AppsRoutes } from 'model/constants/routes';
 import * as applicationv1alpha1 from 'model/services/mapi/applicationv1alpha1';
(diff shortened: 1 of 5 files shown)


gusevda (approved):
Fantastic job!

Is it correct that it's not expected for external schemas to also have references to other schemas?

kuosandys:
> Fantastic job!
> 
> Is it correct that it's not expected for external schemas to also have references to other schemas?

No, that's a good point - it's entirely possible that external schema reference other external schema, we just don't currently have any overly complicated schema examples 😅 

I've updated the function for external schema resolution to recursively resolve external schema references into the original parent schema's `$defs`, until all unique URLs found have been resolved. I've also added some more test cases as well for these scenarios.

PR #3973, opened 2022-12-21
Cluster app form fields suggestions

gusevda:
### What does this PR do?
In this PR:
- support for displaying schema examples was added;
- `SelectWidget` was improved to support `oneOf` schema property. It will allow us to provide custom labels for select values, e.g.:
```
"servicePriorityLabel": {
    "type": "string",
    "oneOf": [
        {"const": "lowest", "title": "Lowest"},
        {"const": "medium", "title": "Medium"},
        {"const": "highest", "title": "Highest"}
     ]
}
```

### How does it look like?
Displaying schema examples:
Screenshot 2022-12-20 at 16 59 15

### DIFF
--- a/src/components/UI/JSONSchemaForm/BaseInputTemplate/index.tsx
+++ b/src/components/UI/JSONSchemaForm/BaseInputTemplate/index.tsx
@@ -25,9 +25,16 @@ const BaseInputTemplate: React.FC<WidgetProps> = ({
     }
   };
 
+  const handleSuggestionSelect = (e: {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    suggestion: any;
+  }) => {
+    onChange(e.suggestion);
+  };
+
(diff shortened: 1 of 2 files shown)


src/components/UI/JSONSchemaForm/BaseInputTemplate/index.tsx
```diff
@@ -25,9 +25,16 @@ const BaseInputTemplate: React.FC<WidgetProps> = ({
     }
   };
 
+  const handleSuggestionSelect = (e: {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    suggestion: any;
```
kuosandys, line 30: Would it be possible to type this as a string instead of `any`?

gusevda: good catch, fixed

kuosandys (approved):

PR #3974, opened 2022-12-22
Allow to delete a group prefix in permissions overview

gusevda:
### What does this PR do?
Currently the permission inspection UI in its "Group" mode restricts the user entry to group names starting with `customer:`. Recently we started to allow for additional `Dex` connectors to be configured so the group name prefix could be different. In this PR the group name restriction was removed. We still pre-fill group name input with `customer:` prefix to remind people that most likely they need it, but it's possible to delete it.

Now it's possible to submit permissions form with any group name.

Towards https://github.com/giantswarm/roadmap/issues/1697.


kuosandys (approved):
Looks great! 👍

PR #3977, opened 2022-12-23
Update how we determine UI loading state

kuosandys:
### What does this PR do?

This PR updates the UI loading logic in components to use the new `isLoading` state returned by the `useSWR` hook with swr `v2`. Previously we determined initial loading state by the condition that the component is validating (`true` for initial load and whenever revalidation happens) and in addition that data and error are both `undefined` (to rule out revalidation).

For more explanation for this state, please see [here](https://swr.vercel.app/blog/swr-v2#isloading) and [here](https://swr.vercel.app/docs/advanced/understanding#combining-with-isloading-and-isvalidating-for-better-ux).

Closes https://github.com/giantswarm/roadmap/issues/1807.

### DIFF
--- a/src/components/MAPI/apps/AppList/index.tsx
+++ b/src/components/MAPI/apps/AppList/index.tsx
@@ -75,7 +75,7 @@ const AppList: React.FC<React.PropsWithChildren<{}>> = () => {
   const {
     data: catalogList,
     error: catalogListError,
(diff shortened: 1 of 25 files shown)


mogottsch (approved):

gusevda (approved):
Very nice!

PR #3996, opened 2023-01-12
Support various dex connectors for login

kuosandys:
### What does this PR do?

This PR adds support for selecting various dex connectors on login. Instead of calling dex with a specific `connector_id`, we now use `connector_filter` to filter the list of connectors, and present users with the appropriate selection list.

In order to correctly determine whether the logged in user is an "admin" within happa, we've previously checked whether the user has membership in the group referenced by `window.config.mapiAuthAdminGroup`. Since each identity provider has a different admin group name, this field is now an array of admin groups.

For customers:
After clicking on "Proceed to login", users will now be presented with dex's connector selection page of all customer connectors (connectors starting with `customer`).

For GS staff:
After navigating to `/admin-login`, users will now be presented with dex's connector selection page of all GS connectors (connectors starting with `giantswarm`).

Closes https://github.com/giantswarm/roadmap/issues/1845.

Verify that logging in presents you with the dex connector selection page, and that logging in using any GS connectors provides you with admin features within happa (i.e. you can see internal app catalogs; you have the "Experiments" selection available to you in the dropdown menu located at the top right corner of the screen).


mogottsch (approved):
I could login with both connectors and had admin features :+1:

anvddriesch (approved):

anvddriesch (approved):
I don't feel too confident reviewing the code but it worked for me, too, when I tried it :)

gusevda (approved):
Looks good to me!

PR #3999, opened 2023-01-13
Replace old testing installation in README.md

mogottsch:
### What does this PR do?

Replaces references to old `ginger` cluster with `gauss`

None.

`ginger` was torn down.

No.

No.


kuosandys (approved):
Nice 👍 Thanks for being vigilant!

PR #4008, opened 2023-01-18
Test schema for JSON schema form

gusevda:
### What does this PR do?

- test schema with all the supported features was added to the json schema form prototype;
- `asterisk` sign was added to the required fields labels.

### How does it look like?

Test schema is selected:
Screenshot 2023-01-18 at 16 09 44

Provider schema is selected:
Screenshot 2023-01-18 at 16 09 52

Asterisk sign in the required field label:
Screenshot 2023-01-18 at 15 54 12

### DIFF
--- a/src/components/MAPI/clusters/CreateCluster/CreateClusterAppForm.tsx
+++ b/src/components/MAPI/clusters/CreateCluster/CreateClusterAppForm.tsx
@@ -127,6 +127,11 @@ enum FormDataPreviewFormat {
   Yaml,
 }
 
+// TODO: replace test schema URL with the correct one
+// after merging prototype branch into the main branch
+const testSchemaURL =
+  'https://raw.githubusercontent.com/giantswarm/happa/cluster-app-creation-form-prototype/src/components/UI/JSONSchemaForm/test.schema.json';
+
 const Prompt: React.FC<React.PropsWithChildren<{}>> = ({ children }) => {
   return <Line prompt={false} text={children} />;
 };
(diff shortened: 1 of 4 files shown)


gusevda:
@kuosandys @marians I think it could be useful to display a link to the schema that is used to render a form. It can be used to see how some properties should be configured. We can display the link below the form near the "Form data preview" section. What do you think?

kuosandys (approved):
Nice! Visually distinguishing the required fields is also a good touch 👍 

Regarding your question about adding a link to the schema used to generate the form - I think it's a great idea. Not an expert on UX but maybe it would be more intuitive to place it near the top where we select the provider and branch?

marians:
Looks good for now!

I agree that showing the link to the schema source would fit best right where it's selected.

marians:
One suggestion: Let's make the test schema conformant to our schema requirements. That means:

- Run it through `schemalint normalize`.
- Make it use the latest draft 2020-12. See https://github.com/giantswarm/rfc/pull/55/files#diff-db8a80d0b569aff6d936c759c43c3406d9766b3d93365a5722bad029836a34f8R51

marians:
Another suggestion: Could we introduce top level objects along the lines of

- String properties
- Numeric properties
- Boolean properties

so that navigating the schema becomes easier?

marians:
And another thought: Should the test suite schema also contain things we do not support? Examples:

- Property with multiple `type` values
- Object property without `properties` and `additionalProperties`

In our generated form these should not be displayed.

Then, do we want an example of a property using `anyOf`, demonstrating our treatment of using the first member that is not deprecated?

More things we could showcase:

- Description that contains markdown or HTML (which must not be interpreted).
- Description that is very long
- Title that is very long (so that it breaks into two lines)
- Property without a `title` and with a very long property name (which could extend the width of the column)

gusevda:
> And another thought: Should the test suite schema also contain things we do not support? Examples:
> 
> * Property with multiple `type` values
> * Object property without `properties` and `additionalProperties`
> 
> In our generated form these should not be displayed.
> 
> Then, do we want an example of a property using `anyOf`, demonstrating our treatment of using the first member that is not deprecated?
> 
> More things we could showcase:
> 
> * Description that contains markdown or HTML (which must not be interpreted).
> * Description that is very long
> * Title that is very long (so that it breaks into two lines)
> * Property without a `title` and with a very long property name (which could extend the width of the column)

Some of these I can add. Some are not supported yet, e.g.
>  ... a property using `anyOf`, demonstrating our treatment of using the first member that is not deprecated
I suggest we add it to the test schema when this functionality is implemented.

marians:
@gusevda Just a note: None of my comments above should block this PR. Feel free to merge this any time.

marians (approved):

PR #4013, opened 2023-01-19
UI schema for cluster app creation form

kuosandys:

This PR adds a way for us to define UI schema to be used in the cluster app creation form.

Currently, UI schema are defined separately for each provider, with one UI schema for each major version. This is based on [previous discussions](https://github.com/giantswarm/roadmap/issues/1181#issuecomment-1340797851) we had about breaking changes in cluster apps, and we can adjust the granularity of this support if required.

A preliminary UI schema has been defined for each provider (including `Azure`, a new addition). More information about what can be defined via UI schema may be found in the [docs](https://react-jsonschema-form.readthedocs.io/en/latest/api-reference/uiSchema/), but the most relevant functionality for us would be:
- the ability to define the ordering of object properties
- overriding the default form component with our own React component
- making a field as read-only/disabling editing
- overriding title and/or description
- autofocus on fields

### DIFF
--- a/src/components/MAPI/clusters/CreateCluster/schemaUtils.ts
+++ b/src/components/MAPI/clusters/CreateCluster/schemaUtils.ts
@@ -53,11 +53,6 @@ const uiSchemaProviderCloudDirector: Record<string, GenericObjectType> = {
         '*',
       ],
     },
-    clusterName: {
-      'ui:options': {
-        widget: ClusterNameWidget,
-      },
-    },
   },
 };
 
@@ -133,17 +128,30 @@ export const getDefaultFormData = (
   organization: string
 ) => {
   switch (provider) {
-    case 'VSphere':
-      return {
-        cluster: {
-          name: generateUID(5),
-        },
-      };
-    default:
+    case 'AWS':
+    case 'Azure':
+    case 'GCP':
+    case 'Open Stack':
       return {
         clusterName: generateUID(5),
         organization,
       };
+
+    case 'Cloud Director':
+      return {
+        organization,
+      };
+
+    case 'VSphere':
+      return {
+        cluster: {
+          name: generateUID(5),
+          organization,
+        },
+      };
+
+    default:
+      return {};
   }
 };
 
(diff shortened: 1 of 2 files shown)


src/components/MAPI/clusters/CreateCluster/schemaUtils.ts
```diff
@@ -0,0 +1,173 @@
...
+        'controlPlane',
+        '*',
+      ],
+    },
+    clusterName: {
+      'ui:options': {
+        widget: ClusterNameWidget,
+      },
+    },
+  },
```
kuosandys, line 23: An example of a UI schema for `cluster-aws`, defined for major version `v0`

src/components/MAPI/clusters/CreateCluster/schemaUtils.ts
```diff
@@ -0,0 +1,173 @@
...
+    'ui:options': {
+      order: [
+        'clusterDescription',
+        'cluster',
+        'organization',
+        'controlPlane',
+        '*',
+      ],
+    },
+    clusterName: {
```
gusevda, line 56: Do we need `clusterName` here? I don't see it in the CloudDirector schema.

src/components/MAPI/clusters/CreateCluster/schemaUtils.ts
```diff
@@ -0,0 +1,173 @@
...
+      return {
+        cluster: {
+          name: generateUID(5),
+        },
+      };
+    default:
+      return {
+        clusterName: generateUID(5),
+        organization,
+      };
```
gusevda, line 146: What about CloudDirector schema? There is no `clusterName` field, or am I missing something?

src/components/MAPI/clusters/CreateCluster/schemaUtils.ts
kuosandys: No we don't - missed this one

src/components/MAPI/clusters/CreateCluster/schemaUtils.ts
kuosandys: Good catch - updating the cases here 👍

gusevda (approved):
Looks great to me!

PR #4018, opened 2023-01-20
Json schema form empty values

gusevda:
### What does this PR do?
In this PR, empty values are being omitted from the form data, including:
- empty arrays;
- empty objects;
- NaN values;
- undefined values;
- null values.

I didn't find such functionality in the `react-jsonschema-form` library, so it happens separately. `formData` object that is passed to the library still contains everything, and there is a new `cleanFormData` object that is used on form submit and in form data preview.

Also, `BaseInputTemplate` component was fixed, so when there is an empty input - value will be `undefined` instead of `''`. If in some cases we want an empty value to be an empty string, we would need to configure it via `uiSchema`, e.g.:
```
    "clusterDescription": {
        "ui:emptyValue": ""
    }
```

### How does it look like?

Before:
Screenshot 2023-01-20 at 15 57 58

After:
Screenshot 2023-01-20 at 15 58 29


kuosandys (approved):
Works great 👍  thanks for explaining the changes to `BaseInputTemplate`!

PR #4034, opened 2023-01-25
Move cluster creation prototype

kuosandys:
### What does this PR do?
Screenshot 2023-01-24 at 16 42 28

This PR moves the cluster creation prototype to the path `/organizations/ORGANIZATION/clusters/new-via-app-bundles`. While the only link to this is from the "Experiments" page (accessible only by admins i.e. GS staff), the form is also accessible to non-admins. This decision was made with the aim of facilitating conversations with customers about the new cluster creation process in the web UI.

Screenshot 2023-01-24 at 16 41 23

The provider and branch selections are also persisted in the path in the form of `/organizations/ORGANIZATION/clusters/new-via-app-bundles/PROVIDER/BRANCH`. Note that omitting the provider and branch from the path defaults to the loading of the test schema (same as the defaulting behaviour now).

Towards https://github.com/orgs/giantswarm/projects/270/views/1?pane=issue&itemId=6196449.

### Reviewers:
As we wanted to merge the prototype branch into the main branch after the merging of this PR, please think about whether the page and path names proposed are appropriate.


gusevda (approved):
Looks great to me! I like how it defaults to the test schema if additional URL parameters are omitted!

Regarding path names, I would like to see provider names in the URL in a "kebab-case", so the
```
/organizations/giantswarm/clusters/new-via-app-bundles/Open%20Stack/main
```
becomes
```
/organizations/giantswarm/clusters/new-via-app-bundles/open-stack/main
```
what do you think?

marians (approved):

PR #4040, opened 2023-01-26
Add JSON schema form visual improvements

gusevda:
### What does this PR do?

Visual improvements for hierarchical object and array fields according to the [design specs](https://www.figma.com/file/DsMPSXboZwWQjEryN5Y8wv/Happa-generated-hierarchical-form?node-id=0%3A1&t=RSAys5SdSeNqhUUC-0).

### How does it look like?

Screenshot 2023-01-26 at 14 15 53

### DIFF
--- a/src/components/UI/JSONSchemaForm/ArrayFieldItem/index.tsx
+++ b/src/components/UI/JSONSchemaForm/ArrayFieldItem/index.tsx
@@ -1,0 +1,62 @@
+import { ArrayFieldTemplateItemType } from '@rjsf/utils';
+import { Box } from 'grommet';
+import { normalizeColor } from 'grommet/utils';
+import React from 'react';
+import styled from 'styled-components';
+
+import ArrayFieldItemActions from '../ArrayFieldItemActions';
+
+const ArrayFieldItemW = styled(Box)`
+  position: relative;
+
+  &::before {
+    content: '';
+    position: absolute;
+    left: 0;
+    top: 10px;
+    bottom: 10px;
+    border-left: 1px solid ${({ theme }) => normalizeColor('border', theme)};
+  }
+`;
+
+const ArrayFieldItemActionsWrapper = styled.div`
+  display: flex;
+  justify-content: center;
+  flex-shrink: 0;
+  width: 35px;
+  padding-top: 10px;
+`;
+
+const ArrayFieldItem: React.FC<ArrayFieldTemplateItemType> = ({
+  disabled,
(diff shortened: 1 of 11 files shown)


marians:
It would be nice to have an example in our test schema where array items are objects with a title. Like this:

```json
                "arrayOfObjectsWithTitle": {
                    "items": {
                        "properties": {
                            "age": {
                                "type": "number"
                            },
                            "name": {
                                "type": "string"
                            }
                        },
                        "type": "object",
                        "title": "Array item object",
                        "description": "Description of array item object"
                    },
                    "title": "Array of objects with title",
                    "type": "array"
                },
```

marians:
Why is there an asterisk on the generated object title `arrayOfObjects-0` here?

image

gusevda:
> Why is there an asterisk on the generated object title `arrayOfObjects-0` here?

The library mark array items as required for some reason. I would need to check what we can do about it.

gusevda:
> It would be nice to have an example in our test schema where array items are objects with a title.

I added an example into the test schema. When an array item has custom title, it's being used without index postfix. If we still want to have index added, I can look into it. It's better to do it in a separate ticket, I think.

src/components/UI/JSONSchemaForm/ArrayFieldItem/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+  hasMoveUp,
+  hasRemove,
+  hasToolbar,
+  readonly,
+  onDropIndexClick,
+  onReorderClick,
+  children,
+}) => {
+  return (
+    <ArrayFieldItemW direction='row'>
```
kuosandys, line 43: Just a small suggestion - could we have a more descriptive name than `ArrayFieldItemW` ? I assumed the `W` stands for wrapper, but I wasn't sure if the abbreviation was intentional

kuosandys (approved):
Works well for me 👍 My comment above about the naming shouldn't be blocking

src/components/UI/JSONSchemaForm/ArrayFieldItem/index.tsx
gusevda: Thank you for noticing, it's definitely unintentional. Will fix it.

marians (approved):
Good improvement as it is!

marians:
Referencing the original issue: https://github.com/giantswarm/roadmap/issues/1181

PR #4053, opened 2023-01-31
JSON schema form visual improvements

gusevda:
### What does this PR do?

1. Index postfix `No. 1`, `No. 2` was added to the array field item titles.
2. Checkbox styles were changed so the label is displayed inline with the checkbox.

3. Several visual improvements were added:
- Clickable area of accordion labels is now limited to the content of a label. Before, the whole header row was clickable.
- Clickable area of an array "Add item" button was fixed. It used to take the whole width of the container.
- Fields labels and descriptions were refactored so their templates are reused between components.

### How does it look like?
Array field when items don't have a title:
Screenshot 2023-01-31 at 16 18 24

Array field when items have a title:
Screenshot 2023-01-31 at 16 20 45

Boolean field:
Screenshot 2023-02-01 at 15 26 50


marians:
Just noticed that in this version title and description of boolean fields are not shown. Not sure it's in this branch or in the main branch

image

Here is the latest release in comparison:

image

gusevda:
As part of our experimentations with custom widgets we added `ToggleWidget` that can be used for `boolean` fields instead of simple checkbox. Currently it wasn't used anywhere and I'm thinking if we should keep it or remove it.
Here is how it looks like
Screenshot 2023-02-01 at 10 57 51
and to use it, this `uiSchema` settings should be set:
```
disabled: {
        'ui:widget': 'toggle',
        'ui:options': {
          label: false,
        },
},
```

@marians, @kuosandys what do you think?

marians:
We haven't made our own spec for this. Anyway, I think that a checkbox is better suited than the toggle control. The latter makes me wonder whether e. g. I can simply click it, or have to click and drag. +1 for using the checkbox. However I can't figure out yet what is title, description, and what happens if both are missing.

gusevda:
Ok, thanks, I'm going to delete `ToggleWidget` from the code base. @marians regarding title and description, here is how it looks like for this schema (first field is boolean without title and description, second one - with title and description)
```
"active": {
    "type": "boolean"
},
"enabled": {
    "description": "Boolean field with title and description.",
    "title": "Enabled",
    "type": "boolean"
}
```
Screenshot 2023-02-01 at 15 26 50

marians:
That looks good to me!

kuosandys (approved):

PR #4062, opened 2023-02-02
JSON schema form number field improvements

gusevda:
### What does this PR do?

Our `NumberPicker` component behaves incorrectly when the input is empty:
1.  the component displays `0` as default value, which could be misleading.
2. if a user manually clears the input, `+` and `-` buttons stop working.

In this PR:
- the behaviour of `NumberPicker` component was changed to be more aligned with browser's default `<input type="number" />` component.
- `BaseInputTemplate` component was fixed to process empty values for number fields.


kuosandys (approved):
👍

PR #4080, opened 2023-02-08
Use idPrefix and idSeparator only from formContext

gusevda:
### What does this PR do?

Recently we started to use `formContext` to pass some values between form components including `idPrefix` and `idSeparator`. But in some places these values are still being imported from the `utils` directly. In this PR I did some refactoring to take these values from the `formContext` everywhere. And this way we are aligned with the `react-jsonschema-form` library's default behaviour, when a user can provide different `idSeparator` or `idPrefix` values via the props.


gusevda:
@kuosandys in [this commit](https://github.com/giantswarm/happa/pull/4080/commits/1a1d392a18c76cc8a20aa2985d354747962eec52) I changed `formContext` type when it comes from props. In our case it can never be `undefined` so we can have less checks in the code. Do you think it's a valid improvement? I'm not sure.

kuosandys:
> @kuosandys in [this commit](https://github.com/giantswarm/happa/pull/4080/commits/1a1d392a18c76cc8a20aa2985d354747962eec52) I changed `formContext` type when it comes from props. In our case it can never be `undefined` so we can have less checks in the code. Do you think it's a valid improvement? I'm not sure.

I think in our case it makes sense 👍 It makes the code easier to read and more ergonomic

kuosandys (approved):
Thanks for keeping an eye out for opportunities to improve the codebase!

PR #4084, opened 2023-02-09
Allow removing empty objects and arrays from form fields while editing

kuosandys:
### What does this PR do?

This PR changes how we process form data to allow removing empty objects and arrays from the form data while editing. This fixes the bug we saw where optional object fields with required properties resulted in an empty object that the user had no way of clearing without refreshing the page.

The `cleanDeepWithException` function extends the `cleanDeep` function we use for form cleaning on submission, by allowing us to specify exceptions to the "cleaning". This is necessary to make a distinction between truly empty arrays `[]` and arrays with uninitialized values (e.g. `[undefined]`, `[{}]`), otherwise any array form field containing uninitialized values will be removed, and as a result we won't be able to add items to array form fields.

This PR also includes a small improvement for how we display validation errors after form submission - previously, we displayed all errors (existing and new) after submission. Now, we only display the existing errors. This aligns the form editing experience before and after submission.

#### Bug concerning arrays with `minItems` defined
Currently, the default form data generated will always populate arrays with items when `minItems` is defined, and it does not take into account whether the array field is required. This means we cannot reliably clean it from the form data, and this should be fixed upstream.

Until this is fixed, users cannot remove items from an array if it has the minimum number of items. To communicate this and to avoid confusion, we disable the "Delete item" button from the array item menu in this case:
Screenshot 2023-02-09 at 15 50 36


gusevda (approved):
Looks good to me! Good idea to reuse `cleanDeep` function that we use on form submission!

Have you thought about adding some tests for the `cleanDeepWithException` function?

gusevda (approved):
Looks great!

PR #4095, opened 2023-02-14
JSON schema form field improvements: multi-line string input and long titles/descriptions

kuosandys:
### What does this PR do?

This PR adds some improvements to the JSON schema form UI component in the following ways:
#### 1. Long titles and descriptions are wrapped, limited to the same max-width as the input field
Screenshot 2023-02-14 at 16 50 32

#### 2. Multi-line string input is supported by replacing `<input type="text"/>` with `<textarea/>`

As any string field as defined in JSON schema could potentially include new line characters, we opted to support this feature for all string fields. The appearance of the input fields remain the same as before, until a multi-line string is inputed.
Screenshot 2023-02-14 at 16 54 26

Screenshot 2023-02-14 at 15 53 19

The height is limited to ~10 rows, with horizontal and vertical scrolling enabled if the content does not fit. To align with the default appearance of input fields from `grommet`, scrollbars have also been hidden.

Multi-line string input results in the following form data, as show in the previews:
Screenshot 2023-02-14 at 15 53 48
Screenshot 2023-02-14 at 15 53 36

Closes https://github.com/giantswarm/roadmap/issues/2005
Towards https://github.com/giantswarm/roadmap/issues/1181


src/components/UI/JSONSchemaForm/BaseInputTemplate/index.tsx
```diff
@@ -55,17 +53,18 @@ const BaseInputTemplate: React.FC<WidgetProps> = ({
       onBlur={handleBlur}
     />
   ) : (
-    <TextInput
+    <TextArea
```
gusevda, line 56: I think we still need to support basic input field for some cases. It can be rendered for `number` fields (when type is `number` in schema). I guess it also can be rendered for some other input types, like `email`, `password`, etc. 

Maybe we can render `textarea` only when `inputProps.type === 'text'`?

gusevda (changes requested):
One more thing - I see that there are two borders around `textarea` inputs.

marians (approved):
LGTM!

gusevda (approved):
Looks great!

PR #4102, opened 2023-02-16
Preprocess schema

gusevda:
### What does this PR do?

This PR adds support for `oneOf` and `anyOf` keywords with our "own treatment". To simplify the UI, we decided to restrict how `oneOf` and `anyOf` keywords can be used. They can be used for two use cases:
1. to specify validation constraints via subschemas;
2. to declare one or more subschemas as deprecated.

For the first use case, we don't need additional controls in the UI, because possibility to change a field's type is not needed to handle validation constraints. In this PR, I render empty component instead of the default "subschema selection" component to simplify the UI.

For second use case - I introduced a preprocessing step for the schema. In this step, for any object that has `oneOf` or `anyOf` keywords, I select first not deprecated subschema, and merge it with the object's schema. This way, an object with multiple subschemas becomes a simple object without `oneOf` or `anyOf` keywords.

I also fixed our `traverseJSONSchemaObject` helper function that we use to traverse a schema. Previously it was performing an action only for "leaf" nodes of the schema, now it does it for every node.

### How does it look like?
Subschemas defined with anyOf
Before:
Screenshot 2023-02-16 at 16 53 33

After:
Screenshot 2023-02-16 at 16 54 12

Subschemas used for validation purposes
Before:
Screenshot 2023-02-16 at 16 53 16

After:
Screenshot 2023-02-16 at 16 52 23


kuosandys (approved):
Looks great! Thanks for adding tests as well 🙂

PR #4106, opened 2023-02-17
Refactor UI schema and initial form data

kuosandys:
### What does this PR do?

This PR refactors how we specify UI schema and initial form data. Previously, we've added the functionality to determine the correct UI schema to use, based on a given cluster app's version. The initial form data is also tightly coupled to the cluster app's structure, and thus should also be determined based on the cluster app's version.

I've also aligned how we specify UI schema to use the `ui.NAME` syntax instead of `ui:options.NAME`. While these two syntaxes are equivalent ([more info](https://rjsf-team.github.io/react-jsonschema-form/docs/api-reference/uiSchema#uixxx-or-uioptionsxxx)), the first yields a more compact structure, and at a glance provides an easier way to distinguish UI schema properties from schema properties.

Towards https://github.com/giantswarm/roadmap/issues/1181


gusevda (approved):
Looks good!

PR #4112, opened 2023-02-21
Support removing schema fields

kuosandys:
### What does this PR do?

This PR adds support for removing schema fields during the schema preprocessing step. This allows us to specify fields to omit from JSON schema for both the form data and the form UI.

An example use case for this functionality is the `internal` field for cluster app schema. The removal of this field during preprocessing has also been implemented by this PR.

Towards https://github.com/giantswarm/roadmap/issues/2052 and https://github.com/giantswarm/roadmap/issues/1181.

### DIFF
--- a/src/components/MAPI/clusters/CreateCluster/__tests__/schemaUtils.ts
+++ b/src/components/MAPI/clusters/CreateCluster/__tests__/schemaUtils.ts
@@ -1,6 +1,8 @@
+import { RJSFSchema } from '@rjsf/utils';
 import cleanDeep, { CleanOptions } from 'clean-deep';
+import testSchema from 'UI/JSONSchemaForm/test.schema.json';
 
-import { cleanDeepWithException } from '../schemaUtils';
+import { cleanDeepWithException, preprocessSchema } from '../schemaUtils';
 
 describe('schemaUtils', () => {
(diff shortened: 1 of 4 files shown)


src/components/MAPI/clusters/CreateCluster/__tests__/schemaUtils.ts
```diff
@@ -1,6 +1,8 @@
+import { RJSFSchema } from '@rjsf/utils';
 import cleanDeep, { CleanOptions } from 'clean-deep';
+import testSchema from 'UI/JSONSchemaForm/test.schema.json';
```
gusevda, line 3: Maybe we can mock the test schema somehow? I think test schema will be modified often, and it will break the tests unexpectedly.

gusevda (approved):
Looks good to me. Two small comments that should not block merging.
Thanks for adding tests for schemaUtils!

src/components/MAPI/clusters/CreateCluster/__tests__/schemaUtils.ts
kuosandys: Sounds like a good idea, I'll create a separate mock schema for tests 👍

PR #4116, opened 2023-02-22
Allow hiding schema fields

kuosandys:
### What does this PR do?

This PR allows hiding schema fields in the JSON schema form UI via the `ui:widget: hidden` UI schema configuration. Note that this differs from the mechanism in https://github.com/giantswarm/happa/pull/4112, in that it only hides the display of the field but still allows values for that field in form data.

Towards https://github.com/giantswarm/roadmap/issues/2052 and https://github.com/giantswarm/roadmap/issues/1181


gusevda (approved):
Looks good

PR #4117, opened 2023-02-22
Fix displaying values passed to TextArea component

kuosandys:
### What does this PR do?

Fixes a bug introduce in https://github.com/giantswarm/happa/pull/4095 so the value passed to `<TextArea />`  component is actually displayed.

No tracking issue, we saw this during refinement.


gusevda (approved):
Looks good. Thank for a quick fix!

PR #4119, opened 2023-02-23
Use provider flavour for CAPI-specific logic

kuosandys:
### What does this PR do?

This PR uses the provider "flavour" to determine whether an installation is a CAPI installation. Previously we derived this information from the provider kind, but as we now have a field we can use from `config`, we refactor instances where we check whether the provider is CAPI to use the value exposed via `window.config.info.general.providerFlavor`. Exposing this value also helps us to determine whether we should offer cluster creation via cluster resources or via app bundles.

Closes https://github.com/giantswarm/roadmap/issues/2029, towards https://github.com/giantswarm/roadmap/issues/1181


gusevda (approved):
Looks good!

PR #4132, opened 2023-03-01
Preprocess default values for the JSON schema form

gusevda:
### What does this PR do?
Our current approach when we clean form data on every change conflicts with how `react-jsonschema-form` library treats default values defined in a schema. If no value for a field entered by a user and there is default value for the property in a schema, this value is being automatically populated into the form data. It causes a problem when a user deletes a value from a field that has default value (or deletes all items from an array field that has default items). For example, we have a default item in a schema for an array field, and user deletes this item, we get an empty array that we immediately  delete from the form data during our form data cleaning process (on every form data change). On the next render, `react-jsonschema-form` library detects absence of the array field, and automatically adds it back with the default value. We get in a situation when it's not possible to delete default values.

In this PR I changed how the default values defined in a schema are treated. Before the first render of a form, I extract all the default values defined in a schema and put them into the initial form data, `default` definitions are being completely deleted from the schema. As a result, we have all the default values initially available in the form data, but in future manipulations with the form, default values play no role.

Fixes https://github.com/giantswarm/roadmap/issues/2063.


kuosandys:
Works great 👍 I was wondering if it would be possible to combine schema default values preprocessing with the other preprocessing steps? That way it's a bit easier to keep track of all the preprocessing we're doing to the schema

gusevda:
> Works great 👍 I was wondering if it would be possible to combine schema default values preprocessing with the other preprocessing steps? That way it's a bit easier to keep track of all the preprocessing we're doing to the schema

Thanks for bringing it up. I wanted to suggest to separate our current preprocessing even further. Currently it happens in the `CreateClusterAppBundles` component, but it contains preprocessing steps that are necessary for any schema. When we use `JSONSchemaForm` component for other apps for example, we would most probably need to duplicate this logic. My suggestion is to move preprocessing steps that are critical for `JSONSchemaForm` to the `JSONSchemaForm` level. There is one preprocessing step, where we delete some schema properties (`internal` at the moment) that seems like specific to cluster apps form implementation, and I think it's ok to keep it on the CreateClusterAppBundles level (maybe we'll have other preprocessing steps in the future that only applied for cluster apps). But, we could also move this step and pass fields that should be deleted as a prop to `JSONSchemaForm` component. I also like the idea to keep this logic in one place. 

The same suggestion is applied for form data cleaning that happens on every change. I think it's generic functionality that will be applied for other apps as well, and we should move it to the JSONSchemaForm level.

@kuosandys  What do you think?

kuosandys (approved):

PR #4134, opened 2023-03-02
JSON schema form select boxes empty option

gusevda:
### What does this PR do?
In this PR, an empty option was added to select boxes, that are rendered for fields with enums. This allows a user to unset a field after a value was selected.

### How does it look like?

Before:
Screenshot 2023-03-02 at 17 20 32

After:
Screenshot 2023-03-02 at 17 20 10

Closes https://github.com/giantswarm/roadmap/issues/2065.

### DIFF
--- a/src/components/UI/JSONSchemaForm/SelectWidget/index.tsx
+++ b/src/components/UI/JSONSchemaForm/SelectWidget/index.tsx
@@ -1,10 +1,13 @@
 import { EnumOptionsType, WidgetProps } from '@rjsf/utils';
-import React from 'react';
+import { Box } from 'grommet';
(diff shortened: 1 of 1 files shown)


kuosandys:
Works well, though I was wondering if the reduced line height of the empty option is deliberate? To me it looks a bit out of alignment with the other options, especially on hover.

marians:
I tested and find the height of 30 to appear a bit too flat. 35 works for me. 40 would also work for me, but in fact it appears like a gapibng hole, so I consider 35 ox better.

marians (approved):
Unblocking

kuosandys (approved):

PR #4135, opened 2023-03-02
Refactor NumberPicker component

gusevda:
### What does this PR do?
There are two problems related to `NumberPicker` component and JSON schema form:

Form widget treats invalid values not correctly. How to reproduce: 
1. Open test schema form -> Numeric fields -> Integer with min/max values.
2. Put value `5` into the input. Check that `integerLimit` is present in the form data and equals 5.
3. Click `+` button. `integerLimit` value is set to `undefined` and deleted from form data.

Value inside `NumberPicker` component state can be out of sync with the `value` prop. How to reproduce:
1. On local machine in development mode.
2. Open test schema form -> Numeric fields -> Integer. Ensure that the input is empty and there is no `integer` value in the form data.
3. Go to `test.schema.json` in code editor and set any default value for the `.numericFields.integer` property. e.g. `"default": 5`.
4. Go back to the browser and without page reload you can see that `integer` value is present in form data, but the form field is still empty.

In this PR these two problems were fixed.

There is no issue tracking this problems.


kuosandys:
Thanks for taking the time to refactor and to fix the bugs in the `NumberPicker` widget 👍 However, I think the changes causes a different bug in the node pool creation form > scaling range configuration. Steps to reproduce:
1. Click on "Add node pool"
2. Under "Scaling range", change the minimum value to something valid.
3. The scaling range is invalid, and the "Create" button is disabled:
Screenshot 2023-03-03 at 19 39 54

It's also invalid if I only change the maximum value. However if I change both values, then the range is valid. Maybe it's caused by removing the `usePrevious` hooks?

gusevda:
@kuosandys thanks for catching it, will try to fix

kuosandys (approved):
Just tested again, and everything works as described 👍

PR #4144, opened 2023-03-08
Support creating cluster creation based on app bundles

kuosandys:
### What does this PR do?

In order to test cluster creation via app bundles end-to-end, we need to add the functionality to create app resources, specifically:
- cluster app App CR (from `cluster-PROVIDER` AppCatalogEntries)
- cluster app userconfig ConfigMap CR
- default apps bundle App CR (from `default-apps-PROVIDER` AppCatalogEntries) 
- default apps bundle userconfig ConfigMap CR

This PR adds the capability to template and create the above resources. These resources are created from AppCatalogEntries available in a production environment, and the values schema used to template the cluster creation form is read from the AppCatalogEntry's values schema annotation. This differs significantly from how we have been fetching values schema for our schema tester form, and as such we create a separate component for the creation form, and share it between the cluster app schema tester page and the actual cluster creation page.

It is now possible to create clusters via app bundles on CAPI installations by clicking on the "Launch New Cluster" button. Note that this is currently only visible to admins (GS staff; there is no change for non-admin users who do not see this button). The JSON output of the form data preview has also been removed according to https://gigantic.slack.com/archives/C02FJACQ3D4/p1676468609464159

Regarding cluster app / default apps versions:
As discussed in https://github.com/giantswarm/roadmap/issues/2031, until we can safely determine compatible cluster app and default apps versions, we will simply select the latest version of each AppCatalogEntry for creation.

The schema tester can still be found in the "Experiments" section, or at the route `account-settings/experiments/cluster-app-schema-tester`.

Towards https://github.com/giantswarm/roadmap/issues/1181, and closes https://github.com/giantswarm/roadmap/issues/2069


gusevda (approved):
Looks awesome!

PR #4154, opened 2023-03-10
Add support for CAPZ resources

kuosandys:
Closes https://github.com/giantswarm/giantswarm/issues/26160.

In order to support CAPZ in the web UI, we first need to fetch CAPZ resources. This PR adds the types and client functions necessary to do so, and updates the utility functions we use to fetch these resources. Permission hooks have also been updated to determine permissions with these new resources.


src/components/MAPI/utils.ts
```diff
@@ -1460,7 +1554,7 @@ function getInstallationBaseURL(providerFlavor: ProviderFlavors): URL {
   return audienceURL;
 }
 
-export function isCAPZCluster(cluster: Cluster): boolean {
+export function isCAPZAlphaCluster(cluster: Cluster): boolean {
   if (cluster.spec?.infrastructureRef?.kind !== capzv1beta1.AzureCluster) {
     return false;
   }
```
kuosandys, line 1560: Making a distinction here that this function tests whether a cluster is an "alpha" CAPZ cluster (created from the `v20.0.0-alpha` release and found on vintage clusters)

gusevda (approved):
Looks good

PR #4165, opened 2023-03-16
JSON schema form payload cleaning

gusevda:
### What does this PR do?

This PR introduces payload cleaning based on default values to the JSON schema form. The new cleaning is applied only for the final payload that is being submitted and displayed in the "Form data preview" below the form. Form data that is used by `react-jsonschema-form` library stays the same as it was before, this way we can see the default values in the form while editing.

Additionally, a few refactoring changes were added:
- schema preprocessing steps were moved from the `CreateClusterAppBundlesForm` to the `JSONSchemaForm` component, since these steps are not specific to cluster apps;
- a new `SimpleNumberPicker` component was added. It's very similar to the `NumberPicker` component we have but has few differences: it's controlled and doesn't manage its own value in a state; it doesn't perform its own validation of the value. This changes are important for the `JSONSchemaForm` performance since they decrease amount of heavy form change events. In the future I suggest we consider possibility to reuse `SimpleNumberPicker` in other places instead of `NumberPicker`;
- `CheckboxWidget` component were fixed to handle value prop.

### How does it look like?

Payload preview for the Azure schema form before the changes:
Screenshot 2023-03-16 at 13 41 34

... after the changes:
Screenshot 2023-03-16 at 13 40 06

Closes https://github.com/giantswarm/roadmap/issues/2057.


kuosandys (commented):
Thanks for moving the form data preprocessing to a more suitable place! Also, really awesome that you've included documentation for the new UI components 👍 

I have a concern regarding the payload cleaning, as it did not function quite as I had expected from reading the linked issue. Specifically the behaviour [mentioned](https://github.com/giantswarm/roadmap/issues/2057#issuecomment-1441388478):
> Case 1
> - String property with app level default value: "Test cluster"
> - User changes this to the empty string ""
> - Resulting user value should be: ""

As I understand, in order to override the default value here, we need to submit an empty string as a value for that property. Currently, we don't include the field in the submitted form data at all. I also believe we would need similar handling for arrays (i.e. if the default value of an array field contains a single item and the user removes it, we should submit an empty value).

Please let me know if I've misunderstood the intended behaviour - also happy to chat about this in a call if it's easier!

gusevda:
@kuosandys you are right, thanks for checking. I forgot to change empty values for strings from `undefined` to an `empty string`. For arrays, I added a change to use an `empty array` as an implicit default value, but there are some other issues with the array fields that I see. Will try to fix.

kuosandys (approved):
Tested again and works well 👍 Thanks for handling the empty values!

PR #4175, opened 2023-03-21
Display provider credentials for CAPZ

kuosandys:
### What does this PR do?

This PR adds support for displaying provider credentials info in the cluster details overview for CAPZ clusters. The `PROVIDER` widget should now be displaying credentials info in the same way as it does for vintage Azure.

In order to reduce duplicate implementations and to have an easier way to keep track of how credentials are fetched for each provider/provider flavour, I've moved the related logic to a utility function, similar to the ones we have for fetching other provider-specific resources. The provider credentials widget now contains only the logic required for displaying credential details. A similar abstraction has been made for the `usePermissionsForProviderCredentials` permissions hook.

Closes https://github.com/giantswarm/giantswarm/issues/26267


gusevda (approved):
Looks good to me. Thanks for putting the efforts into refactoring and tests update.

PR #4199, opened 2023-03-30
Reduce reliance on API version of CRDs

kuosandys:
### What does this PR do?

This PR reduces `happa`'s reliance on the API version of CRDs when determining provider-specific logic. Instead of checking for a match with the entire API version string on the resource (e.g. `infrastructure.cluster.x-k8s.io/v1beta1`), we check only for a match with the API group (e.g. `infrastructure.cluster.x-k8s.io`). This allows us to accommodate for arbitrary versions of resources within an API group, and rely on conversion webhooks to determine the correct version.

No change, except this should fix cluster list fetching on CAPA installations.

Closes https://github.com/giantswarm/roadmap/issues/2210


gusevda (approved):
Looks good!

PR #4204, opened 2023-04-03
Re-arrange AWS cluster creation form properties, hide some

marians:
### What does this PR do?

This PR adds configuration to improve the AWS CAPI cluster creation form by

- Hiding unwanted properties
- Sorting properties
- Customizing certain properties (instance type, cluster name)

See below

### How does it look like?

image

Issue https://github.com/giantswarm/roadmap/issues/1181

No

No

### DIFF
--- a/src/components/MAPI/clusters/CreateClusterAppBundles/schemaUtils.ts
+++ b/src/components/MAPI/clusters/CreateClusterAppBundles/schemaUtils.ts
@@ -53,24 +53,72 @@ const formPropsProviderCAPA: Record<string, FormPropsPartial> = {
   0: {
     uiSchema: {
       'ui:order': [
-        'clusterName',
-        'clusterDescription',
-        'organization',
+        'metadata',
+        'providerSpecific',
         'controlPlane',
+        'connectivity',
+        'nodePools',
         '*',
       ],
-      bastion: {
-        instanceType: {
-          'ui:widget': InstanceTypeWidget,
-        },
+      baseDomain: {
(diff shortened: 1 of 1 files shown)


src/components/MAPI/clusters/CreateClusterAppBundles/schemaUtils.ts
```diff
@@ -53,24 +53,72 @@ const formPropsProviderCAPA: Record<string, FormPropsPartial> = {
...
+        'ui:widget': 'hidden',
+      },
+      managementCluster: {
+        'ui:widget': 'hidden',
+      },
+      metadata: {
+        'ui:order': ['name', 'description', '*'],
+        name: {
+          // TODO: fix crash
+          //'ui:widget': ClusterNameWidget,
```
marians, line 109: @gusevda Can you please help me with this one? If I uncomment line 109, the app crashes ("can't read property length of undefined", if I remember correctly).

gusevda: you should also change `formData` property below to match the new structure. It should be
```
formData: (clusterName, organization) => {
    return {
        metadata: {  
            name: clusterName,
            organization,
        }
    };
},
```

marians: Thanks, that worked.

kuosandys (approved):
Looks great!

PR #4205, opened 2023-04-04
JSON schema form fixes

gusevda:
### What does this PR do?

- The logic behind touched fields was changed. Now, after some fields were unset and removed from the form data they are also being removed from the touched fields array. This way we don't display validation errors until fields are touched again.
- `isException` check in the payload cleaning function was moved and now happens after cleaning. This way we can perform cleaning for array items, which is missing in current implementation.
- `BaseInputTemplate`, `CheckboxWidget`, `SelectWidget` components were changed. Now, in case of empty input, implicit default value (`''` for strings, `0` for numbers and `false` for booleans) is used only if the field has a default value. This is needed to override the default value. If the field doesn't have a default value, `undefined` is being used as empty value. This way a field can be unset (deleted) by a user.
- A new transformation was added for validation errors. If there are errors for a field that uses `oneOf` property for validation, those errors are being combined into one - `must be equal to one of the allowed values`. The same error message is used for `enum` fields validation.

Closes https://github.com/giantswarm/roadmap/issues/2242, https://github.com/giantswarm/roadmap/issues/2243.


kuosandys (approved):
Nice improvement with the error transformation!

PR #4234, opened 2023-04-14
Display worker nodes details for CAPZ

kuosandys:
### What does this PR do?

This PR adds support for displaying details for worker nodes for CAPZ:
Screenshot 2023-04-14 at 16 55 03

Autoscaling min/max configuration is currently not supported for CAPZ `MachineDeployments` and as such, these columns have been hidden from the web UI in the same way as we do for CAPG.

Closes https://github.com/giantswarm/giantswarm/issues/26609, towards CAPZ support in the web UI.


gusevda (approved):
Looks good!

PR #4246, opened 2023-04-19
Increase CircleCI resource class for building FE docs

kuosandys:
The `build-fe-docs` job exits in CircleCI with code 137 - Out of memory. This PR bumps the resource class used to build FE docs from the default `medium` to `large`.


gusevda (approved):
LGTM

PR #4253, opened 2023-04-20
Include cluster ID in default app resource name for CAPI

kuosandys:
### What does this PR do?

This PR changes the default name used when installing apps on CAPI clusters so that it is prepended with the cluster ID (e.g. when installing `demo-app` on cluster `asdf123`, the default name used will be `asdf123-demo-app`). This is necessary to avoid name collisions, as app resources for CAPI clusters are installed in the organization namespace rather than the workload cluster namespace.

The default names used for the app's ConfigMap and Secret also reflect this change.

Users should now be able to install the same app on different workload clusters in the same organization for CAPI without running into name collision issues.

Closes https://github.com/giantswarm/roadmap/issues/2371.


gusevda (approved):
Looks good!

PR #4269, opened 2023-04-26
Increase NGINX header size for happa API

gusevda:
### What does this PR do?

When a user is a member of many AD groups authentication header can become bigger than a limit NGINX sets by default (8KB). As a result requests to the API are being aborted. In this PR, header size limit was increased for happa API ingress.


kuosandys (approved):

ubergesundheit (approved):
LGTM

PR #4280, opened 2023-05-02
Preprocess schema to handle pattern properties

gusevda:
### What does this PR do?

This PR adds support for schema properties with `additionalProperties` and `patternProperties` keywords to the `JSONSchemaForm` component. `react-jsonschema-form` library doesn't support `patternProperties` at the moment, so this functionality was implemented with additional schema preprocessing. Schema properties that contain `patternProperties` or `additionalProperties` are being transformed into array properties, that are already supported by the `JSONSchemaForm` component. On every form change, form data that corresponds to transformed properties is being transformed back from arrays into objects.

Additionally, form data cleaning of default values was changed. Previously, default values were extracted from the schema at the beginning and were passed into `cleanPayload` function as an object. There was an issue with this approach when for some properties default values were identified incorrectly. 
For example, for such schema:
```
{
    type: 'object',
    properties: {
        objectProperty: {
            type: 'object',
            properties: {
                stringProperty: {
                    type: 'string',
                    default: 'def',
                }
            }
        }
    }
}
```
default values were extracted as:
```
{
    objectProperty: {
        stringProperty: 'def',
    }
}
```
and then, inside of `cleanPayload` function, `objectProperty` property was treated like a property with default value, which is wrong.
In this PR `cleanPayload` function was changed, and now default value for each property is being taken directly from the schema.

Closes https://github.com/giantswarm/roadmap/issues/2295.


kuosandys:
Thanks for adding support for pattern properties in the creation form! One thing I noticed while testing: In order to overwrite default values for pattern properties and additional properties, I think we should be submitting `{}` instead of `[]`.
Screenshot 2023-05-02 at 13 02 00
Please correct me if my understanding of this is wrong 😅

gusevda:
> Thanks for adding support for pattern properties in the creation form! One thing I noticed while testing: In order to overwrite default values for pattern properties and additional properties, I think we should be submitting `{}` instead of `[]`. Screenshot 2023-05-02 at 13 02 00 Please correct me if my understanding of this is wrong 😅

I think you are right, let me check that

kuosandys (approved):
Works well, thanks for looking into clearing the default fields as well!

PR #4301, opened 2023-05-10
Migrate master taint to control-plane taint

mogottsch:
### What does this PR do?

This PR replaces the deprecated `node-role.kubernetes.io/master` taint with `node-role.kubernetes.io/control-plane`.
[deprecation](https://github.com/kubernetes/enhancements/tree/master/keps/sig-cluster-lifecycle/kubeadm/2067-rename-master-label-taint)

None.

Towards https://github.com/giantswarm/roadmap/issues/2389

The affected changes are only in tests.
From what I understand, the taint was not used in the test and therefore this has no impact at all.
Please confirm that this has no unintended consequences.

No.

As it only affects tests, I think this does not have to be mentioned in the release notes.


kuosandys:
> From what I understand, the taint was not used in the test and therefore this has no impact at all.
Please confirm that this has no unintended consequences.

I can confirm that we can replace the label with no impact. Thank you for keeping an eye out for deprecated labels!

Just a question - should we remove the deprecated `node-role.kubernetes.io/master` label? I figured that was the intention, but saw that they were still present and wanted to double check.

mogottsch:
I wasn't sure about that.
In the other repositories, where there are tolerations to the taints, I saw people keeping both to keep them compatible with older versions.

In this case, I guess we could also remove the old one. What do you prefer?

kuosandys:
> I wasn't sure about that. In the other repositories, where there are tolerations to the taints, I saw people keeping both to keep them compatible with older versions.

Ah I see. There's no harm doing it either way, but let's keep it consistent across repositories.

kuosandys (approved):

PR #4392, opened 2023-06-05
Align ui-schemas

gusevda:
### What does this PR do?

Several UI schemas were not in sync with corresponding provider schemas. In this PR:
- `defaultMachinePools` setting was deleted from uiSchema for CAPA;
- UI schemas for CAPA and CAPZ were changed to use custom instance type selector in `nodePools` settings.

Additionally, the way we check schema type was improved to allow using multiple types, e.g. `["string", null]`. Previously we only checked it for numeric values, but it also in use in Cloud Director schema:

```
"kubeadm": {
      "type": [
            "null",
            "object"
      ],
      ...
},
```

Closes https://github.com/giantswarm/roadmap/issues/2511.


marians (approved):
LGTM if it works for you

PR #4560, opened 2023-11-28
Add PolicyExceptions for k8s 1.25

kubasobon:
### What does this PR do?

It adds required Kyverno PolicyExceptions, so that happa keeps running on K8s
v1.25. In the long term it would be better to apply security recommendations to
both deployments and remove the exception.


mproffitt (approved):

kubasobon:
@gusevda Can you take a look :pray:

PR #4568, opened 2024-01-12
Prepare Happa for k8s 1.25

allanger:
### What does this PR do?

I'm removing PSP and replacing it with PSS, so the chart is compatible with k8s 1.25

Users will be able to install it on k8s 1.25

Reviewers should check that I've set PSS as we expect them to be set


.circleci/config.yml
```diff
@@ -1,6 +1,6 @@
 version: 2.1
 orbs:
-  architect: giantswarm/architect@4.35.6
+  architect: giantswarm/architect@4.38.0
```
marians, line 3:
```suggestion
  architect: giantswarm/architect@5.0.0
```

helm/happa/values.yaml
```diff
@@ -1,5 +1,8 @@
 name: happa
 namespace: giantswarm
+
+registry:
+  domain: gsoci.azurecr.io
```
marians, line 5: This is redundant

```suggestion
```

marians:
CI fails with this error

```
helm/happa/values.yaml
  101:1     error    duplication of key "registry" in mapping  (key-duplicates)
```

because `.registry` was in fact already set in the values.yaml file.

allanger:
> CI fails with this error
> 
> ```
> helm/happa/values.yaml
>   101:1     error    duplication of key "registry" in mapping  (key-duplicates)
> ```
> 
> because `.registry` was in fact already set in the values.yaml file.

Yes, I see, but before it was failing with another error, probably because of an outdated archorb version

uvegla (approved):

PR #4570, opened 2024-01-20
Control plane nodes widget fix

gusevda:
### What does this PR do?

On CAPI to calculate "ready" and "total" amount of nodes for the "Control Plane" widget we go over the list of control plane nodes. Since during cluster creadtion they appear one after another we don't know the total amount in advance. This makes the widget to display not fully correct information. It could display "Control plane node ready" and after some time "1 of 2 control plane nodes ready" - see screenshot below. In this PR, I changed the way total amount of nodes is calculated. I additionally check the `KubeadmControlPlane` resource associated with the cluster, and get value from there (`.spec.replicas`).

### How does it look like?

Group 4

Closes https://github.com/giantswarm/roadmap/issues/2947.

### DIFF
--- a/.github/workflows/zz_generated.check_values_schema.yaml
+++ b/.github/workflows/zz_generated.check_values_schema.yaml
@@ -1,6 +1,6 @@
 # DO NOT EDIT. Generated with:
 #
-#    devctl@6.18.0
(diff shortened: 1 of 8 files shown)


marians (approved):
Great!

PR #4587, opened 2024-02-09
NGINX modules ndk,lua + route for teleport

ssyno:
### What does this PR do?

Relevant [issue](https://github.com/giantswarm/roadmap/issues/3220)

Update nginx server image to support NDK & LUA

Allow us to expose an endpoint `/jwt-token` via nginx, that reads the `Teleport-Jwt-Assertion` header and write as response.

### DIFF
--- a/nginx/nginx.conf
+++ b/nginx/nginx.conf
@@ -63,7 +63,6 @@ http {
       add_header Content-type text/plain;
 
       content_by_lua_block {
-        ngx.say(ngx.req.get_headers()['User-Agent'])
         ngx.say(ngx.req.get_headers()['Teleport-Jwt-Assertion'])
       }
     }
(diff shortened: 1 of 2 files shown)


nginx/nginx.conf
```diff
@@ -56,6 +59,14 @@ http {
 
     include headers.conf;
 
+    location /jwt-token {
+      add_header Content-type text/plain;
+
+      content_by_lua_block {
+        ngx.say(ngx.req.get_headers()['User-Agent'])
```
tuladhar, line 66:
```suggestion
```

tuladhar, line 66:
```suggestion
```

tuladhar (changes requested):

tuladhar (approved):

PR #4591, opened 2024-02-13
Adapt fe-docs chart for new operations cluster

marians:
Towards https://github.com/giantswarm/giantswarm/issues/29470

This PR attempts to make the fe-docs chart deployable on k8s 1.25

Included changes:

- Adds security details to values.yaml which are already applied in the chart
- Updates schema for values.yaml
- Adds ci values
- Sets explicit NGINX configuration for fe-docs
- Updates nginx-unprivileged to 1.25
- Sets separate liveness probe path

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,10 +63,10 @@ workflows:
       - architect/push-to-app-catalog:
           executor: app-build-suite
           name: push-happa-to-control-plane-app-catalog
-          app_catalog: "control-plane-catalog"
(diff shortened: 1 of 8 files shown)


marians:
FYI: I changed the CircleCI config to publish dev builds for fe-docs, too. Might revert this after testing.

marians:
The app still fails to deploy because of `mkdir() "/tmp/proxy_temp" failed (30: Read-only file system)`. I'll try to remove the need for a cache dir, since fe-docs is all static content anyway.

PR #4611, opened 2024-05-02
Update PolicyExceptions to v2beta1

fhielpos:
## Description

We need to update the PolicyExceptions apiVersion to v2beta1 as v2alpha1 is being removed in the next Kyverno release.


marians:
@fhielpos Can we close this? Happa is supposed to be retired this year.

PR #4618, opened 2024-07-18
Support releases for CAPA

gusevda:
### What does this PR do?

This PR adds support for CAPI releases to cluster creation form. If releases are available for the provider, then Release version selector is available and Cluster App version selector is hidden. If releases are not available for the provider, then Cluster App version selector is available and Release version selector is not (the same as we had it before).

### How does it look like?
Releases selector:

Screenshot 2024-07-18 at 11 35 39

Screenshot 2024-07-18 at 11 36 16

User values with `global.release.version` set to selected release:
Screenshot 2024-07-18 at 11 36 48

App manifest. If release is selected, then `spec.version` is set to an empty string:
Screenshot 2024-07-18 at 11 37 19

Towards https://github.com/giantswarm/roadmap/issues/3536.


mproffitt (approved):
LGTM

PR #4635, opened 2024-11-12
Extend allowed origins

ljakimczuk:
### What does this PR do?

Extends allowed origins for Happa API ingress.

Support for more allowed origins.

### How does it look like?

N/A

See this [issue](https://github.com/giantswarm/giantswarm/issues/32009)

A review of the proposed change.

Nope.

IDK, should it?


ljakimczuk:
@gusevda yes, please

PR #4800, opened 2026-06-08
feat(analytics): track only authenticated pageviews via TelemetryDeck JS SDK

marians:
### What does this PR do?

Replaces the auto-tracking TelemetryDeck **Web SDK** with the manual **JS SDK** (`@telemetrydeck/sdk`), so pageview signals fire **only for authenticated, in-app navigation**.

- Removes the Web SDK `<script>` from `src/index.ejs` and adds `@telemetrydeck/sdk` (pinned exact).
- `TelemetryService` — singleton wrapper (mirrors `ErrorReporter`), disabled in `development` (same gate as Sentry), hashes the logged-in user's email as `clientUser`.
- `normalizePath` — maps concrete paths to route patterns (e.g. `/organizations/acme/clusters/abc123` → `/organizations/:orgId/clusters/:clusterId`) and returns `null` for unknown paths, so org/cluster names and bot/probe paths never reach analytics.
- `usePageViewTracking` hook wired into `Layout` (the authenticated root). Each signal carries `route`, `host` (`window.location.hostname`), `installation`, and `environment`.
- Adds a repo-root `CLAUDE.md` documenting non-obvious workspace conventions encountered while building this (test naming, `window.config` in tests, `GlobalEnvironment` values, dual `src`/`src/components` import roots, Node-20/yarn engine pin, RR v5).

No visible UI change. Analytics now reflect real, authenticated usage instead of being polluted by unauthenticated bot/probing requests (e.g. `/horde/login.php`). As a side effect, in-app SPA route changes are now tracked properly (the Web SDK mostly only fired once on load).

### How does it look like?

No user-facing UI. Behavior is observable only in the TelemetryDeck dashboard / network traffic (signals to `telemetrydeck.com` with the payload described above).

The Web SDK auto-fired a pageview on load for *any* path the SPA shell was served on, so bot scans for paths like `/horde/login.php` showed up as pageviews. Filtering on authentication is the simplest robust filter — bots never authenticate, and the login/logout/OAuth-callback pages render outside `Layout` so they're excluded by construction.

- Sanity-check the **authenticated-only** filtering approach and the signal payload (note: hashed email is sent as `clientUser`; `host`/`installation` identify the deployment).
- Confirm the TelemetryDeck **appID** carried over (`746BB2DF-…`) is correct.
- Skim `normalizePath` route-pattern matching (uses `matchPath`, "fewer params wins" tie-break).

No.

Borderline (not user-visible, but changes what telemetry is sent). Suggested label: `kind/change`.

### DIFF
--- a/src/utils/telemetry/TelemetryService.ts
+++ b/src/utils/telemetry/TelemetryService.ts
@@ -1,19 +1,11 @@
 import TelemetryDeck from '@telemetrydeck/sdk';
+import { isExternalReportingEnabled } from 'utils/config';
 
 /**
  * TelemetryDeck application ID. Not a secret; it only identifies the analytics
  * app to send signals to.
  */
 const TELEMETRY_APP_ID = '746BB2DF-C3C7-4DE8-82A7-9848CEEE9585';
-
-/**
- * Whether analytics should be sent. Disabled in development so local traffic
- * never pollutes the dashboard (mirrors the Sentry gate in
- * `src/components/index.tsx`).
- */
-function isTrackingEnabled(): boolean {
-  return window.config.environment !== 'development';
-}
 
 /**
  * A singleton wrapper around the TelemetryDeck SDK for sending pageview
@@ -38,7 +30,7 @@ class TelemetryService {
    * `clientUser` in sync. The SDK hashes `clientUser` before sending.
    */
   private getClient(clientUser: string): TelemetryDeck | null {
-    if (!isTrackingEnabled()) {
+    if (!isExternalReportingEnabled()) {
       return null;
     }
 
--- a/src/utils/telemetry/TelemetryService.ts
+++ b/src/utils/telemetry/TelemetryService.ts
@@ -1,0 +1,80 @@
+import TelemetryDeck from '@telemetrydeck/sdk';
+
+/**
+ * TelemetryDeck application ID. Not a secret; it only identifies the analytics
+ * app to send signals to.
+ */
+const TELEMETRY_APP_ID = '746BB2DF-C3C7-4DE8-82A7-9848CEEE9585';
+
+/**
+ * Whether analytics should be sent. Disabled in development so local traffic
+ * never pollutes the dashboard (mirrors the Sentry gate in
+ * `src/components/index.tsx`).
+ */
+function isTrackingEnabled(): boolean {
+  return window.config.environment !== 'development';
+}
+
+/**
+ * A singleton wrapper around the TelemetryDeck SDK for sending pageview
+ * signals. Tracking is disabled in development, mirroring the Sentry gate in
+ * `src/components/index.tsx`.
+ */
+class TelemetryService {
+  private static _instance: TelemetryService | null = null;
+
+  public static getInstance(): TelemetryService {
+    if (!TelemetryService._instance) {
+      TelemetryService._instance = new TelemetryService();
+    }
+
+    return TelemetryService._instance;
+  }
+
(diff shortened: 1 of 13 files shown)


src/utils/telemetry/__tests__/TelemetryService.ts
```diff
@@ -0,0 +1,80 @@
...
+  );
+});
+
+afterEach(() => {
+  setEnvironment(originalEnvironment);
+});
+
+describe('TelemetryService', () => {
+  it('is a singleton', () => {
+    expect(new TelemetryService()).toEqual(TelemetryService.getInstance());
```
marians: This test doesn't actually verify the singleton. `toEqual` is deep equality, and both `new TelemetryService()` and `getInstance()` are fresh objects with `client: null`, so they compare equal regardless of identity. The assertion would still pass if `getInstance()` were changed to `return new TelemetryService()` on every call (i.e. not a singleton at all). To test the actual contract, call `getInstance()` twice and assert reference identity:

```suggestion
    expect(TelemetryService.getInstance()).toBe(TelemetryService.getInstance());
```

src/utils/telemetry/TelemetryService.ts
```diff
@@ -0,0 +1,80 @@
...
+ * app to send signals to.
+ */
+const TELEMETRY_APP_ID = '746BB2DF-C3C7-4DE8-82A7-9848CEEE9585';
+
+/**
+ * Whether analytics should be sent. Disabled in development so local traffic
+ * never pollutes the dashboard (mirrors the Sentry gate in
+ * `src/components/index.tsx`).
+ */
+function isTrackingEnabled(): boolean {
```
marians: The `window.config.environment !== 'development'` gate is now duplicated — the Sentry init in `src/components/index.tsx:32` uses the exact same condition, and the doc comment here explicitly ties the two together. If the "what counts as a non-tracking environment" rule ever changes (e.g. a new staging env is added), both sites must be updated in lockstep. Consider extracting a single shared predicate (e.g. `utils/config/isTelemetryEnvironment()`) and using it in both places.

src/utils/telemetry/usePageViewTracking.ts
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Sends a TelemetryDeck pageview signal for each authenticated navigation.
+ *
+ * Tracking is gated on a logged-in user, so bot probes and unauthenticated
+ * pages (login, logout, OAuth callback) are never tracked. The effect runs
+ * once the user becomes known and again on every route change, which covers
+ * both the initial pageview and subsequent in-app navigation.
+ */
+export function usePageViewTracking(): void {
```
marians, line 17: `TelemetryService` and `normalizePath` each got a test file, but the hook — which holds the actual gating logic (early-return on missing `email`, early-return on `normalizePath` returning `null`, firing on `[pathname, email]` changes) — has none. That branching is exactly where a regression would land (e.g. tracking before login, or double-firing). A small `@testing-library/react` `renderHook` test mocking `TelemetryService.getInstance().trackPageView` would cover it.

marians (commented):
Code review (medium effort, precision-focused). The change is well-structured and the correctness angles came back clean — `normalizePath`'s "fewer params wins" tie-break is correct (verified `/clusters/new` resolves to the static route over both `:clusterId` variants), selecting `?.email` as a primitive keeps the effect from re-firing on store-object churn, and the `.catch()` on the signal is right. Posted 3 quality findings inline; none are blocking.

QuentinBisson (approved):

PR #4808, opened 2026-06-10
fix(ci): unbreak build-fe-docs and chart tests

marians:
### What does this PR do?

Reverts two dependency bumps that were merged to `main` and broke CI for **all** open PRs:

- **`brace-expansion` v5** (#4781) → back to `^2.0.2`. v5 is ESM-only and breaks the CommonJS `require(...).default` interop that Storybook's bundled `minimatch@9.0.3` relies on, crashing the `build-fe-docs` job with `TypeError: (0 , brace_expansion_1.default) is not a function`. `^2.0.2` still includes the ReDoS fix (GHSA / CVE-2025-5889).
- **`pytest` v9** (#4747) → back to `==8.4.2`. `pytest-helm-charts==1.3.4` (latest) requires `pytest<9`, so v9 makes `ats/Pipfile.lock` unresolvable and fails the `execute chart tests` job (`Your Pipfile.lock is out of date … Aborting deploy`). No `pytest-helm-charts` release supports pytest 9 yet.

Also adds Renovate guardrails (`allowedVersions`) so neither bump is re-proposed until our toolchain can support it.

None — CI/test-tooling only. No product code changes.

The pytest v9 bump was security-labelled (GHSA-6w46-j5rx-g56g, `/tmp/pytest-of-{user}` handling, fixed only in 9.0.3). That advisory is local-multi-user DoS/privesc; this is a test-only dependency running in ephemeral single-user KinD CI containers, so the practical exposure is negligible. We can move to pytest 9 once `pytest-helm-charts` supports it.

Confirm the trade-off on holding pytest at 8.x is acceptable.

### Verification

- `yarn storybook:build` → `Preview built`, succeeds (failed before).
- `pipenv install --deploy` in `python:3.12.7` → exit 0, no hash mismatch (failed before).

No.

### DIFF
--- a/ats/Pipfile
+++ b/ats/Pipfile
@@ -5,6 +5,6 @@ verify_ssl = true
 
 [packages]
 pytest-helm-charts = "==1.3.4"
-pytest = "==9.0.3"
+pytest = "==8.4.2"
 pykube-ng = "==23.6.0"
 pytest-rerunfailures = "==11.1.2"
--- a/package.json
+++ b/package.json
@@ -252,7 +252,7 @@
     "underscore": "1.13.8",
     "ws": "^8.17.1",
     "string-width": "4.2.3",
-    "brace-expansion": "^5.0.0",
+    "brace-expansion": "^2.0.2",
     "flatted": "^3.3.3",
     "mdast-util-to-hast": "^13.2.1",
     "nanoid": "^5.0.0",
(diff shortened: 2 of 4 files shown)


teemow:
@marians reverting dependency updates is not going to work. renovate will bring them up again and they will be merged again. 

The more important things to focus on:
 * why can't we simply fix it forward with the new version?
 * why did CI not block the merge in the first place?

